PR #4, opened 2017-03-17
Subworkflows are siloed. Validation fixes to allow for multiple workflows with resources of the same name.

crunk1:
Subworkflows are now siloed, meaning that each workflows' vars and resource names will not collide with/refer to other workflows' vars and resource names. 
Workflows do not modify subworkflows, except for the subworkflows' GCE and GCS locations. This makes it possible for parent workflows to make their subworkflows run in the same GCS Bucket, GCE Project and Zone.

Workflows' prerun checks, populate/validation/sources uploads, are all done at the parent workflow's prerun() instead of at the subworkflows' step run time. For example, a workflow w and a subworkflow sw: w.prerun() will run sw.prerun(). This prevents w from running until sw, only to find that sw has a population/validation/source issue.
Also, modified validation to allow for per-workflow (instead of the previously global) resource enumeration and checking. Since daisy accepts multiple workflow templates, this needed to happen regardless of the workflow/subworkflow discussion.


daisy/daisy.go
```diff
@@ -65,7 +65,8 @@ func main() {
 
 	var wfs []*workflow.Workflow
 	for _, path := range flag.Args() {
-		wf, err := workflow.ReadWorkflow(path)
+		wf := workflow.New(ctx)
```
adjackura: i think this needs to be context.Background() so each workflow get their own

daisy/workflow/workflow.go
```diff
@@ -577,6 +564,14 @@ func (w *Workflow) uploadSources() error {
 	return nil
 }
 
+// New instantiates a new workflow.
+func New(ctx context.Context) *Workflow {
+	var w Workflow
+	w.Ctx, w.Cancel = context.WithCancel(ctx)
+	w.id = randString(5)
```
crunk1: FromFile will call New() for subworkflows.

daisy/workflow/workflow.go
```diff
@@ -577,6 +564,14 @@ func (w *Workflow) uploadSources() error {
 	return nil
 }
 
+// New instantiates a new workflow.
+func New(ctx context.Context) *Workflow {
```
crunk1: Mind if I punt this to future discussion/PRs?

daisy/workflow/common.go
```diff
@@ -123,6 +120,16 @@ func substitute(s reflect.Value, replacer *strings.Replacer) {
 		if !f.CanSet() {
 			continue
 		}
+
+		// Don't recurse on subworkflows. Let subworkflows do their own substitutions.
+		switch f.Interface().(type){
+		case *Workflow:
```

daisy/workflow/workflow.go
```diff
@@ -111,85 +112,125 @@ func (s *Step) realStep() (step, error) {
...
+		return nil, errors.New("multiple step types defined")
 	}
 	return result, nil
 }
 
 func (s *Step) run(w *Workflow) error {
 	realStep, err := s.realStep()
 	if err == nil {
-		return realStep.run(w)
+		err = realStep.run(w)
```
crunk1: Done.

daisy/workflow/workflow.go
```diff
@@ -111,85 +112,125 @@ func (s *Step) realStep() (step, error) {
...
+	}
+	return s.wrapRunError(err)
 }
 
-func (s *Step) validate() error {
+func (s *Step) validate(w *Workflow) error {
 	realStep, err := s.realStep()
 	if err == nil {
-		return realStep.validate()
+		err = realStep.validate(w)
```
crunk1: Done.

daisy/daisy.go
crunk1: Done.

daisy/workflow/workflow.go
adjackura: Yep, no need to do this here.

daisy/workflow/workflow.go
adjackura: Ah I missed that bit.

daisy/workflow/common.go
```diff
@@ -133,6 +131,12 @@ func substitute(s reflect.Value, replacer *strings.Replacer) {
 
 		raw := f.Interface()
 		switch raw.(type) {
+		case *Workflow:
+			switch s.Interface().(type) {
+			case SubWorkflow:
```
adjackura: I want to say you should add a default case here that just does this:
substitute(e, replacer)

Not that it is use currently

daisy/workflow/step_sub_workflow_test.go
```diff
@@ -1,169 +0,0 @@
-//  Copyright 2017 Google Inc. All Rights Reserved.
```
adjackura: So the tests in this file are important and are not replicated in any other tests, particularly the referencing of resources before and after subworkflow runs. This test as is will fail currently but we will want it once we hammer out how subworkflows and workflows pass resources to each other. 
Maybe leave the file but comment out the tests with a TODO to make them work?

daisy/workflow/common.go
crunk1: That is incorrect. For one, e doesn't even exist at this point. Also, to be more versatile to changes to SubWorkflow, the default case should look more like the default case below. After futzing around with this for a good 15 minutes, I find it is best to revert this change back to my previous commit.

daisy/workflow/step_sub_workflow_test.go
crunk1: Done.

PR #13, opened 2017-03-27
Add better logging of JSON syntax errors

adjackura:
JSON syntax errors now return a helpful error:
2017/03/27 11:07:39 JSON syntax error in line 34: invalid character '"' after object key:value pair 
          "metadata": {
          ^

Fixed substitute not running on WaitForInstancesStopped and WaitForInstancesSignal steps

### DIFF
--- a/daisy/workflow/common.go
+++ b/daisy/workflow/common.go
@@ -162,6 +162,12 @@ func substitute(s reflect.Value, replacer *strings.Replacer) {
 		case *compute.Client, *storage.Client, context.Context, context.CancelFunc:
 			// We specifically do not want to change fields with these types.
 			continue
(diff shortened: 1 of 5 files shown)


daisy/workflow/common.go
```diff
@@ -162,6 +162,18 @@ func substitute(s reflect.Value, replacer *strings.Replacer) {
 		case *compute.Client, *storage.Client, context.Context, context.CancelFunc:
 			// We specifically do not want to change fields with these types.
 			continue
+		case *WaitForInstancesStopped:
```
crunk1: Didn't we have specific logic for handling slices vs structs?

PR #14, opened 2017-03-28
Hierarchical resource references. 

crunk1:
Added get*() methods and a test for them. Resources are queried at the local workflow level, then the parent, then grandparent, etc.

### DIFF
--- a/daisy/workflow/step_create_disks.go
+++ b/daisy/workflow/step_create_disks.go
@@ -72,11 +72,20 @@ func (c *CreateDisks) run(w *Workflow) error {
 			if !cd.ExactName {
 				name = w.genName(cd.Name)
 			}
-			imageLink := resolveLink(cd.SourceImage, w.imageRefs)
-			if imageLink == "" {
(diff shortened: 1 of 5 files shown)


daisy/workflow/step_create_disks.go
```diff
@@ -72,11 +72,20 @@ func (c *CreateDisks) run(w *Workflow) error {
...
-			if imageLink == "" {
-				e <- fmt.Errorf("unresolved image %q", cd.SourceImage)
+
+			// Get the source image link.
+			var imageLink string
+			var image *resource
+			var err error
+			if isLink(cd.SourceImage) {
+				imageLink = cd.SourceImage
+			} else if image, err = w.getImage(cd.SourceImage); err == nil {
```
adjackura: } else if image, err := w.getImage(cd.SourceImage); err == nil {

no need to declare them above if they are not used outside this scope

crunk1: err is returned if err != nil

PR #15, opened 2017-03-29
Fix cancel, add logging of serial output

adjackura:
Don't use context.WithCancel for handling stopping of workflows, this caused issues because the cleanup function relied on that context.
Stream serial port 1 to GCS when WaitForInstancesStopped is called
Remove steps that are better suited for subworkflows


daisy/workflow/step_wait_for_instances_stopped.go
```diff
@@ -15,13 +15,58 @@
...
 	"fmt"
+	"path"
 	"sync"
+	"time"
 )
 
 // WaitForInstancesStopped is a Daisy WaitForInstancesStopped workflow step.
 type WaitForInstancesStopped []string
 
+func steamSerial(w *Workflow, name string) {
```
crunk1: s/steam/stream

daisy/workflow/step_wait_for_instances_stopped.go
```diff
@@ -15,13 +15,58 @@
...
+	var buf bytes.Buffer
+	var seenErr bool
+	for {
+		tick := time.Tick(1 * time.Second)
+		select {
+		case <-w.Ctx.Done():
+		case <-tick:
+			resp, err := w.ComputeClient.GetSerialPortOutput(w.Project, w.Zone, name, 1, start)
+			if err != nil {
+				// This is most likely because the instance has stopped, don't log but try one more time.
```
crunk1: Would it be better to explicitly check to see if the instance stopped?

daisy/workflow/test_common.go
```diff
@@ -71,15 +71,15 @@ func init() {
 }
 
 func testWorkflow() *Workflow {
-	ctx, cancel := context.WithCancel(context.Background())
+	cancel := make(chan struct{})
```
crunk1: Why not using the Cancel? I thought cancelling a "parent" context would cascade to "children" contexts, cancelling them as well. In short, how does cancel work?

daisy/workflow/workflow.go
```diff
@@ -224,8 +217,8 @@ func (s *Step) wrapValidateError(e error) error {
 // Workflow is a single Daisy workflow workflow.
 type Workflow struct {
 	// Populated on New() construction.
-	Ctx    context.Context    `json:"-"`
-	Cancel context.CancelFunc `json:"-"`
+	Ctx    context.Context `json:"-"`
```
crunk1: I can't remember, do these need to be public?

daisy/workflow/workflow.go
```diff
@@ -315,26 +308,27 @@ func (w *Workflow) FromFile(file string) error {
...
 	}
 	w.logger.Print("Uploading sources")
 	if err := w.uploadSources(); err != nil {
 		w.logger.Printf("Error uploading sources: %v", err)
-		w.Cancel()
+		close(w.Cancel)
 		return err
 	}
 	w.logger.Print("Running workflow")
+	defer w.cleanup()
```
crunk1: We could maybe even put this at the top of Run(). Say, for example, we want to clean up artifacts from uploadSources().

daisy/workflow/test_common.go
adjackura: For canceling the running workflow it works fine, the issue is we also use this context for cleanup which happens after a cancel call.

daisy/workflow/workflow.go
adjackura: Yeah, daisy.go uses them

daisy/workflow/workflow.go
adjackura: Moved it right after validate in case we add any of that.

PR #16, opened 2017-03-30
Add scopes to CreateInstance

adjackura:


daisy/compute/compute.go
```diff
@@ -284,10 +285,14 @@ func (i *Instance) Insert() (*compute.Instance, error) {
...
 
+	scopes := []string{"https://www.googleapis.com/auth/devstorage.read_only"}
+	if len(i.scopes) > 0 {
+		scopes = i.scopes
+	}
 	serviceAccounts := []*compute.ServiceAccount{
 		{
 			Email:  "default",
-			Scopes: []string{"https://www.googleapis.com/auth/devstorage.full_control"},
+			Scopes: scopes,
```
crunk1: If scopes gets passed in as []string{}, what happens?

PR #17, opened 2017-03-30
Don't use '_' in JSON workflows

adjackura:
This will make workflows match the code more closely, also the API uses camel case so it would be good to match that too.
Also


daisy/workflow/test.workflow
```diff
@@ -2,68 +2,68 @@
   "name": "some-name",
   "project": "some-project",
   "zone": "us-central1-a",
-  "gcs_path": "gs://some-bucket/images",
+  "gcsPath": "gs://some-bucket/images",
```
crunk1: Is the JSON unmarshaller case sensitive?

PR #20, opened 2017-03-31
Make sources upload recursive, allow for blank SizeGb if SourceImage is set

adjackura:


daisy/workflow/sources.go
```diff
@@ -0,0 +1,109 @@
+package workflow
```
crunk1: Any tests for this?

adjackura: Added tests

daisy/workflow/sources_test.go
```diff
@@ -0,0 +1,57 @@
...
+	tests := []struct {
+		desc    string
+		sources map[string]string
+		err     string
+		gcs     []string
+	}{
+		{"normal local file to GCS", map[string]string{"local": testPath}, "", []string{w.sourcesPath + "/local"}},
+		{"normal local folder to GCS", map[string]string{"local": dir}, "", []string{w.sourcesPath + "/local/test"}},
+		{"normal GCS obj to GCS", map[string]string{"gcs": "gs://gcs/file"}, "", []string{w.sourcesPath + "/gcs"}},
+		{"normal GCS bkt to GCS", map[string]string{"gcs": "gs://gcs/folder"}, "", []string{w.sourcesPath + "/gcs/object", w.sourcesPath + "/gcs/folder/object"}},
```
crunk1: Where does the source gs://gcs/folder/object get set? And gs://gcs/folder/folder/object?

I'm assuming this is analogous to cp -R gs://gcs/folder gs://<sourcesPath>/gcs.

daisy/workflow/sources_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{"dne local path", map[string]string{"local": "./this/file/dne"}, "stat ./this/file/dne: no such file or directory", nil},
+		{"dne GCS path", map[string]string{"gcs": "gs://gcs/path/dne"}, `source "gs://gcs/path/dne" is not a GCS bucket or object`, nil},
+		{"GCS path, no object", map[string]string{"gcs": "gs://folder"}, "", []string{w.sourcesPath + "/gcs/object", w.sourcesPath + "/gcs/folder/object"}},
```
crunk1: Again, where are ./object and ./folder/object coming from? Not saying there is an issue, I'm just trying to follow the logic.

daisy/workflow/step_create_disks.go
```diff
@@ -47,11 +47,15 @@ func (c *CreateDisks) validate(w *Workflow) error {
 			return fmt.Errorf("cannot create disk: image not found: %s", cd.SourceImage)
 		}
 
-		_, err := strconv.ParseInt(cd.SizeGB, 10, 64)
-		if err != nil {
+		if _, err := strconv.ParseInt(cd.SizeGB, 10, 64); cd.SizeGB != "" && err != nil {
```
crunk1: Does ParseInt("", 10, 64) not return an error? Does it return 0 or something?

adjackura: technically it will return 0 but only because that's what happens when you declare an int, the error it gives is this: strconv.ParseInt: parsing "": invalid syntax

daisy/workflow/sources.go
```diff
@@ -0,0 +1,118 @@
...
+		// GCS to GCS.
+		if bkt, prefix, err := splitGCSPath(origPath); err == nil {
+			if prefix == "" {
+				if err := w.recursiveGCS(bkt, prefix, dst); err != nil {
+					return err
+				}
+				continue
+			}
+			src := w.StorageClient.Bucket(bkt).Object(path.Clean(prefix))
+			// If this is a GCS 'directory' we will get ErrObjectNotExist.
```
adjackura: Yep, fixed so in this case the user would have to have the gcs path as gs://bucket/object/

daisy/workflow/sources.go
```diff
@@ -0,0 +1,114 @@
...
+		return err
+	}
+	return gcs.Close()
+}
+
+func (w *Workflow) uploadSources() error {
+	for dst, origPath := range w.Sources {
+		// GCS to GCS.
+		if bkt, objPath, err := splitGCSPath(origPath); err == nil {
+			if objPath == "" || strings.HasSuffix(objPath, "/") {
```
crunk1: What do you think about forcing people to specify GCS "folders" by making the trailing "/" required? (instead of doing your ErrObjectNotExist check below)

daisy/workflow/step_create_disks.go
crunk1: In that case, do you need the cd.SizeGB != ""... oh, I get it now... nvm.

daisy/workflow/sources.go
adjackura: Yeah that does make things more straightforward.

PR #21, opened 2017-04-01
Cleaned up substitute(). Substitute is more versatile and can work on more types now.

crunk1:


daisy/workflow/common.go
```diff
@@ -97,97 +93,60 @@ func splitGCSPath(p string) (string, string, error) {
...
 }
 
-// substitute iterates through the public fields of the struct represented
-// by s, if the field type matches one of the known types the provided
-// replacer.Replace is run on all string values replacing the original value
-// in the underlying struct.
-// Exceptions: will not change Vars fields or Workflow fields of SubWorkflow types.
-func substitute(s reflect.Value, replacer *strings.Replacer) {
-	if s.Kind() != reflect.Struct {
+// substitute analyzes an element for string values and replaces
```
adjackura: I would mention replacer.Replace here, this function does take any strings for replacement. Also mention the iteration  depending on the element handed in.

PR #25, opened 2017-04-05
Change SSD flag to DiskType in CreateDisks

adjackura:
Drop the SSD bool option in CreateDisks and replace it with a string option of DiskType
compute.CreateDisk can take either the full path to the disk type '"zones/my-zone/diskTypes/pd-standard" or just the type "pd-standard" and will create the full path.


daisy/README.md
```diff
@@ -85,7 +85,7 @@ PD disk created from a source image, the second is blank PD SSD.
     {
       "name": "disk2",
       "sizeGb": "200",
-      "ssd": true
+      "diskType": "pd-ssd"
```
crunk1: The API uses "type".

adjackura: Done

PR #26, opened 2017-04-07
Resolve SourceFile in CreateDisks and some cleanups/TODOs

adjackura:
- Resolve SourceFile in CreateDisks
  - SourceFile can be either GCS path or a file reference from sources.
- Test cleanups
- Change FromFile to NewFromFile
- Allow for storage and compute endpoint overrides

### DIFF
--- a/daisy/workflow/sources.go
+++ b/daisy/workflow/sources.go
@@ -1,12 +1,12 @@
 package workflow
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
-	"fmt"
 
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/iterator"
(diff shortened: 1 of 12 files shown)


daisy/workflow/sources.go
```diff
@@ -32,6 +31,13 @@ func (w *Workflow) recursiveGCS(bkt, prefix, dst string) error {
 	return nil
 }
 
+func (w *Workflow) sourceExists(s string) bool {
+	if _, ok := w.Sources[s]; ok {
```
crunk1: _, ok := w.Sources[s]
return ok

daisy/workflow/step_create_instances.go
```diff
@@ -60,7 +60,7 @@ func (c *CreateInstances) validate(w *Workflow) error {
 		}
 
 		// Startup script checking.
-		if !sourceExists(ci.StartupScript) {
+		if ci.StartupScript != "" && !w.sourceExists(ci.StartupScript) {
```
crunk1: Are you making it so startup script HAS to be in sources?

daisy/workflow/workflow.go
```diff
@@ -663,6 +613,57 @@ func New(ctx context.Context) *Workflow {
 	return &w
 }
 
+// NewFromFile reads and unmarshals a workflow file.
+// Recursively reads subworkflow steps as well.
+func NewFromFile(ctx context.Context, file string) (*Workflow, error) {
```
crunk1: Me gusta.

PR #29, opened 2017-04-07
Update WaitForInstancesSignal, remove WaitForInstancesStopped

adjackura:
WaitForInstancesSignal now can wait for an instance to stop or for for specific serial output data.
CreateInstances automatically streams serial port 1 to GCS

If SerialOutput is set the step will not complete until a line in the serial output matches
SuccessMatch or FailureMatch. A match with FailureMatch will cause the step
to fail.


daisy/workflow/step_wait_for_instances_signal.go
```diff
@@ -14,20 +14,125 @@
...
+	SerialOutput *SerialOutput
+}
+
+func waitForSerialOutput(w *Workflow, name string, port int64, success, failure string, interval time.Duration) error {
+	w.logger.Printf("WaitForInstancesSignal: watching serial port %d, SuccessMatch: %q, FailureMatch: %q.", port, success, failure)
+	var start int64
+	for {
+		tick := time.Tick(interval)
+		select {
+		case <-w.Ctx.Done():
```
crunk1: Same as my question above.

daisy/workflow/step_wait_for_instances_signal.go
```diff
@@ -14,20 +14,125 @@
...
+		return nil
+	}
 }
 
 func (s *WaitForInstancesSignal) validate(w *Workflow) error {
 	// Instance checking.
 	for _, i := range *s {
-		if !instanceValid(w, i) {
-			return fmt.Errorf("cannot wait for instance signal. Instance not found: %s", i)
+		if !instanceValid(w, i.Name) {
```
crunk1: Do we still need instanceValid? Can we do 

if _, ok := w.instanceRefs.get(i.Name); !ok {}

crunk1: If we can get rid of instanceValid, we don't have to change it in this PR; file a bug against me.

adjackura: I'll leave that up to you, you have a better handle on how the validation logic works

daisy/workflow/step_create_instances.go
```diff
@@ -47,6 +50,41 @@ type CreateInstance struct {
...
 
+func logSerialOutput(w *Workflow, name string, port int64) {
+	logsObj := path.Join(w.logsPath, fmt.Sprintf("%s-serial-port%d.log", name, port))
+	w.logger.Printf("CreateInstances: streaming instance %q serial port %d output to gs://%s/%s.", name, port, w.bucket, logsObj)
+	var start int64
+	var buf bytes.Buffer
+	for {
+		tick := time.Tick(1 * time.Second)
+		select {
+		case <-w.Ctx.Done():
```
adjackura: Done

daisy/workflow/step_wait_for_instances_signal.go
adjackura: Done

PR #34, opened 2017-04-17
Ensure gcs objects don't contain '\', add timestamp to scratchpath

adjackura:
ensure gcs objects don't contain '\' (the case when windows paths get translated to GCS objects)
add timestamp to scratchpath

### DIFF
--- a/daisy/workflow/common.go
+++ b/daisy/workflow/common.go
@@ -74,7 +74,7 @@ func isLink(s string) bool {
 
 func randString(n int) string {
 	gen := rand.New(rand.NewSource(time.Now().UnixNano()))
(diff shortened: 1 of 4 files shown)


daisy/workflow/common.go
```diff
@@ -74,7 +74,7 @@ func isLink(s string) bool {
 
 func randString(n int) string {
 	gen := rand.New(rand.NewSource(time.Now().UnixNano()))
-	letters := "abcdefghijklmnopqrstuvwxyz"
+	letters := "bdghjlmnpqrstvwxyz0123456789"
```
crunk1: What's the reason behind these characters?

PR #41, opened 2017-05-05
Add descriptions to resources

adjackura:
Daisy will add a default description to instances, disks, and images. A custom description can be set in the workflow.


daisy/compute/compute.go
```diff
@@ -90,7 +90,7 @@ func (c *Client) operationsWait(project, zone, name string) error {
 }
 
 // CreateDisk creates a GCE persistant disk.
```
crunk1: persistEnt

PR #45, opened 2017-05-10
Bug fixes and feature additions

adjackura:
Populate WaitForInstancesSignal intervals
Fix index error if json error is the first position
Add USER autovar and add it to resource descriptions
Add validate flag to daisy that will just run validation
WaitForInstancesSignal can wait for only FailureMatch

### DIFF
--- a/daisy/workflow/workflow.go
+++ b/daisy/workflow/workflow.go
@@ -33,6 +33,8 @@ import (
 	"sync"
 	"time"
 
+	"os/user"
+
 	"cloud.google.com/go/storage"
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
 	"google.golang.org/api/option"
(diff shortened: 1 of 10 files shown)


daisy/workflow/workflow.go
```diff
@@ -428,20 +457,33 @@ func (w *Workflow) populateStep(step *Step) error {
...
 	vars := map[string]string{}
 	for k, v := range w.Vars {
 		vars[k] = v
 	}
+
 	var replacements []string
 	for k, v := range vars {
 		replacements = append(replacements, fmt.Sprintf("${%s}", k), v)
 	}
+	for k, v := range autovars {
```
crunk1: We might want to put this for loop above the previous for loop. Reason? Until now, user vars overrode autovars. With this autovars for loop AFTER the vars for loop, the autovars ID, DATE, DATETIME, TIMESTAMP, and USERNAME **can't** be overridden. That's fine, but inconsistent since the other autovars NAME, ZONE, PROJECT, etc. **can** be overridden.

adjackura: Done

PR #50, opened 2017-05-11
Allow creating of instances and disks in any project and zone.

adjackura:
Actually allow partial URL's
Add more tests, clean up existing ones.

### DIFF
--- a/daisy/workflow/step_create_disks.go
+++ b/daisy/workflow/step_create_disks.go
@@ -100,7 +100,7 @@ func (c *CreateDisks) run(w *Workflow) error {
 			var imageLink string
 			var image *resource
 			var err error
-			if cd.SourceImage == "" || (isLink(cd.SourceImage) && imageValid(w, cd.SourceImage)) {
+			if cd.SourceImage == "" || imageURLRegex.MatchString(cd.SourceImage) {
 				imageLink = cd.SourceImage
 			} else if image, err = w.getImage(cd.SourceImage); err == nil {
 				imageLink = image.link
--- a/daisy/workflow/step_create_disks.go
+++ b/daisy/workflow/step_create_disks.go
@@ -38,6 +38,10 @@ type CreateDisk struct {
 	// Optional description of the resource, if not specified Daisy will
 	// create one with the name of the project.
 	Description string `json:",omitempty"`
+	// Zone to create the instance in, overrides workflow zone.
+	Zone string
+	// Project to create the instance in, overrides workflow project.
+	Project string
 	// Should this resource be cleaned up after the workflow?
 	NoCleanup bool
 	// Should we use the user-provided reference name as the actual
@@ -82,11 +86,21 @@ func (c *CreateDisks) run(w *Workflow) error {
 				name = w.genName(cd.Name)
 			}
 
+			zone := w.Zone
+			if cd.Zone != "" {
+				zone = cd.Zone
+			}
+
+			project := w.Project
+			if cd.Project != "" {
+				project = cd.Project
+			}
+
 			// Get the source image link.
 			var imageLink string
 			var image *resource
 			var err error
-			if cd.SourceImage == "" || isLink(cd.SourceImage) {
+			if cd.SourceImage == "" || (isLink(cd.SourceImage) && imageValid(w, cd.SourceImage)) {
 				imageLink = cd.SourceImage
 			} else if image, err = w.getImage(cd.SourceImage); err == nil {
 				imageLink = image.link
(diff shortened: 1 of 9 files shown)


daisy/workflow/step_create_disks.go
```diff
@@ -38,6 +38,10 @@ type CreateDisk struct {
 	// Optional description of the resource, if not specified Daisy will
 	// create one with the name of the project.
 	Description string `json:",omitempty"`
+	// Zone to create the instance in, overrides workflow zone.
```
crunk1: Not necessary, but I've been capitalizing Zone, Project, etc. so it matches the Workflow struct's fields. Take it or leave it.

daisy/workflow/step_create_disks.go
```diff
@@ -82,11 +86,21 @@ func (c *CreateDisks) run(w *Workflow) error {
...
+			if cd.Project != "" {
+				project = cd.Project
+			}
+
 			// Get the source image link.
 			var imageLink string
 			var image *resource
 			var err error
-			if cd.SourceImage == "" || isLink(cd.SourceImage) {
+			if cd.SourceImage == "" || (isLink(cd.SourceImage) && imageValid(w, cd.SourceImage)) {
```
crunk1: Is the "isLink" necessary here anymore since you are using URL regexes in the *Valid functions?

daisy/workflow/step_create_images.go
```diff
@@ -91,12 +91,17 @@ func (c *CreateImages) run(w *Workflow) error {
...
+				project = ci.Project
+			}
+
 			// Get source disk link, if applicable.
 			var diskLink string
 			if ci.SourceDisk != "" {
 				var disk *resource
 				var err error
-				if isLink(ci.SourceDisk) {
+				if isLink(ci.SourceDisk) && diskValid(w, ci.SourceDisk) {
```
crunk1: Ditto.

daisy/workflow/step_create_instances.go
```diff
@@ -100,6 +104,33 @@ func (c *CreateInstances) validate(w *Workflow) error {
 			if !diskValid(w, d) {
 				return fmt.Errorf("cannot create instance: disk not found: %s", d)
 			}
+			// Ensure disk is in the same project and zone.
+			if !isLink(d) {
```
crunk1: Do you need this check here? https://golang.org/pkg/regexp/#Regexp.FindStringSubmatch says that 'match' will be nil if 'd' doesn't match the URL regex.

daisy/workflow/step_create_instances.go
```diff
@@ -100,6 +104,33 @@ func (c *CreateInstances) validate(w *Workflow) error {
 			if !diskValid(w, d) {
 				return fmt.Errorf("cannot create instance: disk not found: %s", d)
 			}
+			// Ensure disk is in the same project and zone.
```
crunk1: Good idea.

daisy/workflow/step_create_instances.go
```diff
@@ -143,7 +184,7 @@ func (c *CreateInstances) run(w *Workflow) error {
 			for i, sourceDisk := range ci.AttachedDisks {
 				var disk *resource
 				var err error
-				if isLink(sourceDisk) {
+				if isLink(sourceDisk) && diskValid(w, sourceDisk) {
```
crunk1: Ditto.

daisy/workflow/step_create_instances.go
```diff
@@ -143,7 +184,7 @@ func (c *CreateInstances) run(w *Workflow) error {
...
 					// Real link.
 					inst.AddPD("", sourceDisk, false, i == 0)
 				} else if disk, err = w.getDisk(sourceDisk); err == nil {
```
crunk1: We will have to do the same zone/project checking on this as you did above.

daisy/workflow/step_create_disks.go
adjackura: Yes, imageValid will return true on a disk reference. This section I need to know if it is a link

daisy/workflow/step_create_disks.go
adjackura: Done

daisy/workflow/step_create_instances.go
adjackura: Done

daisy/workflow/step_create_disks.go
crunk1: What about imageURLRegex.Match(cd.SourceImage) instead? isLink might be obsoleted by your previous PR.

daisy/workflow/step_create_instances.go
adjackura: This needs to be done at validation time. To do that we need to store project and zone in nameSet, re reuse refmap instead but with a noop call so the compute service isn't called in validate.

crunk1: That's a good idea.

crunk1: We could do it for partialURLs too, actually.

daisy/workflow/step_create_disks.go
adjackura: Done

daisy/workflow/step_create_images.go
adjackura: Done

PR #53, opened 2017-05-12
Clean up RHEL/CentOS build workflows.

crunk1:


daisy_workflows/image_build/centos/centos6.wf.json
```diff
@@ -4,20 +4,17 @@
...
       "SubWorkflow": {
         "Path": "./centos.wf.json",
         "Vars": {
           "image-dest": "${image-dest}",
-          "installer-iso": "${SOURCESPATH}/installer.iso",
+          "installer-iso": "${installer-iso}",
           "license-id": "1000206",
           "release": "centos6",
-          "wf-name": "build-centos6"
+          "wf-name": "centos6"
```
adjackura: What is this used for? I think just using a standard subworkflow name (instead of a var) would work fine.

crunk1: I don't like how long the resource names were getting: installerprep-build-rhel6-abcde vs installerprep-rhel6-abcde
Also, when the build VM was running, it was build-build-rhel6-abcde.

crunk1: My examples were wrong. They would have been installerprep-build-rhel-abcde and build-build-rhel-abcde, which illustrates another reason to allow the subworkflows' names to be overridable.

PR #54, opened 2017-05-13
Clean up cleanup on cancel, validate step and workflow names

adjackura:
WaitForInstancesSignal doesn't error when instance is stopped or stopping
Add tests for WaitForInstancesSignal
Fixed bug on subworkflow failure not ending workflow
Don't log errors after a workflow is canceled
Subworkflows are named after their step name

### DIFF
--- a/daisy/workflow/test.workflow
+++ b/daisy/workflow/test.workflow
@@ -8,7 +8,7 @@
     "machine_type": "n1-standard-1"
   },
   "steps": {
-    "create disks": {
+    "create-disks": {
       "createDisks": [
         {
           "name": "bootstrap",
@@ -37,7 +37,7 @@
         }
       ]
     },
-    "${bootstrap_instance_name} stopped": {
+    "${bootstrap_instance_name}-stopped": {
       "timeout": "1h",
       "waitForInstancesSignal": [{"name": "${bootstrap_instance_name}", "stopped": true, "interval": "1s"}]
     },
@@ -51,10 +51,10 @@
         }
       ]
     },
-    "postinstall stopped": {
+    "postinstall-stopped": {
       "waitForInstancesSignal": [{"name": "postinstall", "stopped": true}]
     },
-    "create image": {
+    "create-image": {
       "createImages": [
         {
           "name": "image-from-disk",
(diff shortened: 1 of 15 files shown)


daisy/workflow/test.workflow
```diff
@@ -51,30 +51,30 @@
...
     },
-    "sub workflow": {
+    "sub-workflow": {
       "subWorkflow": {
         "path": "./test_sub.workflow"
       }
     }
   },
   "dependencies": {
-    "create disks": [],
```
crunk1: I'm not 100% on enforcing step names. I know you have dreams of doing per-step namespacing, but I'm not sold on it yet. Your call.

daisy/workflow/validate.go
```diff
@@ -141,6 +142,9 @@ func (w *Workflow) validateRequiredFields() error {
 	if w.Name == "" {
 		return errors.New("must provide workflow field 'Name'")
 	}
+	if !rfc1035Rgx.MatchString(strings.ToLower(w.Name)) {
```
crunk1: What do you think about having the workflow not care about this? (I'm thinking in terms of decoupling workflow logic from resource stuff) We could have resource name generation sanitize the workflow name. That way, name enforcement is left to resource creation steps' implementation.

daisy/workflow/test.workflow
adjackura: This was mainly for sake of consistency, I figure it's better to enforce it everywhere we think we may use step names in GCE resources in the future. This PR does use subworkflow names as the name for the subworkflow so it was easy to add for everything.

daisy/workflow/validate.go
adjackura: I like the idea of moving this type of logic into the steps themselves. right now we kinda have an interface for validate and run, we can add one for populate, although there would be a lot of shared code.

crunk1: I've wondered about a Populate() method, but it might not be necessary in this case. We would just have to change genName() to sanitize the workflow name.

adjackura: So I tried a sanitize() function and saw some potential issues. It works fine for GCE resources but gets funky for GCS resources. do we just sanitize everything even though GCS can take the special characters? Do we not sanitize GCS paths and just live with potential white space?
Also what do we do about needing a letter as the first character, print out any numbers? ASlow what about the special characters? 


func sanitize(n string) string {
	noSpaces := strings.Replace(n, " ", "-", -1)
	return strings.ToLower(strings.Replace(noSpaces, "_", "-", -1))
}

crunk1: Not completely sure what is the best way for this. You decide. If it turns out to be too much of a hassle, we can talk more about it.

PR #55, opened 2017-05-15
Update README

zmarano:


README.md
```diff
@@ -4,9 +4,13 @@
 
 Tools for Google Compute Engine image management.
 
-## Daisy (WIP)
+## Daisy
 
-Daisy is a GCE workflow tool. Documentation and code still very much a work in progress.
+Daisy is a GCE workflow tool for images.
```
crunk1: Are we still going with the "for images"? It does more than images.

zmarano: Yeah ok good point. Changed.

PR #58, opened 2017-05-17
Moved Step logic into step.go. Moved Workflow.stepDepends to Step.depends.

crunk1:


crunk1:
This is a part of a series of changes coming as part of my validation and resources rewrite.

daisy/workflow/step.go
```diff
@@ -0,0 +1,154 @@
...
+
+type stepImpl interface {
+	run(s *Step) error
+	validate(s *Step) error
+}
+
+// Step is a single daisy workflow step.
+type Step struct {
+	name string
+	w    *Workflow
```
adjackura: Is this the only field in Step that a stepImpl needs? If so would it make more sense to put this into step that way you don't have to pass it into stepImpl.run and stepImpl.validate?

crunk1: No, the actual *Step will be used. Our ctep types, or stepImpls, only use *Workflows currently. But I am making changes to validate methods that will register *Steps as a resource creator, deleter, or user. Without passing *Step, the validate methods have no idea what *Step they belong to.

adjackura: Ok, I'll wait to see once you have all the changes in.

PR #60, opened 2017-05-19
Moved resources out of workflow.go. Added cleanup hooks.

crunk1:


daisy/workflow/resources_test.go
```diff
@@ -0,0 +1,130 @@
+package workflow
+
+import (
+	"github.com/kylelemons/godebug/pretty"
```
adjackura: this should be separated from the other imports

crunk1: Done.

PR #62, opened 2017-05-19
Establish a styleguide for daisy workflows. Open to suggestions.

zmarano:


daisy_workflows/styleguide.md
```diff
@@ -0,0 +1,61 @@
...
+## Workflow file naming
+Use `.wf.json` for workflow names and prefer underscores in file names.
+
+For example: `my_awesome_build.wf.json`.
+
+## Workflow naming
+Use dashed names for workflow names. `Name": "my-awesome-build"`
+
+## Step naming
+Don't use camel case for step naming as this looks too much like the Steps
```
adjackura: maybe make this text similar to above:
+Use dashed names for workflow names: `create-build-disk`, don't use camel case for step naming as this looks too much like the Steps themselves.

daisy_workflows/styleguide.md
```diff
@@ -0,0 +1,61 @@
...
+underscores for resource names. Short hand is acceptable as long as you are
+consistent throughout.
+
+* Disks:
+`"Name": "disk-debian-build"`
+
+* Instances:
+`"Name": "instance-debian-builder"` or `"Name": "inst-debian-builder"`
+
+* Ephermeral Images (images you need during a workflow but don't care about
```
adjackura: Make this ephemeral note above as it applies to both instances and disks depending on the workflow. Mention the ExactName field to be clear

daisy_workflows/styleguide.md
```diff
@@ -0,0 +1,61 @@
...
+        "SizeGb": "10",
+        "Type": "pd-ssd"
+      }
+    ]
+  }
+}
+```
+
+## Resource naming
+Name resources with a prefix denoting the resource type. This will make it much
```
crunk1: I can see what you're suggesting here, but this makes resource names longer -- not the end of the world, but undesirable.

FYI, semi-relevant side note: it is OK to have a disk "foo" AND an instance "foo" simultaneously. There isn't a naming collision.

daisy_workflows/styleguide.md
crunk1: I think the "ephemeral" discussion isn't necessary here. This is starting to delve into Daisy functionality rather than workflow cleanliness.

daisy_workflows/styleguide.md
zmarano: Done

daisy_workflows/styleguide.md
zmarano: True in both cases. However when you have a long workflow and you are trying to figure out the Dependencies- its so much easier to have these things annotated. I can go either way.

daisy_workflows/styleguide.md
zmarano: How's this

PR #63, opened 2017-05-20
Modify how vars are parsed to allow for additional fields

adjackura:
Vars can now be marked as required and have a description while still allowing for the old var style:
"Vars": {
  "some-normal-var": "some-string"
  "some-required-var": {"Required": true, "Description": "this is a description"}
}

Fix logging for subworkflows

### DIFF
--- a/daisy/workflow/workflow.go
+++ b/daisy/workflow/workflow.go
@@ -32,10 +32,11 @@ import (
 	"strings"
 	"time"
 
+	"sync"
+
 	"cloud.google.com/go/storage"
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
 	"google.golang.org/api/option"
-	"sync"
 )
 
 const defaultTimeout = "10m"
@@ -100,7 +101,7 @@ type Workflow struct {
 	logsPath       string
 	outsPath       string
 	username       string
-	gcsLogging     bool
+	gcsLogWriter   io.Writer
 	ComputeClient  *compute.Client `json:"-"`
 	StorageClient  *storage.Client `json:"-"`
 	id             string
@@ -117,6 +118,7 @@ func (w *Workflow) addCleanupHook(hook func() error) {
 
 // Validate runs validation on the workflow.
 func (w *Workflow) Validate() error {
+	w.gcsLogWriter = ioutil.Discard
 	if err := w.validateRequiredFields(); err != nil {
 		close(w.Cancel)
 		return fmt.Errorf("error validating workflow: %v", err)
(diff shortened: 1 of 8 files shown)


daisy/workflow/workflow.go
```diff
@@ -223,20 +238,52 @@ func (w *Workflow) populateStep(step *Step) error {
...
+		// Don't overwrite existing vars (applies to subworkflows).
+		if _, ok := w.vars[k]; ok {
+			continue
+		}
+		var sv string
+		if err := json.Unmarshal(v, &sv); err == nil {
+			w.vars[k] = vars{Value: sv}
+			continue
+		}
+		var vv vars
```
crunk1: Nit: in practice, we've been doing "vs" to denote plural.

daisy/workflow/workflow.go
```diff
@@ -223,20 +238,52 @@ func (w *Workflow) populateStep(step *Step) error {
...
+			}
+			w.vars[k] = vv
+			continue
+		}
+		return fmt.Errorf("cannot unmarshal Var %q, value: %s", k, v)
+	}
+	return nil
+}
+
 func (w *Workflow) populate() error {
```
crunk1: I've been trying to parse the changes here. Is anything going to change with how autovars vs vars get resolved?

daisy/workflow/workflow.go
```diff
@@ -247,17 +294,12 @@ func (w *Workflow) populate() error {
 		"USERNAME":  w.username,
 	}
 
-	vars := map[string]string{}
-	for k, v := range w.Vars {
-		vars[k] = v
-	}
-
 	var replacements []string
```
crunk1: why not map[string]string?

daisy/workflow/workflow.go
adjackura: This isn't plural, well 'vars' is but that may be a bad struct name. sv above was string value, this was vars value

daisy/workflow/workflow.go
adjackura: not sure what that block was here at all, i just use w.vars directly

daisy/workflow/workflow.go
adjackura: No, at least there shouldn't be.

daisy/workflow/workflow.go
crunk1: No, I mean why isn't replacements map[string]string?

crunk1: Oh, nvm.

PR #64, opened 2017-05-22
Added WFDIR autovar.

crunk1:


daisy/workflow/workflow.go
```diff
@@ -291,6 +291,8 @@ func (w *Workflow) populate() error {
 		"DATETIME":  now.Format("20060102150405"),
 		"TIMESTAMP": strconv.FormatInt(now.Unix(), 10),
 		"USERNAME":  w.username,
+		"WFDIR":     w.workflowDir,
+		"CWD":       os.Getwd(),
```
adjackura: This returns an error

crunk1: Done.

PR #66, opened 2017-05-25
Implemented CreateDisks/Images with GCE Go client Disk/Image structs.

crunk1:


daisy/compute/compute.go
```diff
@@ -32,6 +30,10 @@ import (
 type Client struct {
 	hc  *http.Client
 	raw *compute.Service
+
+	OperationsWaitFake func(project, zone, name string) error
+	CreateDiskFake     func(project, zone string, d *compute.Disk) error
+	CreateImageFake    func(project string, i *compute.Image) error
```
adjackura: I don't think I like this, I don't like the idea of exposing testing features publicly. I much prefer the style of using a testClient

daisy/workflow/common.go
```diff
@@ -68,6 +71,15 @@ func filter(ss []string, s string) []string {
 	return result
 }
 
+func getGCSAPIPath(p string) (string, error) {
+	var b, o string
+	var e error
+	if b, o, e = splitGCSPath(p); e != nil {
+		return "", e
+	}
```
adjackura: b, o, err := splitGCSPath(p)
if err != nil {
  return "", err
}

daisy/workflow/step_create_disks.go
```diff
@@ -57,17 +53,26 @@ func (c *CreateDisks) validate(s *Step) error {
...
-			return fmt.Errorf("cannot create disk: SizeGB and SourceImage not set: %s", cd.SourceImage)
+		// Prepare field values: Disk.Name, Disk.Type, name, Project, Zone
+		cd.name = cd.Name
+		if !cd.ExactName {
+			cd.Name = w.genName(cd.name)
 		}
+		cd.Project = stringOr(cd.Project, w.Project)
+		cd.Zone = stringOr(cd.Zone, w.Zone)
+		if cd.Type != "" && !strings.Contains(cd.Type, "/") {
+			cd.Type = fmt.Sprintf("zones/%s/diskTypes/%s", cd.Zone, cd.Type)
```
adjackura: I think we should note that we do this in the CreateDisk documentation somehow

daisy/workflow/common.go

daisy/workflow/step_create_disks.go
crunk1: Good call.

PR #68, opened 2017-05-26
Validate missing step in Dependencies, parse SizeGb from string

adjackura:


daisy/workflow/step_create_disks.go
```diff
@@ -30,6 +31,9 @@ type CreateDisks []*CreateDisk
 type CreateDisk struct {
 	compute.Disk
 
+	// Size of this disk.
+	SizeGb string `json:",omitempty"`
```
crunk1: Wait, so can we not do int64?

PR #88, opened 2017-06-20
Added populate to steps. Pulled populate logic out of validate methods.

crunk1:


daisy/workflow/step_attach_disks.go
```diff
@@ -24,7 +24,14 @@ type AttachDisk struct {
 	Disk, Instance string
 }
 
+func (a *AttachDisks) populate(s *Step) error {
+	return nil
+}
+
 func (a *AttachDisks) validate(s *Step) error {
+	if err := a.populate(s); err != nil {
```
adjackura: Why is this in step.validate? step.populate should be called from w.populateStep (theres code there that probably should be replaced with this anyway)

crunk1: Removed.

PR #92, opened 2017-06-27
Use a custom unmarshaller for vars, fix IncludeWorkflow autovars

adjackura:
Autovars can now be used in Vars


daisy/workflow/workflow.go
```diff
@@ -243,6 +253,7 @@ func (w *Workflow) populateStep(step *Step) error {
 	return nil
 }
 
+/*
```
zmarano: Take this out?

PR #96, opened 2017-06-28
Ctx should not be in the Workflow struct

adjackura:


daisy/daisy_test.go
```diff
@@ -17,6 +17,7 @@ package main
 import (
 	"reflect"
 	"testing"
+	"context"
```
crunk1: alphabetize

PR #97, opened 2017-06-28
Fix IncludeWorkflow population

adjackura:


daisy/workflow/workflow.go
```diff
@@ -336,7 +336,6 @@ func (w *Workflow) populate(ctx context.Context) error {
 func (w *Workflow) NewIncludedWorkflow() *Workflow {
 	iw := New()
 	iw.Cancel = w.Cancel
-	iw.parent = w
```
crunk1: Should we just delete the parent field?

adjackura: Doesn't SubWorkflow use it?

crunk1: It doesn't NEED it. However, your logger uses it to build "grandparent.parent.child" log prefixes.

adjackura: We can decide what we want to do with it later

PR #98, opened 2017-06-29
Revamp imports to be two distinct workflows

zmarano:
One for importing and one for translation. Create an image for the import workflow instead of a disk. Fix a bunch of bugs. Also, fix annoying log output for gcs log paths in Daisy output.


daisy_workflows/image_import/debian/translate_debian_8.wf.json
```diff
@@ -0,0 +1,54 @@
...
+        }
+      },
+      "Timeout": "60m"
+    },
+    "create-image": {
+      "CreateImages": [
+        {
+          "Name": "${image_name}",
+          "SourceDisk": "disk-deb-8-import",
+          "Project": "${PROJECT}",
```
adjackura: This is redundant, did we want to allow users to specify this?

zmarano: SGTM- I don't think we need to worry about specifying it- unless someone asks for that.

PR #100, opened 2017-06-29
Actually populate WaitForInstancesSignal

adjackura:


daisy/workflow/step_wait_for_instances_signal.go
```diff
@@ -92,49 +92,49 @@ func waitForSerialOutput(w *Workflow, name string, port int64, success, failure
...
 			ws.Interval = defaultInterval
 		}
-		interval, err := time.ParseDuration(ws.Interval)
+		var err error
+		ws.interval, err = time.ParseDuration(ws.Interval)
 		if err != nil {
 			return err
 		}
-		ws.interval = interval
+		(*w)[i] = ws
```
crunk1: Why not convert WaitForInstancesSignal []InstancesSignal to []*InstancesSignal? That's what I've been doing in the other cases.

PR #101, opened 2017-06-29
Skip blank sources in IncludeWorkflow

adjackura:


daisy/workflow/step_includeworkflow.go
```diff
@@ -60,6 +60,9 @@ func (i *IncludeWorkflow) populate(ctx context.Context, s *Step) error {
 
 	// Copy Sources up to parent resolving relative paths as we go.
 	for k, v := range i.w.Sources {
+		if v == "" {
```
crunk1: Should we remove the key-value pair entirely? Same for workflows in general. We have a "sourceExists" method that checks if a key is there if I remember correctly.

PR #102, opened 2017-06-30
Create a GCS bucket if no GCSPath is set, guess at project and zone

adjackura:


daisy/workflow/workflow.go
```diff
@@ -64,6 +65,28 @@ func (l *gcsLogger) Write(b []byte) (int, error) {
 	return n, err
 }
 
+func daisyBkt(ctx context.Context, client *storage.Client, project string) (string, error) {
+	dBkt := project + "-daisy-bkt"
+	it := client.Buckets(ctx, project)
+	for {
```
crunk1: for bucketAttrs, err := it.Next(); err != iterator.Done; bucketAttrs, err = it.Next() {
}

PR #106, opened 2017-07-14
Add userguide for imports.

zmarano:


daisy_workflows/import_userguide.md
```diff
@@ -0,0 +1,275 @@
...
+```
+gcloud compute ssh daisy-control
+```
+
+### Step 3: Convert Virtual Disk to Compute Engine Image
+
+Now inside the VM,run the workflow titled import_image.wf.json to convert the virtual disk file in GCS to a Compute Engine Image.
+
+```
+daisy -variables source_disk_file=YOUR_VIRTUAL_DISK_FILE \\
```
adjackura: do these line continuations work with daisy? seems like parsing the args would fail as these variables need to be one arg

zmarano: Fixed (docs-md conversion mishap)

PR #108, opened 2017-07-14
Fix WaitForInstancesSignal again

adjackura:
STOPPED is a status too
Fix IncluedWorkflow issues caused by previous PR

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -208,7 +208,7 @@ func (c *client) InstanceStatus(project, zone, name string) (string, error) {
 	return inst.Status, nil
 }
 
-// InstanceStopped checks if a GCE instance is in a 'TERMINATED' state.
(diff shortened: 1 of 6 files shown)


daisy/compute/compute.go
```diff
@@ -217,7 +217,7 @@ func (c *client) InstanceStopped(project, zone, name string) (bool, error) {
 	switch status {
 	case "PROVISIONING", "RUNNING", "STAGING", "STOPPING":
```
crunk1: What would happen if we return true when status == "STOPPING"? Can we get a performance gain out of that (by unblocking Daisy sooner)?

adjackura: Stopping should actually never be hit as I think that only would happen if you call a stop. Shutdown from inside the VM would just go from Running to Stopped (I think)

PR #109, opened 2017-07-15
Cleanup of linux image builds.

zmarano:


daisy_workflows/image_build/debian/build.py
```diff
@@ -17,25 +17,13 @@
 
 Parameters (retrieved from instance metadata):
 
-bootstrap-vz-manifest: The version of bootstrap-vz to retrieve and use.
```
adjackura: I've still been using dashes for metadata values, not underscores

zmarano: Its personal preference- but to keep consistency with daisy variables, it makes it easier (and less error prone) to keep them consistent.

zmarano:
Yeah I agree- that's going to be a second cleanup PR. Actually the EL builds still don't work right here.

PR #111, opened 2017-07-17
Buffer GCS logs, fix issues with caching zone and machinetype

adjackura:


daisy/workflow/test_common_test.go
```diff
@@ -102,6 +105,25 @@ func newTestGCEClient() (*compute.TestClient, error) {
 		}
 	}))
 
+	c.GetProjectFn = func(project string) (*compute.Project, error) {
```
crunk1: Cool, I like this. I thought you just didn't like my shorthand p and z variables :).

PR #113, opened 2017-07-19
Extending DeleteResources to also validate instances' autodeleted disks.

crunk1:


daisy/workflow/step_delete_resources.go
```diff
@@ -16,6 +16,7 @@ package workflow
 
 import (
 	"context"
+	"google.golang.org/api/compute/v1"
```
adjackura: This import should be after the builtins (with a space)

crunk1: Done.

PR #116, opened 2017-07-26
Error struct

crunk1:


daisy/workflow/err.go
```diff
@@ -0,0 +1,36 @@
+package workflow
+
+import "fmt"
+
+var (
+	errCodeToName = []string{""}
+	errNameToCode = map[string]uint{"": 0}
+)
+
+type err struct {
```
adjackura: Maybe just change `code uint` to `errType string`

You can skip the whole string indexing and just use the string thats given
Then you can have constants for the types:
const (
  validateErr = "Validate"
  runErr = "Run"
  ...
)

Then maybe change the errf function to:
func newErrf(errType, format string, a ...interface{}) *err {
  return &err{msg: fmt.Sprintf(format, a...), errType: errType}
}

crunk1:
A `cast() error` method has been added to both Error and Errors due to the following:

https://golang.org/doc/faq#nil_error
"An interface value is nil only if the inner value and type are both unset, (nil, nil)."

PR #118, opened 2017-07-31
Move daisy library from workflow to daisy

adjackura:


daisy/daisy/daisy.go
```diff
@@ -13,7 +13,7 @@
 //  limitations under the License.
 
 // Daisy is a GCE workflow tool.
-package main
```
crunk1: Should this package change? Shouldn't it still be main?

PR #120, opened 2017-08-03
Make the buffered logger concurrency safe

adjackura:


daisy/workflow.go
```diff
@@ -66,6 +66,23 @@ func (l *gcsLogger) Write(b []byte) (int, error) {
 	return n, err
 }
 
+type bufferedLogger struct {
```
crunk1: Maybe name it syncedWriter? Or something like that?

PR #123, opened 2017-08-04
Allow workflow variables to be set using flags to daisy

adjackura:
Any arguments passed to daisy in the format -var:<text> will be parsed as flags and then passed to workflows as Vars.

For example to pass the var 'myVar' with value 'myValue' you use the flag -var:myVar=myValue

### DIFF
--- a/daisy/daisy/daisy.go
+++ b/daisy/daisy/daisy.go
@@ -121,7 +121,7 @@ func parseWorkflow(ctx context.Context, path string, varMap map[string]string, p
 
 func addFlags(args []string) {
 	for _, arg := range args {
-		if arg[0] != '-' || len(arg) <= 1 {
+		if len(arg) <= 1 || arg[0] != '-' {
 			continue
 		}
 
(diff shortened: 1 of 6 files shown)


daisy/daisy/daisy.go
```diff
@@ -108,15 +119,49 @@ func parseWorkflow(ctx context.Context, path string, varMap map[string]string, p
...
+			continue
+		}
+
+		numMinuses := 1
+		if arg[1] == '-' {
+			numMinuses = 2
+		}
+
+		name := arg[numMinuses:]
+		if len(name) == 0 || name[0] == '-' || name[0] == '=' || !strings.HasPrefix(name, varFlagPrefix) {
```
crunk1: You should only need 

if !strings.HasPrefix(name, varFlagPrefix) {
}

daisy/daisy/daisy.go
```diff
@@ -108,15 +119,49 @@ func parseWorkflow(ctx context.Context, path string, varMap map[string]string, p
...
+			continue
+		}
+
+		for i, r := range name {
```
crunk1: Why not strings.SplitN(name, "=", 1)?

Issue #132, opened 2017-08-23
Required variables are not validated


zmarano:
Meaning, they are not actually required and you can run workflows without specifying them.

PR #134, opened 2017-08-24
Fix Issue #132. Also some commenting and clean up.

crunk1:


daisy/workflow.go
```diff
@@ -100,22 +100,25 @@ func daisyBkt(ctx context.Context, client *storage.Client, project string) (stri
 	return dBkt, nil
 }
 
-type vars struct {
+// _var is a type with a flexible JSON representation. A _var can be represented
+// by either a string, or by this struct definition. A _var that is represented
+// by a string will unmarshal into the struct: {Value: <string>, Required: false, Description: ""}.
+type _var struct {
```
adjackura: Please no underscores, maybe daisyVar or workflowVar or wVar

crunk1: Done. Went with wVar.

PR #138, opened 2017-08-30
Setting up Prow configuration and prowjob container images. Also, some basic functional tests (not currently being run).

crunk1:
Copied a lot of Prow code/config over from https://github.com/kubernetes/test-infra/tree/master/prow.

This is not done yet, but this is the back bone of our Prow set up.

Added some logging. See https://pastebin.com/8fdw89iT for example logs.


daisy/prow_images/build/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM debian:stretch
```
adjackura: Whats this container used for?

daisy/prow_images/functional_tests/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM debian:stretch
```
adjackura: Maybe see of you can use a smaller image, maybe 'python-2.7-alpine', you'll need to update the package installs but it should have everything.

daisy/prow_images/functional_tests/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+ENV GOOGLE_APPLICATION_CREDENTIALS /etc/service-account/service-account.json
+ENV GCLOUD_VERSION 168.0.0
+ENV PATH "/google-cloud-sdk/bin:${PATH}"
+
+RUN apt-get update
+RUN apt-get upgrade -y
+RUN apt-get install -y \
+    git \
+    golang \
```
adjackura: Don't use the package, it's crazy outdated. Pull the tar and install (assuming a golang container wont work, installing golang might be easier that installing python)

daisy/prow_images/functional_tests/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    python \
+    python-pip \
+    wget
+
+RUN pip install google-auth google-cloud-storage
+
+RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_VERSION-linux-x86_64.tar.gz && \
```
adjackura: Why do you download gcloud? you don't use it at all as you use the google-auth google-cloud-storage libs

PR #139, opened 2017-08-31
Only lock when accessing baseResourceMap

adjackura:


daisy/resources.go
```diff
@@ -74,20 +74,18 @@ func (rm *baseResourceMap) cleanup() {
 }
 
 func (rm *baseResourceMap) delete(name string) error {
```
crunk1: We could potentially have a race condition with multiple things trying to delete the same resource. I don't know if we'll ever run into this, but this function is now thread-unsafe.

crunk1: Specifically, two threads can execute line 84 simultaneously.

PR #145, opened 2017-09-04
translate_el.sh: use guestfs to mount disk

helen-fornazier:
Use guestmount to mount the disk. This tool should mount the disk
independently if LVM is being used or if /boot is in a separate
partition


helen-fornazier:
hold this PR for now, it seems that grub2-mkconfig generates a different grub config if the disk is mounted with guestmount, I am verifying this.

helen-fornazier:
Using virt-customize or a library as python-guestfs seems to work better, I am closing this PR

PR #146, opened 2017-09-06
Apply GCS ACLs on a recursive copy, add populate and validate for CopyGCSObjects

adjackura:
Populate function ensures Role is uppercase
Validate function checks:
- read access to source 
- write access to destination
- ACLRule.Entity is not empty
- ACLRule.Entity can actually be used
- ACLRule.Role matches a known value

Reorganize GCS specific code.

### DIFF
--- a/daisy/step_copy_gcs_objects.go
+++ b/daisy/step_copy_gcs_objects.go
@@ -31,10 +31,17 @@ type CopyGCSObjects []CopyGCSObject
 // CopyGCSFile copies a GCS file from Source to Destination.
 type CopyGCSObject struct {
 	Source, Destination string
-	ACLRules            []storage.ACLRule
+	ACLRules            []*storage.ACLRule
 }
 
-func (c *CopyGCSObjects) populate(ctx context.Context, s *Step) error { return nil }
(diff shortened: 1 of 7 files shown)


daisy/step_copy_gcs_objects.go
```diff
@@ -31,14 +31,88 @@ type CopyGCSObjects []CopyGCSObject
...
+		// Check if source bucket exists and is readable.
+		if !readableBkts.contains(sBkt) {
+			if _, err := s.w.StorageClient.Bucket(sBkt).Attrs(ctx); err != nil {
+				return fmt.Errorf("error reading bucket %q: %v", sBkt, err)
+			}
+			readableBkts.add(sBkt)
+		}
 
-func recursiveGCS(ctx context.Context, w *Workflow, sBkt, sPrefix, dBkt, dPrefix string) error {
+		// Check if destination bucket exists and is readable.
```
crunk1: Why does destination bucket need to be readable? Why not just a writable check?

adjackura: The errors you get from the read check are much more readable, there's a nice difference between 'access denied' and 'bucket does not exist' and with an access denied you get a lot of useful info about the user

PR #148, opened 2017-09-07
translate_el: use python-guestfs

helen-fornazier:
This tool should mount the disk independently if LVM is being used or if
/boot is in a separate partition and perform customizations in the VM
image.

Also, I tested with guestmount tool, but it seems that running
grub2-mkconfig through chroot makes the roofs to point to /dev/fuse
instead of the disk, so a full virtualisation should be used. One of the
options is to use virt-customize, but I find using python-guest to be
better, since it is not bash, we can handle errors better and have more
interoperability.


zmarano:
Thanks @helen-fornazier! Generally I think this is a great idea! I'll take a closer look later today and get back with any comments. I actually think we can use this to make a common python library and merge some of the other distro translation flows to use common logic.

daisy_workflows/image_import/enterprise_linux/translate_el.sh
```diff
@@ -15,131 +15,13 @@
 
 set -x
 
+# Download translate_el.py
```
zmarano: Technically, we could just do all of this in the python script and nix the bash script entirely- in that case we would just set the startup script to the python script. Otherwise, I think this look good.

helen-fornazier: I use the bash script to install python-guestfs and python-pycurl before running the python script
Do you have any suggestion in how we can do this without the bash script? So we can get rid of it?

zmarano: Oh. I see the problem. Yeah ok, we'll keep the bash script because we need the python dependencies installed. Never mind my comment :)

helen-fornazier:
Something is weird, your branch `master` was merged into my branch `translate_el`, but my branch `translate_el` wasn't merged into your `master` branch

zmarano:
Maybe just git pull the master back into your branch and merge your commit back into it.

zmarano:
We'll just squash and merge this- you may have to reset your git branch afterwards but that should be fine.

PR #157, opened 2017-09-18
Modify image tutorial.

jeffmendoza:
Let me know what you think of the directory structure. Should this be under `daisy_workflows`? It is good to separate educational material from full working actually-used examples. Looks like `daisy_workflows/examples` should me morphed into tutorials.

### DIFF
--- a/daisy_tutorials/modify_image/README.md
+++ b/daisy_tutorials/modify_image/README.md
@@ -1,0 +1,197 @@
+# Modifying a base image using Daisy
+
+Daisy workflows [have many
+capabilities,](https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/daisy#workflow-config-overview)
+and it is hard to know where to start. This tutorial will demonstrate
+writing a very simple workflow to modify a GCE base image.
+
+The overall steps of the workflow will be:
+
+1. Create a new instance
+2. Run a provisioning script to modify the instance
+3. Create a new disk image from the instance
+4. Delete the instance
+
+## Provisioning the instance
+
+What you use to provision the instance is open and not dependant on
+Daisy. In this simple case, we are using a simple shell script to
+install Emacs. Because we are creating a CentOS image, we simply call
+`yum install`. The provisioning script is named `emacs_install.sh`.
+
+```shell
+#!/bin/bash
+
+yum -y install emacs-nox
+shutdown -h now
+```
+
+The script shuts down the instance after it is done. We will use this
+later in the Daisy workflow to know when provisioning is done.
+
+## Start writing the workflow
+
+First we will start with a Name:
+
+```
+{
+    "Name": "emacs-image",
+```
+
+Then, list the sources. We will need the provisioning script above.
+
+```
+    "Sources": {
+        "emacs_install.sh": "./emacs_install.sh"
+    },
+```
+
+## Create and Provision the Instance
+
+Now we will start writing steps in the workflow. We will create the
+disk first, as we need to name it and perform operations on it
+later. We give the disk a name, specify the base image, and specify
+the SSD type for speed. The name of the step `create-disks` is
+arbitrary.
+
+```
+    "Steps": {
+        "create-disks": {
+            "CreateDisks": [
+                {
+                    "Name": "instance-disk",
+                    "SourceImage": "projects/centos-cloud/global/images/family/centos-7",
+                    "Type": "pd-ssd"
+                }
+            ]
+        },
+```
+
+Next create the instance using the above disk, and specify the startup
+script that we wrote to provision the instance.
+
+```
+        "create-inst-install": {
+            "CreateInstances": [
+                {
+                    "Name": "inst-install",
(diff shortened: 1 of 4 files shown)


daisy_tutorials/modify_image/README.md
```diff
@@ -0,0 +1,197 @@
...
+Next create the instance using the above disk, and specify the startup
+script that we wrote to provision the instance.
+
+```
+        "create-inst-install": {
+            "CreateInstances": [
+                {
+                    "Name": "inst-install",
+                    "Disks": [{"Source": "instance-disk"}],
+                    "MachineType": "n1-standard-4",
```
zmarano: Probably go with n1-standard-1 here (only because 4 CPU's to install a package is a little bit much and we like to err on the side of using less resources for the example cases).

adjackura:
One thing to note about the daisy_workflows directory is that it gets included in the daisy container (as /workflows/), unsure if it would be useful to include the examples/tutorials in the container. If we did want to include them we could modify the container build to include the directory, or use daisy_workflows/tutorials instead of daisy_tutorials/

daisy_tutorials/modify_image/emacs_image.wf.json
```diff
@@ -0,0 +1,57 @@
+{
+    "Name": "emacs-image",
+    "Sources": {
+        "emacs_install.sh": "./emacs_install.sh"
+    },
+    "Steps": {
+        "create-disks": {
+            "CreateDisks": [
+                {
+                    "Name": "instance-disk",
```
zmarano: To be consistent with the styleguide, name the disk: "disk-install"

daisy_tutorials/modify_image/README.md
```diff
@@ -0,0 +1,197 @@
...
+`yum install`. The provisioning script is named `emacs_install.sh`.
+
+```shell
+#!/bin/bash
+
+yum -y install emacs-nox
+shutdown -h now
+```
+
+The script shuts down the instance after it is done. We will use this
```
zmarano: This is great as a basic example. I wonder if we should do a more complex example with a few different steps in the script- using serial port output and success/failure cases to show how to fail or succeed on different parts of your script. Obviously- in a more advanced example tutorial.

daisy_tutorials/modify_image/README.md
jeffmendoza: Fixed

daisy_tutorials/modify_image/emacs_image.wf.json
jeffmendoza: Fixed.

daisy_tutorials/modify_image/README.md
jeffmendoza: Thanks, yes sounds like a good second tutorial.

daisy_tutorials/modify_image/README.md
```diff
@@ -0,0 +1,197 @@
...
+The overall steps of the workflow will be:
+
+1. Create a new instance
+2. Run a provisioning script to modify the instance
+3. Create a new disk image from the instance
+4. Delete the instance
+
+## Provisioning the instance
+
+What you use to provision the instance is open and not dependant on
```
crunk1: s/dependant/dependent

daisy_tutorials/modify_image/README.md
```diff
@@ -0,0 +1,197 @@
...
+                    "ExactName": true
+                }
+            ]
+        },
+```
+
+Note the `NoCleanup`, Daisy automatically cleans up any resources
+created during a run. We don't want Daisy to delete our image though!
+Specifying `NoCleanup` will instruct Daisy to leave the
+image. `ExatcName` tells Daisy to use the `Name` as-is. Normally Daisy
```
crunk1: s/ExcatName/ExactName

daisy_tutorials/modify_image/README.md
```diff
@@ -0,0 +1,197 @@
...
+adds a suffix to resource names to avoid name colissions.
```
crunk1: s/colissions/collisions

daisy_tutorials/modify_image/README.md
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+    },
+```
+
+And that completes the steps for customizing the image!
+
+## Ordering steps.
+
+Daisy will try to run steps in parallel for efficiency, so we need to
```
crunk1: Nice, explanation here!

daisy_tutorials/modify_image/README.md
jeffmendoza: done

daisy_tutorials/modify_image/README.md
jeffmendoza: done

daisy_tutorials/modify_image/README.md
jeffmendoza: done

PR #167, opened 2017-09-22
Allow foo.com:bar project naming

adjackura:


daisy/step_create_instances_test.go
```diff
@@ -542,6 +542,7 @@ func TestCreateInstanceValidateMachineType(t *testing.T) {
 		shouldErr bool
 	}{
 		{"good case", fmt.Sprintf("projects/%s/zones/%s/machineTypes/%s", testProject, testZone, testMachineType), false},
+		{"good case2", fmt.Sprintf("projects/%s/zones/%s/machineTypes/%s", testProject, testZone, testMachineType), false},
```
crunk1: How is this different than "good case"?

daisy/validate.go
```diff
@@ -24,8 +24,9 @@ import (
 )
 
 var (
-	rfc1035    = "[a-z]([-a-z0-9]*[a-z0-9])?"
-	rfc1035Rgx = regexp.MustCompile(fmt.Sprintf("^%s$", rfc1035))
+	rfc1035     = "[a-z]([-a-z0-9]*[a-z0-9])?"
+	rfc1035Proj = "[a-z]([-.:a-z0-9]*[a-z0-9])?"
```
crunk1: rfc1035Proj is technically a misnomer. ":" is not RFC 1035 compliant.

PR #170, opened 2017-09-23
translate_ubuntu: use python-guestfs

helen-fornazier:
This tool should mount the disk independently if LVM is being used or if
/boot is in a separate partition and perform customizations in the VM
image.

### DIFF
--- a/daisy_workflows/image_import/ubuntu/translate_ubuntu.sh
+++ b/daisy_workflows/image_import/ubuntu/translate_ubuntu.sh
@@ -15,108 +15,36 @@
 
 set -x
 
+# Download translate_el.py
 URL="http://metadata/computeMetadata/v1/instance/attributes"
-UBU_RELEASE="$(curl -f -H Metadata-Flavor:Google ${URL}/ubuntu_release)"
-INSTALL_GCE="$(curl -f -H Metadata-Flavor:Google ${URL}/install_gce_packages)"
-MNT="/mnt/imported_disk"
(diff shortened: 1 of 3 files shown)


daisy_workflows/image_import/ubuntu/translate_ubuntu.sh
```diff
@@ -15,108 +15,36 @@
...
-       failsafe:
-         primary: http://ports.ubuntu.com/ubuntu-ports
-         security: http://ports.ubuntu.com/ubuntu-ports
+GCS_SOURCE_PATH="$(curl -f -H Metadata-Flavor:Google ${URL}/daisy-sources-path)"
+gsutil cp "${GCS_SOURCE_PATH}/translate_ubuntu.py" /tmp/translate_ubuntu.py
+
+# Install dependencies
+DEBIAN_FRONTEND=noninteractive apt-get -y install python-guestfs python-pycurl tinyproxy
+
+# cloud-init command uses GCE API, as libguestfs executes inside qemu, it uses a
```
zmarano: How does this work? GCE doesn't support nested qemu.

helen-fornazier: Qemu can be used to run an appliance instead of a full VM, there is a nice explanation here https://rwmj.wordpress.com/2011/07/07/how-does-libguestfs-live-work/
And the official docs http://libguestfs.org/guestfs-internals.1.html

zmarano: Ahh ok. Interesting.

PR #173, opened 2017-09-26
Add additional checks for resources existing in GCE

adjackura:
Checks for dependency existence of any non workflow created resources
Checks for naming collisions of workflow created resources
Lookups are cached to limit the amount of API calls needed.

### DIFF
--- a/daisy/disk.go
+++ b/daisy/disk.go
@@ -168,6 +168,8 @@ var diskCache struct {
 	mu     sync.Mutex
 }
 
+// diskExists should only be used during validation for existing GCE disks
+// and should not be relied or populated for daisy created resources.
 func diskExists(client compute.Client, project, zone, disk string) (bool, error) {
 	diskCache.mu.Lock()
 	defer diskCache.mu.Unlock()
--- a/daisy/disk.go
+++ b/daisy/disk.go
@@ -17,6 +17,9 @@ package daisy
 import (
 	"fmt"
 	"regexp"
+	"sync"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
 )
 
 var (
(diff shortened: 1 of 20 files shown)


daisy/disk.go
```diff
@@ -159,3 +162,31 @@ func (dr *diskRegistry) registerAllDetachments(iName string, s *Step) error {
 	}
 	return errs.cast()
 }
+
+var diskCache struct {
```
crunk1: Comment about diskCache and diskExists only being used at validation time?

daisy/image.go
```diff
@@ -42,3 +48,40 @@ func (ir *imageRegistry) deleteFn(res *resource) error {
...
 }
+
+var imagesCache struct {
+	exists map[string][]string
+	mu     sync.Mutex
+}
+
+func imageExists(client compute.Client, project, family, name string) (bool, error) {
+	if family != "" {
+		if _, err := client.GetImageFromFamily(project, family); err != nil {
```
crunk1: Add another cache map for families?

daisy/image.go
```diff
@@ -42,3 +48,40 @@ func (ir *imageRegistry) deleteFn(res *resource) error {
 	}
 	return nil
 }
+
+var imagesCache struct {
```
crunk1: Comment about imagesCache and imageExists only being used during validation?

daisy/resources.go
```diff
@@ -89,6 +90,27 @@ func (r *baseResourceRegistry) get(name string) (*resource, bool) {
 	return res, ok
 }
 
+func resourceExists(client compute.Client, url string) (bool, error) {
+	if !strings.HasPrefix(url, "projects/") {
+		return false, fmt.Errorf("partial GCE resource URL %q needs leading \"projects/PROJECT/\"", url)
+	}
+	switch {
```
crunk1: Not a fan of the switch case, but we can prob remove this once we move to resources/zonal resources.

daisy/resources_test.go
```diff
@@ -112,7 +113,7 @@ func TestResourceMapDelete(t *testing.T) {
 }
 
 func TestResourceMapConcurrency(t *testing.T) {
-	rm := baseResourceRegistry{}
+	rm := baseResourceRegistry{w: testWorkflow()}
```
crunk1: s/rm/rr

crunk1: Here and elsewhere.

PR #178, opened 2017-09-28
translate_rhel: use Debian as the translator host

helen-fornazier:
Use the same platform for all VMs to do the translation, this will
ensure we install the good python dependencies and run the same
libguestfs, only the commands inside libguestfs will change.
This patch fixes fails in RHEL where the translate_el.sh script uses apt
to install the dependencies for translate_el.py script.


helen-fornazier:
I see, hold this PR then, I had a problem in my RHEL image, I'll test this better.

zmarano:
I may have a better solution soon but at the moment this was the only way.

helen-fornazier:
I'll send another patch with another approach. I am closing this PR for now.

PR #179, opened 2017-09-28
translate_el: Fix translation of RHEL licensed

helen-fornazier:
RHEL licensed doesn't use Debian as host for the translation to allow
the system to download the google-rhui-client-rhel package.
Fix the bash script to install the dependencies using yum if available
and also fix the package installation process of the rhui.

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate_el.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate_el.py
@@ -136,14 +136,11 @@ def translate():
 
   if rhel_license == 'true':
     if 'Red Hat' in g.cat('/etc/redhat-release'):
-      g.command([
-          'yum', 'install', '-y', '--downloadonly', '--downloaddir=/tmp',
-          'google-rhui-client-rhel%s' % el_release])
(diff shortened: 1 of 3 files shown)


zmarano:
Helen, as of now (due to unrelated problems) the rhui client package is available in the Google Cloud repo for RHEL 6 and RHEL 7. Therefore we should be able to install it in the chroot after adding the cloud repo files.

helen-fornazier:
Yes, this patch do exactly as before, i.e, download the rhui in the host, transfer it to the image and with libguestfs (instead of directly chroot) and install the package in the imported image

zmarano:
Ok that works- what I mean though is that we can standardize on the Debian box to do the work because the rhui client package is now available in the chroot environment.

helen-fornazier:
ah, right, now it is available in the [google-cloud](https://packages.cloud.google.com/yum/repos/google-cloud-compute-el7-x86_64) repository right? i can see it in the list. Great, let me try to do that then, so we can use the same box in all the imports.

zmarano:
Yup, cool that should make things a bit easier.

PR #191, opened 2017-10-10
Add OverWrite field to CreateImage, follow NextPageToken for List* functions

adjackura:


daisy/compute/compute.go
```diff
@@ -277,12 +277,23 @@ func (c *client) GetMachineType(project, zone, machineType string) (*compute.Mac
 }
 
 // ListMachineTypes gets a list of GCE MachineTypes.
-func (c *client) ListMachineTypes(project, zone string) (*compute.MachineTypeList, error) {
-	mt, err := c.raw.MachineTypes.List(project, zone).Do()
-	if shouldRetryWithWait(c.hc.Transport, err, 2) {
-		return c.raw.MachineTypes.List(project, zone).Do()
+func (c *client) ListMachineTypes(project, zone string) ([]*compute.MachineType, error) {
```
crunk1: I'm gonna be annoying and say "Can you refactor some of this logic into another method? There is a lot of copied code."

Maybe some method that does a generic paginated List that returns an untyped slice that you can then cast?

daisy/compute/compute.go
```diff
@@ -349,12 +382,23 @@ func (c *client) GetDisk(project, zone, name string) (*compute.Disk, error) {
 }
 
 // ListDisks gets a list of GCE Disks.
-func (c *client) ListDisks(project, zone string) (*compute.DiskList, error) {
-	d, err := c.raw.Disks.List(project, zone).Do()
-	if shouldRetryWithWait(c.hc.Transport, err, 2) {
-		return c.raw.Disks.List(project, zone).Do()
+func (c *client) ListDisks(project, zone string) ([]*compute.Disk, error) {
+	var dl []*compute.Disk
```
crunk1: Nit: other methods are using a different naming standard here. Wanna make them match?

var ds []*compute.Disk
...
for dl, err := ...

daisy/compute/compute.go
```diff
@@ -394,12 +449,23 @@ func (c *client) GetNetwork(project, name string) (*compute.Network, error) {
 }
 
 // ListNetworks gets a list of GCE Networks.
-func (c *client) ListNetworks(project string) (*compute.NetworkList, error) {
-	n, err := c.raw.Networks.List(project).Do()
-	if shouldRetryWithWait(c.hc.Transport, err, 2) {
-		return c.raw.Networks.List(project).Do()
+func (c *client) ListNetworks(project string) ([]*compute.Network, error) {
+	var nl []*compute.Network
```
crunk1: Ditto about the naming standard nit.

daisy/resources_test.go
```diff
@@ -189,7 +189,7 @@ func TestResourceMapRegisterCreation(t *testing.T) {
 	}
 
 	// Test duplication create.
-	if err := rr.registerCreation("foo", r, nil); err == nil {
+	if err := rr.registerCreation("foo", r, nil, false); err == nil {
 		t.Error("should have returned an error, but didn't")
 	}
```
crunk1: Mind adding tests exercising overwrite?

crunk1: Also, do we have good coverage for registerExisting?

daisy/step_create_images.go
```diff
@@ -143,17 +147,24 @@ func (c *CreateImages) run(ctx context.Context, s *Step) error {
 		wg.Add(1)
 		go func(ci *CreateImage) {
 			defer wg.Done()
-
-			project := strOr(ci.Project, w.Project)
```
crunk1: I assume this was some leftover thing before populate(). If so, thanks for the cleanup.

daisy/step_create_images.go
```diff
@@ -143,17 +147,24 @@ func (c *CreateImages) run(ctx context.Context, s *Step) error {
...
-
 			// Get source disk link if SourceDisk is a daisy reference to a disk.
 			if d, ok := disks[w].get(ci.SourceDisk); ok {
 				ci.SourceDisk = d.link
 			}
 
+			// Delete existing if OverWrite is true.
+			if ci.OverWrite {
```
crunk1: Do we need to do this for disks and instances?

We could run into issues here when the existing disk is attached... Hmm. For now, we could let the workflow fail, I suppose.

adjackura: I specifically only implemented this for images as that could get more complicated

daisy/compute/compute.go
adjackura: I got this working but really hated it, I think being clever with reflect is unidiomatic:
https://blog.golang.org/laws-of-reflection

'It's a powerful tool that should be used with care and avoided unless strictly necessary.'

PR #201, opened 2017-10-20
E2E tests for Linux. Ubuntu 14.04 and 16.04, CentOS 6 and 7.

zmarano:


crunk1:
Just FYI, the /go test will not kick these tests off for you. /go test only runs unit tests. We will have to wait until the periodic E2E run before we know if these tests are good.

PR #212, opened 2017-10-30
create docs directory, add a bunch of docs.

ericdand:
This PR takes all of the config file spec out of the main README.md and drops it into its own page in `docs/`. It also adds three additional pages of documentation.

Before this PR is accepted, we need to do three things:

1. Get a "GitHub Pages" site going for this project,
2. somehow test that the relative linking I do in these docs actually works, and
3. settle on a naming scheme for Daisy documentation. (e.g. should they all begin with "Daisy: "?)

More on the relative linking: I link from `/docs/*.md` pages to the examples directory using `../daisy_workflows/...`, as well as to other Markdown files. Relative links to `../somewhere` places may or may not work from GitHub Pages pages. (I couldn't find any documentation saying conclusively that it would or would not work.)


docs/daisy-reusing-workflows.md
```diff
@@ -0,0 +1,87 @@
...
+
+Instead of requiring you to copy-paste large sections of workflows to perform
+similar tasks, Daisy provides two different step types for workflow reuse:
+`IncludeWorkflow` and `SubWorkflow`. Both step types will cause Daisy to run the
+workflow they specify, but how Daisy does this is different:
+
+### IncludeWorkflow
+
+`IncludeWorkflow` will run the steps of the included workflow in parallel with
+the parent workflow's steps. A workflow included with `IncludeWorkflow` will
```
zmarano: Technically, its not always in parallel unless the dependency map says so. You can also run subWorkflows in parallel the same way. The primary point is that the resource namespaces are shared and therefore it is up to you to sort out collisions. It is useful however for sharing resources between multiple workflows (a good example is a disk that os created in a top level workflow and you want to perform actions on that disk in an included workflow that has common functionality to perform actions on a disk provided as a variable).

ericdand: Right. Thanks for pointing this out. I'll make it clearer that that's the case.

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+your image, write a script that performs all of them.
+
+All of the files that your startup script needs to access (which are not already
+present on the machine) should be listed in the "Sources" field of your Daisy
+workflow. Sources can be paths to local files, or to files in a Storage bucket
+using the URL format `gs://<bucket>/<file>`.
+
+Files listed in the "Sources" field of your Daisy workflow are all copied to a
+special location in Google Cloud Storage. For more information on getting the
+URL of this location and downloading files from it, see "Passing Data to
```
crunk1: Link "Passing Data..." to the doc.

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+`gcesysprep.bat`. All other OS scripts should end by shutting down the instance.
+
+## A Typical Workflow
+
+The typical Daisy image creation workflow has five steps:
+
+    1. A CreateDisks step which creates at least one disk. The SourceImage for
+       one of these disks should be your base image.
+    2. A CreateInstances step which creates an instance out of the disks the
+       previous step created. The first disk in this step's "Disks" field must
```
crunk1: Here and everywhere else: I like to put field names in inline code blocks, e.g. \`Disks\` = `Disks`. Take it or leave it.

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+       be bootable. This step also specifies the StartupScript.
+    3. A WaitForInstancesSignal step which waits for the startup script to
+       complete. WaitForInstancesSignal has a field called "Stopped" to listen
```
crunk1: Another example for my above comment: \`Stopped\`

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+       for the instance shutting down. It can also be made to listen for
+       specific output on the serial port.
+    4. A CreateImages step which actually creates an image from the disk once
+       the instance has shut down. Make sure you specify "NoCleanup" on this
```
crunk1: \`NoCleanup\`

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+       step, or your image will be deleted when the workflow ends!
+    5. A DeleteResources step that deletes the instance you created in the
+       CreateInstances step.
+
+Each step should depend on the step before it. Use the "Dependencies" field of
```
crunk1: \`Dependencies\`

docs/daisy-reusing-workflows.md
```diff
@@ -0,0 +1,90 @@
+# Reusing Workflows
+
+Instead of requiring you to copy-paste large sections of workflows to perform
+similar tasks, Daisy provides two different step types for workflow reuse:
+`IncludeWorkflow` and `SubWorkflow`. Both step types will cause Daisy to run the
+workflow they specify, but how Daisy does this is different:
+
+### IncludeWorkflow
+
+`IncludeWorkflow` will run the steps of the included workflow in parallel with
```
crunk1: I'm pretty sure IncludeWorkflows run as a single step (like how you described SubWorkflow), but @adjackura or @zmarano might know better than me. By the way, both a SubWorkflow step and an IncludeWorkflow step run in parallel to other steps with which they have no dependency relation.

The biggest (probably only) difference is that IncludeWorkflows share the same resource namespace as the parent workflow. SubWorkflows do not.

docs/daisy-reusing-workflows.md
```diff
@@ -0,0 +1,90 @@
...
+### SubWorkflow
+
+`SubWorkflow` runs the subworkflow specified as one step. Subworkflows do not
+have access to their parent's resources. This means that any workflow can be
+used as a subworkflow without fear of resource name collisions. Additionally,
+the subworkflow inherits its parent's `Project`, `Zone`, `GCSPath`, and
+`OAuthPath` fields. Even if these fields are set in the subworkflow, they will
+be overwritten with the parent's values.
+
+Unlike `IncludeWorkflow`, which is considered complete as soon as the included
```
crunk1: Same as above. I'm pretty sure IncludeWorkflows run like SubWorkflows: they run as a step and do not complete until their workflow completes.

The only difference is the shared resource namespace.

docs/daisy-migrating-existing-workflows.md
ericdand: Sounds good. Let's do that.

docs/daisy-reusing-workflows.md
ericdand: For sure, any and all steps will be run in parallel if there are no dependencies. But given that, I was talking with Andrew yesterday about this, and he explained to me (though I may have misunderstood) that `IncludeWorkflow` and `SubWorkflow` steps are considered "done" (for the purposes of dependencies) at different times: `IncludeWorkflow` immediately once all the tasks have been included (as soon as the dependency tree is rebuilt to fit the newly included workflow), and `SubWorkflow` only once all of that workflow's tasks have actually finished.

@adjackura , can you confirm?

@zmarano also commented on this same paragraph, calling for more clarity on how dependencies affect when tasks are run. I'll take another pass at it and see if I can't make it better.

ericdand:
Okay, docs updated after a talk with Zach and Scott. They corrected me on my idea of how `IncludeWorkflow` works.

@adjackura when you get to this, you can ignore all comments up to here. Just read the current revision of the docs and give your thoughts. ;)

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+  5. Write a Daisy workflow which creates at least one disk (the first disk
+     should be your base image), starts an instance from that disk, runs your
+     script, waits for the script to finish, then takes an image from the disk.
+
+The rest of this page describes these steps in greater depth.
+
+## Selecting a Base Image
+
+Your base image can come from anywhere, but if you're starting with an OS that
+is available on Compute Engine, then you can use what Daisy calls a "partial
```
adjackura: Partial URL is a term from the API, a Daisy workflow is essentially just a workflow of API calls (the json is mostly just passed through after some minor setup/validation). A partial URL is just the url without the common https://www.googleapis.com/compute/v1/

See sourceImage here:
https://cloud.google.com/compute/docs/reference/latest/disks

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+
+    projects/<project>/global/images/family/<family-name>
+
+You will need to find the correct project and family name for your chosen image
+family. To do this, navigate in the Cloud Console to Compute Engine -> Images,
+then click on the name of the instance you would like to use. On this page will
+be listed the image's family. To find the project, click on "Equivalent REST"
+and find the "selfLink" field. The second part of the path is the project name.
+Note that the project name is not the name of your own project, it is the name
+of the special project which contains your chosen base image. For example, the
```
adjackura: Drop the 'special' from special project, we don't know where they are sourcing their image (could be their project even)

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+
+The biggest thing to keep in mind is that Daisy only allows you to run one
+script on startup. This script can download and run other scripts, but it must
+be the one to do it. Once you have listed all of the steps you take to set up
+your image, write a script that performs all of them.
+
+All of the files that your startup script needs to access (which are not already
+present on the machine) should be listed in the "Sources" field of your Daisy
+workflow. Sources can be paths to local files, or to files in a Storage bucket
+using the URL format `gs://<bucket>/<file>`.
```
adjackura: Technically gcs stores objects

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+
+Files listed in the "Sources" field of your Daisy workflow are all copied to a
+special location in Google Cloud Storage. For more information on getting the
```
adjackura: Instead of 'special location' maybe 'scratch directory for your workflow'

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,110 @@
...
+    4. A CreateImages step which actually creates an image from the disk once
+       the instance has shut down. Make sure you specify `NoCleanup` on this
+       step, or your image will be deleted when the workflow ends!
+    5. A DeleteResources step that deletes the instance you created in the
+       CreateInstances step.
+
+Each step should depend on the step before it. Use the `Dependencies` field of
+your Daisy workflow to specify dependencies. If no dependencies are specified,
+then Daisy will try to run all the steps in parallel, which may give you some
+*interesting* results.
```
adjackura: We have validation that will catch things like trying to start an instance from a disk that that step does not depend on

docs/daisy-passing-data.md
```diff
@@ -0,0 +1,91 @@
...
+There are three ways data can be made accessible to a machine created by
+the CreateInstances step:
+
+1. Short strings and flags can be passed using the `Metadata` field.
+2. A single script can be run using the `StartupScript` field.
+3. Files can be retrieved from the workflow's `Sources` field.
+
+## Retrieving Metatdata
+
+Metadata can be though of as "Cloud environment variables". Metadata is a simple
```
adjackura: Thought

docs/daisy-passing-data.md
```diff
@@ -0,0 +1,91 @@
...
+see [here](https://cloud.google.com/compute/docs/storing-retrieving-metadata)
+for documentation on working with metadata.
+
+## Working with Sources
+
+The "Sources" field in your Daisy workflow can contain links to local files, or
+files in a Google Cloud Storage bucket using the URL scheme
+`gs://bucket-name/file-name`. Everything listed in Sources is copied to a
+special bucket created by Daisy. Daisy sets the special metadata key
+`daisy-sources-path` to a `gs://` path to the Sources bucket.
```
adjackura: Scratch bucket

docs/daisy-migrating-existing-workflows.md
ericdand: Cool, added a link to https://cloud.google.com/compute/docs/reference/latest/disks#sourceImage to the doc and made it sound less like partial URLs are unique to Daisy.

docs/daisy-migrating-existing-workflows.md
ericdand: done

docs/daisy-migrating-existing-workflows.md
ericdand: changed.

docs/daisy-migrating-existing-workflows.md
ericdand: Added a note about that behavior.

docs/daisy-passing-data.md
ericdand: good catch.

docs/daisy-passing-data.md
ericdand: done

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,115 @@
+# Migrating Image Creation Workflows to Daisy
```
zmarano: I'm thinking we should call this something other than "migrating" - only because it can be confused with VM and image migration (which is a thing we are doing with Daisy). Maybe we frame this as just using Daisy for image creation workflows. I get the point about assuming there are previous workflows for other tools but there are so many variations that we may as well just be general here.

A useful future guide however would be how to convert a packer workflow into a daisy workflow as that is one of the more common tools used.

docs/daisy-migrating-existing-workflows.md
```diff
@@ -0,0 +1,115 @@
...
+URL for the Ubuntu 16.04 LTS image family is:
+
+    projects/ubuntu-os-cloud/global/images/family/ubuntu-1604-lts
+
+You can use this partial URL in the "SourceImage" field in a "CreateDisks" Daisy
+step, or anywhere else that Daisy accepts partial URLs. By using an image
+family, you ensure that you will always run your workflow on the most up-to-date
+base image available.
+
+## Translating your Current Workflow
```
zmarano: Ditto about Translating, as that is a term we are using to translate imported images to GCE images.

docs/daisy-migrating-existing-workflows.md
ericdand: True, this page may be too specialized to be generally useful. I did indeed write it with a certain audience in mind.

I'll change it to be more generally about using Daisy to create images.

docs/daisy-migrating-existing-workflows.md
ericdand: Ack

docs/daisy-migrating-existing-workflows.md
ericdand: Okay, done. It actually took fewer changes than I thought -- I barely had to change anything. The tone of the doc is now a little more open to the possibility that you're currently building your images by hand, rather than assuming you already had an automated solution.

The title of the page (and thus the name of the file) has been changed to "Automating Image Creation with Daisy".

PR #220, opened 2017-11-08
Add workflow support for Oracle Linux

tomlanyon:
This adds support for building Oracle Linux 6 and 7 images.

I haven't included license information for the resulting images – do you have any guidance on where the IDs used in the CentOS and RHEL images come from?

Cleaned up a couple of generic EL pieces – simplified the ISO label handling and fixed an issue which was preventing the Anaconda logs being captured.


daisy_workflows/image_build/enterprise_linux/kickstart/ol7-post.cfg
```diff
@@ -0,0 +1,122 @@
...
+
+# 3
+%post
+# Configure the system to run in GCE.
+
+# Install GCE supplied disk expansion package.
+yum install -y gce-disk-expand
+
+# Install packages needed by GCE.
+yum install -y google-compute-engine google-config google-compute-engine-init
```
zmarano: Do you know if python-boto is available? That was one of the issues I had in the past.

tomlanyon: Yes, it's in the ol7_addons repo which is enabled above.

PR #224, opened 2017-11-09
Docs rearrangement + enhancement

ericdand:
This PR does two main things:

1. Fixes a couple of links in `daisy_workflows/README.md`, and
2. Addresses most of the feedback I wrote down while originally exploring the Daisy documentation.

There are also a few small typo/wording corrections here and there.

I also moved the glossary from the `daisy/README.md` to the workflow config spec, which probably should have been done in the first place, when I split the spec into its own document.


ericdand:
@zmarano There actually is a link -- it just tiny and easy to miss. The first paragraph ends in "Documentation can be found here."

I'll make a new "Docs" heading to make it more obvious.

PR #229, opened 2017-11-10
image_import: increase libguestfs memory to 4G

helen-fornazier:
libguestfs limits the memory to 500MB, this was causing the post install
script from google-compute-engine-oslogin package to run out of memory.
Increasing it to 4G solves the problem.


zmarano:
@helen-fornazier  You'll have to update your client before we can merge all of these. If we time it right, we can just do it all at once.

helen-fornazier:
do you mean to rebase my tree?

Issue #233, opened 2017-11-10
DeleteResources should ignore resources that don't exist


zmarano:
When using DeleteResources, if a resource does not exist (maybe because you manually deleted it), the workflow should continue, validation should pass, and other steps and resource deletions should still be able to occur.

Issue #242, opened 2017-11-14
Image status should be taken into account when validating image resources


zmarano:
Images can have multiple states including OBSOLETE- which means that the image resource exists but you can't actually use it to create disks or instances. Validation should take the state into account.

PR #246, opened 2017-11-15
Only log errors when a resource DNE in DeleteResources

adjackura:
Fail validation if an image is marked OBSOLETE or DELETED

Fixes issue #233 and #242

### DIFF
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -83,7 +83,7 @@ func (l *syncedWriter) Flush() error {
 }
 
 func daisyBkt(ctx context.Context, client *storage.Client, project string) (string, error) {
-	dBkt := project + "-daisy-bkt"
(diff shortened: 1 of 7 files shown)


daisy/workflow.go
```diff
@@ -83,7 +83,7 @@ func (l *syncedWriter) Flush() error {
 }
 
 func daisyBkt(ctx context.Context, client *storage.Client, project string) (string, error) {
-	dBkt := project + "-daisy-bkt"
+	dBkt := strings.Replace(project, ":", "-", -1) + "-daisy-bkt"
```
crunk1: What is this?

adjackura: Ah, forgot about that. This sanitizes the old style org projects (foo.com:my-project) before creating the daisy bucket.

PR #250, opened 2017-11-23
Changing daisy internal errors to dErrs.

crunk1:


daisy/error.go
```diff
@@ -21,58 +21,135 @@ import (
 
 const (
 	untypedError     = ""
+	multiError       = "MultiError"
+	fileIOError      = "FileIOError"
 	resourceDNEError = "ResourceDoesNotExist"
+
+	apiError    = "APIError"
+	apiError404 = "APIError404"
```
adjackura: Drop this as its unused, also a API 404 is resourceDNEError in our current code

daisy/project.go
```diff
@@ -37,7 +37,7 @@ func projectExists(client compute.Client, project string) (bool, error) {
 		if apiErr, ok := err.(*googleapi.Error); ok && apiErr.Code == http.StatusNotFound {
 			return false, nil
 		}
-		return false, err
+		return false, typedErr(apiError, err)
```
adjackura: My issue with this is you wont be able to check the api error in calling functions. You don't need to change this PR but maybe we should just put the API error code as the type? Or something like that so we don't have to have a daisy error type for every possible api error we get but can also still check (or probably just print out) what it is.

PR #252, opened 2017-12-06
import precheck: Windows 2008 R2 SHA2 driver signing check.

crunk1:


import_precheck/check_sha2.go
```diff
@@ -0,0 +1,65 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import (
+	"runtime"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/packages"
```
adjackura: fix your imports

import_precheck/check_sha2.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	pkgs, errs := packages.GetInstalledPackages()
+	if errs != nil {
+		return nil, fmt.Errorf("GetInstalledPackages errors:\n* %s", strings.Join(errs, "\n* "))
+	}
+
+	var pkgInfo *packages.PkgInfo
+	var kb string
+	for _, pkg := range append(pkgs["qfe"], pkgs["wua"]...) {
```
adjackura: for _, pkg := range append(pkgs["qfe"], pkgs["wua"]...) {
  if pkg.Version == SHA2WINDOWS2008R2KB || pkg.Version == WINDOWS2008R2ROLLUPKB {
    r.Info(fmt.Sprintf("Windows Update %s containing SHA2 driver signing support found: %v", kb, pkg))
    return r, nil
  } 
}
r.Fatal("SHA2 driver signing support not found.")
return r, nil

import_precheck/check_sha2.go
crunk1: Done.

import_precheck/check_sha2.go
crunk1: Done.

PR #260, opened 2017-12-13
Add gce_inventory_agent and instance_inventory binaries.

adjackura:
gce_inventory_agent gathers and writes instance inventory to guest attributes.
instance_inventory pretty prints out an instances inventory data, this is mainly meant to be used as a testing tool during the POC phase.

### DIFF
--- a/cli_tools/instance_inventory/main.go
+++ b/cli_tools/instance_inventory/main.go
@@ -1,0 +1,135 @@
+//  Copyright 2017 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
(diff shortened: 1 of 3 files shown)


cli_tools/instance_inventory/main.go
```diff
@@ -0,0 +1,135 @@
...
+		log.Fatalf("error creating HTTP client: %v", err)
+	}
+	service, err := compute.New(hc)
+	if err != nil {
+		log.Fatalf("error creating compute client: %v", err)
+	}
+
+	ats := []string{"Timestamp", "Hostname", "LongName", "ShortName", "Version", "KernelVersion", "InstalledPackages", "PackageUpdates", "Errors"}
+
+	//ii := map[string]string{}
```
crunk1: Delete.

cli_tools/instance_inventory/main.go
```diff
@@ -0,0 +1,135 @@
...
+	var ii sync.Map
+	var wg sync.WaitGroup
+	for _, at := range ats {
+		wg.Add(1)
+		go func(at string, eg *sync.WaitGroup) {
+			defer wg.Done()
+			value, err := getAttribute(service, *project, *zone, *instance, at)
```
crunk1: Doesn't `golint` or `go vet` yell at you when you use a loop variable (`at` in this case) in an inner go routine? Will the loop variable change?

crunk1: Looks like it will probably be OK. https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

adjackura: I am not using a loop variable, it would change and cause all sorts of issues, thats why I'm passing it through to the nested function (I just happen to use the same varaible name)

PR #264, opened 2017-12-19
New prow set up. More generic test infra; built to test all of compute-image-tools.

crunk1:


test-infra/prow/cluster/sinker_deployment.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: sinker
-        image: gcr.io/k8s-prow/sinker:0.21
+        image: gcr.io/k8s-prow/sinker:0.24
```
adjackura: Will we have to keep updating these images? Is there a release tag we could use instead?

crunk1: They don't have release tags yet.

PR #272, opened 2017-12-22
Initial commit of the gce_image_publish tool

adjackura:
This is a preliminary commit so that I can get feedback on anything I may be missing. From my testing this works as expected for publish and rollback of images (at least the same as our current tool).
Docs and additional unit tests still need to be written.

The idea is you write a publish template json file and point this tool at that, the template contains all the options you need to publish a set of images. Source images can be either a GCS image file or a GCE image, a daisy workflow is generated for the publish task.

### DIFF
--- a/cli_tools/gce_image_publish/main.go
+++ b/cli_tools/gce_image_publish/main.go
@@ -46,23 +46,40 @@ var (
 )
 
 type publish struct {
-	Name            string
-	WorkProject     string
-	SourceProject   string
-	SourceGCSPath   string
-	PublishProject  string
+	// Name for this publish template.
+	Name string
+	// Project to perform the work in.
+	WorkProject string
+	// Project to source images from, should not be used with SourceGCSPath.
+	SourceProject string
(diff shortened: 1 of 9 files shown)


cli_tools/gce_image_publish/main.go
```diff
@@ -0,0 +1,405 @@
...
+const gcsImageObj = "root.tar.gz"
+
+var (
+	now            = time.Now().UTC()
+	oauth          = flag.String("oauth", "", "path to oauth json file")
+	sourceVersion  = flag.String("source_version", now.Format("20060102"), "version on source image")
+	publishVersion = flag.String("publish_version", *sourceVersion, "version for published image if different from source")
+	skipDup        = flag.Bool("skip_duplicates", false, "skip publishing an image that already exists")
+	rollback       = flag.Bool("rollback", false, "rollback image publish")
+	noConfirm      = flag.Bool("noConfirm", false, "don't ask for confirmation")
```
illfelder: I think this should have underscores in the flag name.

PR #273, opened 2018-01-04
Daisy: Add {Create,Delete}Network to daisy/compute. Merge Create* and Delete* tests.

crunk1:


ericdand:
Looks good, but what's up with the Prow stuff in 79db840? Should that be its own PR, or is it related to the rest of what you're doing here?

crunk1:
The Prow stuff was needed to get tests to run on this PR.

PR #290, opened 2018-01-19
Allow simpler GuestOsFeatures

adjackura:


daisy/image.go
```diff
@@ -115,6 +117,23 @@ func (i *Image) MarshalJSON() ([]byte, error) {
 	return json.Marshal(*i)
 }
 
+type guestOsFeatures []string
+
+// UnmarshalJSON unmarshals GuestOsFeatures.
+func (g *guestOsFeatures) UnmarshalJSON(b []byte) error {
```
crunk1: We could leverage similar logic on Create Instances probably. Not needed in this PR obviously.

daisy/image.go
```diff
@@ -139,9 +158,20 @@ func (i *Image) populate(ctx context.Context, s *Step) dErr {
 		}
 	}
 	i.link = fmt.Sprintf("projects/%s/global/images/%s", i.Project, i.Name)
+	i.populateGuestOSFeatures(s.w)
 	return errs
 }
 
+func (i *Image) populateGuestOSFeatures(w *Workflow) {
```
crunk1: Any validation needed for guest os features?

adjackura: Sadly I can't easily validate features without hardcoding them into our code.

PR #291, opened 2018-01-19
Fix Debian build workflows to correctly handle Google Cloud Repos.

zmarano:
- Also cleanup unnecessary tar exports and other oddities.


daisy_workflows/image_build/debian/debian.wf.json
```diff
@@ -52,17 +47,10 @@
     "wait": {
```
adjackura: Our other build workflows create the image in GCE and don't export the image to GCS

PR #295, opened 2018-01-20
Daisy: Added CreateNetworks and Network validation logic.

crunk1:
Also some other code/test cleanup.


daisy/step_create_networks.go
```diff
@@ -0,0 +1,71 @@
...
+
+// CreateNetworks is a Daisy CreateNetwork workflow step.
+type CreateNetworks []*Network
+
+func (c *CreateNetworks) populate(ctx context.Context, s *Step) dErr {
+	var errs dErr
+	for _, n := range *c {
+		errs = addErrs(errs, n.populate(ctx, s))
+	}
+	return nil
```
adjackura: return errs?

daisy/step_create_networks.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (c *CreateNetworks) validate(ctx context.Context, s *Step) dErr {
+	var errs dErr
+	for _, n := range *c {
+		errs = addErrs(errs, n.validate(ctx, s))
+	}
+	return nil
```
adjackura: return errs?

daisy/step_create_networks.go
crunk1: You can't err out if you don't return them.
roll-safe-think-about-it

PR #304, opened 2018-01-26
Workflows: Add export for additional format types using qemu-img

adjackura:


daisy_workflows/export/README.md
```diff
@@ -1,38 +1,77 @@
...
+format. 
+
+## GCE raw disk image
+`image_export.wf.json` and `disk_export.wf.json` export a raw image in
+a tar.gz, this is the 'native' GCE format and the resulting image can 
+be imported directly to GCE.
+
 Required vars:
-+ `source_disk` GCE disk to export
++ `source_image` GCE image to export
```
crunk1: Looks like you repurposed this disk export workflow documentation to fit image export. Do we have documentation for disk export anywhere now?

adjackura: No, I figured the image export was more clear, and it looks very similar to disk export. We can add disk export at a later time but it would require some additional comments about ensuring disks exist in the same zone/project

PR #308, opened 2018-01-29
Daisy: add ${SOURCE:source} variable types to insert source content in workflows

adjackura:


daisy/validate.go
```diff
@@ -113,7 +108,9 @@ func (w *Workflow) validateVarsSubbed() dErr {
 		switch v.Interface().(type) {
 		case string:
 			if match := unsubbedVarRgx.FindStringSubmatch(v.String()); match != nil {
-				return errf("Unresolved var %q found in %q", match[0], v.String())
+				if !sourceVarRgx.MatchString(v.String()) {
```
crunk1: If we are iterating over unsubbedVars here, why do we even need validateVarsSubbed?

daisy/instance.go
```diff
@@ -181,6 +181,19 @@ func (i *Instance) populateMetadata(w *Workflow) dErr {
 		i.Metadata["windows-startup-script-url"] = i.StartupScript
 	}
 	for k, v := range i.Metadata {
+		if match := sourceVarRgx.FindStringSubmatch(v); match != nil {
```
crunk1: Why are we doing special var subbing logic in populateMetadata?

crunk1: Why not part of the overall workflow var subbing logic?

daisy/step_includeworkflow.go
```diff
@@ -81,6 +81,12 @@ func (i *IncludeWorkflow) populate(ctx context.Context, s *Step) dErr {
 	}
 	substitute(reflect.ValueOf(i.Workflow).Elem(), strings.NewReplacer(replacements...))
 
+	// We do this here, and not in validate, as embeded startup scripts could
```
crunk1: Nit: embedded.

daisy/workflow.go
```diff
@@ -394,6 +394,11 @@ func (w *Workflow) populate(ctx context.Context) dErr {
 	}
 	substitute(reflect.ValueOf(w).Elem(), strings.NewReplacer(replacements...))
 
+	// We do this here, and not in validate, as embeded startup scripts could
```
crunk1: Nit: embedded.

daisy/step_includeworkflow.go
```diff
@@ -81,6 +81,16 @@ func (i *IncludeWorkflow) populate(ctx context.Context, s *Step) dErr {
 	}
 	substitute(reflect.ValueOf(i.Workflow).Elem(), strings.NewReplacer(replacements...))
 
+	// We do this here, and not in validate, as embedded startup scripts could
+	// have what we think are daisy variables.
+	if err := i.Workflow.validateVarsSubbed(); err != nil {
+		return err
```
crunk1: instead of doing the if err := __; err != nil { return err } thing, you can do errs = addErrs(errs, ___). If you want.

crunk1: Same comment for other changes, too.

adjackura: These are pretty much hard fails, that's why these functions don't already do that. For example failure to populate something could cause a string of other failures, so better to just return the error when you get it.

crunk1:
I'd still recommend tests.

PR #311, opened 2018-01-30
Start implementation of Linux integration tests

helen-fornazier:
Implementing boot tests on the image given by the source_image variable

### DIFF
--- a/daisy_workflows/image_test/README.md
+++ b/daisy_workflows/image_test/README.md
@@ -1,0 +1,163 @@
+# GCE Image Integration Test
+These are workflows used by the GCE team to automate integration tests
+with Linux images and GCE
+
+# Test Cases
+
+Boot
+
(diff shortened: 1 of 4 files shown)


helen-fornazier:
The way I was viewing this was to use daisy workflows to handle the creation of instances to be tested.
For the Metadata SSH tests for example, we would need to generate two instances (A and B), the instance A would create the ssh keys and add it in to the metadata for the instance B, A would try to login to B and in case of success or failure it would signal to the daisy workflow. With daisy we can trigger several tests in parallel too.
Please let me know if this make sense or if you had another architecture in mind.

zmarano:
Yup this is exactly what we were looking at doing. The one thing that we should think about (going forward here) is creating a library of common functionality for the more advanced tests to share (similar to the utils.py library we are using for image import and image builds). But otherwise, this looks good.

PR #314, opened 2018-01-31
Daisy: Add AttachDisks step type

### LINKED ISSUE #247: Create new step type(s) to attach and detach disks from instances
We need a new step type with a name like `AttachDisks` to attach disks, and another, call it `DetachDisks` to detach disks. These should attach/detach disks from running instances.

adjackura:
Fixes #247

### DIFF
--- a/daisy/step_attach_disks.go
+++ b/daisy/step_attach_disks.go
@@ -57,9 +57,10 @@ func (a *AttachDisks) validate(ctx context.Context, s *Step) (errs dErr) {
 		}
 
 		ir, err := s.w.instances.regUse(ad.Instance, s)
-		if err != nil {
+		addErrs(errs, err)
+		if ir == nil {
 			// Return now, the rest of this function can't be run without ir.
-			return addErrs(errs, err, errf("cannot attach disk: instance %q not found in registry", ad.Instance))
+			return addErrs(errs, errf("cannot attach disk: instance %q not found in registry", ad.Instance))
 		}
 
 		if diskURLRgx.MatchString(ad.Source) {
@@ -67,9 +68,10 @@ func (a *AttachDisks) validate(ctx context.Context, s *Step) (errs dErr) {
 		}
 
 		dr, err := s.w.disks.regUse(ad.Source, s)
-		if err != nil {
+		addErrs(errs, err)
+		if dr == nil {
 			// Return now, the rest of this function can't be run without dr.
-			return addErrs(errs, err, errf("cannot attach disk: disk %q not found in registry", ad.Source))
+			return addErrs(errs, errf("cannot attach disk: disk %q not found in registry", ad.Source))
 		}
 
 		// Ensure disk is in the same project and zone.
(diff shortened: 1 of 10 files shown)


daisy/step_attach_disks.go
```diff
@@ -0,0 +1,128 @@
...
+		if !checkDiskMode(ad.Mode) {
+			fmt.Println("adding err")
+			errs = addErrs(errs, errf("cannot attach disk: bad disk mode: %q", ad.Mode))
+		}
+		if ad.Source == "" {
+			errs = addErrs(errs, errf("cannot attach disk: AttachedDisk.Source is empty"))
+		}
+
+		ir, err := s.w.instances.regUse(ad.Instance, s)
+		if err != nil {
```
crunk1: This doesn't mean that ir doesn't exist. This could also err if it is determined that this step doesn't have the proper relationships with the creator and/or deleter of ir. e.g. this step doesn't depend on the creator step of ir.

daisy/step_attach_disks.go
```diff
@@ -0,0 +1,128 @@
...
+			// Return now, the rest of this function can't be run without ir.
+			return addErrs(errs, err, errf("cannot attach disk: instance %q not found in registry", ad.Instance))
+		}
+
+		if diskURLRgx.MatchString(ad.Source) {
+			ad.Source = extendPartialURL(ad.Source, ir.Project)
+		}
+
+		dr, err := s.w.disks.regUse(ad.Source, s)
+		if err != nil {
```
crunk1: Same.

daisy/step_attach_disks.go
```diff
@@ -0,0 +1,128 @@
...
+			// Return now, the rest of this function can't be run without dr.
+			return addErrs(errs, err, errf("cannot attach disk: disk %q not found in registry", ad.Source))
+		}
+
+		// Ensure disk is in the same project and zone.
+		disk := namedSubexp(diskURLRgx, dr.link)
```
crunk1: We might be able to refactor this logic into regAttach. I think CreateInstances might do something similar when validating attached disks.

crunk1: We can do this in a separate PR perhaps.

daisy/step_attach_disks.go
```diff
@@ -0,0 +1,130 @@
...
+
+		ir, err := s.w.instances.regUse(ad.Instance, s)
+		addErrs(errs, err)
+		if ir == nil {
+			// Return now, the rest of this function can't be run without ir.
+			return addErrs(errs, errf("cannot attach disk: instance %q not found in registry", ad.Instance))
+		}
+
+		if diskURLRgx.MatchString(ad.Source) {
+			ad.Source = extendPartialURL(ad.Source, ir.Project)
```
crunk1: This is a matter of opinion perhaps, but I've started moving the "if URL match, extend URL" logic into populate instead of validate.

If it can't be put in populate, could you put a comment here?

PR #317, opened 2018-02-01
Use the same builder image for Debian image builds.

zmarano:
This fixes an issue with how the filesystem is being created on the
resulting image. The e2fs version in Debian 8 does not support
metadata_checksums. If the partition is created from a Debian 9 e2fs
toolset, fsck will fail on Debian 8 boots.
systemd-fsck[121]: /dev/sda1 has unsupported feature(s): metadata_csum


zmarano:
Also, fix GCS test bucket references.

PR #328, opened 2018-02-08
image_test: Add metadata-ssh tests

helen-fornazier:
First implementation of metadata-ssh tests
Include linux_common for metadata ssh tests to allow reuse in other
tests

### DIFF
--- a/daisy_workflows/image_test/linux_common/bootstrap.py
+++ b/daisy_workflows/image_test/linux_common/bootstrap.py
@@ -1,0 +1,58 @@
+#!/usr/bin/python
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


daisy_workflows/image_test/linux_common/bootstrap.py
```diff
@@ -0,0 +1,58 @@
...
+    subprocess.check_call(['mkdir', test_dir])
+    subprocess.check_call(
+        ['gsutil', '-m', 'cp', '-r', os.path.join(test_gcs_dir, '*'), 
+        test_dir])
+    logging.info('Making test script %s executable.', full_test_script)
+    subprocess.check_call(['chmod', '+x', test_script], cwd=test_dir)
+    logging.info('Running %s.', full_test_script)
+    subprocess.check_call([full_test_script], cwd=test_dir)
+  except Exception as e:
+    print('TestFailed: error: ')
```
adjackura: If you can try to print the basic error message on the same line as TestFailed so that daisy will log the reason (digging into the gcs logs for the trace is fine)

PR #340, opened 2018-02-21
Add a diagnostics tool to help collect logs on Windows.

geekster777:
The tool allows one to collect a series of logs and save them to a zip
file. When supplied with a Signed Url, the tool can upload the logs to a
GCS Bucket. The tool currently supports Windows only.


cli_tools/diagnostics/main_windows.go
```diff
@@ -0,0 +1,241 @@
...
+	traceStart.run()
+	time.Sleep(10 * time.Minute)
+	paths := runAll([]runner{
+		traceStop,
+	})
+	logs <- logFolder{"Trace", paths}
+
+}
+
+func gatherLogs(trace bool) []logFolder {
```
adjackura: This could be simplified into something like this:

runFuncs := []func(logs chan logFolder) {
  gatherSystemLogs
  gatherDiskLogs
  ...
}
if trace {
  runFuncs = append(runFuncs, gatherTraceLogs)
}
folderCount := len(runFuncs)

for _, run := range runFuncs {
  go run(ch)
}

cli_tools/diagnostics/main_windows.go
```diff
@@ -0,0 +1,241 @@
...
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	sys32 = `C:\Windows\System32`
+)
+
+type runner interface {
```
adjackura: runner and cmd should probably be in main.go so that linux can use the same style

cmd.run() probably could also be generalized

cli_tools/diagnostics/main_windows.go
```diff
@@ -0,0 +1,241 @@
...
+type wmiQuery struct {
+	class          string
+	namespace      string
+	outputFileName string
+}
+
+func (command cmd) run() (string, error) {
+	outPath := filepath.Join(tmpFolder, command.outputFileName)
+
+	cmdPath := filepath.Join(sys32, command.name)
```
adjackura: Don't do this here, use the full path when creating the cmd so that you can run files not just in sys32

cli_tools/diagnostics/main_windows.go
```diff
@@ -0,0 +1,241 @@
...
+		if err == nil {
+			break
+		}
+	}
+	if err != nil {
+		return outPath, err
+	}
+
+	header := fmt.Sprintf("Queried wmi objects [%s] from namespace %s\n\n", query.class, query.namespace)
+	_, err = outFile.WriteString(header)
```
adjackura: if _, err := ...; err != nil {

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,153 @@
...
+		if arg == "--trace" {
+			trace = true
+		} else {
+			signedURL = arg
+		}
+	}
+
+	paths := gatherLogs(trace)
+	zipFile := filepath.Join(tmpFolder, "logs.zip")
+	err := zipFiles(paths, zipFile)
```
adjackura: if err := ...; err != nil {

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		log.Fatalf("Error zipping files: %v", err)
+	}
+
+	if signedURL != "" {
+		err = uploadToSignedURL(zipFile, signedURL)
+		if err != nil {
+			log.Fatalf("Error uploading to signed url: %v. Logs can be found at %s", err, zipFile)
+		} else {
```
adjackura: unnecessary else

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,153 @@
...
+			log.Print("Logs uploaded to the supplied url successfully.")
+		}
+	} else {
+		knownZipPath, err := moveZipFile(zipFile)
+		if err != nil {
+			log.Fatalf("Error moving logs to well known directory. They can be found instead at: %s", zipFile)
+		} else {
```
adjackura: unnecessary else

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func main() {
+	if tmpFolderErr != nil {
+		log.Panic("Error creating a temporary folder. Exiting")
+	}
+
+	trace := false
+	signedURL := ""
+	for _, arg := range os.Args[1:] {
```
adjackura: use the flags library

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,153 @@
...
+		return "", err
+	}
+
+	knownZipPath := filepath.Join(currDir, filepath.Base(path))
+	return knownZipPath, os.Rename(path, knownZipPath)
+}
+
+func main() {
+	if tmpFolderErr != nil {
+		log.Panic("Error creating a temporary folder. Exiting")
```
adjackura: avoid panic, for this:

var tmpFolder string

main() {
var err error
tmpFolder, err = ioutil.TempDir("", "diagnostics")
if err != nil {
  log.Fatal(err)
}
...
}

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,153 @@
...
+	signedURL := ""
+	for _, arg := range os.Args[1:] {
+		if arg == "--trace" {
+			trace = true
+		} else {
+			signedURL = arg
+		}
+	}
+
+	paths := gatherLogs(trace)
```
adjackura: you may want to think about setting up an error channel so that gatherLogs can return an error so you can exit non-0 at the end of main

cli_tools/diagnostics/main_windows.go
geekster777: Done

cli_tools/diagnostics/main_windows.go
geekster777: Done

cli_tools/diagnostics/main_windows.go
geekster777: Done

cli_tools/diagnostics/main_windows.go
geekster777: Fixed this everywhere I've seen it. 

Just for clarity, you only do this if the value(s) being returned are being used exclusively within the scope of the if statement?

cli_tools/diagnostics/main.go
geekster777: Done

cli_tools/diagnostics/main.go
geekster777: Done

cli_tools/diagnostics/main.go
geekster777: Done

cli_tools/diagnostics/main.go
geekster777: Done

cli_tools/diagnostics/main.go
geekster777: Done

cli_tools/diagnostics/main.go
geekster777: I added an error channel within gatherLogs, but it just returns a generic error if the channel is not empty. Let me know if there's a more go-centric way of accomplishing this.

cli_tools/diagnostics/main_windows.go
adjackura: Yes, the if statement scopes the variable declaration

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,172 @@
...
+				continue
+			}
+
+			if _, err = io.Copy(zf, file); err != nil {
+				log.Printf("Error saving contents of file %s with error %v\n", path, err)
+				err = errNonFatal
+			}
+		}
+	}
+	return err
```
adjackura: Reason for this err return? I dont see you setting it at all. Why not just return nil?

cli_tools/diagnostics/main.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func zipFiles(logs []logFolder, outputPath string) error {
+	newFile, err := os.Create(outputPath)
+	if err != nil {
+		return err
+	}
+	defer newFile.Close()
+	writer := zip.NewWriter(newFile)
+	defer writer.Close()
```
adjackura: You should catch the errors at close (here and for newFile). You can do this using named returns:

func zipFiles(logs []logFolder, outputPath string) (err error) {
...
defer func() {
	cErr := writer.Close()
	if cErr != nil && err == nil {
		err = cErr
	}
}()
}

cli_tools/diagnostics/main_windows.go
```diff
@@ -0,0 +1,248 @@
...
+	wg.Wait()
+	close(ch)
+	close(errs)
+
+	for path := range ch {
+		paths = append(paths, path)
+	}
+
+	if len(errs) > 0 {
+		errCh <- errNonFatal
```
adjackura: for this function you should probably return all the errors concatenated as one, leave the errNonFatal logic up to the caller

cli_tools/diagnostics/main_windows.go
```diff
@@ -0,0 +1,248 @@
...
+	return outPath, err
+}
+
+func runAll(commands []runner, errCh chan error) []string {
+	paths := make([]string, 0, len(commands))
+	ch := make(chan string, len(commands))
+	errs := make(chan error, len(commands))
+	var wg sync.WaitGroup
+
+	for _, command := range commands {
```
adjackura: Your doing a lot of concurrent runs in this tool making it very resource intensive, maybe serialize runAll? I'm on the fence about serializing gatherLogs too.

cli_tools/diagnostics/main.go
geekster777: Cleaned up the namespace.

cli_tools/diagnostics/main.go
geekster777: Updated everywhere I close a file.

cli_tools/diagnostics/main_windows.go
geekster777: Done.

cli_tools/diagnostics/main_windows.go
geekster777: Changed runAll to be serial. Keeping gatherLogs in concurrent as some commands like tracing will take multiple minutes.

PR #362, opened 2018-03-08
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/daisy_workflows/image_import/ubuntu/translate.py
+++ b/daisy_workflows/image_import/ubuntu/translate.py
@@ -107,9 +107,9 @@ def DistroSpecific(g):
 
   # Try to reset the network to DHCP.
   if ubu_release == 'trusty':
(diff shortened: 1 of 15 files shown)


daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -107,9 +107,9 @@ def DistroSpecific(g):
 
   # Try to reset the network to DHCP.
   if ubu_release == 'trusty':
-    g.write('/etc/network/intefaces', trusty_network)
```
jsoref: #341 introduced this, and I don't think it would work...

zmarano: You're right, please change it to /etc/network/interfaces

PR #375, opened 2018-03-28
Fix Debian build and take full .tar.gz path as image_dest arg.

zmarano:
- This is more consistent with the other image builds.

### DIFF
--- a/daisy_workflows/image_build/debian/build.py
+++ b/daisy_workflows/image_build/debian/build.py
@@ -81,7 +81,7 @@ def main():
 
   # Inject Google Cloud test repo plugin if using staging or unstable repos.
   # This is used to test new package releases in images.
-  if repo is not 'stable':
+  if repo != 'stable':
     utils.Status('Adding Google Cloud test repos plugin for bootstrapvz.')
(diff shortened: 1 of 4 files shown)


daisy_workflows/image_build/debian/build.py
```diff
@@ -102,9 +102,8 @@ def main():
   # Upload tar.
   image_tar_gz = '/target/disk.tar.gz'
   if os.path.exists(image_tar_gz):
-    image_tar_gz_dest = os.path.join(image_dest, 'root.tar.gz')
-    utils.Status('Saving %s to %s' % (image_tar_gz, image_tar_gz_dest))
-    utils.Gsutil(['cp', image_tar_gz, image_tar_gz_dest])
+    utils.Status('Saving %s to %s' % (image_tar_gz, image_dest))
```
ericdand: Nit: prefer this:

    'Saving {} to {}'.format(image_tar_gz, image_dest)

to this:

    'Saving %s to %s' % (image_tar_gz, image_dest)

because it's Python 3-compatible.

ericdand:
Before we check this in -- there's something wrong with the Prow Python scripts. They're not even building. Wanna take a look?

PR #380, opened 2018-04-18
Add basic Cloud Logging support.

ryanwe:
Add Cloud Logging support to daisy.

- Cloud Logging is the default.
    - Logs will go to GCS when specified by the flag or when CloudLogging
      is unavailable.
- Removes stdout logging unless specified by a flag.
  - This was undesirable when running via gcloud.
- Cloud logs are structured JSON logs.
 - Stdout and GCS logs are still strings for compatibility. However, the timestamp now contains timezone.

Cloud Logs example
screenshot from 2018-04-19 12-09-15

### DIFF
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -16,14 +16,11 @@
 package daisy
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
-	"log"
 	"os"
 	"path"
 	"path/filepath"
@@ -33,6 +30,7 @@ import (
 	"sync"
 	"time"
 
+	"cloud.google.com/go/logging"
 	"cloud.google.com/go/storage"
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
 	"google.golang.org/api/iterator"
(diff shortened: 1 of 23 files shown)


chunjiali:
I also think we should use cloud logging as the default instead of logging to both (cloud logging and gcs).

ryanwe:
Cool. So to recap, let me know if this sounds right:
- The default is cloud logging.
  - If cloud logging is unavailable (no api access, etc.) we fallback to using GCS.
  - If the new flag `-log-gcs true` is set, we use GCS instead.
  - The Stackdriver "LogName" will be `daisy-workflow-[workflow.id]`
- We will update all our logs to make them structured.
  - I'll look into what the actual structure of the logs can be to make it easy to track sub-workflows, steps, etc.
  - Serial logging and GCS logging will flatten the structured logs to maintain the current log schema. (Or is that not necessary? I'm not sure how much string matching might already be in place around these log entries for monitoring purposes.)
- I'll leave the serial output logs alone for now. We can fold them in afterwards if we decide that's what we want to do.

adjackura:
Yeah that sounds good. Keeping the current logging would be best so as not
to confuse users, writing a stringer for the structured log format you come
up with shouldn't be too difficult. If anything this will make the logging
logic way more understandable, also keep in mind that console logs may need
to look a bit different, workflows themselves log and the Daisy binary logs
as well (cloud logger probably only needs to be in the workflow logging
logic)

ryanwe:
I ended up adding another flag for enabling/disabling console logs which should make Daisy play nice with gcloud. I updated the PR summary to reflect these decisions and the changes made.

daisy/workflow.go
```diff
@@ -478,6 +436,11 @@ func (w *Workflow) NewIncludedWorkflow() *Workflow {
 	return iw
 }
 
+// ID is the unique identifyier for this Workflow.
+func (w *Workflow) ID() string {
```
ericdand: What is the purpose of adding a randomized ID to a workflow, which is only ever exposed in logs? Just to make the logs easier to parse?

ryanwe: Yes. The id already existed and is used as the suffix in the GCS bucket name. It's now also a suffix in the log name, eg. `projects/[project]/logs/daisy-workflow-[workflow.ID()]`. I wanted to add it to the initial console logs to make it easier to find the GCS bucket and Cloud Log. So now when you kick off a new workflow you see `[Daisy] Running workflow "[workflow name]" (id=abcde)` instead of just the name.

PR #381, opened 2018-04-23
Cloud Logging - test permissions when creating client.

ryanwe:
- Test Cloud Logging permissions when creating client by pinging the service.
- Update documentation.
- Add more tests.

### DIFF
--- a/daisy/README.md
+++ b/daisy/README.md
@@ -25,6 +25,15 @@ Other use-case examples:
 Daisy documentation can be found
 [here](https://googlecloudplatform.github.io/compute-image-tools/daisy.html).
 
+# Logging
+
+Daisy will send logs to Cloud Logging. If the API is disabled or the
+account Daisy is running is under does not have permission to write log entries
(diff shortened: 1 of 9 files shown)


daisy/README.md
```diff
@@ -25,6 +25,15 @@ Other use-case examples:
 Daisy documentation can be found
 [here](https://googlecloudplatform.github.io/compute-image-tools/daisy.html).
 
+# Logging
+
+Daisy will send logs to Cloud Logging. If the API is disabled or the
```
ericdand: It would be good to make "Cloud Logging" a link to the docs page for Cloud Logging (https://cloud.google.com/logging/), so people can quickly find out what it is, especially since we often call it "Stackdriver Logging" externally.

ericdand: Also, consider adding a note about the logging flags here: https://googlecloudplatform.github.io/compute-image-tools/daisy-installation-usage.html

PR #386, opened 2018-04-27
Daisy: set MTUs in Windows-for-containers image.

pjh:
Adds a step to set the vEthernet MTU inside a container. This change is
then reflected in all subsequent containers based on the
windowsservercore and nanoserver images.

Fixes a bug where the host vEthernet MTU was not being set correctly
because the interface did not exist yet.

### DIFF
--- a/daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
+++ b/daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
@@ -50,7 +50,13 @@ try {
     Restart-Computer -Force
   }
   else {
-    Write-Host 'Setting vEthernet MTU to 1460'
+    # For some reason the docker service may not be started automatically on the
+    # first reboot, although it seems to work fine on subsequent reboots. The
+    # docker service must be running or else the vEthernet interface may not be
+    # present.
+    Restart-Service docker
+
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
```diff
@@ -50,15 +50,20 @@ try {
     Restart-Computer -Force
   }
   else {
-    Write-Host 'Setting vEthernet MTU to 1460'
+    # For some reason the docker service may not be started automatically on the
+    # first reboot, although it seems to work fine on subsequent reboots. The
+    # docker service must be running or else the vEthernet interface may not be
+    # present.
+    Restart-Service docker
```
functicons: Instead of restarting docker, can we verify it is running? Something like:

```
if docker is not running:
  restart docker
if docker is not running:
  exit with error
Write-Host '...'
```

adjackura: That's additional logic with really no benefit. Docker seems to have issues starting reliably right after it is installed (even if it says it's started we have had issues connecting to it), unsure if this is simply a timing issue or what but simply restarting the service fixes it. Subsequent commands will fail with errors if docker isn't running so no need to add an additional check.

PR #389, opened 2018-05-01
 Daisy: Add GCSPath to DeleteResources step

adjackura:


zmarano:
Rather, the DeleteResources step

PR #397, opened 2018-05-04
daisy: Add StopStep

helen-fornazier:
Step to stop an instance without deleting it.
The StopStep is similar to DeleteResources for instances, but allow
adding WaitForInstanceSignal without making it a dependency to the
Delete step.

Depend on #396

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -39,6 +39,7 @@ type Client interface {
 	DeleteDisk(project, zone, name string) error
 	DeleteImage(project, name string) error
 	DeleteInstance(project, zone, name string) error
+	StopInstance(project, zone, name string) error
 	DeleteNetwork(project, name string) error
 	DeprecateImage(project, name string, deprecationstatus *compute.DeprecationStatus) error
 	GetMachineType(project, zone, machineType string) (*compute.MachineType, error)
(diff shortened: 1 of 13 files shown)


daisy/step_stop_instances.go
```diff
@@ -0,0 +1,77 @@
...
+	w := s.w
+	e := make(chan dErr)
+
+	for _, i := range st.Instances {
+		wg.Add(1)
+		go func(i string) {
+			defer wg.Done()
+			w.Logger.StepInfo(w, "StopInstances", "stopping instance %q.", i)
+			if err := w.instances.stop(i); err != nil {
+				if err.Type() == resourceDNEError {
```
adjackura: This should never happen, just go ahead and remove the resourceDNEError check and log

daisy/compute/compute.go
```diff
@@ -39,6 +39,7 @@ type Client interface {
 	DeleteDisk(project, zone, name string) error
 	DeleteImage(project, name string) error
 	DeleteInstance(project, zone, name string) error
+	StopInstance(project, zone, name string) error
```
adjackura: You'll also need to add the test implementation to test_client, thats probably why the tests are failing

daisy/step.go
```diff
@@ -48,6 +48,7 @@ type Step struct {
 	CreateInstances        *CreateInstances        `json:",omitempty"`
 	CreateNetworks         *CreateNetworks         `json:",omitempty"`
 	CopyGCSObjects         *CopyGCSObjects         `json:",omitempty"`
+	StopInstances          *StopInstances          `json:",omitempty"`
```
adjackura: Would you also need startinstances? If you don't need it for you tests there's no need to add it now, though we proabbly should make an issue to track it.

helen-fornazier: I don't need startinstances for my tests, I'll make an issue then

daisy/compute/compute.go
helen-fornazier: done, could you check please?

helen-fornazier:
sorry, codecov wasn't configured in my repo, I'll fix those checks and update the PR

PR #398, opened 2018-05-04
image_test: Add tests for metadata-scripts (basic)

helen-fornazier:
Basic tests for:
- Test integrity of the scripts
- Test if the machine doesn't crash after executing a junk file
- Test when no script is available

Depend on #396 and #397


zmarano:
That would be because gsutil doesn't work on SLES and the current incarnation of the Linux guest uses gsutil to pull startup-script-url scripts. So instead of using StartupScript you would have to populate the "startup-script" metadata with a script of some sort... we can punt on SLES testing for now if you want.

zmarano:
Actually, you can do this instead of using StartupScript and it will pull in the contents of the script:

  "Sources": {
    "startup_script.sh": "./startup_script.sh"
  },
...
    "create-instance": {
      "CreateInstances": [
        {
          "Name": "inst",
          "Disks": [{"Source": "disk"}],
          "MachineType": "n1-standard-2",
          "MetaData": {
            "block-project-ssh-keys": "TRUE",
            "startup-script": "${SOURCE:startup_script.sh}"
          }
        }
      ]
    },

helen-fornazier:
> Actually, you can do this instead of using StartupScript and it will pull in the contents of the script:

@zmarano Yes, but I need to change a bit the architecture of this test, as the SOURCE tag doesn't accept a source file that doesn't exist and I use this feature.
Is it ok if I do this in another PR? So I can advance in the FreeBSD testing infra and then come back for this issue latter.
We also have this gsutil issue in other parts, I want to modify bootstrap.py to not depend on gsutil too.

helen-fornazier:
I am closing this, we are going to re-work on this from a branch in collabora-gce project.

PR #403, opened 2018-05-09
Daisy: Log serial output to cloud logging

adjackura:
Clean up log text
Make GCS links clickable instead of using the gs:// path

### DIFF
--- a/daisy/logger.go
+++ b/daisy/logger.go
@@ -75,6 +75,8 @@ func (l *daisyLog) StepInfo(w *Workflow, stepName, stepType, format string, a ..
 		StepName:       stepName,
 		StepType:       stepType,
 		Message:        fmt.Sprintf(format, a...),
+
+		Type: "Daisy",
 	}
 	l.writeLogEntry(entry)
 }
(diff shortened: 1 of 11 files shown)


daisy/logger.go
```diff
@@ -107,11 +109,13 @@ type logEntry struct {
 	StepName       string    `json:"stepName,omitempty"`
 	StepType       string    `json:"stepType,omitempty"`
 	Message        string    `json:"message"`
+
+	Type string `json:"type"`
```
ericdand: This field doesn't seem to be ever used. It's just set to "Daisy" every time. Do you intend to use it in the future or something? Or am I just missing where it's used?

daisy/step_create_instances.go
```diff
@@ -31,47 +32,71 @@ type CreateInstances []*Instance
...
+	// Write the output to cloud logging only after instance has stopped.
+	// Type assertion check is needed for tests not to panic.
+	// Split if output is too long for log entry (100K max, we leave a 2K buffer).
+	dl, ok := w.Logger.(*daisyLog)
+	if ok {
+		ss := strings.SplitAfter(buf.String(), "\n")
+		var str string
+		for i, s := range ss {
+			str += s
+			if i+1 < len(ss) && len(str)+len(ss[i+1]) < 98*1024 {
```
ericdand: Small nit: this could be simpler and more readable with logic like this:

```
if len(str) + len(s) > 98*1024 {
  dl.cloudLogger.Log(...)
  str = s
} else {
  str += s
}
```
No array lookahead, and eliminates the need for a "continue".

daisy/logger.go
adjackura: These are fields that get sent to cloud logging, most don't get used directly in Daisy

PR #404, opened 2018-05-11
Add linux configuration integration tests

gut:
Currently working on:
rhel-6
centos-6
rhel-7
centos-7
debian-8
debian-9

Detected sshd_config mismatch on the following images:
ubuntu-1404-lts
ubuntu-1604-lts
ubuntu-1710
ubuntu-1804-lts
(https://github.com/GoogleCloudPlatform/compute-image-packages/issues/604)

Couldn't test SUSE because of missing gsutils. This will require changes
in bootstrap.py (will be sent in another PR)

### DIFF
--- a/daisy_workflows/image_test/configuration/configuration-test.py
+++ b/daisy_workflows/image_test/configuration/configuration-test.py
@@ -336,9 +336,6 @@ class DebianTests(GenericDistroTests):
     if not self.IsPackageInstalled(package_name):
       raise Exception('%s package is not installed' % package_name)
 
-    # service returns zero if service exists and is running
-    utils.Execute(['service', 'unattended-upgrades', 'status'])
-
     # check unattended upgrade configuration
     command = ['unattended-upgrade', '-v']
     rc, output = utils.Execute(command, capture_output=True)
--- a/daisy_workflows/image_test/configuration/configuration-test.py
+++ b/daisy_workflows/image_test/configuration/configuration-test.py
@@ -1,0 +1,349 @@
+#!/usr/bin/env python2
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


gut:
I signed it!

daisy_workflows/image_test/configuration/configuration-test.py
```diff
@@ -0,0 +1,349 @@
...
+      if actual_sshd_configs[key] != sshd_desired_configs[key]:
+        raise Exception('Sshd key "%s" should be "%s" and not "%s"' % (
+            key, sshd_desired_configs[key], actual_sshd_configs[key]))
+
+  def TestPackageManagerConfig(self):
+    """
+    Ensure apt/yum repos are setup for GCE repos.
+    """
+    # The google-cloud-sdk package should only be found if GCE repo is setup
+    self.EnsurePackageIsInstalled('google-cloud-sdk')
```
gut: actually I see that this is not the best way of verifying it... (rhel-6 has gcloud/gsutil binaries installed by other means than rpm). 
I'll rewrite this check to really check if `packages.cloud.google.com` is used by apt/yum

gut: done

gut:
> I am wondering if we should stub out the distro specific configuration tests into their own files given that there will be many (and more as time moves on) and this will become quite long.

No problem. I'll split those classes to different files.

gut:
Is the failed test prow/presubmit/flake8 a concern? It does not look to me.

zmarano:
No- just ignore that for now. We need to go back and clean up all the rest of the python code at some point. If you see anything in the flake8 results that pertains to the code you are changing, you should fix that however.

gut:
ps: updated PR description. Now redhat/centos 6 and debian 8 is passing

daisy_workflows/image_test/configuration/ubuntu.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python2
```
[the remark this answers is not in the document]
gut: And on the other hand, on rhel-6, there is no Python3.

I thought it would be easier to make a bootstrap script that actually installs python2 if we're already developing all scripts using python2.

It's hard to keep everything compatible, I agree. Please tell me if you have other thoughts.

zmarano:
I'm going to merge this so you can iterate on the other cases.

PR #423, opened 2018-05-23
daisy: upload sources from IncludeWorkflow

helen-fornazier:
The uploadSources function uploads files from the SubWorkflow step but
IncludeWorkflow step was missing.

### DIFF
--- a/daisy/sources.go
+++ b/daisy/sources.go
@@ -180,6 +180,11 @@ func (w *Workflow) uploadSources(ctx context.Context) dErr {
 				return err
 			}
 		}
+		if step.IncludeWorkflow != nil {
+			if err := step.IncludeWorkflow.Workflow.uploadSources(ctx); err != nil {
+				return err
+			}
+		}
 	}
 	return nil
 }


adjackura:
We treat includeworklows differently, they are merged with their parent.

See https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/step_includeworkflow.go#L117

helen-fornazier:
right, I thought that only the namespace would be shared and my tests where not working using IncludeWorkflow, but I can change it to subworkflow.
Sorry about the noise

adjackura:
If you have workflows that you expect to pass and don't I can look into it, I know the subworkflow/includeworkflow sharing system is a bit complicated and we have had bugs in the past

helen-fornazier:
@adjackura No need, I can use SubWorflow.
The workflow pass with SubWorkflow but it doesn't pass if I use IncludeWorkflow for two reasons:
1) It can't find the files to download (I didn't investigate if it should find it thought, I need to take a look)
2) I use the FULLNAME variable to deduct the name of the resources, but FULLNAME is constructed with the name of all parents, maybe it should behave differently when used with IncludeWorkflow (so it matches the suffix of the resources).

PR #425, opened 2018-05-24
image_test: replace FULLNAME by RealName

helen-fornazier:
The RealName of a resource is composed by the name of the parent
workflows, but the genName function truncates it if too big, and it
can do a different truncation per resource, which means that using
FULLNAME variable is not reliable to deduce the RealName of the
resources.

It becomes a problem when we have several subworkflows where the
RealName of the resources are bigger, which breaks usages like the
oslogin test, where we need to pass the RealName of an instance
to another using metadata.


helen-fornazier:
I would suggest that we do something similar to Source Vars, but for resources, so we could have a tag ${RESOURCE:instance-name}
This would be reliable and we could deprecate FULLNAME (not really useful)

adjackura:
Could you create an issue for that suggestion, I agree this would be pretty useful.

PR #428, opened 2018-05-25
Fix copying serial port output from Instance in another Project.

tomlanyon:
Prior to this change, creating an Instance in a Project other than that of the
Workflow results in silently failing to capture the serial console output.


adjackura:
Looks like you need to run gofmt

tomlanyon:
Fixed, thanks

PR #434, opened 2018-06-03
Network integration image test

gut:


gut:
_Note: since google/google-api-python-client#499, the python test that authenticate are actually broken. I'll fix it afterwards for all image_tests, ok?_

daisy_workflows/image_test/network/network-tester.py
```diff
@@ -0,0 +1,95 @@
...
+
+  # Ensure routes are added when enabling IP Forwarding
+  compute = utils.GetCompute(discovery, oauth2client.client.GoogleCredentials)
+
+  # (Negative test)
+  # The testee instance has IP Forward disabled, hence it should not accept the
+  # traffic from a different IP even if the router is sending traffic to it.
+  # This was a test to guarantee that behavior.
+  try:
+    TestForwardingRule(MM(compute, testee), testee)
```
gut: forgot to raise an exception here: the `TestForwardingRule` should not work on this case. Hang on..

gut: done

gut:
rebased PR and adjusted content

adjackura:
I feel like a lot of this logic could and should be moved into the daisy workflow. I don't like the idea of maintaining an additional compute library just for tests when daisy can set all that up. 

For example if we added steps to add forwarding rules, that way we don't need to add all the logic in for cleanup. This may require a test case for the positive case and the negative but thats ok if it means less code.

gut:
> I feel like a lot of this logic could and should be moved into the daisy workflow. I don't like the idea of maintaining an additional compute library just for tests when daisy can set all that up.

That's right. It seems like a bunch of reimplementation, right? There is daisy, the python library, this `utils.py`... I may reanalize that as you suggested.

> For example if we added steps to add forwarding rules, that way we don't need to add all the logic in for cleanup. This may require a test case for the positive case and the negative but thats ok if it means less code.

I can start by implementing this part but I guess we need to reassess what is being done on this python tests.

If that works well, then I'd rather send a new pull request than to update this one, if that's ok for you. If so, please close this without merging.

gut:
@zmarano btw, Is this test wrongly defined?
`Ensure routes are added when enabling IP Forwarding` 

I expect it to be
`Ensure routes are added when enabling ForwardingRule`

right?

gut:
oh, I just noticed some confusing terms:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google_compute_engine/networking/network_daemon.py#L116

this `forwardedIPs` list is actually the list of `ForwardingRules` IPs and not the `canIpForward` network parameter as on: https://cloud.google.com/vpc/docs/using-routes#canipforward

gut:
This PR was not closed so I just updated this one with the ForwardingRule and TargetInstances being created on daisy. It's much better this way, thanks for the idea!

gut:
ops, I completely forgot to extend the go tests. Hang on

gut:
@adjackura the PR is once again ready to be reviewed

daisy/compute/compute.go
```diff
@@ -271,6 +306,24 @@ func (c *client) CreateDisk(project, zone string, d *compute.Disk) error {
 	return nil
 }
 
+func (c *client) CreateForwardingRule(project, region string, ti *compute.ForwardingRule) error {
```
adjackura: fr *compute.ForwardingRule

daisy/compute/compute.go
```diff
@@ -192,21 +203,45 @@ func (c *client) BasePath() string {
 	return c.raw.BasePath
 }
 
-func (c *client) operationsWait(project, zone, name string) error {
+type OperationGetterFunc func() (*compute.Operation, error)
```
adjackura: Does this type need to be public?

daisy/compute/compute.go
adjackura: Also docstring

daisy/compute/compute.go
gut: No. I'll change it to be unexported.

gut:
done. Thanks

PR #435, opened 2018-06-03
Fix oauth2 deprecation

gut:
Fix broken API after google/google-api-python-client#499


gut:
Note: #434 and this PR have a special relationship: After one of them is merged, the other one needs to be updated. Please don't merge them at once as something will be broken.

The motivation for sending this one while #434 is still open is to have this reviewed in other to push things faster. After one of them is merged I'll promptly update the other one.

PR #438, opened 2018-06-07
Multi NIC integration test

gut:
Note: for some reason the internal DNS server (metadata.google.internal)
is not registering the hostname from the secondary NIC so this test only
works if the common subnetwork is setup on the primary NIC.


gut:
rebase done.

adjackura:
It would be better if the tests did not depend on any existing resources to exist (where it makes sense).
For the multinic tests it would be better to create the networks that test needs (daisy should automatically tear them down at the end of the workflow). I can increase network quota as needed

gut:
oh, ok, so then I could add Network creation on daisy. I'll get my hands on it soon.

PR #439, opened 2018-06-08
Add daisy_test_runner tool 

adjackura:
The daisy_test_runner tool runs a test suite defined in a test template and outputs the results as junit xml
Daisy logging libraries we're slightly tweaked to allow efficient logging
Turn the image test workflow into a daisy_test_runner test suite
Add test-runner prowjob that runs daisy_test_runner
Tweak the prowjob wrapper a bit to provide better logging


helen-fornazier:
@adjackura this is great!
Could the required service accounts in the test projects for oslogin tests as described here https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/daisy_workflows/image_test#setup ? It seems this is the current reason why they are not passing.

PR #447, opened 2018-06-15
daisy_test_runner: add project locking logic

adjackura:


cli_tools/daisy_test_runner/main.go
```diff
@@ -351,7 +367,125 @@ type test struct {
...
+		locked = false
+		md, err = getCommonInstanceMetadata(client, project)
+		if err != nil {
+			return nil, err
+		}
+
+		for i, mdi := range md.Items {
+			if mdi != nil && strings.HasPrefix(mdi.Key, prefix) {
+				if isExpired(*mdi.Value) {
+					md.Items = delItem(md.Items, i)
```
ryanwe: minor: this would read a little clearer if you put `locked = false` here instead of flipping it back and forth until it expires.

cli_tools/daisy_test_runner/main.go
```diff
@@ -351,7 +367,125 @@ type test struct {
...
+				} else {
+					locked = true
+					r := rand.Intn(10) + 5
```
ryanwe: minor: it might be helpful to add a small comment. eg. "Sleep for a random interval between 5 and 15 seconds before checking again."

cli_tools/daisy_test_runner/main.go
```diff
@@ -351,7 +367,125 @@ type test struct {
...
+					time.Sleep(time.Duration(r) * time.Second)
+					break
```
ryanwe: Can you move the return statement here?

cli_tools/daisy_test_runner/main.go
adjackura: done

PR #448, opened 2018-06-18
Remove gsutils dependency on python scripts and fix bootstrap.py

gut:


gut:
> Please have them confirm that here in the pull request.

Of course I confirm it's ok for @helen-fornazier to make changes. All users with write-access to collabora-gce/compute-image-tools belong to the same team and are here to help

PR #450, opened 2018-06-19
Metadata script integration test

gut:


gut:
@adjackura Hi, One curious thing I found when running this test on a faulty manner:
```
[metadata-script-integrity-test.test-metadata-script-integrity.test-shutdown-check-log.wait-exec]: 2018-06-19T15:39:18-03:00 WaitForInstancesSignal: Instance "inst-metadata-scripts-integrity-metadata-script-integrit-rdt08": watching serial port 1, SuccessMatch: "449cacffded6c4b0d73cbe97e3521b15", StatusMatch: "TestStatus:".
[metadata-script-integrity-test.test-metadata-script-integrity.test-shutdown-check-log.wait-exec]: 2018-06-19T15:39:20-03:00 WaitForInstancesSignal: Instance "inst-metadata-scripts-integrity-metadata-script-integrit-rdt08" stopped, not waiting for serial output.
[metadata-script-integrity-test.test-metadata-script-integrity.test-shutdown-check-log]: 2018-06-19T15:39:20-03:00 Step "wait-exec" (WaitForInstancesSignal) successfully finished.
```
Does it really makes sense that, if the machine stopped but the step `WaitForInstancesSignal` is running, it should just be ok? In my opinion you should not be expecting from that machine so it should actually return an error on https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/step_wait_for_instances_signal.go#L111 instead of just logging that message.

Are there other use cases where actually this behavior could be dismissed?

adjackura:
The idea was that if the machine is stopped we don't need to wait for
signal anymore (we had not envisioned workflows hat stop then start a
machine)
What we probably can do just keep waiting until that instance is deleted,
simply ignoring errors waiting the instance to start up again. We could
also record last state and if it is stopped just wait until it starts to
resume watching serial output

gut:
Interesting approaches. On my case they would work due to a timeout error. I explain:

The machine was not going to restart. What happened is that there was no SuccessMatch found for the shutdown-script and the machine ended up turning off. Amazingly, instead of having that step failing, it succeeded.

adjackura:
The timeout would be the best way to signal that your step failed. If we
modify the step to not complete on stopped instance (unless that's a signal
specified) then a step timeout failure would occurred.

PR #452, opened 2018-06-19
Fix WaitForInstanceSignal on stopped instance: don't give up!

gut:
If WaitForInstanceSignal has a SerialOutput parameter, then even if the
machine is STOPPED, it should keep trying to grab the SerialOutput (or
until timeout occurs).

This is not the case if the parameter Stopped is used.


gut:
@adjackura follow up of https://github.com/GoogleCloudPlatform/compute-image-tools/pull/450#issuecomment-398536916

gut:
lint failure fixed

PR #453, opened 2018-06-20
daisy e2e: move e2e tests to new test runner

adjackura:
Added object registry to daisy to allow creating images off of GCS objects daisy has created.
Add buffering to test wrapper GCS logger


daisy_integration_tests/README.md
```diff
@@ -0,0 +1,11 @@
+# Daisy E2E Tests
+
+This directory contains end to end tests for Daisy. These tests are run
+periodically against HEAD.
```
ryanwe: Just to clarify, is that `master/HEAD`? Is it possible to run these on demand against a specific branch or PR? Say I have a large change and I'd like to run the e2e tests before merging, is that possible? If so, can you include those instructions here?

PR #455, opened 2018-06-21
Port boot and multi-nic integration tests to windows

gut:
Also added network, metadata-script and multi-nic tests to prow testing.

### DIFF
--- a/test-infra/prow/config.yaml
+++ b/test-infra/prow/config.yaml
@@ -166,3 +166,34 @@ periodics:
     - name: compute-image-tools-test-service-account
       secret:
        secretName: compute-image-tools-test-service-account
+ - name: windows-image-tests
+   interval: 3h
+   agent: kubernetes
+   spec:
+    containers:
+    - image: gcr.io/compute-image-tools-test/test-runner:latest
+      args:
+      - "-out_path=/artifacts/junit.xml"
+      - "-projects=\
+          compute-image-test-pool-001,\
+          compute-image-test-pool-002,\
+          compute-image-test-pool-003"
+      - "-zone=us-central1-c"
+      - "-var:images=\
+          projects/windows-cloud/global/images/family/windows-2016-core"
+      - "-var:alias_ips=\
+          10.128.3.160"
+      - "daisy_workflows/image_test/windows_images.test.gotmpl"
+      env:
+      - name: REPO_OWNER
+        value: GoogleCloudPlatform
+      - name: REPO_NAME
+        value: compute-image-tools
+      volumeMounts:
+      - name: compute-image-tools-test-service-account
+        mountPath: /etc/compute-image-tools-test-service-account
+        readOnly: true
+    volumes:
+    - name: compute-image-tools-test-service-account
+      secret:
+       secretName: compute-image-tools-test-service-account
--- a/test-infra/prow/config.yaml
+++ b/test-infra/prow/config.yaml
@@ -143,6 +143,15 @@ periodics:
           projects/rhel-cloud/global/images/family/rhel-7,\
           projects/freebsd-org-cloud-dev/global/images/family/freebsd-11-2,\
           projects/freebsd-org-cloud-dev/global/images/family/freebsd-12-0-snap"
+      - "-var:alias_ips=\
+          10.128.3.128,\
+          10.128.3.132,\
+          10.128.3.136,\
+          10.128.3.140,\
+          10.128.3.144,\
+          10.128.3.148,\
+          10.128.3.152,\
+          10.128.3.156"
       - "daisy_workflows/image_test/linux_images.test.gotmpl"
       env:
       - name: REPO_OWNER
--- a/cli_tools/daisy_test_runner/main.go
+++ b/cli_tools/daisy_test_runner/main.go
@@ -59,6 +59,7 @@ var (
 	parallelCount = flag.Int("parallel_count", 0, "TestParallelCount")
 
 	funcMap = map[string]interface{}{
+		"zip":     zip,
 		"mkSlice": mkSlice,
 		"mkMap":   mkMap,
 		"split":   strings.Split,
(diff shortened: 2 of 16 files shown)


gut:
@adjackura Hi. Some things to notice:

1. I didn't add the **os-login** and **metadata-ssh** tests on windows on purposes. Please tell me if you disagree.
1. The ip aliases I defined on the `daisy_workflows/image_test/linux_images.test.gotmpl` and `daisy_workflows/image_test/windows_images.test.gotmpl` files should be modified to match the machine's subnetwork. Please advise if they are not correct or if it's on a conflict zone.
1. I hope the change on `common.substituteSourceVars` is something good to have. I kept making the mistake of defining a variable before the script imported as `${SOURCE:...}` and it takes a while to actually look into the metadata and see that it replaced everything with only the source script (I usually assumed it was my script's mistake). I considered throwing an error if the string containing a `${SOURCE:...}` has something else but I ended up improving the functionality to support that and also having multiple SOURCE scripts, if referenced. Ok?

gut:
> Merging #455 into master will increase coverage by 0.03%.
The diff coverage is 100%.

100%? what the hell? that's some strange report...

test-infra/prow/config.yaml
```diff
@@ -143,6 +143,15 @@ periodics:
           projects/rhel-cloud/global/images/family/rhel-7,\
           projects/freebsd-org-cloud-dev/global/images/family/freebsd-11-2,\
           projects/freebsd-org-cloud-dev/global/images/family/freebsd-12-0-snap"
+      - "-var:alias_ips=\
```
adjackura: I'm not a fan of having to specify a list of alias ips as long as the list of images we test, it seems very cumbersome and something we are very likely to mess up and forget when we add new images. Is the reason for these addresses simply so we can run these tests at the same time? If that's the case I'd say it is probably better to add a new type of lock to the test runner. A duplicate test lock or something, I think a new field something like this for these tests:

"CustomProjectLock": "alias-ip"

Setting this would lock the project with this tag so only one of this type of test will run in the project at a time. If this would work for this set of tests and not require this list of ips here I'll work on adding the additional locking functionality

adjackura:
Also, I think we should add a readme in each test folder describing what the test tests, what it does and how it works on linux/windows. The existing readme could probably be chopped up and moved to each specific folder.  

Not needed for this PR though, just something to think about as we add more tests.

test-infra/prow/config.yaml
gut: Well, I don't think that would solve the problem actually: Some IP still needs to be given (even if it's just 1). There are always a lot of IP space on the subnetwork so I didn't think it was a concern.

Additionally, I added a check to guarantee that the `len(images) == len(alias_ips)` ( https://github.com/GoogleCloudPlatform/compute-image-tools/pull/455/files#diff-8c48497c3da51f59e852bb3a7a93cea5R8 , ok, it's not really a check but you will see no test at all if it's inconsistent. That's something...).

Another idea is to change the `images` list to have the mapping of the alias_ip explicitly per line (then I'd rewrite the splitting, but at least it'd be 1 list, which is much harder to make mistakes).

Or do you rather prefer having that lock? I'll have to investigate that and I'd ask you to supply the alias-ip that can be used.

gut:
> Also, I think we should add a readme in each test folder describing what the test tests, what it does and how it works on linux/windows. The existing readme could probably be chopped up and moved to each specific folder.

> Not needed for this PR though, just something to think about as we add more tests.

Good idea. I think the existing image_test/README.md is deprecated already. I'll check that soon

test-infra/prow/config.yaml
adjackura: I still expect us to need to specify at least 1 ip to use, its just the list of IPs that must match the length of the list of images I don't like. Especially if we are running 2 or more test suites with overlapping IP ranges. 

I'm fine with giving a list to use, we could do something fancy and have the lock use the IP as the key so that we could give multiple test suites the same list of IPs (could be more or less than the number of images) and they could all safely run in the same project

gut: ok, so there will be a list of IPs that could be used and a lock mechanism to restraint its availability. Understood. I'll take a close look on implementing that

adjackura: We can add a rand function to the template that takes a list and returns a random entry

adjackura: I can add the lock if you want, you can just modify the PR to assume that CustomProjectLock exists.

Something like:
{{$alias_ip := rand $alias_ips -}}
...
...
"test-network [{{$image}}]": {
  "Path": "./network/network.wf.json",
  "Vars": {
     "source_image": "{{$image}}",
    "alias_ip": "{{$alias_ip}}"
  },
  "CustomProjectLock": "alias-ip-{{$alias_ip}}"
},

gut: alright. Will do then

gut:
Now it's using the CustomProjectLock.

1. Please advise about the IPs to be used on `alias_ips` list
1. Do you want me to clear the "zip" method that's now not being used?

Also, I'm still waiting for you position on this one:
> I didn't add the os-login and metadata-ssh tests on windows on purposes. Please tell me if you disagree.

adjackura:
Yeah go ahead and remove that zip method. You can replace it with rand if you want, should be simple, something like:

func randItem(args ...[]string) string {
rand.Seed(time.Now().UnixNano())
		return args[rand.Intn(len(args)]
}

os-login and ssh tests are not applicable on windows right now so no need to add them

adjackura:
Also I think the correct ip range is 10.128.0.0/20

gut:
randItem: done
ip range: the 10.128.3.128 matches 10.128.0.0/20 so I guess they are safe to use (as IP allocation is normally allocated in a "lowest available" policy, I chose a high one to avoid collisions).

adjackura:
Ok cool, tests may be failing for a bit until i get the locking working and some other stuff squared away with permisions

gut:
great, thanks!

PR #457, opened 2018-06-25
 test runner: fix up some bugs in e2e tests, clean up sources

adjackura:


cli_tools/daisy_test_runner/main.go
```diff
@@ -554,13 +554,16 @@ func runTestCase(ctx context.Context, test *test, tc *junitTestCase, errors chan
 
 	start := time.Now()
 	fmt.Printf("[TestRunner] Running test case %q\n", tc.Name)
-	if err := test.testCase.w.Run(ctx); err != nil {
+	err = test.testCase.w.Run(ctx)
+	tc.Time = time.Since(start).Seconds()
+	tc.SystemOut = test.testCase.logger.buf.String()
```
ryanwe: minor: this seems like a violation of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). Not a big deal but it can lead to code being brittle and having to change anytime you refactor or update the implementation. Consider adding a function so this can be something like `test.testCase.getLogBuffer()` or something. But I also see this is just a refactor so your call.

daisy/workflow.go
```diff
@@ -547,7 +547,7 @@ func (w *Workflow) runStep(ctx context.Context, s *Step) dErr {
 	case err := <-e:
 		return err
 	case <-timeout:
-		return errf("step %q did not stop in specified timeout of %s", s.name, s.timeout)
+		return errf("step %q did not complete in specified timeout of %s", s.name, s.timeout)
```
ryanwe: nit: maybe "did not complete within the specified ..."

test-infra/prowjobs/wrapper/main.go
```diff
@@ -165,16 +165,19 @@ func main() {
 		log.Fatal(err)
 	}
 
-	logFile := bufio.NewWriter(&gcsLogger{client: client, object: "build-log.txt", ctx: ctx})
+	logfile := &gcsLogger{client: client, object: "build-log.txt", ctx: ctx}
+	buffered := bufio.NewWriter(logfile)
```
ryanwe: minor: would `buffer` be clearer?

PR #458, opened 2018-06-26
Improve integration test documentation

gut:


gut:
hang on just a sec: I noticed that the `docs/daisy-workflow-config-spec.md` file can also be improved. It's missing the TargetInstance and ForwardingRule steps.

gut:
> hang on just a sec: I noticed that the docs/daisy-workflow-config-spec.md file can also be improved. It's missing the TargetInstance and ForwardingRule steps.

Done

PR #467, opened 2018-06-29
Improve multi nic test by creating networks on daisy

gut:


gut:
btw, check your quotas: this test will require a lot of subnetworks and networks if they run in parallel.

gut:
great, coverage check is satisfied now :-)

daisy/step_delete_resources.go
```diff
@@ -265,7 +278,22 @@ func (d *DeleteResources) run(ctx context.Context, s *Step) dErr {
 		}(d)
 	}
 
-	// Delete disks after instances.
+	// Delete subnetworks after instances.
```
adjackura: You'll need to have a separate wait after deleting subnetworks, subnetworks can't be deleted until all instances that reference them are deleted (done with the wait above), but then networks can't be deleted until all subnetworks are deleted

gut: done. I took this chance to make it an auxiliary function to avoid triplicating the code

daisy/resource.go
```diff
@@ -47,17 +47,22 @@ type Resource struct {
...
 
 func (r *Resource) populateWithRegion(ctx context.Context, s *Step, name, region string) (string, string, dErr) {
-	errs := r.populateHelper(ctx, s, name)
-	return r.RealName, strOr(region, s.w.Zone[:len(s.w.Zone)-2]), errs
+	errs := r._populateHelper(ctx, s, name)
+	return r.RealName, strOr(region, getRegionFromZone(s.w.Zone)), errs
 }
 
-func (r *Resource) populateHelper(ctx context.Context, s *Step, name string) dErr {
+func (r *Resource) _populateHelper(ctx context.Context, s *Step, name string) dErr {
```
adjackura: Don't start function names with an underscore

daisy/step_delete_resources.go
adjackura: Thanks, I was going to mention something like that but figured we could clean it up later.

daisy/resource.go
gut: done

gut: is there a better way of marking that interface as being private? In python the underscores are normally used for protected and private scope visibilities.

adjackura: lowercase is private, Uppercase is public. Also its actually enforces unlike in python.

PR #468, opened 2018-06-29
Network-test: add ip alias while running ported to windows

gut:


daisy_workflows/image_test/network/network-tester.py
```diff
@@ -39,28 +43,51 @@ def GetHostname(addr, timeout=10):
...
+  info = {
+      'aliasIpRanges': alias_ip_ranges,
+      # a previous fingerprint is needed in order to change the instance
+      'fingerprint': MD.GetInstanceIfaces(instance)[0][u'fingerprint']
+  }
+  MD.Wait(MD.SetInstanceIface(instance, info))
+
+  # The operation is completed but it might not guarantee that the agent has
+  # already applied the rule. For that reason, wait
+  time.sleep(30)
```
gut: I didn't like this but I didn't come up with a better method. Anyone?

daisy_workflows/image_test/network/network-testee.bat
```diff
@@ -0,0 +1,40 @@
+@echo off
```
adjackura: Just make this startup script powershell to begin with.

gut: lol, you're right

adjackura:
Does this test pass on Windows? We don't setup ip aliasing in the Windows agent like we do on Linux

gut:
> Does this test pass on Windows? We don't setup ip aliasing in the Windows agent like we do on Linux

Nope, it doesn't route traffic from the aliased IP (I raised https://github.com/GoogleCloudPlatform/compute-image-windows/issues/144 ). If you want me to silently change that behavior for windows, I can do it, but I was hoping to have it fixed so that it also behaves as Linux does.

adjackura:
Updated that issue with some background. I would just skip this test on Windows for now.

gut:
done.

daisy_workflows/image_test/network/network-tester.py
adjackura: We could watch the serial port, but thats kinda complicated, this sleep is more than enough. If we wanted to clean it up we can later,

PR #470, opened 2018-07-02
Fix disk test on debian/ubuntu

gut:
Improve shell script code to debian/ubuntu default shell acceptance.

Change disk resize test as success/fail case to only reporting status:
Some distros are using GPT partition table so the 2048GB partition
maximum size is not an issue anymore (it could overflow to the beginning
of the disk).


gut:
even though this PR is approved, I had to do more changes to make it work on WIndows so I've integrated these changes on the PR #473 and I'm closing this.

PR #474, opened 2018-07-07
daisy_workflows: use logging instead of wrappers

helen-fornazier:
Standarize logs by setting the formats for every logging level in the
root logger.

As a consequance, this fixes the following bug when calling the
wrappers:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file utils.py, line 51
```

### DIFF
--- a/daisy_workflows/image_build/debian/build.py
+++ b/daisy_workflows/image_build/debian/build.py
@@ -26,6 +26,7 @@ image_dest: The Cloud Storage destination for the resultant image.
 
 import collections
 import json
+import logging
 import os
 import shutil
 import urllib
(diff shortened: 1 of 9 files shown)


daisy_workflows/image_build/debian/build.py
```diff
@@ -62,32 +63,32 @@ def main():
     raise ValueError(
         'Metadata "google_cloud_repo" must be one of %s.' % REPOS)
 
-  utils.LogStatus('Bootstrap_vz manifest: %s' % bvz_manifest)
-  utils.LogStatus('Bootstrap_vz version: %s' % bvz_version)
-  utils.LogStatus('Google Cloud repo: %s' % repo)
+  logging.info('Bootstrap_vz manifest: %s' % bvz_manifest)
```
zmarano: These workflows use these message to report status (https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_build/debian/debian.wf.json#L61)... so all of these should be prefaced with "BuildStatus:" to make that work. Likewise, errors are "BuildFailed:".

helen-fornazier: All the logging.info will be prefixed by "BuildStatus", but thought LogFormatter at  https://github.com/GoogleCloudPlatform/compute-image-tools/pull/474/commits/73f23f4124a399d190be292ca944adba86e422a5#diff-000c28015d35ee8c5a7977072608a5dcR345

zmarano: Oh awesome- sorry I missed that before :)

PR #475, opened 2018-07-11
oslogin-ssh: Use google api python library instead of gcloud

gut:


gut:
@adjackura do you agree with https://github.com/GoogleCloudPlatform/compute-image-tools/pull/475/commits/5b62de831cf83d935b290ad347dc37f511a11862 implementation? . I hoped to see older keys being removed on https://console.cloud.google.com/compute/metadata/sshKeys due to this expiration, but I don't see the old ones being removed.

gut:
At codecov : are you serious?

adjackura:
No one will remove the metadata keys for you, they are just metadata entries, the expiry only applies to the agent ingesting the keys. The test should remove any metadata keys it sets up.

If your referring to Max's comment, I believe he was speaking about the oslogin api. You set keys in the api, it has nothing to do with metadata IIUC.

Also don't worry so much about codecov :p

gut:
> The test should remove any metadata keys it sets up.

They are removed [here](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/475/files#diff-bbc3b52fb44811403200e3df6e7aa9f8R229)

> f your referring to Max's comment, I believe he was speaking about the oslogin api. You set keys in the api, it has nothing to do with metadata IIUC.

yes, I was thinking on the oslogin api but I got confused with the metadata. No worries then :-). Thanks

PR #476, opened 2018-07-11
Daisy: Add resize and detach disk steps

gut:


daisy/step_resize_disks.go
```diff
@@ -0,0 +1,98 @@
...
+	"sync"
+
+	"google.golang.org/api/compute/v1"
+)
+
+// ResizeDisks is a Daisy ResizeDisks workflow step.
+type ResizeDisks []*resizeDisk
+
+type resizeDisk struct {
+	Disk
```
adjackura: resizeDisk should be public, also just take disk name and DisksResizeRequest, (something like you have with DetachDisks):

type ResizeDisk struct {
Disk string

DisksResizeRequest *compute.DisksResizeRequest
}

validate should then use regUse

It's a bit wordy and complicated but allows for the API to add new fields to DisksResizeRequest and daisy just works with them.

gut: oh, I missed that one. Sorry. Just a sec...

gut: I chose to not add the DiskResizeRequest as member but rather to inherit from it, ok? By doing that, the definition of the resize disk step seems more pleasant to see:
```json
"resizes": {
  "ResizeDisks": [
    {
      "Name": "disk1",
      "SizeGb": "128"
    }
  ]
}
```

daisy/step_resize_disks.go
```diff
@@ -0,0 +1,100 @@
...
+		dr, err := s.w.disks.regUse(rd.Name, s)
+		if dr == nil {
+			// Return now, the rest of this function can't be run without dr.
+			return addErrs(errs, errf("cannot resize disk: %v", err))
+		}
+		// Reference the actual name of the disk
+		rd.Name = dr.RealName
+
+		pre := fmt.Sprintf("cannot resize disk %q", rd.Name)
+		if rd.SizeGb <= 0 {
```
adjackura: The API wont allow you to decrease the size of the disk either. We should add a check for that as well here. Feel free to just add a TODO comment for this.

daisy/step_resize_disks.go
```diff
@@ -0,0 +1,100 @@
...
+
+// ResizeDisks is a Daisy ResizeDisks workflow step.
+type ResizeDisks []*ResizeDisk
+
+// ResizeDisk is used to resize a GCE disk.
+type ResizeDisk struct {
+	compute.DisksResizeRequest
+	// Name of the disk to be resized
+	Name          string
+	Project, Zone string
```
adjackura: I think all you should have here is Name, or Disk (see start/stop instance step and deleteresources)
I think what we've settled on is to allow either the daisy name or the uri of the resource.

daisy/step_resize_disks.go
gut: can we check the current disk size on `validate`? I'm afraid the disk information would not be there (if you give an image, you don't know how big it would be at that point).

I'm considering doing that on `run`, ok?

adjackura: Doing it in run is a bit pointless as at that point you can just make the API call. Having this in populate would be a bit tough without recording info on disk sizes in the disk registry

PR #477, opened 2018-07-12
Fix metadata-script variable naming

gut:


gut:
@adjackura sorry for this small typo. Here is the fix

PR #484, opened 2018-07-20
Implement debian image build via FAI tool

gut:
This workflow is separated from the ones of bootstrap-vz because debian-cloud's
scripts don't support jessie.

Below the deviations from the generated image and the GCE official jessie image:
* sshd's PermitRootLogin is the default prohibit-password instead of desired NO
* NTP is not installed (so it's not configured to use metadata.google.internal server)
* cmdline arguments:
  * no biosdevname=0
  * no scsi_mod.use_blk_mq=Y
  * console=ttyS0,115200 instead of console=ttyS0,38400n8


gut:
Also note that buster is not working at the moment due to missing oslogin package on google repository. Whenever that dependency is solved, the scripts should work for creating a buster image.

daisy_workflows/linux_common/utils.py
```diff
@@ -65,6 +65,11 @@ def PipInstall(package_list):
   return Execute(['pip', 'install', '-U'] + package_list)
 
 
+def DpkgInstall(package_list):
+  """Install a package from .deb file via dpkg."""
+  return Execute(['dpkg', '-i'] + package_list)
```
zmarano: Would it be better to use apt-get here... apt-get install ./package.deb so that dependencies are taken into account.

daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -0,0 +1,118 @@
...
+      'debian_version', raise_on_not_found=True)
+  image_dest = utils.GetMetadataAttribute('image_dest',
+      raise_on_not_found=True)
+  outs_path = utils.GetMetadataAttribute('daisy-outs-path',
+      raise_on_not_found=True)
+
+  logging.info('fai-cloud-images version: %s' % fai_cloud_images_version)
+  logging.info('debian version: %s' % debian_version)
+
+  # First, update fai-client as the older version on stretch will not work
```
zmarano: I think there was an upstream version in a deb repo here: https://fai-project.org/download/

gut: oh, look at that! Great idea! I'll use that

daisy_workflows/linux_common/utils.py
zmarano: True- even better :)

gut:
just wait a sec, the output image is not booting...

gut:
there you go

gut:
I guess flake8 is wrongly flagging this PR, right?

gut:
btw, I hope to fix those deviations directly on debian-cloud's scripts so there would be no change here.

gut:
note: approved but not merged

PR #487, opened 2018-07-24
Fix oslogin test keys' lifetime to 30 minutes

gut:


daisy_workflows/image_test/oslogin-ssh/oslogin_slave_tester.sh
```diff
@@ -35,32 +35,34 @@ exec {fd_stdout}>&1
...
-    if [ ! -f \${KEY}.pub ]; then
-      ssh-keygen -t rsa -N '' -f daisy-key -C "\$(uname -n)"
+    if [ ! -f ${KEY}.pub ]; then
+      ssh-keygen -t rsa -N '' -f daisy-key -C "$(uname -n)"
     fi
-    gcloud compute os-login ssh-keys add --key-file=\${KEY}.pub
-    cat \${KEY}.pub >&"\$fd_stdout"
+    cat ${KEY}.pub >&"$fd_stdout"
+    NKEYS=$(gcloud compute os-login ssh-keys list | grep -v FINGERPRINT | wc -l)
+    echo "Total number of keys currently on os-login: $NKEYS"
```
gut: I added this echo to be able to check if os-login's keys is a crowded place or not. Specially if this number grows after a while

gut:
I'm not expecting that this change will solve the current errors of `TestFailed: error: SSH command failed when expected to succeed` when adding keys through `gcloud compute os-login ssh-keys add --key-file=daisy-key.pub`, which yields `ERROR: (gcloud.compute.os-login.ssh-keys.add) INVALID_ARGUMENT: Request contains an invalid argument.`, however, If it's because os-login keys repository is crowded, it'll be easily spotted now.

PR #488, opened 2018-07-24
gce_inventory_agent: add debian packages

adjackura:


cli_tools/gce_inventory_agent/debian/changelog
```diff
@@ -0,0 +1,5 @@
+google-compute-engine-inventory (1.1.0) stable; urgency=low
```
zmarano: Version should follow the standard Debian format major.minor.minor-release: 1.1.0-1

PR #489, opened 2018-07-25
Super basic placeholder of osconfig agent

adjackura:


package_library/packages_windows.go
```diff
@@ -182,6 +169,162 @@ func wuaUpdates(query string) ([]PkgInfo, error) {
...
 
+// InstallWUAUpdates installs all WUA updates that match query.
+func InstallWUAUpdates(query string) error {
+	connection := &ole.Connection{Object: nil}
+	if err := connection.Initialize(); err != nil {
+		return err
+	}
+	defer connection.Uninitialize()
+
+	unknown, err := oleutil.CreateObject("Microsoft.Update.Session")
```
ryanwe: `unknown` seems weird. Could it be something like `oleSessionObj` instead?

package_library/packages_windows.go
```diff
@@ -182,6 +169,162 @@ func wuaUpdates(query string) ([]PkgInfo, error) {
...
+	defer updtsRaw.Clear()
+
+	updts := updtsRaw.ToIDispatch()
+	count, err := updts.GetProperty("Count")
+	if err != nil {
+		return err
+	}
+	defer count.Clear()
+
+	updtCnt, _ := count.Value().(int32)
```
ryanwe: super minor: if this `updtCnt` is the value you care about, it would probably read a little better to call `count` `countRaw` and then `updtCnt` can just be count. Or cleaner still, you can factor this out and call `count,err := getUpdateCount(updts)`

package_library/packages_windows.go
```diff
@@ -182,6 +169,162 @@ func wuaUpdates(query string) ([]PkgInfo, error) {
...
+	if updtCnt == 0 {
+		logger.Infof("No updates to install")
+		return nil
+	}
+
+	logger.Infof("%d Windows updates available", updtCnt)
+
+	var msg string
+	for i := 0; i < int(updtCnt); i++ {
+		if err := func() error {
```
ryanwe: Just curious: what does this immediately invoked anonymous function get you? Is it just the per-iteration `defer`als? If so, would it be cleaner to factor that into a separate function call?

package_library/packages_windows.go
```diff
@@ -182,6 +169,162 @@ func wuaUpdates(query string) ([]PkgInfo, error) {
...
+	}
+	installer := installerRaw.ToIDispatch()
+	defer installerRaw.Clear()
+
+	if _, err := installer.PutProperty("Updates", updates); err != nil {
+		return fmt.Errorf("error calling PutProperty Updates on IUpdateInstaller: %v", err)
+	}
+
+	logger.Infof("Installing updates")
+	if _, err := installer.CallMethod("Install"); err != nil {
```
ryanwe: Does this return an object that can be used to track progress or logs? If so, it might be worth logging that value for now to show that we have it and will use it. It seems like that could be valuable. The same for `Download`.

adjackura: It doesn't return anything as this is the synchronous method (that empty val is indeed empty), there is an async method we could use to track progress, it uses callback functions that I really didn't want to get into implementing right now. Added TODO as this could be worth spending time on in the future

PR #490, opened 2018-07-25
Improve tests stability

gut:


gut:
Will this PR have its commits squashed and no information copied on its merge body? If so I'll copy each commit description to the PR body.

adjackura:
Yeah, explicitly add the firewalls you need, some of the projects may have the defaults but I wouldn't count on it.

Also I can do a rebase and merge if you want to keep each commit, but you'll have to take care of squashing any cleanup commits

gut:
> Also I can do a rebase and merge if you want to keep each commit, but you'll have to take care of squashing any cleanup commits

This time I would recommend keeping the history of this changes similarly of how this one got inside: https://github.com/GoogleCloudPlatform/compute-image-tools/commit/83bc03b3f2214d30250b343f9cdc8315cc4afccc

After I noticed that the changes were logged, then I stopped sending incremental commits after reviews and always squashed them to remove the fixups. I don't know what is usually happening on merging but it didn't seem to be standardized.

gut:
> Yeah, explicitly add the firewalls you need

Done

PR #494, opened 2018-07-28
Auto expand disks when importing them

gut:


gut:
I detected this shortcoming on https://cloud-images.ubuntu.com/releases/16.04/release/ubuntu-16.04-server-cloudimg-amd64.tar.gz, which had a minimal disk with virtually no space left. After the regular importing, the file system wasn't enlarged so translating it failed (there is some apt-get work which needs space).

gut:
ops. Just noticed that the https://cloud-images.ubuntu.com/releases/16.04/release/ubuntu-16.04-server-cloudimg-amd64.tar.gz is a "Cloud Image/EC2 tarball" variant, which in the end is not what these scripts are for (image didn't boot despite successfully running the whole import_and_translate.wf.json). If I use the https://cloud-images.ubuntu.com/releases/16.04/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img "QCOW2 disk image file for use with QEMU and KVM", there is no need to worry about partition enlargement as it already grows to disk size without this patch.

I guess I can close this one as it's not a bug / unwanted feature.

PR #495, opened 2018-07-31
image_tests: add firewall rule to allow ssh communication

gut:


gut:
@adjackura Hi, could you please review this one?

PR #496, opened 2018-08-02
Add suse image import

gut:


gut:
Notes:
* Leap 42: Somehow the systemd is not running the google-startup-scripts.service even after enabling them by hand. I have to take a closer look.
* Leap 15: There is no google-cloud-sdk (so no gsutil). For this reason the startup-scripts can't run if they have to download something from a bucket.

gut:
@zmarano Hi, can you please review this one as well?

PR #502, opened 2018-08-09
prow: Remove deprecated Debian 8 from linux image list

gut:


gut:
official docs already reflect this: (no debian-8 found) https://cloud.google.com/compute/docs/images#images_without_shielded_vm_name_support

PR #521, opened 2018-08-22
Configuration: check if google server is being used on NTP client

gut:
Before it was only checking the configuration but it's much more
reliable to verify if the client is actually using it as the client can
be configured from DHCP, overriding the actual configuration

### DIFF
--- a/daisy_workflows/image_test/configuration/linux/generic_distro.py
+++ b/daisy_workflows/image_test/configuration/linux/generic_distro.py
@@ -204,10 +204,41 @@ class GenericDistroTests(object):
     """
     Ensure that the NTP server is set to metadata.google.internal.
     """
-    # Below, not the most pythonic thing to do... but it's the easiest one
-    command = ['grep', '^server \+metadata.google.internal', '/etc/ntp.conf']
-    utils.Execute(command)
-    # if the above command returned any lines, then it found a match
+    def CheckNtpRun(cmd):
+      """
+      Run @cmd and check, if successful, whether google server is found on
+      output.
+
+      Args:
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_test/configuration/linux/generic_distro.py
```diff
@@ -204,10 +204,41 @@ def TestNTPConfig(self):
...
+
+      Return value:
+        bool. True if client exists and google server is found. False
+        otherwise.
+      """
+      try:
+        rc, out = utils.Execute(cmd, raise_errors=False, capture_output=True)
+        if rc == 0:
+          # ntp client found on system
+          if out.find('metadata.google') >= 0:
```
zmarano: metadata.google.internal is what we should look for (for instance, metadata.google.com would not be right)

gut: I didn't do it because ntpq does not report that (I've got to verify if the NTP server given from DNS is actually returning something else (unlikely) or if simply ntpq's output is trimmed (likely)):
debian-9:
```
$ ntpq -p
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
*metadata.google 71.79.79.71      2 u  180 1024  377    0.498   -0.023   0.050
```
RHEL-7:
```
$ chronyc sources
210 Number of sources = 1
MS Name/IP address         Stratum Poll Reach LastRx Last sample               
===============================================================================
^* metadata.google.internal      2   9   377   408    +25us[  +48us] +/-  796us
```

zmarano: Ahh gotcha.

PR #523, opened 2018-08-23
Fix Debian FAI build.

zmarano:


gut:
This PR is deprecated: there is another change needed for using FAI tool that is renaming the repository name from fai-cloud-images to debian-cloud-images (the tarball will be extracted with an unexpected folder name otherwise). I recommend closing this issue in favor of #533 , which solve both problems and implement UEFI arguments.

PR #524, opened 2018-08-29
Remove code related to Debian 8 (jessie)

gut:
Debian 8 reached end of life and the public image was already removed.
I updated the scripts to not rely on that image while keeping import
and build script for supporting the community which still uses it.


zmarano:
For now, lets leave the Debian 8 build and translate code around. We still need to support it as a working OS on GCE even if it is no longer distributed.

gut:
I agree for for translating, but not building. Either way, it's your call.
Please note the [debian 8 build script uses a debian 8 public image](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/524/commits/2e44515da486a7ad61c2e60286d57ddd758f0c68#diff-ad401eea08f4c799e7c91fa9305361e5R19), that's broken. If you say it's safe to just use stretch instead, I can comfortably change that.

zmarano:
I'll deal with the Debian 8 build cleanup later on- we can still build it if necessary using the deprecated image.

PR #525, opened 2018-08-29
 Separate metadata script tests, add windows specialize scripts

adjackura:
Cleanup tests a bit to make them more efficient 
Drop Linux test number to keep under quotas
Make it possible to create an instance with no scopes attached
Fix disk auto deletion validation


gut:
@adjackura Hi, I just missed the update of this test documentation, namely the https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_test/metadata-script/README.md . 

A short glimpse over the changes made me feel like nothing changed functionally besides adding sysprep tests on windows, but you'd be the best to comment it.

adjackura:
Yeah, no functional changes, just addition of the sysprep test (just another form of startup scripts). The test failures I believe are actually linux guest environment issues. I actually made it possible for daisy to create an instance with no scopes, so the unauthenticated gs:// url download started failing on linux as the script runner doesnt fall back as it should:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google_compute_engine/metadata_scripts/script_retriever.py#L114

PR #528, opened 2018-08-30
Solve quota issues for windows too

gut:
See #526


gut:
@adjackura sorry. I just realized I missed windows but the original PR was merged.

adjackura:
Also note I have a PR that reduces the number of concurrent Linux tests to also help with quota. We can fine tune it more if we have test overruns

PR #537, opened 2018-09-06
Fix FAI script reference for building buster images

gut:
Now it points to the correct buster repository.


gut:
@zmarano that's the commit from your accepted merge request on debian-cloud-images

PR #538, opened 2018-09-07
Move all image_tests out of daisy_workflow directory

gut:


gut:
Hi @adjackura . I couldn't really test how prow will behave with this. Please check.

adjackura:
Looks ok, will watch the next few runs closely to make sure nothing is messed up

PR #539, opened 2018-09-08
Fix linux_common location due to #538 change

gut:


gut:
@adjackura This should fix test-grid issues.

PR #540, opened 2018-09-08
Fix configuration's gcloud test: OSError was raised when not found

gut:
I guess this issue has not been noticed before as all distros that we
currently test have a gcloud/gsutil command (google-cloud-sdk package).
But that's not a requirement so without this patch, it'd as an
unexpected exception would be thrown.

This was possibly not happening before when subprocess.check_call or
subprocess.check_output was being used on utils.Execute.


gut:
Those unittests check is broken, isn't it?

PR #542, opened 2018-09-11
 Daisy: Fix instance deletion before its creation

gut:


gut:
This one fix the root cause that led to #541 keeping instances on the project when it ran.

daisy/instance.go
```diff
@@ -371,13 +372,25 @@ func newInstanceRegistry(w *Workflow) *instanceRegistry {
...
 
+// Sleep function is mocked on testing.
+var SleepFn = time.Sleep
+
 func (ir *instanceRegistry) deleteFn(res *Resource) dErr {
 	m := namedSubexp(instanceURLRgx, res.link)
-	err := ir.w.ComputeClient.DeleteInstance(m["project"], m["zone"], m["instance"])
-	if gErr, ok := err.(*googleapi.Error); ok && gErr.Code == http.StatusNotFound {
-		return typedErr(resourceDNEError, err)
+	for {
```
adjackura: If the instance is deleted outside of the workflow then this will just spin forever, Is this because of a race condition between a create step, a failure, and a cleanup?

gut: > If the instance is deleted outside of the workflow then this will just spin forever

A daisy instance removed by someone else or on a different daisy step? Is that a concern?

> Is this because of a race condition between a create step, a failure, and a cleanup?

Yes. The creation step takes a while, and while it's creating, the CopyGCSObject fails and triggers the cleanup. The cleanup runs this instance deletion code but despite finding information about the instance in the internal registry, the googleapi will not find the instance for removing (the creation didn't finish, the instance is not even on PROVISIONING state!). That was why I implemented this wait mechanism.

gut:
sorry @daisy-bot , I can't solve gocheck and unittests checks. My PR is unrelated to those.

daisy/instance.go
adjackura: Deleting resources is definitely a thing that can happen, I would be ok with just a limit on retries here for now, maybe later we can come up with a better state management system for active compute calls.

daisy/instance.go
```diff
@@ -371,13 +373,26 @@ func newInstanceRegistry(w *Workflow) *instanceRegistry {
...
-	err := ir.w.ComputeClient.DeleteInstance(m["project"], m["zone"], m["instance"])
-	if gErr, ok := err.(*googleapi.Error); ok && gErr.Code == http.StatusNotFound {
-		return typedErr(resourceDNEError, err)
+	for i := 1; i < 4; i++ {
+		if _, err := ir.w.ComputeClient.GetInstance(m["project"], m["zone"], m["instance"]); err != nil {
+			// Can't remove an instance that was not even yet created!
+			// However as the command was already submitted, wait.
+			SleepFn((time.Duration(rand.Intn(1000))*time.Millisecond + 1*time.Second) * time.Duration(i))
+			continue
+		}
```
adjackura: Just end the for loop here, always end the function with a DeleteInstance call and whatever error may result

adjackura:
Ignore the gocheck error, I should have fixed that before, I'll have to take another look.

adjackura:
Looks like the tests are failing because of a new URL option in the API, I'll have to fix that separately.

PR #545, opened 2018-09-14
Use latest FreeBSD 11.2 images on prow tests

gut:
The Google Compute Agent was updated on FreeBSD 11.2 but prow was only testing
the release image. Tests should be focusing on the latest builds.


gut:
E.g: with freebsd-11-2-snap instead of freebsd-11-2, I see the network and oslogin-ssh tests passing

PR #547, opened 2018-09-14
Update agent with patch runner logic

adjackura:
Sync in latest autogen library (ignore the first three files)
Send PatchPolicies to patchManager

Patches get registered then sleep until they are in a patch window
The then get sent on a patch channel to the patchrunner which will run them one at a time
Basic logic for saving state and continuing on reboot is implemented as well


cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
 type patchWindow struct {
-	start, end time.Time
+	Name   string
+	Policy *patchPolicy
+
+	Start, End time.Time
+
+	t      *time.Timer
+	cancel chan struct{}
+	mx     sync.RWMutex
```
ryanwe: Seeing a mutext on the patch window has me wondering. If every patch window has its own mutex, what do they use to avoid conflicts with each other? Also, if every patch window needs a mutex, that seems like a lot of possible contention and threads/agents that it might be hard to avoid race condition bugs. I still wonder if it's not possible to reduce the amount of concurrency to simplify what is going on.

I'm still not up to speed with Go's concurrency semantics, but in a traditional threaded model, I'd see this agent having 2 threads - 1 for periodically calling the service and 1 for running OsConfig and PatchPolicy updates. Yes, there's some lock contention between the two, but for the most part, they wait and operate independently. And the shared state can be minimized and accessed in a small and controlled way.

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
 func (w *patchWindow) in() bool {
+	select {
+	case <-w.cancel:
+		return false
+	default:
+	}
+
 	now := time.Now()
-	return now.After(w.start) && now.Before(w.end)
+	return now.After(w.Start) && now.Before(w.End)
```
ryanwe: very minor: you could also check if `now == w.Start`.

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+
+func patchManager(efps []*osconfigpb.LookupConfigsResponse_EffectivePatchPolicy) {
+	// Deregister any existing patchWindows that no longer exist.
+	var ppns []string
+	for _, pp := range efps {
+		ppns = append(ppns, pp.GetFullName())
+	}
+
+	var toDeregister []*patchWindow
+	aw.mx.Lock()
```
ryanwe: shouldn't the Unlock() be deferred here in case there's a panic or the code is refactored to return early?

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+	for _, w := range aw.windows {
+		if !strIn(w.Name, ppns) {
+			toDeregister = append(toDeregister, w)
+		}
+	}
+
+	fmt.Printf("DEBUG: list of patchWindows to deregister: %q\n", toDeregister)
+	for _, w := range toDeregister {
+		defer w.deregister()
```
ryanwe: why is this, and the other actions below, deferred?

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+	w.t = time.AfterFunc(w.Start.Sub(time.Now()), func() { rc <- w })
+	fmt.Println(w.Name, "register done")
+}
+
+// deregister stops a patch windows timer and removes it from activeWindows
+// but does not cancel any current runs.
+func (w *patchWindow) deregister() {
+	fmt.Println(w.Name, "deregister")
+	w.t.Stop()
+	aw.delete(w.Name)
```
ryanwe: It seems unconventional that an item in a list, deletes itself from that list. Maybe this is something that you're going to change or maybe there's a good reason for it. But it's different from what I'd usually expect to see and can be confusing when following the code.

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+
+		w, err := newPatchWindow(pp)
+		if err != nil {
+			log.Print("ERROR:", err)
+			continue
+		}
+		fmt.Println("DEBUG: patchWindow to create:", w.Name)
+		defer w.register()
+	}
+	aw.mx.Unlock()
```
ryanwe: Maybe one way to simplify the locking is to look at it like this: if the agent is running a patch or updating the OsConfig, it doesn't need to call the service or update `aw.windows`. And if it's in the process of updating `aw.windows`, then no patch should start. Not sure if that helps, but it's a mental model that might simplify things.

osinfo/osinfo_linux.go
```diff
@@ -78,7 +78,7 @@ func parseOsRelease(path string) (*DistributionInfo, error) {
 func parseEnterpriseRelease(path string) (*DistributionInfo, error) {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
-		return &DistributionInfo{ShortName: "linux"}, fmt.Errorf("unable to obtain release info: %v", err)
+		return &DistributionInfo{ShortName: Linux}, fmt.Errorf("unable to obtain release info: %v", err)
```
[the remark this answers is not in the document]
adjackura: This is part of the OS, if this is linux it defaults to linux, the *_linux.go files only get compiled for linux, same for *_windows.go.

cli_tools/osconfig_agent/patch_window.go
adjackura: Usually yes, a panic exist so that's not an issue but early returns are. I specifically don't defer here as I need to unlock before function end, and cars is taken to not have anything return early before then.

cli_tools/osconfig_agent/patch_window.go
adjackura: So they run after the lock is no longer held.

cli_tools/osconfig_agent/patch_window.go
adjackura: Ack, this is probably way more complicated than it needs to be, a single lock probably can take care of all this as no two things need to happen at once. I'll look at that in the next PR

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+}
+
+func loadState(state string) (*patchWindow, error) {
```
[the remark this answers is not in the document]
adjackura: If you mean file, then yes I can do that

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+
+	w.mx.RLock()
+	defer w.mx.RUnlock()
+
+	d, err := json.Marshal(w)
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(state, d, 0600)
```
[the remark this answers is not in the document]
adjackura: Technically that would require assumptions of the underlying file system or OS I don't want to make. I also don't want to have to force a sync before rename or anything and the cross platform nature of the agent makes things even harder, i'm pretty sure I'd have to use MoveFileEx on Windows. 

I don't think this is a huge issue though as if there is we don't really care about the previous contents of the file as any time we want to write the previous contents are no longer relevant.,

package_library/packages.go
```diff
@@ -20,9 +20,27 @@ import (
...
+	// AptExists indicates whether apt is installed.
+	AptExists bool
+	// YumExists indicates whether yum is installed.
+	YumExists bool
+	// ZypperExists indicates whether zypper is installed.
+	ZypperExists bool
+	// GemExists indicates whether gem is installed.
+	GemExists bool
+	// PipExists indicates whether pip is installed.
+	PipExists bool
```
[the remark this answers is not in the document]
adjackura: Note the package, this isnt part of osconfig agent

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -15,26 +15,306 @@
...
+	// Deregister any existing patchWindows that no longer exist.
+	var ppns []string
+	for _, pp := range efps {
+		ppns = append(ppns, pp.GetFullName())
+	}
+
+	var toDeregister []*patchWindow
+	aw.mx.Lock()
+	for _, w := range aw.windows {
+		if !strIn(w.Name, ppns) {
```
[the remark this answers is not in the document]
adjackura: Ack, will address in the next PR

cli_tools/osconfig_agent/osconfig.go
```diff
@@ -16,20 +16,64 @@ package main
...
+	}
+	if packages.YumExists {
+		configTypes = append(configTypes, osconfigpb.LookupConfigsRequest_YUM)
+	}
+	if packages.GooGetExists {
+		configTypes = append(configTypes, osconfigpb.LookupConfigsRequest_GOO)
+	}
+	if info.ShortName == osinfo.Windows {
+		configTypes = append(configTypes, osconfigpb.LookupConfigsRequest_WINDOWS_UPDATE)
+	}
```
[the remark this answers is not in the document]
adjackura: Ack, will address in the next PR

cli_tools/osconfig_agent/patch_window.go
adjackura: Added todo to look at this once the logic is more complete

PR #549, opened 2018-09-19
Add compute-image-package build and test workflow for debian-9

gut:


adjackura:
I'm still working out how this will best serve us. Right now I see two scenarios we want to test:

Test prerelease packages, this PR doesnt help much there and I think this is the bigger thing we would like to test, we do this well with windows right now but our linux build process needs to be updated to handle this. Auto building packages is the best way to handle this, then we just need to build packages updates into the existing tests via a flag. Or autobuild images based on prerelease packages, this is probably much simpler from a test perspective.

Test packages based on a specific PR or repo, this is what I see this PR focusing on. In this case I don't think these tests need to run automatically at all, we just need a simple command line to run

helen-fornazier:
Hi @zmarano @adjackura,

As we discussed before, the idea here is to have a workflow where we can test a given PR for compute-image-packages, so we can add a CI in compute-image-packages to trigger this when a PR happens.
It will be hard to test a PR against every distro, but at least against Debian is a start, What do you think?

PR #553, opened 2018-09-26
 Add Windows packaging and service setup os osconfig_agent

adjackura:
Updated proto definitions


cli_tools/osconfig_agent/main.go
```diff
@@ -33,6 +40,58 @@ var (
...
+	// Retry forever, increase sleep between retries (up to 5 times) in order
+	// to wait for slow network initialization.
+	var rt time.Duration
+	for i := 1; ; i++ {
+		res, err = client.Do(req)
+		if err == nil {
+			break
+		}
+		if i < 6 {
+			rt = time.Duration(3*i) * time.Second
```
ryanwe: minor: you can simplify this a little by doing something like `rt = time.Duration(math.Min(3 * i, MAX_RETRY_DELAY)) * time.Second`. Then you can remove the `if i < 6` check and you don't need to track the state of `rt` between iterations to understand how long the delay is. It will also make it easier to extend in the future to add exponential backoff and jitter.

cli_tools/osconfig_agent/main.go
```diff
@@ -33,6 +40,58 @@ var (
...
+		}
+		logger.Errorf("error connecting to metadata server, retrying in %s, error: %v", rt, err)
```
ryanwe: minor: it would probably also be helpful to add `i`, the retry count to the error log.

cli_tools/osconfig_agent/main.go
```diff
@@ -42,22 +101,42 @@ func strIn(s string, ss []string) bool {
...
+		res, err := lookupConfigs(ctx, client, res)
+		if err != nil {
+			log.Println("ERROR:", err)
+		} else {
+			runPackageConfig(res)
+			patchManager(res.PatchPolicies)
+		}
+
+		select {
+		case <-ticker.C:
```
ryanwe: Just curious, why do you use a `Ticker` here but `Sleep` for retrying the resource lookup?

cli_tools/osconfig_agent/main.go
```diff
@@ -42,22 +101,42 @@ func strIn(s string, ss []string) bool {
...
-
+func run(ctx context.Context) {
 	client, err := osconfig.NewClient(ctx, option.WithEndpoint(*endpoint), option.WithCredentialsFile(*oauth))
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalln("NewClient Error:", err)
 	}
 
-	res, err := lookupConfigs(ctx, client, *resource)
+	res, err := getResource(*resource)
```
ryanwe: This is for lookup up the instance info right? If so, should it be called that instead?

cli_tools/osconfig_agent/main.go
```diff
@@ -42,22 +101,42 @@ func strIn(s string, ss []string) bool {
...
-		log.Fatal(err)
+		log.Fatalln("getResource error:", err)
 	}
-	patchManager(res.PatchPolicies)
 
-	//runUpdates()
+	patchInit()
+	ticker := time.NewTicker(interval)
+	for {
+		res, err := lookupConfigs(ctx, client, res)
```
ryanwe: You you mean to reuse `res` here? Does `lookupConfigs()` and `getResource()` return the same thing?

cli_tools/osconfig_agent/main.go
```diff
@@ -42,22 +101,42 @@ func strIn(s string, ss []string) bool {
...
+		if err != nil {
+			log.Println("ERROR:", err)
+		} else {
+			runPackageConfig(res)
```
ryanwe: `PackageConfig` isn't really the right name here is it? This call will be setting registry settings as well. And maybe more in the future. How about something like `setOsConfig()`? 

Also, we might want to tweak our api so this would be something like `setOsConfig(res.osConfig)`. But I don't think we need to worry about that right now.

cli_tools/osconfig_agent/main.go
```diff
@@ -42,22 +101,42 @@ func strIn(s string, ss []string) bool {
...
+			patchManager(res.PatchPolicies)
```
ryanwe: also for consistency, this should probably be `runPatchManager()` or `setPatchPolicies()` depending on what you go with for the PackageConfig.

cli_tools/osconfig_agent/packages_windows.go
```diff
@@ -16,60 +16,76 @@ package main
...
+	buf.WriteString("# Repo file managed by Google OSConfig agent\n")
 	for _, repo := range repos {
 		buf.WriteString(fmt.Sprintf("\n- name: %s\n", repo.Name))
 		buf.WriteString(fmt.Sprintf("  url: %s\n", repo.Url))
 	}
-	if err := ioutil.WriteFile(googetRepoFile, buf.Bytes(), 0600); err != nil {
-		log.Printf("Error writing GooGet repo file: %v", err)
-	}
+
+	return ioutil.WriteFile(googetRepoFile, buf.Bytes(), 0600)
```
ryanwe: minor: for FileMode, consider a const like `OWNER_READ_WRITE = 0600`

cli_tools/osconfig_agent/patch_window.go
```diff
@@ -274,7 +274,9 @@ func patchRunner() {
 			reboot := pw.run()
 			if reboot {
 				fmt.Println("DEBUG: reboot requested", pw.Name)
-				// TODO: Actually reboot.
+				if err := rebootSystem(); err != nil {
+					fmt.Println("ERROR: error running reboot:", err)
+				}
 				os.Exit(0)
```
ryanwe: This only happens if we fail to reboot. Should we log the fact that we're exiting so it doesn't look like a crash?

cli_tools/osconfig_agent/main.go
adjackura: Sleep just sleeps, this waist for the first channel to send something. In the normal case the ticker will just tick and the loop will continue, but if the service is told to restart this will break the service loop. This is mostly needed for Windows services but in the future we may add more cleanup that needs to happen here.

cli_tools/osconfig_agent/main.go
adjackura: No, it doesn't matter as this is in a different scope but I renamed to avoid confusion.

cli_tools/osconfig_agent/main.go
adjackura: Ack, changed to setOsConfig. And yeah I think that api change would make sense.

PR #558, opened 2018-10-03
 osconfig: support rebootConfig, windows update classifications and excludes

adjackura:
Write log to guest attributes, clean up locking logic


cli_tools/osconfig_agent/updates_windows.go
```diff
@@ -34,16 +38,190 @@ func rebootRequired() (bool, error) {
...
+	}
+	kbArticleIDsCount, _ := kbArticleIDsCountRaw.Value().(int32)
+
+	fmt.Printf("DEBUG: filtering out KBs: %q\n", excludes)
+	for i := 0; i < int(kbArticleIDsCount); i++ {
+		kbRaw, err := kbArticleIDs.GetProperty("Item", i)
+		if err != nil {
+			return err
+		}
+		if _, ok := excludes[kbRaw.ToString()]; ok {
```
ryanwe: consider separating the COM logic from the filter logic. Ideally, the filter logic would read cleanly with the COM logic abstracted away.

cli_tools/osconfig_agent/updates_windows.go
```diff
@@ -34,16 +38,190 @@ func rebootRequired() (bool, error) {
...
+	defer updateCollObj.Release()
+
+	updateColl, err := updateCollObj.IDispatch(ole.IID_IDispatch)
+	if err != nil {
+		return err
+	}
+	defer updateColl.Release()
+
+	class := make(map[string]struct{})
+	for _, c := range pp.WindowsUpdate.Classifications {
```
ryanwe: is this additional filtering?

go/packages/packages_linux.go
```diff
@@ -498,15 +497,15 @@ func installedGEM() ([]PkgInfo, error) {
...
+		fmt.Println("No gems installed.")
 		return nil, nil
 	}
 
 	var pkgs []PkgInfo
 	for _, ln := range lines[2:] {
 		pkg := strings.Fields(ln)
 		if len(pkg) < 2 {
-			logger.Errorf("%q does not represent a gem", ln)
+			fmt.Printf("%q does not represent a gem\n", ln)
```
ryanwe: Might be a good time to add the beginnings of a custom logger even if the logger just calls `fmt.Printf()`

adjackura: Did this for the main binary, these libraries still need to have something passed in which was a bit more than I wanted to setup now

PR #563, opened 2018-10-25
gce_inventory_agent: Update the format for WUA and QFE packages

adjackura:


go/packages/packages.go
```diff
@@ -40,11 +40,44 @@ var (
...
 // PkgInfo describes a package.
 type PkgInfo struct {
 	Name, Arch, Version string
 }
 
+// WUAPackage describes a Windows Update Agent package.
+type WUAPackage struct {
+	Title          string
+	Description    string
+	Categories     []string
```
ryanwe: As per our conversation, let's also add a non-localized category id in the form of "Security", "Critical", etc.

PR #564, opened 2018-10-25
Added subnet variable to image and disk export workflows

zoran15:
Added subnet variable to image and disk export workflows. Passing this variable to CreateInstance step. 

Specifying subnet is necessary when a GCP project has custom network setup.


zoran15:
Will re-do with correct author set

PR #569, opened 2018-10-29
Refactor shutdown image tests

lucaskanashiro:
These tests were not reliable because some times the instance powered off
before the logs had been processed. Now, we delayed a bit the instance
shutdown and the tests start to watch the serial output before the
shutdown process starts.

Besides that we add some small fixes related to FreeBSD.

### DIFF
--- a/image_test/metadata-script/shutdown-script-integrity.wf.json
+++ b/image_test/metadata-script/shutdown-script-integrity.wf.json
@@ -3,14 +3,17 @@
   "Vars": {
     "source_image": {"Required": true, "Description": "Image to be tested"},
     "shutdown_msg": {"Required": true, "Description": "Hash of shutdown script"},
+    "wait_msg": {"Required": true, "Description": "Message to stop instance"},
     "shutdown_script_name": {"Required": true, "Description": "Shutdown script of the created instance"},
+    "startup_script_name": {"Required": true, "Description": "Startup script of the created instance"},
     "instance_url": "integrity-url",
     "instance_gcs": "integrity-gcs",
     "instance_public_gcs": "integrity-public-gcs",
     "instance_metadata": "integrity-metadata"
   },
   "Sources": {
-    "shutdown_file.ps1": "${shutdown_script_name}"
+    "shutdown_file.ps1": "${shutdown_script_name}",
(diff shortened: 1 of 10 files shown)


image_test/metadata-script/shutdown-script-integrity.wf.json
```diff
@@ -3,14 +3,16 @@
...
     "shutdown_script_name": {"Required": true, "Description": "Shutdown script of the created instance"},
     "instance_url": "integrity-url",
     "instance_gcs": "integrity-gcs",
     "instance_public_gcs": "integrity-public-gcs",
     "instance_metadata": "integrity-metadata"
   },
   "Sources": {
-    "shutdown_file.ps1": "${shutdown_script_name}"
+    "shutdown_file.ps1": "${shutdown_script_name}",
+    "startup-script": "startup-script-for-shutdown-tests.sh"
```
adjackura: This will fail on windows, see above how the script is a variable, that way you pass it in for both windows and linux. The naming is a bit odd but just make this startup_script.ps1 (both windows and linux). Underscores are just what we use elsewhere, the extension just needs to be whatever scipt you use on windows (bat or ps1), linux will just ignore it.

image_test/metadata-script/shutdown-script-integrity.wf.json
```diff
@@ -65,6 +73,11 @@
           "Source": "${SOURCESPATH}/shutdown_file.ps1",
           "Destination": "${SOURCESPATH}/shutdown_file_public.ps1",
           "AclRules": [{"Entity": "allUsers", "Role": "READER"}]
+        },
+        {
+          "Source": "${SOURCESPATH}/startup-script",
+          "Destination": "${SOURCESPATH}/startup-script-public",
```
adjackura: Needs to be startup_script_public.ps1 to support both linux/windows

image_test/metadata-script/startup-script-for-shutdown-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
adjackura: You will need to replicate this for windows

PR #572, opened 2018-10-29
OS Config Patch Window RunOnce support

ryanwe:
Add support for RunOnce patch windows. 

If the window has already expired, which it shouldn't have, an error is returned. If the window already started, the start time will be the actual start time and the agent should begin executing the patch when as soon as it is ready.


cli_tools/osconfig_agent/patch_window_test.go
```diff
@@ -20,6 +20,7 @@ import (
 
 	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/osconfig_agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
 	"github.com/golang/protobuf/ptypes/duration"
+	date "google.golang.org/genproto/googleapis/type/date"
```
adjackura: no need for the named import, just "google.golang.org/genproto/googleapis/type/date"

I'll have to go back through and rename all the pb imports to NAMEpb as is standard

PR #585, opened 2018-11-06
Import diskutils.py only where is needed

lucaskanashiro:
Fix a regression added by fef17d7d6ca556fbe20ca078f98bbcb2bd5f81e0

In configuration image tests we import utils.py and after this commit,
it also imports diskutils.py. This script is trying to install
python-guestfs using apt in all distributions even in non-Debian based
and leading it to a failure.

You can check these failures in [testgrid](https://k8s-testgrid.appspot.com/google-gce-compute-image-tools#linux-image-tests&width=20).

### DIFF
--- a/daisy_workflows/image_import/debian/translate.py
+++ b/daisy_workflows/image_import/debian/translate.py
@@ -24,6 +24,8 @@ install_gce_packages: True if GCE agent and SDK should be installed
 import logging
 
 import utils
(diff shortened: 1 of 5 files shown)


adjackura:
Is this still needed? We fixed the regression and the tests are mostly ok again.

adjackura:
NVM, I was referring to the daisy integration tests not the linux tests

PR #588, opened 2018-11-07
osconfig_agent: Merge in inventory agent functionality

adjackura:
This is really basic for now, it's almost a straight copy paste of the relevant code. This will morph as we update the design of the overall agent.

### DIFF
--- a/cli_tools/osconfig_agent/google-osconfig-agent.goospec
+++ b/cli_tools/osconfig_agent/google-osconfig-agent.goospec
@@ -1,6 +1,6 @@
 {
   "name": "google-osconfig-agent",
-  "version": "0.2.0@1",
+  "version": "0.3.0@1",
   "arch": "x86_64",
   "authors": "Google Inc.",
   "license": "http://www.apache.org/licenses/LICENSE-2.0",
@@ -9,7 +9,7 @@
     "osconfig_agent.exe": "<ProgramFiles>/Google/OSConfig/osconfig_agent.exe"
   },
   "releaseNotes": [
-    "0.1.0 - Initial packaging"
+    "0.3.0 - Merge inventory agent funtionality"
   ],
   "install": {
     "path": "install.ps1"
--- a/cli_tools/osconfig_agent/inventory.go
+++ b/cli_tools/osconfig_agent/inventory.go
@@ -1,0 +1,138 @@
+//  Copyright 2017 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// gce_inventory_agent gathers and writes instance inventory to guest attributes.
+package main
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/go/osinfo"
+	"github.com/GoogleCloudPlatform/compute-image-tools/go/packages"
+)
+
+const (
+	inventoryURL = reportURL + "/guestInventory"
+)
+
+type instanceInventory struct {
+	Hostname          string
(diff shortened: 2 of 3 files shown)


cli_tools/osconfig_agent/google-osconfig-agent.goospec
```diff
@@ -9,7 +9,7 @@
     "osconfig_agent.exe": "<ProgramFiles>/Google/OSConfig/osconfig_agent.exe"
   },
   "releaseNotes": [
-    "0.1.0 - Initial packaging"
```
wj-chen: nit: we don't want to keep this in the list and "backfill" the 0.2.0 notes?

cli_tools/osconfig_agent/inventory.go
```diff
@@ -0,0 +1,138 @@
...
+		return err
+	}
+	if err := b.Close(); err != nil {
+		return err
+	}
+
+	return postAttribute(url, buf)
+}
+
+func writeInventory(state *instanceInventory, url string) {
```
wj-chen: I wonder when will writeInventory be called with an url other than the constant inventoryURL and whether we need this as an input param?

cli_tools/osconfig_agent/google-osconfig-agent.goospec
wj-chen: nit: we don't want to keep this in the list and "backfill" the 0.2.0 notes?

cli_tools/osconfig_agent/inventory.go
```diff
@@ -0,0 +1,138 @@
...
+
+	hs.PackageUpdates, errs = packages.GetPackageUpdates()
+	if len(errs) != 0 {
+		hs.Errors = append(hs.Errors, errs...)
+	}
+
+	return hs
+}
+
+func runInventory() {
```
wj-chen: nit: print a message after run is complete? but I imagine we'll progressively add more and better logging as we evolve this

cli_tools/osconfig_agent/inventory.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func getInventory() *instanceInventory {
+	fmt.Println("Gathering instance inventory.")
+
+	hs := &instanceInventory{}
+
+	hn, err := os.Hostname()
+	if err != nil {
```
wj-chen: I noticed that the error handling portion gets repetitive...is that a go language thing?

cli_tools/osconfig_agent/google-osconfig-agent.goospec
adjackura: Eh, I'm not sure how much I even want to keep here, we could just not have any notes until later. Once we go beta Ill keep each entry, but up to then many entries are mostly useless to end users and after beta you really don't care about what was done pre-release

cli_tools/osconfig_agent/inventory.go
adjackura: For tests we will want to change it (set it to the test server), tests that are currently unimplemented :)

cli_tools/osconfig_agent/inventory.go
adjackura: Yes, thats a quirk of the language, pretty much you are always going to have 'if err != nil {' after each function

cli_tools/osconfig_agent/inventory.go
adjackura: Yeah, the print statements that are in here are mostly debugging placeholders, we still need to implement proper logging

adjackura:
the search function in github is not super great but you can search the current repo. reportURL comes from main.go

Unit tests will still need to be written, the code this was copied from had no unit tests (yeah, I know :(), i figured the copy paste was good enough for this PR then unit tests shortly after

cli_tools/osconfig_agent/inventory.go
```diff
@@ -0,0 +1,138 @@
...
+	for i := 0; i < e.NumField(); i++ {
+		f := e.Field(i)
+		u := fmt.Sprintf("%s/%s", url, t.Field(i).Name)
+		switch f.Kind() {
+		case reflect.String:
+			if err := postAttribute(u, strings.NewReader(f.String())); err != nil {
+				state.Errors = append(state.Errors, err.Error())
+				fmt.Println("ERROR:", err)
+			}
+		case reflect.Map:
```
ryanwe: Why do we encode and compress maps? Are we unable to maintain that structure in the guest attrs?

Why do you post each field separately? Maybe I'm misunderstanding something. But wouldn't it be safer to post the whole inventory to guest attrs in a single call? Otherwise, a random failure would leave the inventory in a possible inconsistent state.

adjackura: That's a good suggestion, we can discuss this in our design discussions and when we rethink how to send this data. To answer you first question, this is actually broken right now, I am working on a fix for our new data structures now (and tests to verify everything)

PR #589, opened 2018-11-09
Add commands to support zypper management

iamsubratp:
-- add implementation for Zypper package manager.
-- add unit test for the zypper commands


cli_tools/osconfig_agent/osconfig_linux.go
```diff
@@ -15,14 +15,18 @@
...
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"strings"
 
 	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/osconfig_agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
-	"github.com/GoogleCloudPlatform/compute-image-tools/go/packages"
+	"github.com/compute-image-tools/go/packages"
```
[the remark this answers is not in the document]
iamsubratp: ah.. I remembered to switch this back before sending the changes out for review. somehow missed to add it.

go/packages/packages_linux.go
```diff
@@ -79,12 +82,56 @@ func InstallYumPackages(pkgs []string) {}
 // RemoveYumPackages removes yum packages.
 func RemoveYumPackages(pkgs []string) {}
 
+// Installs zypper packages
```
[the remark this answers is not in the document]
iamsubratp: done (Y)

go/packages/packages.go
```diff
@@ -23,6 +23,8 @@ import (
 	"github.com/GoogleCloudPlatform/compute-image-tools/go/osinfo"
 )
 
+type RunFunc func(*exec.Cmd) ([]byte, error)
```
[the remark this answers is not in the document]
iamsubratp: per offline discussion, using the functionality of global variables to re-assign the global "run" variable with a mocked version of run for testing purpose.
--implemented the changes.

cli_tools/osconfig_agent/osconfig_linux.go
```diff
@@ -79,8 +83,51 @@ func yumInstalls(pkgs []*osconfigpb.Package) error { return nil }
 
 func yumRemovals(pkgs []*osconfigpb.Package) error { return nil }
 
-func zypperRepositories(repos []*osconfigpb.ZypperRepository) error { return nil }
+func zypperRepositories(repos []*osconfigpb.ZypperRepository) error {
```
adjackura: We will need to add additional entries for at least a few things:
repo_gpgcheck
pkg_gpgcheck
gpgkey
and maybe 'type' since this is zypper

I'm fine with not adding in this PR but add a todo line for the gpg entries:
'// TODO: Write repo_gpgcheck, pkg_gpgcheck, and gpgkey entries.'

go/packages/packages_linux.go
```diff
@@ -116,15 +164,17 @@ func aptUpgrade() error {
 	return nil
 }
 
+// update yum packages
 func yumUpdate() error {
-	if _, err := exec.Command(yum, yumCheckUpdatesArgs...).CombinedOutput(); err != nil {
+	if _, err := exec.Command(yum, yumUpdateArgs...).CombinedOutput(); err != nil {
```
adjackura: Wrap this in run as well? Looks like this was missed for all the yum commands previously.

go/packages/packages_linux.go
```diff
@@ -405,8 +455,19 @@ func GetInstalledPackages() (Packages, []string) {
 			pkgs.Gem = gem
 		}
 	}
+	if exists(zypper) {
```
adjackura: You don't need this check at all, zypper is just a frontend to rpm

go/packages/packages_linux.go
iamsubratp: tthe 'run' is just a wrapper around .CombinedOutput method with some extra logging. I believe it was created for code re-usability.
Over here, i think, we forgot to use run and instead called the combinedoutput directly. i will change it to use run since that makes it consistent with rest of the code and also helps in writing test for this method.

go/packages/packages_linux.go
iamsubratp: so you mean, we essentially will get the same package list from rpmquery and zypper query. yeah, then it makes sense to remove the blob

PR #590, opened 2018-11-12
osconfig_agent: add inventory tests, fix writeInventory function

adjackura:


cli_tools/osconfig_agent/main.go
```diff
@@ -140,8 +140,8 @@ func run(ctx context.Context) {
 		} else {
 			setOsConfig(resp)
 			setPatchPolicies(resp.PatchPolicies)
-			runInventory()
 		}
+		runInventory()
```
wj-chen: Curious about the reason behind this change

cli_tools/osconfig_agent/inventory_test.go
```diff
@@ -0,0 +1,139 @@
...
+	var buf bytes.Buffer
+	io.Copy(&buf, zr)
+	zr.Close()
+
+	var pkgs packages.Packages
+	json.Unmarshal(buf.Bytes(), &pkgs)
+	return pkgs
+}
+
+func TestWriteInventory(t *testing.T) {
```
wj-chen: Do we have the ability to test an unhappy case (check the Error []string field) yet?

cli_tools/osconfig_agent/main.go
adjackura: The inventory run does not require a non error from the osconfig api. This is a bit of a hack to get around projects that don't have osconfig enabled but want to use inventory data. We will have a better solution in the future.

cli_tools/osconfig_agent/inventory_test.go
adjackura: The Error field is going to be removed in favor of reporting errors via some other method, and then we will have to add a test for the various failure modes

PR #597, opened 2018-11-15
linux_common/bootstrap.py: add Ubuntu 18.x support

lucaskanashiro:
Python 2 is not installed by default in Ubuntu 18.x and bootstrap.py
depends on it.

Ubuntu configuration tests are failing in [test grid](https://k8s-testgrid.appspot.com/google-gce-compute-image-tools#linux-image-tests&width=20) because python2 is not available in those images.

### DIFF
--- a/daisy_workflows/image_build/debian/debian.wf.json
+++ b/daisy_workflows/image_build/debian/debian.wf.json
@@ -17,7 +17,7 @@
     "build_files/utils/diskutils.py": "../../linux_common/utils/diskutils.py",
     "build_files/utils/__init__.py": "../../linux_common/utils/__init__.py",
     "build_files/google_cloud_test_repos": "./google_cloud_test_repos/",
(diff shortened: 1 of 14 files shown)


lucaskanashiro:
I discussed with @zmarano yesterday the possibility to migrate this bootstrap script to python 3. However, this script is used by the configuration image tests and it is executed in the tested image, and some distros do not have python 3 installed by default, such as CentOS 6 and 7. So if we migrate to python 3 we still need some workaround to install python 3 in these images. What do you think is the best solution here?

zmarano:
As discussed, the best thing to do here is move to py3 entirely.

lucaskanashiro:
OK then, I'll update to python 3 and keep a similar workaround (to install python 3 when it is not provided by default) to keep image configuration tests passing in CentOS/RHEL images.

PR #602, opened 2018-11-30
Fix file uploads (again) and specify usage.

zmarano:


daisy_workflows/linux_common/utils/common.py
```diff
@@ -300,24 +300,43 @@ def RunTest(test_func):
...
+  """
   # import 'google.cloud.storage' locally as 'google-cloud-storage' pip package
   # is not a mandatory package for all utils users
   from google.cloud import storage
   global CLIENT, BUCKET
 
-  dest_stripped = dest[len('gs://'):]
-  dest_splitted = dest_stripped.split('/')
-  bucket_name, blob_path = dest_splitted[0], '/'.join(dest_splitted[1:])
+  dest_stripped = dest_file[len('gs://'):]
```
illfelder: https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google_compute_engine/metadata_scripts/script_retriever.py

PR #608, opened 2018-12-05
Test new import files & if BYOL activation skipped

rofuentes:


daisy_integration_tests/scripts/post_translate_test.ps1
```diff
@@ -45,16 +71,29 @@ function Check-Activation  {
   }
 }
 
+function Check-SkipActivation {
+  if (!((Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SoftwareProtectionPlatform').KeyManagementServiceName -eq $null)) {
```
adjackura: How about:

$kms = (Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SoftwareProtectionPlatform').KeyManagementServiceName

if ($kms -ne "") {
  throw "KMS server set when it should be blank: $kms"
}

daisy_integration_tests/scripts/post_translate_test.ps1
```diff
@@ -45,16 +71,29 @@ function Check-Activation  {
...
+    throw 'Expected: Activation Skipped, Found: Activated License'
+  }
+}
+
 try {
+  $script:byol = Get-MetadataValue -key 'byol'
```
adjackura: Set the -default flag here to avoid the error in lookup, also the 'script' scoping is unneeded as it's just used in this block and not elsewhere: 
$byol = Get-MetadataValue -key 'byol' -default 'false'

daisy_integration_tests/scripts/post_translate_test.ps1
rofuentes: We are not testing for an empty string, but rather the null value.

daisy_integration_tests/scripts/post_translate_test.ps1
rofuentes: Updated.

PR #611, opened 2018-12-06
Rewrite bootstrap script using shell

lucaskanashiro:
Testing scripts now can migrate to python3, since bootstrap.sh installs
python3 for the images that do not have it installed by default.

This script is not used yet, I am working on another branch migrating the the testing scripts to python3 and making use of this `bootstrap.sh`. When it is done I'll remove `bootstrap.py` (in the next PR).


lucaskanashiro:
This modification was included in #613

PR #613, opened 2018-12-07
Move to python3

lucaskanashiro:


daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
zmarano: I need to make sure the worker image we are using supports this... stand by ;)

lucaskanashiro: ack

daisy_workflows/image_build/enterprise_linux/build_installer.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
lucaskanashiro: I already migrated `daisy_workflow/linux_common/utils` to python 3 in this PR, could you check it?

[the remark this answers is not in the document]
zmarano: I have no idea why I missed that before- yes you did :)

daisy_workflows/linux_common/bootstrap.sh
```diff
@@ -0,0 +1,134 @@
...
+    url="https://storage.googleapis.com/$bucket_target/$bucket_file"
+
+    dir=$(dirname "$file_dest")
+    [ ! -d "$dir" ] && mkdir -p "$dir"
+
+    logger -p daemon.info "Status: Bucket save: $url => $file_dest"
+    curl -H "Metadata-Flavor: Google" -H "Authorization: $token" -X GET "$url" -o "$file_dest"
+}
+
+CheckPython3Installation() {
```
zmarano: Is there a case for this right now? It wouldn't work on EL6 as far as I know anyway. Maybe we should remove it as to not cause problems later on.

image_test/configuration/linux/configuration-test.py
```diff
@@ -20,7 +20,7 @@
 
 def main():
   # detect distros to instantiate the correct class
-  if sys.version_info >= (3, 6):
+  if sys.version_info >= (3, 7):
```
zmarano: I think this was accurate since 3.6 and above have distro.

daisy_workflows/linux_common/bootstrap.sh
lucaskanashiro: Yes, this is needed for image configuration test. I just executed that daisy workflow using a RHEL 6 image and it works. RHEL 6 already has the EPEL repository installed and everything is fine.

image_test/configuration/linux/configuration-test.py
lucaskanashiro: In FreeBSD, using python 3.6, we do not have the module `distro` by default and it raises an error. And since [`platform` was removed only on python 3.8](https://bugs.python.org/issue28167) I only increased this version constraint. If you do not agree with that approach, we can find a way to install  `pip` and then `distro` in FreeBSD via `bootstrap.sh`.

daisy_workflows/linux_common/bootstrap.sh
zmarano: Oh, I didn't know RHEL 6 had py3 in EPEL. But cool.

image_test/configuration/linux/configuration-test.py
zmarano: SGTM

PR #616, opened 2018-12-11
osconfig_agent: add tasker package, rename osconfig to ospackage

adjackura:
The tasker package is a task serialization queue for osconfig_agent
I specifically ignored patch functionality as that will be changing very soon.


cli_tools/osconfig_agent/tasker/tasker.go
```diff
@@ -0,0 +1,48 @@
...
+	tc <- &task{name: name, run: f}
+}
+
+func tasker() {
+	logger.Debugf("Tasker start.")
+	for {
+		logger.Debugf("Waiting for tasks to run.")
+		select {
+		case t := <-tc:
+			logger.Debugf("Tasker running %q.", t.name)
```
ryanwe: Do log messages have a timestamp? It may be useful to know how long tasks take to execute.

cli_tools/osconfig_agent/service/service.go
```diff
@@ -12,30 +12,34 @@
...
+	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/osconfig_agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
 )
 
 var dump = &pretty.Config{IncludeUnexported: true}
 
-// Poll periodically calls the service to pull the latest applicaple configurations.
-func Poll(ctx context.Context) {
-	client, err := osconfigAgent.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
+func run(ctx context.Context) {
+	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
```
ryanwe: Not needed for this change but something I was wondering, can we move this to inside the for loop? That way we reread the endpoint and credentials for each call which allows for them to change w/out restarting the service. It can make it easier to fix/debug problems.

cli_tools/osconfig_agent/tasker/tasker.go
adjackura: Yes, right now all logs use these flags:
log.Ldate | log.Lmicroseconds | log.Lshortfile

https://golang.org/pkg/log/#pkg-constants

cli_tools/osconfig_agent/service/service.go
adjackura: That makes sense especially if we want configs to change without a service restart, I'll work on that for the next PR

PR #621, opened 2018-12-12
 osconfig_tests: add core images, update tests a bit

adjackura:


osconfig_tests/test_suites/inventory/inventory.go
```diff
@@ -219,17 +255,17 @@ func runGatherInventoryTest(ctx context.Context, testSetup *inventoryTestSetup,
 	}
 
 	testCase.Logf("Checking inventory data")
-	// It can take a bit to start collecting data, retry a few times.
+	// It can take a bit to start collecting data, especially on Windows.
```
ryanwe: Since this was an issue, might be worth defining "bit" which IIRC is on the order of 15 minutes.

osconfig_tests/test_suites/inventory/inventory.go
```diff
@@ -219,17 +255,17 @@ func runGatherInventoryTest(ctx context.Context, testSetup *inventoryTestSetup,
...
 	for i := 0; ; i++ {
 		ga, err := client.GetGuestAttributes(inst.Project, inst.Zone, inst.Name, "guestInventory/", "")
-		if err != nil && i > 5 {
+		if err != nil && i > 100 {
```
ryanwe: Can you add a comment that this is to time out after 25 minutes of retry/waiting. Or better still, you could make the code self explanatory, eg.

```
totalRetryTime := i * RETRY_TIME
if err != nil && totalRetryTime < 25 * time.Minutes {
```

Issue #622, opened 2018-12-12
Ubuntu can't download public gcs file when created in daisy


helen-fornazier:
Hello,

I get the following error when executing this simple workflow: https://paste.ee/p/HFEoD

```
startup-script: INFO Found startup-script-url in metadata.
startup-script: INFO Downloading url from gs://main-nucleus-128012-daisy-bkt/daisy-startup-script-linux-20181212-19:58:21-n5hrn/sources/startup_file_public.ps1 to /startup-ddyqfbac/tmpcg39bi_g using gsutil.
chronyd[1620]: Selected source 169.254.169.254
google_metadata_script_runner[1680]: Failure: Could not reach metadata service: Not Found.
startup-script: WARNING Could not download gs://main-nucleus-128012-daisy-bkt/daisy-startup-script-linux-20181212-19:58:21-n5hrn/sources/startup_file_public.ps1 using gsutil. Command '['gsutil', 'cp', 'gs://main-nucleus-128012-daisy-bkt/daisy-startup-script-linux-20181212-19:58:21-n5hrn/sources/startup_file_public.ps1', '/startup-ddyqfbac/tmpcg39bi_g']' returned non-zero exit status 1..
```

But I don't have this issue when I use gcloud with `--no-scopes` and `--no-service-account`:

```
gcloud compute instances create koike-u18-2 \
 --metadata=startup-script-url=gs://main-nucleus-128012-daisy-bkt/daisy-startup-script-linux-20181212-19:58:21-n5hrn/sources/startup_file_public.ps1 \
--image=projects/ubuntu-os-cloud/global/images/family/ubuntu-1804-lts \
--no-service-account --no-scopes
```

where the file startup_file_public.ps1 is public.

I am not entirely sure what is the difference between the instances created by daisy and gcloud, I am still investigating, please let me know if you have already seen this before.

Thanks

helen-fornazier:
If I add the scopes (with the code below) in the daisy workflow it works.

```
          "Scopes": [
            "https://www.googleapis.com/auth/devstorage.read_only"
          ]
```

But when creating the instance in gcloud using `--no-scopes` if works fine, confusing.

adjackura:
I have a feeling this may be a bug in gcloud, would have to do some digging, you can check the instance after its created to see if --no-scopes actually works. That workflow looks like its working correctly, what you have there will create an instance with no scopes, maybe try the UI as well?
[...]

PR #637, opened 2019-01-03
Fix scopes: set default value when empty

helen-fornazier:
The way Scopes was declared, json populates it as an empty list (even if
it is not set in the json object), so Scopes is never set to nil, and
the if statement that sets the default value is never true.

Declare Scopes as a pointer instead.

### DIFF
--- a/daisy/instance.go
+++ b/daisy/instance.go
@@ -88,7 +88,7 @@ type Instance struct {
 	Metadata map[string]string `json:"metadata,omitempty"`
 	// OAuth2 scopes to give the instance. If left unset
 	// https://www.googleapis.com/auth/devstorage.read_only will be added.
-	Scopes []string `json:",omitempty"`
+	Scopes *[]string `json:",omitempty"`
 	// StartupScript is the Sources path to a startup script to use in this step.
 	// This will be automatically mapped to the appropriate metadata key.
 	StartupScript string `json:",omitempty"`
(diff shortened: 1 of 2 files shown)


helen-fornazier:
@adjackura this kinda fix #622 , but it renders the image test useless, it would be nice to have a way to specify no-scopes in daisy, what do you think?

adjackura:
I think you might have this mixed up, the reason we check nil here is so you can specify no scopes. If you leave Scopes and ServiceAccounts blank daisy will set defaults, if you set them as an empty list you will get no scopes and no service accounts. Does setting no service accounts fix your issue?

helen-fornazier:
@adjackura Yes, setting no service accounts fix it

helen-fornazier:
@adjackura I see what you mean, but the way Scopes is declared
Scopes []string `json:",omitempty"`
Json never populates it as nil, it is an empty list, and that if statement will never be true (unless I am misreading the code).
e.g. this code https://paste.ee/p/NGPUE prints `{[] test}` instead of `{<nil> test}`

I updated PR changing the Scopes to a pointer and setting the default value seems to be working now, could you please check?
It turns out that this is unrelated to #622

adjackura:
Sorry for taking so long on this, does this fix fix all the issues you were having?
I see why this works but now I'm wondering why the service accounts work, at least I think they work from my past tests. 

If there still is an issue your having with not being able to specify no scopes and a service account (or use the default), or no scopes and no service account let me know and I can run a few more tests to find the best solution. We may need to override the ServiceAccounts field to be a pointer as well.

helen-fornazier:
@adjackura from what I understood in the [google-api-go-client](https://github.com/googleapis/google-api-go-client/blob/master/compute/v1/compute-gen.go), ServiceAccounts is declared as a pointer:

    ServiceAccounts []*ServiceAccount `json:"serviceAccounts,omitempty"`

adjackura:
Thats a slice of pointers, not the same, a slice is a pointer to an array so that's why it can have a nil value, a slice pointer is a bit redundant but may be required for the json unmarshal. I may have to do some testing to figure this out and make sure we allow for each of the various options. The current unit tests also should be updated to read from json as thats the common use case here.

PR #641, opened 2019-01-07
osconfig_agent: update build scripts, add version info

adjackura:
Add windows build script
Add version info to build
Don't use cgo (use Go resolver, not cgo)
Move common build commands to common function

### DIFF
--- a/cli_tools/google-osconfig-agent/packaging/debian/google-osconfig-agent.lintian-overrides
+++ b/cli_tools/google-osconfig-agent/packaging/debian/google-osconfig-agent.lintian-overrides
@@ -1,0 +1,1 @@
+google-osconfig-agent binary: statically-linked-binary
--- a/cli_tools/google-osconfig-agent/packaging/debian/google-osconfig-agent.lintian-overrides
+++ b/cli_tools/google-osconfig-agent/packaging/debian/google-osconfig-agent.lintian-overrides
@@ -1,1 +1,0 @@
-google-osconfig-agent binary: statically-linked-binary
--- a/cli_tools/google-osconfig-agent/packaging/google-osconfig-agent.spec
+++ b/cli_tools/google-osconfig-agent/packaging/google-osconfig-agent.spec
@@ -33,7 +33,7 @@ Contains the OSConfig agent binary and startup scripts
 %autosetup
 
 %build
-GOPATH=%{gopath} go build -ldflags="-s -w -linkmode=external -X main.version=${VERSION}" -o google_osconfig_agent
+GOPATH=%{gopath} CGO_ENABLED=0 go build -ldflags="-s -w -X main.version=%{_version}" -o google_osconfig_agent
 
 %install
 install -d %{buildroot}%{_bindir}
--- a/cli_tools/google-osconfig-agent/packaging/google-osconfig-agent.spec
+++ b/cli_tools/google-osconfig-agent/packaging/google-osconfig-agent.spec
@@ -33,7 +33,7 @@ Contains the OSConfig agent binary and startup scripts
 %autosetup
 
 %build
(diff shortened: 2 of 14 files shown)


cli_tools/google-osconfig-agent/packaging/setup_deb.sh
```diff
@@ -55,6 +39,6 @@ cd ${NAME}-${VERSION}
 cp -r ${working_dir}/packaging/debian ./
 cp -r ${working_dir}/*.service ./debian/
 
-debuild -us -uc
+debuild -us -uc -e "VERSION=${VERSION}"
```
hopkiw:
```
 dpkg-buildpackage -rfakeroot -us -uc -e VERSION=0.4.2
dpkg-buildpackage: error: unknown option or argument VERSION=0.4.2
```

cli_tools/google-osconfig-agent/packaging/google-osconfig-agent.spec
```diff
@@ -33,7 +33,7 @@ Contains the OSConfig agent binary and startup scripts
 %autosetup
 
 %build
-GOPATH=%{gopath} go build -ldflags="-linkmode=external" -o google_osconfig_agent
+GOPATH=%{gopath} CGO_ENABLED=0 go build -ldflags="-s -w -X main.version=%{_version}" -o google_osconfig_agent
```
hopkiw: use the go11 we downloaded and extracted here

cli_tools/google-osconfig-agent/packaging/setup_goo.sh
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source packaging/common.sh 
+
+echo "Building package"
+$GO get github.com/google/googet/goopack
+GOOS=windows $GO build -ldflags="-s -w -X main.version=${VERSION}" -o google_osconfig_agent.exe
```
hopkiw: for debuild and rpmbuild, the actual compilation steps are considered the responsibility of the underlying build systems. this is `debian/rules` and the `%build` or `%install` macros, respectively. so in keeping with that, i think the `go build` line should be in the `.goospec` file

cli_tools/google-osconfig-agent/packaging/debian/google-osconfig-agent.lintian-overrides
```diff
@@ -0,0 +1 @@
+google-osconfig-agent binary: statically-linked-binary
```
hopkiw: lintian overrides almost always indicate you're doing the wrong thing.. why must this be statically compiled?

cli_tools/google-osconfig-agent/packaging/setup_goo.sh
adjackura: Ack, this was an attempt to have one place for versioning. I'll move this into the goospec until I can implement variables.

cli_tools/google-osconfig-agent/packaging/debian/google-osconfig-agent.lintian-overrides
adjackura: This is an artifact of lintians logic for go packages. Lintian is supposed to skip this check for go packages but the way it skips is checking for the build-depends of golang-go or starts with golang-, since we don't depend on that package (salsa go-team package, part of their build toolchain) we get the error.

https://salsa.debian.org/lintian/lintian/blob/master/checks/binaries.pm#L139

PR #642, opened 2019-01-10
osconfig_agent: Update autogen code for new patch api

adjackura:
Ignore any files under _internal as they are autogenerated (first 7 files)

patchRun.go is copied from patchWindow.go and modified to almost work with the new API, there still are placeholder TODOs that I need to implement but this all does compile and run now.

Minor updates to package builds so the service starts automatically


cli_tools/google-osconfig-agent/patch/patch_run.go
```diff
@@ -0,0 +1,161 @@
...
+			logger.Errorf("osconfig.NewClient Error: %v", err)
+		} else {
+			tasker.Enqueue("Run patch", func() { patchRunner(ctx, client, pr) })
+		}
+	}
+
+	go watcher(ctx)
+}
+
+type patchRun struct {
```
ryanwe: It's a mouthful but the service calls this: InstancePatchStatus and it includes the current step/state of the instance's patch application. If these are the same, and I'm not entirely sure they are, it would be best to be consistent with the naming.

cli_tools/google-osconfig-agent/patch/patch_run.go
```diff
@@ -0,0 +1,161 @@
...
+	StartedAt, EndedAt time.Time `json:",omitempty"`
+	Complete           bool
+	Errors             []string `json:",omitempty"`
+}
+
+type patchJob struct {
+	*osconfigpb.ReportPatchJobInstanceDetailsResponse
+}
+
+// MarshalJSON marchals a patchConfig using jsonpb.
```
ryanwe: typo: marshals

PR #646, opened 2019-01-14
osconfig_agent: update packaging process

adjackura:
Add daisy workflows for the package builds
Auto start on el distros

### DIFF
--- a/cli_tools/google-osconfig-agent/packaging/build_packages_deb.sh
+++ b/cli_tools/google-osconfig-agent/packaging/build_packages_deb.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright 2019 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+function exit_error
+{
+  echo "build failed"
+  exit 1
+}
+
+URL="http://metadata/computeMetadata/v1/instance/attributes"
+GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/daisy-outs-path)
+if [[ -z $GCS_PATH ]]; then
+  echo 'Package build failed gcs-path metadata key blank'
+  exit 1
+fi
+
+apt-get install -y git-core || exit_error
+git clone https://github.com/GoogleCloudPlatform/compute-image-tools.git || exit_error
+cd compute-image-tools/cli_tools/google-osconfig-agent || exit_error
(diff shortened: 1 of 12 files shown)


osconfig_tests/test_suites/inventory/inventory.go
```diff
@@ -62,6 +62,19 @@ gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg
 EOM
 yum -y install google-osconfig-agent
 echo 'inventory install done'`
+var installInventoryYumEL6 = `sleep 10
```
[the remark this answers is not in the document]
adjackura: Yeah, this test specifically is just for inventory so that why i used this variable name here, the agent is the osconfig agent that will handle a number of different scenarios. I renamed to avoid this confusion

cli_tools/google-osconfig-agent/packaging/build_packages_rpm.sh
```diff
@@ -0,0 +1,35 @@
...
+{
+  echo "build failed"
+  exit 1
+}
+
+URL="http://metadata/computeMetadata/v1/instance/attributes"
+GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/daisy-outs-path)
+BASE_REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/base-repo)
+
+# There seems to be some kindof network initialization issue on el6, this sleep fixes that
```
hopkiw: if this is el6 only, add an el6 check

cli_tools/google-osconfig-agent/packaging/build_packages_deb.sh
```diff
@@ -0,0 +1,32 @@
...
+{
+  echo "build failed"
+  exit 1
+}
+
+URL="http://metadata/computeMetadata/v1/instance/attributes"
+GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/daisy-outs-path)
+BASE_REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/base-repo)
+
+apt-get install -y git-core || exit_error
```
hopkiw: you have five lines in a row ending `|| exit_error`
consider `set -e` and `trap` instead

cli_tools/google-osconfig-agent/packaging/build_packages_deb.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+trap "{ echo 'build failed' && exit 1 }" EXIT
```
hopkiw: needs a semicolon after `exit 1` or you'll get a syntax error
trap on exit will always be executed -- this makes the script a permanent failing script

adjackura: ack, this was 'working' but with the syntax error, I'm working on a better version that doesn't wok just because of ordering

cli_tools/google-osconfig-agent/packaging/build_packages_rpm.sh
adjackura: Made this just a retry on fail

PR #648, opened 2019-01-16
Add e2e test structure for package management

iamsubratp:
* Add a wrapper for osconfig service
* added a test case to create OsConfig, list and then delete the osconfig. This is just a sample test case to test the setup and osconfig service APIs. Tests around actual customer use cases to follow

### DIFF
--- a/osconfig_tests/osconfig_server/osconfig_service_wrapper.go
+++ b/osconfig_tests/osconfig_server/osconfig_service_wrapper.go
@@ -12,7 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-// Package compute contains wrappers around the GCE compute API.
+// Package contains wrapper around osconfig service APIs and helper methods
 package osconfig_server
 
 import (
@@ -28,11 +28,21 @@ import (
 
 var dump = &pretty.Config{IncludeUnexported: true}
 
-func CreateOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.CreateOsConfigRequest) (*osconfigpb.OsConfig, error) {
+func getOsConfigClient(ctx context.Context, logger *log.Logger) (*osconfig.Client, error) {
 	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
 
 	if err != nil {
 		logger.Printf("error while creating osconfig client: %s\n", err)
+	}
+	return client, err
+
+}
+
+func CreateOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.CreateOsConfigRequest) (*osconfigpb.OsConfig, error) {
+	client, err := getOsConfigClient(ctx, logger)
+
+	if err != nil {
+		return nil, err
 	}
 
 	logger.Printf("create osconfig request:\n%s\n\n", dump.Sprint(req))
@@ -48,10 +58,10 @@ func CreateOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.Cre
 }
 
 func ListOsConfigs(ctx context.Context, logger *log.Logger, req *osconfigpb.ListOsConfigsRequest) *osconfig.OsConfigIterator {
-	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
+	client, err := getOsConfigClient(ctx, logger)
 
 	if err != nil {
-		logger.Printf("error while creating osconfig client: %s\n", err)
+		return nil
 	}
 
 	logger.Printf("List osconfig request:\n%s\n\n", dump.Sprint(req))
@@ -66,10 +76,10 @@ func ListOsConfigs(ctx context.Context, logger *log.Logger, req *osconfigpb.List
 }
 
 func DeleteOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.DeleteOsConfigRequest) error {
-	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
+	client, err := getOsConfigClient(ctx, logger)
 
 	if err != nil {
-		logger.Printf("error while creating osconfig client: %s\n", err)
+		return err
 	}
 
 	logger.Printf("Delete osconfig request:\n%s\n\n", dump.Sprint(req))
--- a/osconfig_tests/osconfig_server/osconfig_service_wrapper.go
+++ b/osconfig_tests/osconfig_server/osconfig_service_wrapper.go
@@ -12,8 +12,8 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-// Package contains wrapper around osconfig service APIs and helper methods
-package osconfig_server
+// Package osconfigserver contains wrapper around osconfig service APIs and helper methods
+package osconfigserver
 
 import (
 	"context"
@@ -46,6 +46,7 @@ func getOsConfigClient(ctx context.Context, logger *log.Logger) (*osconfig.Clien
 	return client, err
 }
 
+// CreateOsConfig is a wrapper around createOsConfig API
 func CreateOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.CreateOsConfigRequest) (*osconfigpb.OsConfig, error) {
 	client, err := getOsConfigClient(ctx, logger)
 
@@ -65,6 +66,7 @@ func CreateOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.Cre
 	return res, nil
 }
 
+// ListOsConfigs is a wrapper around listOsConfigs API
 func ListOsConfigs(ctx context.Context, logger *log.Logger, req *osconfigpb.ListOsConfigsRequest) *osconfig.OsConfigIterator {
 	client, err := getOsConfigClient(ctx, logger)
 
@@ -76,13 +78,14 @@ func ListOsConfigs(ctx context.Context, logger *log.Logger, req *osconfigpb.List
 
 	resp := client.ListOsConfigs(ctx, req)
 	if resp == nil {
-		logger.Printf("error while listing osconfig:\n%s\n\n", *resp)
+		logger.Printf("error while listing osconfig:\n%v\n\n", resp)
 		return nil
 	}
 
 	return resp
 }
 
+// DeleteOsConfig is a wrapper around deleteOsConfig API
 func DeleteOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.DeleteOsConfigRequest) error {
 	client, err := getOsConfigClient(ctx, logger)
 
@@ -100,8 +103,7 @@ func DeleteOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.Del
 	return ok
 }
 
-// This function will cleanup all the osconfig created under project
-// Assumption is that this project is only used by this test application
+// CleanupOsConfig function will cleanup the osconfig created under project
 func CleanupOsConfig(ctx context.Context, testCase *junitxml.TestCase, logger *log.Logger, osConfig *osconfigpb.OsConfig) {
 
 	logger.Printf("Starting OsConfig cleanup...")
--- a/osconfig_tests/osconfig_server/osconfig_service_wrapper.go
+++ b/osconfig_tests/osconfig_server/osconfig_service_wrapper.go
@@ -1,0 +1,83 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package compute contains wrappers around the GCE compute API.
+package osconfig_server
+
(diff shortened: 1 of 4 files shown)


osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,93 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+var dump = &pretty.Config{IncludeUnexported: true}
+
+func getOsConfigClient(ctx context.Context, logger *log.Logger) (*osconfig.Client, error) {
+	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
+
+	if err != nil {
+		logger.Printf("error while creating osconfig client: %s\n", err)
+	}
+	return client, err
+
```
ryanwe: nit: extra line break

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+	"google.golang.org/api/option"
+)
+
+var dump = &pretty.Config{IncludeUnexported: true}
+
+func getOsConfigClient(ctx context.Context, logger *log.Logger) (*osconfig.Client, error) {
+	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
+
+	if err != nil {
+		logger.Printf("error while creating osconfig client: %s\n", err)
```
ryanwe: Should this cause the tests to fail in some way? It seems like a pretty grave error. Also, if you've seen this happen and have any suggestions on what to do if someone else runs into it, it would be good to comment that here.

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+		logger.Printf("error while creating osconfig:\n%s\n\n", err)
+		return nil, err
+	}
+	logger.Printf("create osconfig response:\n%s\n\n", dump.Sprint(res))
+
+	return res, nil
+}
+
+func ListOsConfigs(ctx context.Context, logger *log.Logger, req *osconfigpb.ListOsConfigsRequest) *osconfig.OsConfigIterator {
+	client, err := getOsConfigClient(ctx, logger)
```
ryanwe: Is there a reason to recreate the client each time instead of using a singleton client? I'm not advocating any change, just curious.

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if err != nil {
+		return nil
+	}
+
+	logger.Printf("List osconfig request:\n%s\n\n", dump.Sprint(req))
+
+	resp := client.ListOsConfigs(ctx, req)
```
ryanwe: no err? What if the client doesn't have access or the service is down?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -0,0 +1,187 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -0,0 +1,187 @@
...
+
+var dump = &pretty.Config{IncludeUnexported: true}
+
+// TODO: Move to the new combined osconfig package, also make this easily available to other tests.
+var installOsConfigAgentDeb = `echo 'deb http://packages.cloud.google.com/apt google-osconfig-agent-stretch-unstable main' >> /etc/apt/sources.list
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+apt-get update
+apt-get install -y google-osconfig-agent
+echo 'osconfig agent install done'`
+var installOsConfigAgentGooGet = `c:\programdata\googet\googet.exe -noconfirm install -sources https://packages.cloud.google.com/yuck/repos/google-osconfig-agent-unstable google-osconfig-agent
```
ryanwe: minor: consider a line break between these vars to make it a little easier to read.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -0,0 +1,187 @@
...
+	logger.Printf("Finished TestSuite %q", testSuite.Name)
+}
+
+func runCreateOsConfigTest(ctx context.Context, testCase *junitxml.TestCase, logger *log.Logger) {
+	defer cleanupOsConfig(ctx, testCase, logger)
+
+	req := &osconfigpb.CreateOsConfigRequest{
+		Parent: "projects/compute-image-test-pool-001",
+		OsConfig: &osconfigpb.OsConfig{
+			Name: "foo",
```
ryanwe: consider "create-osconfig-test-osconfig" or something with a little more meaning to help debug if you're looking at the project.

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
iamsubratp: sure

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
iamsubratp: done

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
adjackura: This is a very serious error, really this shouldn't happen ever unless there's something very wrong (not with osconfig specifically though)

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Package contains wrapper around osconfig service APIs and helper methods
+package osconfig_server
+
+import (
+	"context"
+	"log"
+
+	osconfig "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/cloud.google.com/go/osconfig/apiv1alpha1"
+	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
```
adjackura: have the proto imports separated by a new line, so like:
import (
"context"

"google.golang.org/api/option"

osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
)

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -0,0 +1,187 @@
...
+
+const testSuiteName = "PackageManagementTests"
+
+// TODO: Should these be configurable via flags?
+const testProject = "compute-image-test-pool-001"
+const testZone = "us-central1-c"
+
+var dump = &pretty.Config{IncludeUnexported: true}
+
+// TODO: Move to the new combined osconfig package, also make this easily available to other tests.
```
adjackura: You can drop the todo now

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		testCase.WriteFailure("error while creating osconfig:\n%s\n\n", err)
+	}
+
+	logger.Printf("CreateOsConfig response:\n%s\n\n", dump.Sprint(res))
+}
+
+// This function will cleanup all the osconfig created under project
+// Assumption is that this project is only used by this test application
+func cleanupOsConfig(ctx context.Context, testCase *junitxml.TestCase, logger *log.Logger) {
```
adjackura: Maybe use a key to filter on so this only cleans up configs for this test. Or have a cleanup wrapper around each osconfig you create or some sort of defer function.

osconfig_tests/test_suites/package_management/package_management.go

osconfig_tests/test_suites/package_management/package_management.go

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: changing the project will require some setup in the new project(test-infra, prow, etc). Also, I was talking to Liam today about our efforts on making our tools repo CI/CD. We have plans to redo our setup in order to move towards CI/CD pipeline.
I would suggest, for now lets keep the tests in this project for temporary period and once we get more idea about how our CI/CD   pipeline would look like, we can make the changes.

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
iamsubratp: sure

osconfig_tests/test_suites/package_management/package_management.go

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: a wrapper function more sounds like what we want. will add one. thanks for the suggestion

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
iamsubratp: there is no error being generated from the underlying client call. it returns an iterator in any case.

osconfig_tests/test_suites/package_management/package_management.go
ryanwe: To clarify, I was referring to the `Name` of the OsConfig and not the project where it resides. I'm fine with whatever project(s) we decide on. But if someone stumbles into the project via the console or gcloud, giving these resources identifying names will be helpful.

iamsubratp: ah..ok.. yeah sure. will change the name of the osconfig.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -0,0 +1,155 @@
...
+
+const testSuiteName = "PackageManagementTests"
+
+// TODO: Should these be configurable via flags?
+const testProject = "compute-image-test-pool-001"
+const testZone = "us-central1-c"
+
+var dump = &pretty.Config{IncludeUnexported: true}
+
+var installOsConfigAgentDeb = `echo 'deb http://packages.cloud.google.com/apt google-osconfig-agent-stretch-unstable main' >> /etc/apt/sources.list
```
adjackura: Use the utils library for this

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,122 @@
...
+	ok := client.DeleteOsConfig(ctx, req)
+	if ok != nil {
+		logger.Printf("error while deleting osconfig:\n%s\n\n", ok)
+		return nil
+	}
+	return ok
+}
+
+// CleanupOsConfig function will cleanup the osconfig created under project
+func CleanupOsConfig(ctx context.Context, testCase *junitxml.TestCase, logger *log.Logger, osConfig *osconfigpb.OsConfig) {
```
adjackura: This function should return and error, then the caller should log it. This library should not know about the testcase at all.

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
```diff
@@ -0,0 +1,122 @@
...
+	if resp == nil {
+		logger.Printf("error while listing osconfig:\n%v\n\n", resp)
+		return nil
+	}
+
+	return resp
+}
+
+// DeleteOsConfig is a wrapper around deleteOsConfig API
+func DeleteOsConfig(ctx context.Context, logger *log.Logger, req *osconfigpb.DeleteOsConfigRequest) error {
```
adjackura: Can you just combine these two functions into one called DeleteOsConfig, no reason to have to at this surface

iamsubratp: my idea of having 2 separate functions was that DeleteOsConfig will be used to imitate user behavior where user at certain step of his workflow deletes an osconfig.
Whereas Cleanup is supposed to be a part of test framework which will be used at the end to remove the left over osconfigs which was created by the user. Deleting this osconfig is not supposed be a part of this test. A noob example of such test case is, user creates an osconfig with a specific name and then tries to create another osconfig with same name and he should be able to do so. In this case, cleaning up the actual osconfig after the test completes is not a part of actual test.
I think, i should move the CleanupOsConfig method to a test framework related package.

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
iamsubratp: i agree, the library should not know about the testcase at all. In that case, the method will not be defer'able anymore, which is fine.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: yeah.. will make the changes to use the utils library

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
adjackura: Its still deferable, you can check for and act on errors with a defer, just use an anonymous function

osconfig_tests/osconfig_server/osconfig_service_wrapper.go
adjackura: if you wrap the config like I do for instances you can create a cleanup hook:
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/osconfig_tests/compute/instance.go#L28

iamsubratp: sure, yeah. I can do that. thanks

PR #652, opened 2019-01-18
Fix test results marshalling into different files

iamsubratp:
Currently, the results are concatenated and dumped into a single file
which is parsed and results are shown on testgrid. With multiple testsuites
in place, the xml generated does not have a root element resulting in failure
in parsing the xml.


osconfig_tests/main.go
```diff
@@ -85,23 +81,25 @@ func main() {
 		close(tests)
 	}()
 
-	var testSuites []*junitxml.TestSuite
 	for ret := range tests {
-		testSuites = append(testSuites, ret)
-	}
+		testSuiteOutPath := fmt.Sprintf("%s-%s", ret.Name, *outPath)
```
adjackura: outPath is supposed to be a path, maybe we need to change the flag to outdir? Then this would be filepath.Join

iamsubratp: yeah, changed the outdir.

osconfig_tests/main.go
```diff
@@ -85,23 +81,25 @@ func main() {
 		close(tests)
 	}()
 
-	var testSuites []*junitxml.TestSuite
 	for ret := range tests {
-		testSuites = append(testSuites, ret)
-	}
+		testSuiteOutPath := filepath.Join(*outDir, fmt.Sprintf("%s.xml", ret.Name))
```
adjackura: For clarity:
fmt.Sprintf("%s_junit.xml", ret.Name)

osconfig_tests/main.go
adjackura: Also update config.yaml in test-infra/prow with this new flag

osconfig_tests/main.go
iamsubratp: done

osconfig_tests/main.go
iamsubratp: done

PR #660, opened 2019-01-29
Add package installation test case

iamsubratp:
summary of packageInstallation testcase as done by runner.
* creates osconfig
* creates assignment
* creates a debian VM
* waits for osconfigagent to get installed
* waits for osconfigagent to do package installation
* queries the vm for list of installed packages and redirects the output to the serial console.
* wait until the package name comes up in the serial console.


osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -35,29 +40,20 @@ const testSuiteName = "PackageManagementTests"
...
-baseurl=https://packages.cloud.google.com/yum/repos/google-osconfig-agent-el7-unstable
-enabled=1
-gpgcheck=0
-repo_gpgcheck=1
-gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg
-       https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
-EOM
-yum -y install google-osconfig-agent
-echo 'osconfig agent install done'`
+var (
```
ryanwe: should these be consts?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
 
-	req := &osconfigpb.CreateOsConfigRequest{
-		Parent:   "projects/compute-image-test-pool-001",
-		OsConfig: osConfig,
-	}
+	parent := fmt.Sprintf("projects/%s", testProject)
+	res, err := osconfigserver.CreateOsConfig(ctx, logger, oc, parent)
 
-	logger.Printf("create osconfig request:\n%s\n\n", dump.Sprint(req))
+	defer cleanuposconfig(ctx, testCase, logger, oc)
```
ryanwe: Should this be `cleanupOsConfig`?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
 	if err != nil {
 		testCase.WriteFailure("error while creating osconfig:\n%s\n\n", err)
 	}
 
 	logger.Printf("CreateOsConfig response:\n%s\n\n", dump.Sprint(res))
 }
 
+func runPackageInstallTest(ctx context.Context, testCase *junitxml.TestCase, testSetup *packageManagementTestSetup, logger *log.Logger) {
+
+	osConfig, err := osconfigserver.JsonToOsConfig(packageInstallTestOsConfigString, logger)
```
ryanwe: Why are you specifying the configs as JSON strings instead of using plain structs with type safety?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+		return
+	}
+
+	assign := &osconfigserver.Assignment{Assignment: assignment}
+
+	res, err := osconfigserver.CreateAssignment(ctx, logger, assign, parent)
+	if err != nil {
+		testCase.WriteFailure("error while creating assignment: \n%s\n", err)
+	}
+	_ = res
```
ryanwe: if you don't care about `res`, then you can change line 162 to `_, err := ...` and remove this.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+	defer cleanupassignment(ctx, testCase, logger, assign)
+
+	client, err := daisyCompute.NewClient(ctx)
+	if err != nil {
+		testCase.WriteFailure("error creating client: %v", err)
+		return
+	}
+
+	testCase.Logf("Creating instance with image %q", testSetup.image)
+	testSetup.name = fmt.Sprintf("osconfig-test-%s-%s", path.Base(testSetup.image), "packageinstalltest")
```
ryanwe: Does `packageinstalltest` have to be one lowercase word? It'd be easier to read if it was camel case or used hyphens. Also, why use a constant string as the variable for `fmt.Sprintf()`? Maybe this can be rewritten as

```
testSetup.name = fmt.Sprintf("osconfig-test-%s-package-install", path.Base(testSetup.image))
```

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+
+	i := &api.Instance{
```
ryanwe: consider making this a const that can be cloned and modified for different tests. Or maybe just extract it to a separate function. It's a lot of code and seems like something that will be reused.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+	testCase.Logf("Waiting for agent install to complete")
+	if err := inst.WaitForSerialOutput("osconfig install done", 1, 5*time.Second, 5*time.Minute); err != nil {
+		testCase.WriteFailure("Error waiting for osconfig agent install: %v", err)
+		return
+	}
+
+	testCase.Logf("Agent installed successfully")
+
+	// allow agent to make the lookupconfig call and install the package
+	time.Sleep(1 * time.Minute)
```
ryanwe: Sleeping for a fixed amount of time can lead to slow, flaky tests. Can this wait for serial out too?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+	// TODO refactor to remove hardcoding of package name
+	listPkgCmd := getDebListCmd()
+	sshCmd := getGcloudSshCmd(testZone, testSetup.name, listPkgCmd, logger)
+	out, err := run(sshCmd, logger)
+
+	if err != nil {
+		testCase.WriteFailure("Error running verification command: %v", err)
+		return
+	}
+	_ = out
```
ryanwe: same here

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+	if err := inst.WaitForSerialOutput("osconfig install done", 1, 5*time.Second, 5*time.Minute); err != nil {
+		testCase.WriteFailure("Error waiting for osconfig agent install: %v", err)
+		return
+	}
+
+	testCase.Logf("Agent installed successfully")
+
+	// allow agent to make the lookupconfig call and install the package
+	time.Sleep(1 * time.Minute)
+	// TODO refactor to remove hardcoding of package name
```
ryanwe: nit: move comment to where you've hardcoded the package name.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+		testCase.WriteFailure("Error running verification command: %v", err)
+		return
+	}
+	_ = out
+
+	if err = inst.WaitForSerialOutput("cowsay", 1, 5*time.Second, 5*time.Minute); err != nil {
+		testCase.WriteFailure("Error waiting for assertion: %v", err)
+		return
+	}
+
```
ryanwe: nit: extra line break

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+}
+
+func getGcloudSshCmd(zone string, instance string, pkgManagerCommand string, logger *log.Logger) *exec.Cmd {
+	return exec.Command("gcloud", []string{"compute", "ssh", "--zone", fmt.Sprintf("%s", zone), instance, "--command", fmt.Sprintf("%s %s\n", pkgManagerCommand, writeToSerialConsole)}...)
```
ryanwe: This line seems really long. Not sure if the linter is missing it but I think you need a few line breaks here.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +108,197 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+	if err != nil {
+		testCase.WriteFailure(fmt.Sprintf("error while deleting assignment: %s", err))
+	}
+}
+
+func getDebListCmd() string {
+	cmd := dpkgquery
+	for _, arg := range dpkgqueryArgs {
+		cmd = fmt.Sprintf("%s %s", cmd, arg)
+	}
```
ryanwe: You could replace this with `return fmt.Sprintf("%s %s", dpkgquery, strings.Join(dpkgqueryArgs, " "))`. Though if this is the only place you use these args, it might be clearer to substitute `dpkgquery` and `dpkgqueryArgs` for a const `debListCmd := "/usr/bin/dpkg-query -W"` and remove this function.

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+type Assignment struct {
+	*osconfigpb.Assignment
+}
+
+// CreateAssignment is a wrapper around createAssignment API
+func CreateAssignment(ctx context.Context, logger *log.Logger, assignment *Assignment, parent string) (*Assignment, error) {
+	client, err := GetOsConfigClient(ctx, logger)
+
```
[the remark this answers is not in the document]
iamsubratp: done!

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -0,0 +1,77 @@
...
+		logger.Printf("error while creating assignment:\n%s\n\n", err)
+		return nil, err
+	}
+	logger.Printf("create assignment response:\n%s\n", dump.Sprint(res))
+
+	return &Assignment{Assignment: res}, nil
+}
+
+// Cleanup function will cleanup the assignment created under project
+func (o *Assignment) Cleanup(ctx context.Context, logger *log.Logger) error {
```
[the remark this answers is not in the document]
iamsubratp: o is chosen as an abbreviation for object. changed it to 'a', makes it more readable

osconfig_tests/osconfig_server/osconfig_utils.go
```diff
@@ -0,0 +1,44 @@
...
+// JsonToOsConfig creates an osconfig object from json string
+func JsonToOsConfig(jsonString string, logger *log.Logger) (*osconfigpb.OsConfig, error) {
+
+	var osconfig osconfigpb.OsConfig
+	err := json.Unmarshal([]byte(jsonString), &osconfig)
+
+	return &osconfig, err
+}
+
+//
```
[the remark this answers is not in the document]
iamsubratp: oops.. missed it. added the comment.

osconfig_tests/test_suites/package_management/package_management.go

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: my initial idea was to have json data files for each osconfig, assignment, etc. this way the actual code does not become lengthy making it more readable.

osconfig_tests/test_suites/package_management/package_management.go

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: plan was to move this out to a common method once  i add more test cases, since it will have common prefix. for now i am changing it to as you suggested. will move it out in a different commit.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: please discard the previpus response.

carved it out to a separate function.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: done!

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: spread it across multiple lines.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: oh.. good pick. I earlier had few other arguments, like -f for formatting output, for dpkgquery but eventually realized that i only need -W but forgot to change the method.
will change it to a single var to avoid extra processing.

during the whole experiment with package manager commands, i realized that there is need for a library that can provide standard APIs to perform package manager operations. Otherwise, we might end up writing the same code at multiple places.
I will discuss with @adjackura about it.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: we can definitely make the test_runner wait on some kind of logs generated when the agent makes the lookupconfig call, for example, the agent logs the lookupconfig response.
However, I don't think waiting on a log that is generated by the agent would be robust. Lets imagine we decide to change the log line on which the test_runner depends, then we will have to change the test as well.
I would rather prefer the log line to be generated by the test_runner itself. Unfortunately, the test_runner does not have a way to know when the lookupconfig call is made by the agent.

I am thinking of rather than having a fixed and large sleeptime, I would change it to be a retry'ble command that would timeout after a certain number of tries. The waitforSerialOutput does the same thing.
Let me know if it is fine.

osconfig_tests/test_suites/package_management/package_management.go
adjackura: It's simpler and has better error handling just to have a go file with the various structs you want to use, then you can at least have comments.

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -112,44 +105,202 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+		if totalTime < timeout {
+			logger.Print("Did not get expected response from serial console, will retry")
+		} else {
+			testCase.WriteFailure("Error while assertion: %v", err)
+			break
+		}
+	}
+}
+
+func getGcloudSshCmd(zone string, instance string, pkgManagerCommand string, logger *log.Logger) *exec.Cmd {
```
adjackura: As discussed, don't shell out to gcloud, use startup scripts.
In the future if we need to ssh create an ssh client in go, don't rely on external commands.

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -0,0 +1,76 @@
...
+
+	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
+)
+
+type Assignment struct {
+	*osconfigpb.Assignment
+}
+
+// CreateAssignment is a wrapper around createAssignment API
+func CreateAssignment(ctx context.Context, logger *log.Logger, assignment *Assignment, parent string) (*Assignment, error) {
```
adjackura: I don't think these functions should take a logger function, the logging should just be done by the caller.

iamsubratp: yeah, i agree, it should not. i added them while initially writing the wrapper package. it helped me debugging the apis.

will remove logger from the wrapper package.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: actually yeah, it was annoying at times to fix the json string. i will replace the json strings to structs and still keep all of them in a separate data file.

osconfig_tests/main.go
```diff
@@ -40,7 +39,7 @@ var (
 
 var testFunctions = []func(context.Context, *sync.WaitGroup, chan *junitxml.TestSuite, *log.Logger, *regexp.Regexp, *regexp.Regexp){
 	packagemanagement.TestSuite,
-	inventory.TestSuite,
+	// inventory.TestSuite,
```
adjackura: debug edit?

osconfig_tests/main.go
```diff
@@ -81,6 +80,7 @@ func main() {
 		close(tests)
 	}()
 
+	var testSuites []*junitxml.TestSuite
```
adjackura: Did you mean to populate this?

osconfig_tests/main.go
iamsubratp: oh yeah! just to save some time, i removed it, since test-suite-filter didn't work. will remove it.

osconfig_tests/main.go
iamsubratp: ah.. i had this in a different branch. this was to address a bug. the bug was that the execution of the test_runner never fails even if there is a failure. This is because we first read the testsuites from the channel to create the junit files and the we again try to iterate over the channel to check for failures. but during the second loop, the channel is already empty and so the execution never fails.
I wanted to fix it by dumping the suites to an array and then use it for all iterations.. will send out a different pr for this.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: moved away from creating the objects using json to creating them through structs.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: had a discussion with @adjackura . we decided to use a startup script that keep writing to the serial console and then have the test case poll the serial output for a specific string.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: yet to fix some linter issues.

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req := &osconfigpb.CreateAssignmentRequest{
+		Parent:     parent,
+		Assignment: assignment.Assignment,
+	}
+
+	res, err := client.CreateAssignment(ctx, req)
```
ryanwe: Why don't you check if `err != nil`?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -1,4 +1,3 @@
-//  Copyright 2019 Google Inc. All Rights Reserved.
```
ryanwe: Why is this line deleted?

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+	"time"
+
+	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/compute"
+	osconfigserver "github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/osconfig_server"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/utils"
+	api "google.golang.org/api/compute/v1"
+)
+
+var (
```
ryanwe: Should this be `const`?

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+	pkgTestSetup = []*packageManagementTestSetup{}
+	pkgManagers  = [...]string{"apt"}
+)
+
+var vf = func(inst *compute.Instance, vfString string, port int64, interval, timeout time.Duration) error {
```
ryanwe: What is `vf`? Should this have a different name or a comment?

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := inst.WaitForSerialOutput(vfString, port, interval, timeout); err != nil {
```
ryanwe: How about `return insta.WaitForSerial...`?

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+	desc := "test osconfig creation"
+	for _, pkgManager := range pkgManagers {
+		var oc *osconfigpb.OsConfig
+
+		switch pkgManager {
+		case "apt":
+			pkg := osconfigserver.BuildPackage("cowsay")
+			pkgs := []*osconfigpb.Package{pkg}
+			oc = osconfigserver.BuildOsConfig(testName, desc, osconfigserver.BuildAptPackageConfig(pkgs, nil, nil), nil, nil, nil, nil)
+		}
```
ryanwe: Can you add a `default` case to catch any future mistakes as we add support for additional pkg manager types? It can just log or panic.

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+		var instancename string
+
+		switch pkgManager {
+		case "apt":
+			instancename = fmt.Sprintf("%s-%s", filepath.Base(debianImage), testName)
+			pkg := osconfigserver.BuildPackage("cowsay")
+			pkgs := []*osconfigpb.Package{pkg}
+			oc = osconfigserver.BuildOsConfig(testName, desc, osconfigserver.BuildAptPackageConfig(pkgs, nil, nil), nil, nil, nil, nil)
+			assign = osconfigserver.BuildAssignment(testName, desc, osconfigserver.BuildInstanceFilterExpression(instancename), []string{fmt.Sprintf("projects/%s/osConfigs/%s", testProjectID, oc.Name)})
+		}
```
ryanwe: Can you add the `default` case here too?

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+	testName := "packageinstalltest"
+	desc := "test package installation"
+	for _, pkgManager := range pkgManagers {
+		var oc *osconfigpb.OsConfig
+		var assign *osconfigpb.Assignment
+		var instancename string
+
+		switch pkgManager {
+		case "apt":
+			instancename = fmt.Sprintf("%s-%s", filepath.Base(debianImage), testName)
```
ryanwe: Should `instancename` be `instanceName`?

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,169 @@
...
+			startup: &api.MetadataItems{
+				Key:   "startup-script",
+				Value: &utils.StartupScriptDeb,
+			},
+		}
+		pkgTestSetup = append(pkgTestSetup, &setup)
+	}
+}
+
+func generateAllTestSetup() []*packageManagementTestSetup {
```
ryanwe: Are these test "setups" or the actual tests?

osconfig_tests/osconfig_server/assignment_wrapper.go
iamsubratp: i think, it just got missed. added it.

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return &Assignment{Assignment: res}, nil
+}
+
+// Cleanup function will cleanup the assignment created under project
+func (a *Assignment) Cleanup(ctx context.Context, projectId string) error {
+	client, err := GetOsConfigClient(ctx)
+
```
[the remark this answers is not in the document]
iamsubratp: removed.

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return err
+	}
+
+	deleteReq := &osconfigpb.DeleteAssignmentRequest{
+		Name: fmt.Sprintf("projects/%s/assignments/%s", projectId, a.Name),
+	}
+	ok := client.DeleteAssignment(ctx, deleteReq)
```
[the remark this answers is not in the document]
iamsubratp: done.

osconfig_tests/osconfig_server/osconfig_client.go
```diff
@@ -0,0 +1,36 @@
...
+
+	osconfig "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/cloud.google.com/go/osconfig/apiv1alpha1"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/config"
+	"google.golang.org/api/option"
+)
+
+var osconfigClient *osconfig.Client
+
+func GetOsConfigClient(ctx context.Context) (*osconfig.Client, error) {
+
```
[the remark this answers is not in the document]
iamsubratp: done.

osconfig_tests/osconfig_server/osconfig_client.go
```diff
@@ -0,0 +1,36 @@
...
+	if osconfigClient != nil {
+		return osconfigClient, nil
+	}
+
+	client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
```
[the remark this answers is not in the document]
iamsubratp: done

osconfig_tests/osconfig_server/osconfig_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type OsConfigIterator struct {
+	*osconfig.OsConfigIterator
+}
+
+// CreateOsConfig is a wrapper around createOsConfig API
+func CreateOsConfig(ctx context.Context, oc *OsConfig, parent string) (*OsConfig, error) {
+	client, err := GetOsConfigClient(ctx)
+
```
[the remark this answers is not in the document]
iamsubratp: removed from here and other occurrences

osconfig_tests/osconfig_server/osconfig_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &OsConfig{OsConfig: res}, nil
+}
+
+// ListOsConfigs is a wrapper around listOsConfigs API
+func ListOsConfigs(ctx context.Context, req *osconfigpb.ListOsConfigsRequest) *OsConfigIterator {
+	client, err := GetOsConfigClient(ctx)
+
```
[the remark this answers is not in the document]
iamsubratp: done

osconfig_tests/osconfig_server/osconfig_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+	resp := client.ListOsConfigs(ctx, req)
+
+	return &OsConfigIterator{OsConfigIterator: resp}
+}
+
+// Cleanup function will cleanup the osconfig created under project
+func (o *OsConfig) Cleanup(ctx context.Context, projectId string) error {
+	client, err := GetOsConfigClient(ctx)
+
```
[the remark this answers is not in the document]
iamsubratp: done!

osconfig_tests/osconfig_server/osconfig_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return err
+	}
+
+	deleteReq := &osconfigpb.DeleteOsConfigRequest{
+		Name: fmt.Sprintf("projects/%s/osConfigs/%s", projectId, o.Name),
+	}
+	ok := client.DeleteOsConfig(ctx, deleteReq)
```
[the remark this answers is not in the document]
iamsubratp: done

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: my bad! this is by mistake. reverted it

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: since golang is statically compiled, arrays and slices cannot be constants because arrays and slices are evaluated during runtime.

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: added comment for vf

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: yeah, that makes more sense. changed it

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: yeah sure.. done

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: done

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: added for all the methods

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: my idea of "setup" was creating any resource(and thus the environment) to run a particular test. I wanted these setups to be de-coupled from the actual tests. but the setup is really tied the actual tests in the sense that for a particular test, there is a particular environment required.
that's the reason i had to introduce the "fname" in the setup struct and do all the filtering logic in the test runner.
honestly, i wanted to structure this in a better way so that the test runner did not have to do the filtering based on the function name( test runner need not have to have any context about the testsetup). but since i had already spent too much time figuring this out, in the interest of time, i just went ahead with my current logic.

osconfig_tests/utils/utils.go
```diff
@@ -21,12 +21,24 @@ import (
 )
 
 var (
-	// InstallOSConfigDeb installs the osconfig agent on deb based systems.
-	InstallOSConfigDeb = `echo 'deb http://packages.cloud.google.com/apt google-osconfig-agent-stretch-unstable main' >> /etc/apt/sources.list
+	// TODO: the startup script installs the osconfig agent and then keeps querying
```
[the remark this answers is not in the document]
iamsubratp: i had this in my mind, but i wanted to have a more generic solution for running multiple startup scripts keeping in mind that we might have to change this when we write tests for patching( i am not sure what would change, though)
now i think, instead of having a generic solution to run multiple startup scripts, how about having a generic solution to create multiple startup scripts, like an interface that each test-suite can use and implement their own way of creating a single start-up script.

osconfig_tests/test_suites/package_management/package_management_test_data.go
adjackura: Don't panic, if you need error handling add error handling, or just print an error as this really may not matter much.

osconfig_tests/test_suites/package_management/package_management_test_data.go
adjackura: These should not be globals, generateAllTestSetup should pass in pkgManagers and do the appending for pkgTestSetup. If you want to keep pkgTestSetup up here for easy additions that's fine but that's also a bit redundant, just have each of the addCreate* functions use a for loop, why have it defined twice unless you expect different inputs at different times.

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -0,0 +1,179 @@
...
+	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/compute"
+	osconfigserver "github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/osconfig_server"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/utils"
+	api "google.golang.org/api/compute/v1"
+)
+
+var (
+	pkgTestSetup = []*packageManagementTestSetup{}
+	pkgManagers  = [...]string{"apt"}
```
adjackura: Using arrays isn't idiomatic, use slices.

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: changed it to startup script

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: i wanted to keep the list of pkgmanagers global since i did not want to hide  it in a function. since the list of package managers is static throughout the osconfig package and hence can also be used at other places( in my mind, the list is not just scoped for generating the testsetups, rather its the list of package managers that we support through osconfig). 

but yeah, it makes sense to move the pkgtestsetup to function scope. Let me know if that's ok

adjackura: Having the list global is fine, but the individual addCreate functions should not reference it. Instead generateAllTestSetup should pass it in as an arg for each of the functions. This allows us to run unit tests and refactor in a much more simple matter.

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: changed.

iamsubratp: for now, let it panic. i will add logger and error handling in it

osconfig_tests/utils/utils.go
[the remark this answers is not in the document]
iamsubratp: i left the startupscripts as it is and moved the query string to specific testSetup and package manager. this also allow us to have specific query string based on the package manager and the package that is being used for the test.

PR #663, opened 2019-01-30
Update logic of patch run to report / check status as it goes

ryanwe:
- Patching reports status and stops if the patch job is not active
- Handles restarts across crashes and reboots
- Handles dryruns
- Handles stopping on failure
- Supports running the agent as `ospatch` to facilitate testing.
- Does not *yet* handle retry attempts (that'll be another PR)
- I did a lot of manual testing but this code doesn't lend itself well to unit testing. I need to think more about how to do that in a subsequent PR.

PR #664, opened 2019-01-31
Fix patch notification watcher

ryanwe:
- We can't wait on the metadata key since it might not exist
- Watch on the attributes instead of keep track of changes

### DIFF
--- a/cli_tools/google-osconfig-agent/patch/patch_run.go
+++ b/cli_tools/google-osconfig-agent/patch/patch_run.go
@@ -30,20 +30,24 @@ import (
 
 // Init starts the patch system.
 func Init(ctx context.Context) {
+	savedPatchName := ""
 	// Load current patch state off disk.
 	pr, err := loadState(state)
 	if err != nil {
 		logger.Errorf("loadState error: %v", err)
-	} else if pr != nil && !pr.Complete {
-		client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
-		if err != nil {
-			logger.Errorf("osconfig.NewClient Error: %v", err)
-		} else {
-			tasker.Enqueue("Run patch", func() { patchRunner(ctx, client, pr) })
+	} else if pr != nil {
+		savedPatchName = pr.Job.PatchJobName
+		if !pr.Complete {
+			client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
+			if err != nil {
+				logger.Errorf("osconfig.NewClient Error: %v", err)
+			} else {
+				tasker.Enqueue("Run patch", func() { patchRunner(ctx, client, pr) })
+			}
 		}
 	}
 
-	go watcher(ctx)
+	go watcher(ctx, savedPatchName)
 }
 
 type patchRun struct {
--- a/cli_tools/google-osconfig-agent/patch/watcher.go
+++ b/cli_tools/google-osconfig-agent/patch/watcher.go
@@ -28,9 +28,9 @@ import (
 )
 
 const (
-	defaultEtag         = ""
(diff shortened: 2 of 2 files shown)


cli_tools/google-osconfig-agent/patch/patch_run.go
```diff
@@ -30,20 +30,24 @@ import (
...
-	} else if pr != nil && !pr.Complete {
-		client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
-		if err != nil {
-			logger.Errorf("osconfig.NewClient Error: %v", err)
-		} else {
-			tasker.Enqueue("Run patch", func() { patchRunner(ctx, client, pr) })
+	} else if pr != nil {
+		savedPatchName = pr.Job.PatchJobName
+		if !pr.Complete {
+			client, err := osconfig.NewClient(ctx, option.WithEndpoint(config.SvcEndpoint()), option.WithCredentialsFile(config.OAuthPath()))
```
iamsubratp: may be use the utils method https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/osconfig_tests/osconfig_server/osconfig_service_wrapper.go#L35
unless you want to use a separate client for patch.

Also, i feel i should move the service wrapper one step up in the directory structure since it is not specific to tests.

ryanwe: This is going away in #663 and subsequent PRs. But I'm curious, what is the value in caching the client? I think it's better to create a new client for each API call instead to account for any permissions changes that may occur on the VM. Am I not understanding something about how the client works? Is creating a new client expensive compared to actually making the API request?

cli_tools/google-osconfig-agent/patch/watcher.go
```diff
@@ -88,9 +94,30 @@ func watcher(ctx context.Context) {
...
 			continue
 		}
-		id := strings.Split(md, ",")[0]
-		if id != "" {
-			ackPatch(ctx, id)
+
+		patchNotification, err := metadata.InstanceAttributeValue(metadataPatchKey)
+		if err != nil {
+			switch err.(type) {
+			case metadata.NotDefinedError:
```
iamsubratp: why is notdefinederror a special case? just for logging?  is it the case that the metadata key is eventually going to be available, because if it is not, then does it make sense to retry to fetch the attribuevalue?

cli_tools/google-osconfig-agent/patch/patch_run.go
iamsubratp: correct me if i am wrong, but every  time we create a new client, we create a new connection which ends up consuming server resources and if it is being used for only once, the client holds the server resources until the connection timesout. something where we can optimize.
might not be a great deal for us today, but it might hurt us in future. 

Also for the permissions changes, i am curious to learn if the credential library does refresh the credentials used by the client. It would be weird, if applications break while running, just because the credentials got refreshed.

cli_tools/google-osconfig-agent/patch/watcher.go
iamsubratp: i see... yeah, my initial impression was why can't the block be written as
`if err != nil {
    logger.Errorf("error line...")
    conitnue
}`
but now i understand, we explicitly want to call out that keynotpresent is not really an error for our application. totally makes sense.
thanks for the explanation.

cli_tools/google-osconfig-agent/patch/patch_run.go
ryanwe: I'd have to see the implementation to be sure but I'd be surprised if the client creates a connection when it's initialized. I would think it would do nothing until a request was made and that it would clean up its resources afterwards (possibly caching reusable things like credentials).

I know that the instance identity token is regularly refreshed since they're only valid for a few hours. Any credentials the client uses may also have some expiration but I'd have to see the details of how it's implemented to be sure. My assumption was that the client is just a nice way of making HTTP calls (using gRpc, built in support for auth, type safety, etc.)

PR #669, opened 2019-02-11
osconfig_tests: use larger instances for inventory tests

adjackura:


adjackura:
This is an attempt to deflake the tests, bigger instances means shorter time to inventory so overall the tests run quicker and should be less prone to timeout.

PR #672, opened 2019-02-12
Compare current state with desired state

ryanwe:
This is not yet tied into the actual inventory or package management code. It is just the logic that will be used to determine what updates are needed given the caller already has the `instanceInventory` and the `LookupConfigsResponse`.

The caller is expected to know which package managers they care about and pass the package info for just those packages.

### DIFF
--- a/cli_tools/google-osconfig-agent/inventory/changes.go
+++ b/cli_tools/google-osconfig-agent/inventory/changes.go
@@ -22,6 +22,7 @@ import (
 	osconfigpb "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
 )
 
+// Changes represents the delta between the actual and the desired package installation state.
 type Changes struct {
 	PackagesToInstall []string
 	PackagesToUpgrade []string
--- a/cli_tools/google-osconfig-agent/inventory/changes.go
+++ b/cli_tools/google-osconfig-agent/inventory/changes.go
@@ -1,0 +1,73 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cli_tools/google-osconfig-agent/inventory/changes.go
```diff
@@ -0,0 +1,74 @@
...
+// Changes represents the delta between the actual and the desired package installation state.
+type Changes struct {
+	PackagesToInstall []string
+	PackagesToUpgrade []string
+	PackagesToRemove  []string
+}
+
+// GetNecessaryChanges compares the current state and the desired state to determine which packages
+// need to be installed, upgraded, or removed.
+func GetNecessaryChanges(installedPkgs []packages.PkgInfo, upgradablePkgs []packages.PkgInfo, packageInstalls []*osconfigpb.Package, packageRemovals []*osconfigpb.Package) Changes {
```
adjackura: installedPkgs and upgradablePkgs come in the form []PkgInfo
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/go/packages/packages.go#L66

Should this function just take that and convert here or should the caller convert? Up to you, I don't currently have a string opinion.

cli_tools/google-osconfig-agent/inventory/changes.go
```diff
@@ -0,0 +1,74 @@
...
+	var pkgsToInstall []string
+	var pkgsToUpgrade []string
+
+	for _, pkg := range packageInstalls {
+		_, isInstalled := installedPkgMap[pkg.Name]
+		_, isUpgradable := upgradeablePkgMap[pkg.Name]
+
+		if !isInstalled {
+			pkgsToInstall = append(pkgsToInstall, pkg.Name)
+		} else if isInstalled && isUpgradable {
```
adjackura: isInstalled is a bit redundant

ryanwe: As discussed, I'm leaving this as-is for readability.

cli_tools/google-osconfig-agent/inventory/changes.go
ryanwe: Yep. That's what it does :smile:

PR #674, opened 2019-02-13
Fixed Image importer not loading workflow file due to relative paths not working within Docker image.

zoran15:


cli_tools/gce_vm_image_import/main.go
```diff
@@ -386,13 +388,22 @@ func buildDaisyVars(translateWorkflowPath string) map[string]string {
 	return varMap
 }
 
+func toWorkingDir(dir string) string {
+	wd, err := filepath.Abs(filepath.Dir(os.Args[0]))
```
adjackura: Don't address os.Args[0] in this function, take it as an arg, this will help with writing tests.

zoran15: I've refactored the code and updated unit tests.

PR #679, opened 2019-02-14
Perform regex matching instead of plain string matching

iamsubratp:
In some cases, assertion on an output in serial console becomes tricky
with plain strings.
For example:
to check if a package is installed with rpmquery on centos, i am using
rpmquery -a {packagename}
The result is {packageName}-{version}-{arch}.
Assertion on packagename is ok but this might introduce a bug where
packge is not installed and output is
"package {packagename} is not installed".

### DIFF
--- a/osconfig_tests/compute/instance.go
+++ b/osconfig_tests/compute/instance.go
@@ -40,10 +40,16 @@ func (i *Instance) Cleanup() {
 }
 
 // WaitForSerialOutput waits for a match with regular expression regex on a serial port.
+// if you are using regular expression make sure to check the err
 func (i *Instance) WaitForSerialOutput(regex string, port int64, interval, timeout time.Duration) error {
 	var start int64
 	var errs int
-	var validMatch = regexp.MustCompile(regex)
+	validMatch, err := regexp.Compile(regex)
+
+	if err != nil {
+		return fmt.Errorf("error compiling regex: %v", err)
+	}
+
 	tick := time.Tick(interval)
 	timedout := time.Tick(timeout)
 	for {
--- a/osconfig_tests/compute/instance.go
+++ b/osconfig_tests/compute/instance.go
@@ -49,7 +49,7 @@ func (i *Instance) WaitForSerialOutput(regex string, port int64, interval, timeo
 	for {
 		select {
 		case <-timedout:
-			return fmt.Errorf("timed out waiting to match regular expression:", regex)
+			return fmt.Errorf("timed out waiting to match regular expression: %s", regex)
 		case <-tick:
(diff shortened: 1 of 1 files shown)


hopkiw:
the output for `rpmquery -a <not-installed-package>` is the empty string

iamsubratp:
> the output for `rpmquery -a <not-installed-package>` is the empty string.
right. I cannot assert on empty string read from serial console. also, i cannot assert on "<not-installed-package-name>" on serial console, because this string may be gets populated as a part of execution of the startup-script that i am using to get the status of the package.

i am using `rpmquery -s <not-installed-package>` to confirm that the package is not installed. the output is
'package <not-installed-package> is not installed'

the case where package is not installed, the result is pretty clear and can be programmed. It's the case where the package is installed. 

rpmquery, unlike dpkgquery, does not give an explicit statement that the package is installed, instead it writes the result in  {packageName}-{version}-{arch} format.

osconfig_tests/compute/instance.go
```diff
@@ -38,16 +39,17 @@ func (i *Instance) Cleanup() {
 	}
 }
 
-// WaitForSerialOutput waits to a string match on a serial port.
-func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeout time.Duration) error {
+// WaitForSerialOutput waits for a match with regular expression regex on a serial port.
+func (i *Instance) WaitForSerialOutput(regex string, port int64, interval, timeout time.Duration) error {
 	var start int64
 	var errs int
+	var validMatch = regexp.MustCompile(regex)
```
adjackura: Use compile and then check for and return the error.

iamsubratp: changed it

adjackura:
I think it's better to add more logic to the startup script to have clearer output than introduce unneeded regex to this function. We haven't ever needed complex matching on serial output, instead all our logic is placed into the startup scripts.

iamsubratp:
> I think it's better to add more logic to the startup script to have clearer output than introduce unneeded regex to this function. We haven't ever needed complex matching on serial output, instead all our logic is placed into the startup scripts.

with multiple different ways that i tried to make the startup script easy to construct and at the same time it has good readability( i can see Ryan's concern around creating the stratup script in the structure makes it look a bit messy), I have decided to move the whole startup script creation to a different location so the struct creation will be more cleaner.

Also, with multiple different options available for package-management operations, i got confused with which one to choose. Thanks to Liam, he helped me on how to  decide which one to use.
To perform a particular operation, there is a standard command and then there are bunch of other commands that can provide the same information but those are just side effects of the commands and not really intended for the particular operation.

PR #681, opened 2019-02-15
Expose qemu error for export workflow

dntczdx:
Currently, qemu error details are not shown in daisy log and console. Let's output it there so user/dev can figure out the root cause easily.

It's tested manually since there isn't fancy new feature added.

### DIFF
--- a/daisy_workflows/export/export_disk_ext.sh
+++ b/daisy_workflows/export/export_disk_ext.sh
@@ -37,11 +37,11 @@ SIZE_BYTES=$(qemu-img info --output "json" /dev/sdb | grep -m1 "virtual-size" | 
 SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
 
 echo "GCEExport: Exporting disk of size ${SIZE_GB}GB and format ${FORMAT}."
-qemu-img convert /dev/sdb "/gcs/${GCS_PATH}" -p -O $FORMAT &> qemu_output.txt
-QEMU_ERR=$?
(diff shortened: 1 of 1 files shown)


daisy_workflows/export/export_disk_ext.sh
```diff
@@ -37,9 +37,10 @@ SIZE_BYTES=$(qemu-img info --output "json" /dev/sdb | grep -m1 "virtual-size" |
 SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
 
 echo "GCEExport: Exporting disk of size ${SIZE_GB}GB and format ${FORMAT}."
-qemu-img convert /dev/sdb "/gcs/${GCS_PATH}" -p -O $FORMAT
-if [ $? -ne 0 ]; then
-  echo "ExportFailed: Failed to export disk source to ${GS_PATH}."
+qemu-img convert /dev/sdb "/gcs/${GCS_PATH}" -p -O $FORMAT &> qemu_output.txt
```
hopkiw: i can recommend paradigm we use in some other scripts
```
if ! out=$(qemu-img ...); then
  echo "ExportFailed: qemu-img failed: ${out}"
fi
```

dntczdx: correct me if I am wrong: here "out" will catch all stdout, right?
If that's the case, due that qemu-img will output some info to stdout, we can't use "out" to decide whether it's success. We have to use return code.

dntczdx: Updated as Liam suggested to simplify. Please help to review this version.

PR #682, opened 2019-02-20
Add test cases for rhel-7

iamsubratp:
Needed to change packagemanager list to a tuple since centos/rhel uses the same package manager

### DIFF
--- a/osconfig_tests/test_suites/package_management/package_management_test_data.go
+++ b/osconfig_tests/test_suites/package_management/package_management_test_data.go
@@ -25,13 +25,18 @@ import (
 	api "google.golang.org/api/compute/v1"
 )
 
+type platformPkgManagerTuple struct {
+	platform   string
+	pkgManager string
+}
+
 const (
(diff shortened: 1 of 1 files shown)


osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -25,13 +25,18 @@ import (
 	api "google.golang.org/api/compute/v1"
 )
 
+type platformPkgManagerTuple struct {
+	platform   string
+	pkgManager string
```
adjackura: I don't see pkgManager used anywhere, is there a reason we need this?

iamsubratp: it is used to fetch the startup script, since the startup scripts are dependent on package manager
https://github.com/GoogleCloudPlatform/compute-image-tools/pull/682/files#diff-1f8f3e56a68ead097f839c047dbc57b2R113

PR #685, opened 2019-02-20
osconfg agent: Only take package action when needed

adjackura:
Only install packages if not installed/up to date
Only remove packages if installed
Only edit the repo files if the contents differ from expectation


adjackura:
Most of this code is just reorganized, the new logic is in ospackage.go and the various new test files.

go/packages/gem.go
```diff
@@ -0,0 +1,98 @@
...
+
+var (
+	gem string
+
+	gemListArgs     = []string{"list", "--local"}
+	gemOutdatedArgs = []string{"outdated", "--local"}
+)
+
+func init() {
+	if runtime.GOOS != "windows" {
```
iamsubratp: for my knowledge, why do we need this check?

adjackura: Mainly for future proofing, and to be very clear these package managers are not setup for Windows. We may support gem and pip in the future though.

Functionally there is no difference if I left this check out as nothing currently evaluates this, and even if it did these paths would not exist on Windows.

PR #686, opened 2019-02-21
 osconfig agent: hide features behind build tags

adjackura:
This adds a build tag of public for agent builds with only the inventory feature. This allows us to build packages with just the inventory features but still continue developing and testing the other unreleased features.

go build -tags public > creates an agent that should not have any ospackage or patch code or functionality.

### DIFF
--- a/cli_tools/google-osconfig-agent/main.go
+++ b/cli_tools/google-osconfig-agent/main.go
@@ -19,9 +19,7 @@ import (
 	"context"
 	"flag"
 	"log"
-	"os"
 
-	osconfig "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/_internal/gapi-cloud-osconfig-go/cloud.google.com/go/osconfig/apiv1alpha1"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/config"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/inventory"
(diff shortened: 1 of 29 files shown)


cli_tools/google-osconfig-agent/main.go
```diff
@@ -60,40 +57,32 @@ func main() {
...
 		res, err := config.Instance()
 		if err != nil {
-			logger.Fatalf("Instance error: %v", err)
-		}
-
-		resp, err := service.LookupConfigs(ctx, client, res)
-		if err != nil {
-			logger.Fatalf("LookupConfigs error: %v", err)
+			logger.Close()
+			logger.Fatalf("get instance error: %v", err)
```
iamsubratp: should the logger be closed after the logging?

and other places as well

iamsubratp:
> allows us to build packages with just the inventory features but still continue developing and testing the other unreleased features.

why can't we just park the upcoming features in unstable or staging rapture? one potential reason i can think of is that we only use master branch for this git repo. So, it would be difficult to release features that are developed after a feature  that is not supposed to go live. But, is there anything stopping us from starting to use a development branch? we can do all the PRs on development branch and selectively release features to master.

adjackura:
I think we might be able to use the combination of a new dev branch and a dev set of repos to have a similar split, it'll be a bit of a merge headache as this repo is overloaded though, also we will have to have be very strict about what files you touch in what branch to avoid unnecessary drift.

hopkiw:
i am not a fan of using build magic to hide code, especially in a language-specific way; i would prefer we use a more standardized development method such as feature flags or failing that, vcs solutions

PR #688, opened 2019-02-21
Clarify API authentication error, fixes #368

### LINKED ISSUE #368: Error validating workflow "workflow": error populating workflow: APIError: dialing: missing 'type' field in credentials
I'm not sure what `dialing` is...

I used https://console.cloud.google.com/apis/credentials to create an `OAuth 2.0 client IDs` and then downloaded the resulting JSON file.

rofuentes:
Error message is passed from utility method which creates a secure connection to
a Google API [doc link](https://godoc.org/google.golang.org/api/transport#NewHTTPClient)

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -208,7 +208,7 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (Client, error)
 	opts = append(o, opts...)
 	hc, ep, err := transport.NewHTTPClient(ctx, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("dialing: %v", err)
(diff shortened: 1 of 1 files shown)


daisy/compute/compute.go
```diff
@@ -208,7 +208,7 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (Client, error)
 	opts = append(o, opts...)
 	hc, ep, err := transport.NewHTTPClient(ctx, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("dialing: %v", err)
+		return nil, fmt.Errorf("creating secure connection: %v", err)
```
adjackura: How about:
error creating HTTP API client

PR #690, opened 2019-02-22
osconfig agent: don't run ospackage or osconfig unless set in config file

adjackura:
Code slightly reorganized, main change is OSPackageEnabled and OSPatchEnabled checks before running those features, the idea is they will be options in a config file. Right now I'm not thinking of using command line flags as a config file is easier to setup.
Config file reading will be in a follow up PR
The command line args ospatch and ospackage can still be used to run those features directly


cli_tools/google-osconfig-agent/main.go
```diff
@@ -60,46 +90,33 @@ func main() {
...
 		// Just run inventory and exit.
 		inventory.RunInventory()
-		logger.Close()
-		os.Exit(0)
-	}
-
-	if action == "ospackage" {
+		return
+	case "ospackage":
 		// Just run SetConfig and exit.
```
hopkiw: comment is no longer true, you're calling ospackage.Run instead of osconfig.SetConfig

cli_tools/google-osconfig-agent/main.go
```diff
@@ -60,46 +90,33 @@ func main() {
 	}
 
 	logger.Init(ctx, proj)
+	defer logger.Close()
 
-	action := flag.Arg(0)
-	if action == "inventory" {
+	switch action := flag.Arg(0); action {
+	case "":
+		if err := service.Register(ctx, "google_osconfig_agent", "Google OSConfig Agent", "", run, "run"); err != nil {
```
hopkiw: where did you previously run service.Register?

cli_tools/google-osconfig-agent/main.go
```diff
@@ -46,6 +45,37 @@ func (l *logWritter) Write(b []byte) (int, error) {
 	return len(b), nil
 }
 
+func run(ctx context.Context) {
```
hopkiw: the reason for this being a separate func is not clear.. you're not using it in a goroutine or passing in arguments

cli_tools/google-osconfig-agent/main.go
adjackura: Removed comment as I think this code is pretty clear now

cli_tools/google-osconfig-agent/main.go
adjackura: In service.go, that separation and indirection was not accomplishing anything so I moved it here. 'register' here takes an action 'run' to register the service and run it, really this just runs the code on linux and on windows runs the code with the correct hooks for the windows service system. Othe args of install, uninstall, start in the service library can no longer get passed on the command line as the package installs do that work.

cli_tools/google-osconfig-agent/main.go
adjackura: See service.Register, run is passed in as an arg and invoked by that library.

cli_tools/google-osconfig-agent/main.go
```diff
@@ -60,46 +90,31 @@ func main() {
...
-		if err != nil {
-			logger.Fatalf("NewClient Error: %v", err)
+	switch action := flag.Arg(0); action {
+	case "":
+		if err := service.Register(ctx, "google_osconfig_agent", "Google OSConfig Agent", "", run, "run"); err != nil {
+			logger.Fatalf("service.Register error: %v", err)
 		}
-
+	case "noservice":
+		run(ctx)
```
ryanwe: Should you have a `return` here? It doesn't seem necessary but keeps this consistent with the other cases.

cli_tools/google-osconfig-agent/main.go
```diff
@@ -60,46 +90,31 @@ func main() {
...
-		}
-
-		resp, err := service.LookupConfigs(ctx, client, res)
-		if err != nil {
-			logger.Fatalf("LookupConfigs error: %v", err)
+			logger.Fatalf("get instance error: %v", err)
 		}
-		if err := ospackage.SetConfig(resp); err != nil {
-			logger.Fatalf("SetConfig error: %v", err)
+		if err := ospackage.Run(ctx, res, false); err != nil {
```
ryanwe: It's not clear what `false` means here. Consider using an enum/const or separate functions.

cli_tools/google-osconfig-agent/ospackage/ospackage.go
```diff
@@ -25,14 +26,78 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("osconfig.NewClient Error: %v", err)
+	}
+
+	resp, err := lookupConfigs(ctx, client, res)
+	if err != nil {
+		return fmt.Errorf("LookupConfigs error: %v", err)
+	}
+
+	if enqueue {
```
ryanwe: shouldn't we always enqueue and know that the queue is empty and this will run right away if nothing else is happening?

cli_tools/google-osconfig-agent/main.go
adjackura: Separated functions for the public interface

cli_tools/google-osconfig-agent/ospackage/ospackage.go
adjackura: Right now we have no way to wait on tasks, once we do we can go this route as it would be cleaner

PR #691, opened 2019-02-25
[OsConfig Agent] Make projectID configurable through flags

iamsubratp:


test-infra/prow/config.yaml
```diff
@@ -127,6 +127,8 @@ periodics:
     - image: gcr.io/compute-image-tools-test/osconfig-tests:latest
       args:
       - "-out_dir=/artifacts"
+      - "-test_project_id=compute-image-osconfigagent"
```
adjackura: If this is a new project there's a bunch of stuff we need to do to set it up, let me know and I can walk you through it. Also we should probably create a pool of projects for osconfig tests to run in as we will have locking issues.

osconfig_tests/test_config/project_config.go
```diff
@@ -0,0 +1,45 @@
...
+	tpn, err := cloudresourcemanager.GetProjectNumberFromProjectID(tpid)
+	if err != nil {
+		fmt.Printf("%+v", err)
+		return nil, err
+	}
+
+	return &ProjectConfig{
+		TestProjectID:        tpid,
+		TestProjectNumber:    tpn,
+		ServiceAccountEmail:  fmt.Sprintf("%s-compute@developer.gserviceaccount.com", tpn),
```
adjackura: just use "default", no need to do any of the GetProjectNumberFromProjectID stuff as the osconfig api can take ProjectID just fine. I would just remove this package completly as it's not doing much once you remove the lookup and it's not universally useful accros tests.

iamsubratp: yeah, the reason i introduced the cloudresourcemanager package was because i was under the impression that projectID is not allowed for api calls.
however, i would still insist on keeping the projectconfig struct so that all the project related configurations are clubbed together.
I am thinking of moving it to osconfig_tests/main and so the projectconfig can be created once and used in all the test suites( i could have done this before sending the PR itself)

test-infra/prow/config.yaml
iamsubratp: yeah, i did some setup around iam. But i was not sure, if that is going to be enough.
Btw, if these are some default steps that we always do when we setup a project for compute-image-test-*, would it be better to just write an automation to do the basic setup?

osconfig_tests/test_config/project_config.go
adjackura: Do you still want the project number? It'd be better to use the compute client your already have (GetProject) instead of invoking yet another api client.

iamsubratp: i dont need the projectNumber, will remove it from the struct. but we still need the projectid, serviceaccount, and other information

iamsubratp: ok.. instead of moving it to main package, i thought the best place for the project info is the compute package.

osconfig_tests/main.go
```diff
@@ -5,7 +5,6 @@
 //  You may obtain a copy of the License at
 //
 //      http://www.apache.org/licenses/LICENSE-2.0
-//
```
ryanwe: why did you change this?

osconfig_tests/compute/instance.go
```diff
@@ -89,3 +105,19 @@ func CreateInstance(client daisyCompute.Client, project, zone string, i *api.Ins
 	}
 	return &Instance{Instance: i, client: client, Project: project, Zone: zone}, nil
 }
+
+// GetProject fetches the default service account used for the projectID
+func GetProject(client daisyCompute.Client, projectID, testZone string) (*Project, error) {
```
ryanwe: I didn't realize we used Daisy. Can you remind me why we need it?

iamsubratp: daisy is used for the GCE related actions. Until now, we used it for VM spin up. But since daisy also supports all the compute APIs, i decided to use to to fetch the project details

osconfig_tests/main.go
iamsubratp: this was not deliberate. will remove it.

iamsubratp: done

osconfig_tests/compute/instance.go
ryanwe: Sounds good. I think at some point, we'll want to use Compute directly and remove that dependency. But this is fine for now.

osconfig_tests/test_config/project_config.go
```diff
@@ -0,0 +1,35 @@
...
+	ServiceAccountScopes []string
+}
+
+// GetProject ...
+func GetProject(projectID, testZone string) *Project {
+	return &Project{
+		TestProjectID:       projectID,
+		TestZone:            testZone,
+		ServiceAccountEmail: "default",
+		ServiceAccountScopes: []string{
```
adjackura: We're going to want to make this an arg so that we can ensure things like inventory work without a service account, fine for now though.

iamsubratp: inventory does not use this service account. although it has this information. so, yeah, in future we might want to not provide information that are not required by the different sub-agent

PR #693, opened 2019-02-26
Export: Change "qemu-img" to use disk as buffer, and then copy output to GCS gssutil

dntczdx:
This CL changes the way of using qemu-img:
Original: qemu-img output file to GCS by gcsfuse
Now: qemu-img output file to disk, then copy to GCS by gsutil

It achieves 3 purposes:
1. When qemu-img failed due to disk space or network issue, the workflow will fail. Previously it failed silently and the workflow got a success, and thus output a corrupt image.
2. Much higher success rate for large-size image, due that gsutil is much more reliable than gcsfuse, especially when network is not very stable. For a 198GB image, old method success rate is 29% while new method is 100% per my testing.
3. It made the code ready for next improvement: dynamically enlarge disk space to export >200GB image.

Conclusion: this CL resolved problems for <200GB image export, and technically prepared for >200GB improvement.

Concern of time efficiency: there was a concern that new method may double the time consuming. Below test result shows the concern is invalid:
test

### DIFF
--- a/daisy_workflows/export/export_disk_ext.sh
+++ b/daisy_workflows/export/export_disk_ext.sh
@@ -21,8 +21,8 @@ FORMAT=$(curl -f -H Metadata-Flavor:Google ${URL}/format)
 # Strip gs://
 LOCAL_PATH=${GS_PATH##*//}
 # Create dir for output
-OUTS=${LOCAL_PATH%/*}
-mkdir -p "/gs/${OUTS}"
+OUTS_PATH=${LOCAL_PATH%/*}
+mkdir -p "/gs/${OUTS_PATH}"
 
 # Disk image size info.
 SIZE_BYTES=$(qemu-img info --output "json" /dev/sdb | grep -m1 "virtual-size" | grep -o '[0-9]\+')
@@ -38,7 +38,7 @@ fi
 echo ${out}
 
 if ! out=$(gsutil cp "/gs/${LOCAL_PATH}" "${GS_PATH}"); then
-  echo "ExportFiled: Failed to copy output image to ${GS_PATH}"
+  echo "ExportFiled: Failed to copy output image to ${GS_PATH}, error: ${out}"
   exit
 fi
 echo ${out}
--- a/daisy_workflows/export/export_disk_ext.sh
+++ b/daisy_workflows/export/export_disk_ext.sh
@@ -37,7 +37,7 @@ if ! out=$(qemu-img convert /dev/sdb "/gs/${LOCAL_PATH}" -p -O $FORMAT 2>&1); th
 fi
 echo ${out}
 
-if ! out=$(gsutil cp "/gs/${LOCAL_PATH}" "${GS_PATH}"); then
+if ! out=$(gsutil cp "/gs/${LOCAL_PATH}" "${GS_PATH}" 2>&1); then
   echo "ExportFiled: Failed to copy output image to ${GS_PATH}, error: ${out}"
   exit
 fi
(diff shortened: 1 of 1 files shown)


daisy_workflows/export/export_disk_ext.sh
```diff
@@ -19,17 +19,10 @@ GS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/gcs-path)
...
-mkdir -p "/gcs/${GCS_BUCKET}"
-gcsfuse ${GCS_BUCKET} /gcs/${GCS_BUCKET}
-if [ $? -ne 0 ]; then
-  echo "ExportFailed: Failed to mount ${GCS_BUCKET}."
-fi
-OUTS=${GCS_PATH%/*}
-mkdir -p "/gcs/${OUTS}"
+LOCAL_PATH=${GS_PATH##*//}
+# Create dir for output
+OUTS=${LOCAL_PATH%/*}
```
hopkiw: what is $OUTS used for?

daisy_workflows/export/export_disk_ext.sh
```diff
@@ -38,10 +31,17 @@ SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
...
 echo "GCEExport: Exporting disk of size ${SIZE_GB}GB and format ${FORMAT}."
 
-if ! out=$(qemu-img convert /dev/sdb "/gcs/${GCS_PATH}" -p -O $FORMAT 2>&1); then
+if ! out=$(qemu-img convert /dev/sdb "/gs/${LOCAL_PATH}" -p -O $FORMAT 2>&1); then
   echo "ExportFailed: Failed to export disk source to ${GS_PATH} due to qemu-img error: ${out}"
+  exit
 fi
+echo ${out}
 
+if ! out=$(gsutil cp "/gs/${LOCAL_PATH}" "${GS_PATH}"); then
```
hopkiw: need `2>&1` ?

daisy_workflows/export/export_disk_ext.sh
```diff
@@ -38,10 +31,17 @@ SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
...
+  echo "ExportFiled: Failed to copy output image to ${GS_PATH}"
```
hopkiw: s/Filed/Failed/
you exit here without printing the gsutil command output, is that intentional?

daisy_workflows/export/export_disk_ext.sh
dntczdx: to create dir for outputs in below line. let me rename it to "OUTS_PATH" to clarify.

daisy_workflows/export/export_disk_ext.sh
dntczdx: No I missed that part, thanks :P

dntczdx:
Updated accordingly

adjackura:
This will fail for anything over 200Gb correct? If that's the case we need to be sure we don';t push this in a release. How close is the follow up PR? I would prefer to have the full fix in one PR so there isn't a period of time where export code will simply not work.

daisy_workflows/export/export_disk_ext.sh
dntczdx: true...

dntczdx:
> This will fail for anything over 200Gb correct? If that's the case we need to be sure we don';t push this in a release. How close is the follow up PR? I would prefer to have the full fix in one PR so there isn't a period of time where export code will simply not work.

Hi Andrew, This will fail for anything over 200Gb, just as how it behaves today. It won't fail any case that is passing today.

dntczdx:
> > This will fail for anything over 200Gb correct? If that's the case we need to be sure we don';t push this in a release. How close is the follow up PR? I would prefer to have the full fix in one PR so there isn't a period of time where export code will simply not work.
> 
> Hi Andrew, This will fail for anything over 200Gb, just as how it behaves today. It won't fail any case that is passing today.

to be accurate, it will fail anything >200Gb after compression

PR #703, opened 2019-03-02
Resize buffer disk dynamically

dntczdx:
Today, a large-enough disk (which means, more than 200GB after compression) can't be exported with format converting. This CL fixed the problem.

Originally: Worker instance use a 200GB for both buffering and OS. Now: It creates a separate buffer disk starting from 200GB and will increase 25GB when there is not enough space during the exporing process.
Detail of the resizing logic:
200GB as initial size.
Check available space every 10s.
If available space is less than 25GB, then add 25GB more.
Unless it has reached the max size. Max size is source disk + 5GB.
Resizing logic is host in a separate shell script "disk_resizing_mon.sh", which runs in background.
OS is hosted in a default size disk, which is 10GB now. It helps to isolate buffer size from OS, because OS size can be different in the future.
This design won't cost too much space - it applies new space only when it's required.
It unblocks >200GB scenario for customers. Ideally it can support the max disk size of google cloud - 64TB. We need test to verify how big it can actually handle. Manual test for 400GB+ works well.

Test result 1: perf regression test. Disk size 198GB.
Originally: 88min Now: 78min. Basically the same level.

Test result 2: large disk (700GB before compression, 493GB after compression) test.
Originally: 100% fail. Now: 100% pass

### DIFF
--- a/daisy_workflows/export/disk_export_ext.wf.json
+++ b/daisy_workflows/export/disk_export_ext.wf.json
@@ -20,7 +20,7 @@
     },
     "buffer_disk_size": {
       "Value": "200",
-      "Description": "size of the export instances buffer disk, this disk starts from a larger size to increase PD read speed, and grows on demand"
+      "Description": "size of the export instances buffer disk, this disk starts from a fixed size which can guarantee a nice PD read speed, and grows on demand"
     },
     "export_network": {
       "Value": "global/networks/default",
@@ -41,12 +41,14 @@
         {
           "Name": "disk-${NAME}-os",
           "SourceImage": "${export_instance_disk_image}",
-          "Type": "pd-ssd"
+          "Type": "pd-ssd",
+	  "ExactName": true
         },
         {
-          "Name": "disk-${NAME}-buffer",
+          "Name": "disk-${NAME}-buffer-${ID}",
           "SizeGb": "${buffer_disk_size}",
-          "Type": "pd-ssd"
+          "Type": "pd-ssd",
+	  "ExactName": true
         }
       ]
     },
@@ -57,7 +59,7 @@
           "Disks": [
             {"Source": "disk-${NAME}-os"},
             {"Source": "${source_disk}", "Mode": "READ_ONLY"},
-            {"Source": "disk-${NAME}-buffer"}
+            {"Source": "disk-${NAME}-buffer-${ID}"}
           ],
           "MachineType": "n1-highcpu-4",
           "Metadata": {
@@ -65,7 +67,7 @@
             "gcs-path": "${OUTSPATH}/${NAME}",
             "format": "${format}",
 	    "instance-name": "inst-${NAME}",
-            "buffer-disk": "disk-${NAME}-buffer"
+            "buffer-disk": "disk-${NAME}-buffer-${ID}"
           },
           "networkInterfaces": [
             {
--- a/daisy_workflows/export/disk_export_ext.wf.json
+++ b/daisy_workflows/export/disk_export_ext.wf.json
@@ -18,9 +18,9 @@
       "Value": "projects/compute-image-tools/global/images/family/debian-9-worker",
       "Description": "image to use for the exporter instance"
     },
-    "export_instance_disk_size": {
+    "buffer_disk_size": {
       "Value": "200",
-      "Description": "size of the export instances disk, this disk is unused for the export but a larger size increase PD read speed"
+      "Description": "size of the export instances buffer disk, this disk starts from a larger size to increase PD read speed, and grows on demand"
     },
     "export_network": {
       "Value": "global/networks/default",
@@ -32,15 +32,20 @@
     }
   },
   "Sources": {
-    "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh"
+    "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh",
+    "disk_resizing_mon.sh": "./disk_resizing_mon.sh"
   },
   "Steps": {
     "setup-disks": {
       "CreateDisks": [
         {
-          "Name": "disk-${NAME}",
+          "Name": "disk-${NAME}-os",
           "SourceImage": "${export_instance_disk_image}",
-          "SizeGb": "${export_instance_disk_size}",
+          "Type": "pd-ssd"
+        },
+        {
+          "Name": "disk-${NAME}-buffer",
+          "SizeGb": "${buffer_disk_size}",
           "Type": "pd-ssd"
         }
       ]
@@ -49,12 +54,18 @@
       "CreateInstances": [
         {
           "Name": "inst-${NAME}",
-          "Disks": [{"Source": "disk-${NAME}"}, {"Source": "${source_disk}", "Mode": "READ_ONLY"}],
+          "Disks": [
+            {"Source": "disk-${NAME}-os"},
+            {"Source": "${source_disk}", "Mode": "READ_ONLY"},
+            {"Source": "disk-${NAME}-buffer"}
+          ],
           "MachineType": "n1-highcpu-4",
           "Metadata": {
             "block-project-ssh-keys": "true",
             "gcs-path": "${OUTSPATH}/${NAME}",
-            "format": "${format}"
+            "format": "${format}",
+	    "instance-name": "inst-${NAME}",
+            "buffer-disk": "disk-${NAME}-buffer"
           },
           "networkInterfaces": [
             {
@@ -62,7 +73,7 @@
               "subnetwork": "${export_subnet}"
             }
           ],
-          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control"],
+          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control", "https://www.googleapis.com/auth/cloud-platform"],
           "StartupScript": "${NAME}_export_disk_ext.sh"
         }
       ]
--- a/daisy_workflows/export/disk_export_ext.wf.json
+++ b/daisy_workflows/export/disk_export_ext.wf.json
@@ -39,7 +39,7 @@
     "setup-disks": {
       "CreateDisks": [
         {
-          "Name": "disk-${NAME}-os",
+          "Name": "disk-${NAME}-os-${ID}",
           "SourceImage": "${export_instance_disk_image}",
           "Type": "pd-ssd",
 	  "ExactName": true
@@ -57,7 +57,7 @@
         {
           "Name": "inst-${NAME}",
           "Disks": [
-            {"Source": "disk-${NAME}-os"},
+            {"Source": "disk-${NAME}-os-${ID}"},
             {"Source": "${source_disk}", "Mode": "READ_ONLY"},
             {"Source": "disk-${NAME}-buffer-${ID}"}
           ],
--- a/daisy_workflows/export/disk_export_ext.wf.json
+++ b/daisy_workflows/export/disk_export_ext.wf.json
@@ -18,9 +18,9 @@
       "Value": "projects/compute-image-tools/global/images/family/debian-9-worker",
       "Description": "image to use for the exporter instance"
     },
-    "buffer_disk_size": {
+    "export_instance_disk_size": {
       "Value": "200",
-      "Description": "size of the export instances buffer disk, this disk starts from a fixed size which can guarantee a nice PD read speed, and grows on demand"
+      "Description": "size of the export instances buffer disk, this disk starts from a fixed size which can guarantee an acceptable PD read speed, and grows on demand"
     },
     "export_network": {
       "Value": "global/networks/default",
@@ -33,20 +33,19 @@
   },
   "Sources": {
     "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh",
-    "disk_resizing_mon.sh": "./disk_resizing_mon.sh"
+    "${NAME}_disk_resizing_mon.sh": "./disk_resizing_mon.sh"
   },
   "Steps": {
     "setup-disks": {
       "CreateDisks": [
         {
-          "Name": "disk-${NAME}-os-${ID}",
+          "Name": "disk-${NAME}-os",
           "SourceImage": "${export_instance_disk_image}",
-          "Type": "pd-ssd",
-	  "ExactName": true
+          "Type": "pd-ssd"
         },
         {
           "Name": "disk-${NAME}-buffer-${ID}",
-          "SizeGb": "${buffer_disk_size}",
+          "SizeGb": "${export_instance_disk_size}",
           "Type": "pd-ssd",
 	  "ExactName": true
         }
@@ -57,7 +56,7 @@
         {
           "Name": "inst-${NAME}",
           "Disks": [
-            {"Source": "disk-${NAME}-os-${ID}"},
+            {"Source": "disk-${NAME}-os"},
             {"Source": "${source_disk}", "Mode": "READ_ONLY"},
             {"Source": "disk-${NAME}-buffer-${ID}"}
           ],
@@ -66,8 +65,8 @@
             "block-project-ssh-keys": "true",
             "gcs-path": "${OUTSPATH}/${NAME}",
             "format": "${format}",
-	    "instance-name": "inst-${NAME}",
-            "buffer-disk": "disk-${NAME}-buffer-${ID}"
+            "buffer-disk": "disk-${NAME}-buffer-${ID}",
+	    "resizing-script-name": "${NAME}_disk_resizing_mon.sh"
           },
           "networkInterfaces": [
             {
@@ -75,7 +74,7 @@
               "subnetwork": "${export_subnet}"
             }
           ],
-          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control", "https://www.googleapis.com/auth/cloud-platform"],
+          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control", "https://www.googleapis.com/auth/compute"],
           "StartupScript": "${NAME}_export_disk_ext.sh"
         }
       ]
--- a/daisy_workflows/export/disk_resizing_mon.sh
+++ b/daisy_workflows/export/disk_resizing_mon.sh
@@ -3,53 +3,36 @@
 METADATA_URL="http://metadata/computeMetadata/v1/instance"
 
 MAX_SIZE=$1
+BUFFER_MIN_SIZE=10
 BUFFER_SIZE=25
 INTERVAL=10
 
-echo "Max disk size ${MAX_SIZE}GB, buffer size ${BUFFER_SIZE}GB, starting monitoring available disk buffer every ${INTERVAL}s..."
-while [ 1 ]
-do
-  sleep ${INTERVAL}
+# Prepare parameters for resizing
+ZONE=$(curl "${METADATA_URL}/zone" -H "Metadata-Flavor: Google"| cut -d'/' -f4)
+if [ $? -ne 0 ]; then
+  echo "GCEExport: Failed to get metadata from ${METADATA_URL}/zone, will try again later."
+  continue
+fi
+BUFFER_DISK=$(curl "${METADATA_URL}/attributes/buffer-disk" -H "Metadata-Flavor: Google")
 
+echo "Max disk size ${MAX_SIZE}GB, min buffer size ${BUFFER_SIZE}GB, starting monitoring available disk buffer every ${INTERVAL}s..."
+while sleep ${INTERVAL};
+do
   # Check whether available buffer space is lower than threshold.
   AVAILABLE_BUFFER=$(df -BG /dev/sdc --output=avail | sed -n 2p)
   AVAILABLE_BUFFER=${AVAILABLE_BUFFER%?}
-  if [ ${AVAILABLE_BUFFER} -ge ${BUFFER_SIZE} ]; then
+  if [ ${AVAILABLE_BUFFER} -ge ${BUFFER_MIN_SIZE} ]; then
+    echo "ge!!!! ${AVAILABLE_BUFFER} vs ${BUFFER_MIN_SIZE}  "
     continue
   fi
+  echo "not ge!!!! ${AVAILABLE_BUFFER} vs ${BUFFER_MIN_SIZE}  "
 
-  # Decide the new size, which won't be higher than max size.
-  CURRENT_SIZE=$(lsblk /dev/sdc --output=SIZE | sed -n 2p)
-  CURRENT_SIZE=${CURRENT_SIZE%?}
-  NEXT_SIZE=$(awk "BEGIN {print int(${CURRENT_SIZE} + ${BUFFER_SIZE})}")
-  if [ ${NEXT_SIZE} -gt ${MAX_SIZE} ]; then
-    NEXT_SIZE=${MAX_SIZE}
-  fi
+  # Decide the new size of the device.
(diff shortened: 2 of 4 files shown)


daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -18,9 +18,9 @@
       "Value": "projects/compute-image-tools/global/images/family/debian-9-worker",
       "Description": "image to use for the exporter instance"
     },
-    "export_instance_disk_size": {
+    "buffer_disk_size": {
       "Value": "200",
-      "Description": "size of the export instances disk, this disk is unused for the export but a larger size increase PD read speed"
+      "Description": "size of the export instances buffer disk, this disk starts from a larger size to increase PD read speed, and grows on demand"
```
hopkiw: 'larger' is a comparative word, consider using the real size here in description

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+METADATA_URL="http://metadata/computeMetadata/v1/instance"
+
+MAX_SIZE=$1
+BUFFER_SIZE=25
+INTERVAL=10
+
+echo "Max disk size ${MAX_SIZE}GB, buffer size ${BUFFER_SIZE}GB, starting monitoring available disk buffer every ${INTERVAL}s..."
+while [ 1 ]
```
hopkiw: `while true; do`

hopkiw: or better (since primary purpose of loop is action+sleep), `while sleep ${INTERVAL}; do`

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
...
+  # Decide the new size, which won't be higher than max size.
+  CURRENT_SIZE=$(lsblk /dev/sdc --output=SIZE | sed -n 2p)
+  CURRENT_SIZE=${CURRENT_SIZE%?}
+  NEXT_SIZE=$(awk "BEGIN {print int(${CURRENT_SIZE} + ${BUFFER_SIZE})}")
+  if [ ${NEXT_SIZE} -gt ${MAX_SIZE} ]; then
+    NEXT_SIZE=${MAX_SIZE}
+  fi
+
+  # Prepare parameters for resizing
+  ZONE=$(curl "${METADATA_URL}/zone" -H "Metadata-Flavor: Google"| cut -d'/' -f4)
```
hopkiw: these checks don't need to happen on every loop.. and errors should probably cause a failure, otherwise you're likely to see infinite repetition in logs

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
...
+  if [ $? -ne 0 ]; then
+    echo "GCEExport: Failed to get metadata from ${METADATA_URL}/zone, will try again later."
+    continue
+  fi
+  INSTANCE_NAME_PREFIX=$(curl "${METADATA_URL}/attributes/instance-name" -H "Metadata-Flavor: Google")
+  if [ $? -ne 0 ]; then
+    echo "GCEExport: Failed to get metadata from ${METADATA_URL}/attributes/instance-name, will try again later."
+    continue
+  fi
+  HOST_NAME=$(curl "${METADATA_URL}/hostname" -H "Metadata-Flavor: Google"| cut -d'.' -f1)
```
hopkiw: is there a reason we don't trust the `hostname` command?

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
...
+  fi
+  BUFFER_DISK_PREFIX=$(curl "${METADATA_URL}/attributes/buffer-disk" -H "Metadata-Flavor: Google")
+  if [ $? -ne 0 ]; then
+    echo "GCEExport: Failed to get metadata from ${METADATA_URL}/attributes/buffer-disk, will try again later."
+    continue
+  fi
+  BUFFER_DISK=${BUFFER_DISK_PREFIX}-$(echo ${HOST_NAME} | awk -F"${INSTANCE_NAME_PREFIX}-" '{ print $2}')
+
+  echo "GCEExport: Resizing buffer disk from ${CURRENT_SIZE}GB to ${NEXT_SIZE}GB..."
+  if ! out=$(gcloud compute disks resize ${BUFFER_DISK} --size=${NEXT_SIZE}GB --quiet --zone "${ZONE}" 2>&1); then
```
hopkiw: you only need to capture the output to a variable if you will use it conditionally or in another command..you echo it in both failure and success cases, so just remove the subshell wrapping

hopkiw: or dont echo it on success..

daisy_workflows/export/export_disk_ext.sh
```diff
@@ -14,31 +14,56 @@
...
-SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
+# 2. Round up to the next GB.
+SIZE_OUTPUT_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000) + 1)}")
+# 3. Add 5GB of additional space to max size to prevent the corner case that output
+# file is slightly larger than source disk.
+MAX_BUFFER_DISK_SIZE_GB=$(awk "BEGIN {print int(${SIZE_OUTPUT_GB} + 5)}")
+
+# Prepare buffer disk
+echo "GCEExport: Initializing buffer disk for qemu-img output..."
+sudo arted -a optimal /dev/sdc mklabel gpt
```
hopkiw: parted

hopkiw: there's no error checking on these commands..

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
...
+
+echo "Max disk size ${MAX_SIZE}GB, buffer size ${BUFFER_SIZE}GB, starting monitoring available disk buffer every ${INTERVAL}s..."
+while [ 1 ]
+do
+  sleep ${INTERVAL}
+
+  # Check whether available buffer space is lower than threshold.
+  AVAILABLE_BUFFER=$(df -BG /dev/sdc --output=avail | sed -n 2p)
+  AVAILABLE_BUFFER=${AVAILABLE_BUFFER%?}
+  if [ ${AVAILABLE_BUFFER} -ge ${BUFFER_SIZE} ]; then
```
hopkiw: use double bracket syntax in bash, e.g. `if [[ $AVAILABLE_BUFFER -ge $BUFFER_SIZE ]]; then`

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
...
+    continue
+  fi
+
+  # Decide the new size, which won't be higher than max size.
+  CURRENT_SIZE=$(lsblk /dev/sdc --output=SIZE | sed -n 2p)
+  CURRENT_SIZE=${CURRENT_SIZE%?}
+  NEXT_SIZE=$(awk "BEGIN {print int(${CURRENT_SIZE} + ${BUFFER_SIZE})}")
+  if [ ${NEXT_SIZE} -gt ${MAX_SIZE} ]; then
```
hopkiw: shouldn't you error on this condition?

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,68 @@
...
+  if [ $? -ne 0 ]; then
+    echo "GCEExport: Failed to get metadata from ${METADATA_URL}/hostname, will try again later."
+    continue
+  fi
+  BUFFER_DISK_PREFIX=$(curl "${METADATA_URL}/attributes/buffer-disk" -H "Metadata-Flavor: Google")
+  if [ $? -ne 0 ]; then
+    echo "GCEExport: Failed to get metadata from ${METADATA_URL}/attributes/buffer-disk, will try again later."
+    continue
+  fi
+  BUFFER_DISK=${BUFFER_DISK_PREFIX}-$(echo ${HOST_NAME} | awk -F"${INSTANCE_NAME_PREFIX}-" '{ print $2}')
```
hopkiw: what does this command aim to do?

daisy_workflows/export/export_disk_ext.sh
hopkiw: why create a label if you're putting a filesystem directly on the root disk btw?

daisy_workflows/export/disk_resizing_mon.sh
adjackura: See here, this is much better to pass in from the workflow that expect the startup script to somehow guess at the name:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_import/import_image.sh

adjackura: This is probably what you missing from the workflow:
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_import/import_image.wf.json#L29

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: it's a network call so it can fail.

dntczdx: Since I didn't know whether it can be an overwritten concept in  cloud project. It seems that it isn't, so yes I can use "hostname".

daisy_workflows/export/export_disk_ext.sh
dntczdx: no need for a label, removed.
I didn't add error check for local operations since I assumed they will success (they have lower chance to fail, comparing to network calls), and if not, the following steps will fail due to file system problem.
Do you think it's valuable to add error check? I feel that made the whole process too complex.

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: In error path, the output needs to be embedded in a "ExportFailed:" msg, which is required for daisy to capture. It works as a "thrown exception" somehow. So I do need this variable.

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: no, its purpose is to prevent unused extra space assignment. anyways I need to remove it because of the change of the design

daisy_workflows/export/export_disk_ext.sh
hopkiw: well what is the expected outcome if either mkfs or mklabel fail? what will log output?

daisy_workflows/export/disk_resizing_mon.sh
hopkiw: the requirement is to print 'ExportFailed' at beginning of line on failure. that's orthogonal to the capture strategy here.. you can do:
```
if ! gcloud compute ....; then
  echo ExportFailed: failed to export
fi
```

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: I followed the pattern in the codebase which usually use single '['. It may also have better compatibility. May I know why '[[' is prefered?

hopkiw: [ and [[ are different functionally, and [[ is much newer and follows bash conventions in an easier to understand way. it's also recommended in our style guide: 
https://google.github.io/styleguide/shell.xml#Test,_%5B_and_%5B%5B

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: I understand the point. However, I need to avoid newline after echo. I did a bit research, it seems that different implementation of 'echo' has inconsistent support on this. I want to get rid of the dependency on that. Does that sound good?

daisy_workflows/export/export_disk_ext.sh
dntczdx: if mkfs fail, then mount will also fail. And the whole workflow ultimately failed on qemu outputing step. We can't get an accurate error msg in that case. I was trying to balance the probability of failure v.s. code complexity.

Shell scripts in this code base always contain a serial of linux commands. Per my observation, local command usually doesn't have an error handling logic. I think it's because the probability is low. An example: mounting a pd ssd doesn't have error handling, while mounting a GCS path via gcsfuse has error handling.

So I didn't add error handling for mkfs to keep the code simple and clear. But if mkfs actually has a reasonably-high probability to fail, then I totally agree to add error handling. How do you think?

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: offline synced, will change to double bracket as style guide suggested.

dntczdx:
Addressed comments and changed design of some details

daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -18,9 +18,9 @@
       "Value": "projects/compute-image-tools/global/images/family/debian-9-worker",
       "Description": "image to use for the exporter instance"
     },
-    "export_instance_disk_size": {
+    "buffer_disk_size": {
```
adjackura: We can't go changing these variable names, there somewhat of a public api

daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -32,37 +32,50 @@
...
-    "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh"
+    "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh",
+    "disk_resizing_mon.sh": "./disk_resizing_mon.sh"
   },
   "Steps": {
     "setup-disks": {
       "CreateDisks": [
         {
-          "Name": "disk-${NAME}",
+          "Name": "disk-${NAME}-os-${ID}",
```
adjackura: This disk doesnt need the id or exact name set as it is unused in the script (you just need this for the buffer disk)

daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -32,37 +32,50 @@
     }
   },
   "Sources": {
-    "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh"
+    "${NAME}_export_disk_ext.sh": "./export_disk_ext.sh",
+    "disk_resizing_mon.sh": "./disk_resizing_mon.sh"
```
adjackura: Prepend ${NAME} for this script as well, it makes this source file unique so you can call this workflow multiple times in the same context.

daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -32,37 +32,50 @@
...
+            "buffer-disk": "disk-${NAME}-buffer-${ID}"
           },
           "networkInterfaces": [
             {
               "network": "${export_network}",
               "subnetwork": "${export_subnet}"
             }
           ],
-          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control"],
+          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control", "https://www.googleapis.com/auth/cloud-platform"],
```
adjackura: cloud-platform scope is too broad, all you should need is https://www.googleapis.com/auth/compute

daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -32,37 +32,50 @@
...
+            {"Source": "${source_disk}", "Mode": "READ_ONLY"},
+            {"Source": "disk-${NAME}-buffer-${ID}"}
+          ],
           "MachineType": "n1-highcpu-4",
           "Metadata": {
             "block-project-ssh-keys": "true",
             "gcs-path": "${OUTSPATH}/${NAME}",
-            "format": "${format}"
+            "format": "${format}",
+	    "instance-name": "inst-${NAME}",
```
adjackura: This appears to be unused, also this is not the actual instance name as it will have a daisy generated name if ExactName isnt set

daisy_workflows/export/disk_export_ext.wf.json
```diff
@@ -18,9 +18,9 @@
       "Value": "projects/compute-image-tools/global/images/family/debian-9-worker",
       "Description": "image to use for the exporter instance"
     },
-    "export_instance_disk_size": {
+    "buffer_disk_size": {
       "Value": "200",
-      "Description": "size of the export instances disk, this disk is unused for the export but a larger size increase PD read speed"
+      "Description": "size of the export instances buffer disk, this disk starts from a fixed size which can guarantee a nice PD read speed, and grows on demand"
```
zoran15: "nice speed" is not a technical term. Find a better adjective, such as "acceptable" or similar.

daisy_workflows/export/image_export_ext.wf.json
```diff
@@ -18,9 +18,9 @@
       "Value": "projects/compute-image-tools/global/images/family/debian-9-worker",
       "Description": "image to use for the exporter instance"
     },
-    "export_instance_disk_size": {
+    "buffer_disk_size": {
       "Value": "200",
-      "Description": "size of the export instances disk, this disk is unused for the export but a larger size increase PD read speed"
+      "Description": "size of the export instances buffer disk, this disk starts from a fixed size which can guarantee a nice PD read speed, and grows on demand"
```
zoran15: Same comment for "nice speed".

dntczdx:
Addressed comments, please have a look again :D

daisy_workflows/export/export_disk_ext.sh
hopkiw: i think the logic for where to add error handling is related partly to failure risk, i.e. 'ls' command is unlikely to fail for files you have created. It is also related to logic of the program. In this case i'm not sure what the result will be, so I suggest we either clarify that we have confidence of the failure case (what will the failure mode be, how will it be presented) or add add'l checks.

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,50 @@
...
+BUFFER_SIZE=25
+INTERVAL=10
+
+# Prepare parameters for resizing
+ZONE=$(curl "${METADATA_URL}/zone" -H "Metadata-Flavor: Google"| cut -d'/' -f4)
+BUFFER_DISK=$(curl "${METADATA_URL}/attributes/buffer-disk" -H "Metadata-Flavor: Google")
+
+echo "Max disk size ${MAX_SIZE}GB, min buffer size ${BUFFER_SIZE}GB, starting monitoring available disk buffer every ${INTERVAL}s..."
+while sleep ${INTERVAL};
+do
```
hopkiw: nitpick but `do` should be on same line as `while`

daisy_workflows/export/disk_resizing_mon.sh
```diff
@@ -0,0 +1,50 @@
...
+  # Check whether available buffer space is lower than threshold.
+  AVAILABLE_BUFFER=$(df -BG /dev/sdc --output=avail | sed -n 2p)
+  AVAILABLE_BUFFER=${AVAILABLE_BUFFER%?}
+  if [[ ${AVAILABLE_BUFFER} -ge ${BUFFER_MIN_SIZE} ]]; then
+    continue
+  fi
+
+  # Decide the new size of the device.
+  CURRENT_DEVICE_SIZE_BYTES=$(lsblk /dev/sdc --output=size -b | sed -n 2p)
+  CURRENT_DEVICE_SIZE=$(awk "BEGIN {print int(((${CURRENT_DEVICE_SIZE_BYTES}-1)/1073741824) + 1)}")
```
hopkiw: nitpick: make 1073741824 a constant

daisy_workflows/export/disk_resizing_mon.sh
hopkiw: there are two `echo`, a binary and a bash built-in. they differ in handling of escapes, but not in handling of newlines. if you're concerned, specify `/bin/echo`, but i think it's not related to the substance of my comment. if you always want output, simply remove `out=$()` wrapping

daisy_workflows/export/export_disk_ext.sh
```diff
@@ -17,28 +17,51 @@ set -x
...
+DISK_RESIZING_MON_LOCAL_PATH=/gs/${DISK_RESIZING_MON}
+echo "GCEExport: Copying disk size monitor script..."
+if ! out=$(gsutil cp "${DISK_RESIZING_MON_GCS_PATH}" "${DISK_RESIZING_MON_LOCAL_PATH}" 2>&1); then
+  echo "ExportFailed: Failed to copy disk size monitor script. Error: ${out}"
+  exit
+fi
+echo ${out}
+
+echo "GCEExport: Launching disk size monitor in background..."
+bash ${DISK_RESIZING_MON_LOCAL_PATH} ${MAX_BUFFER_DISK_SIZE_GB} &
```
hopkiw: np no need for bash prefix

daisy_workflows/export/export_disk_ext.sh
dntczdx: Sure! let me add clarification here

dntczdx:
addressed all comments

daisy_workflows/export/disk_resizing_mon.sh
dntczdx: In this script, "out" must be in the same line with "ExportFailed", otherwise daisy can't show it to user on console. I do need the $out var to achieve this.

daisy_workflows/export/export_disk_ext.sh
hopkiw: why did you resolve this?

daisy_workflows/export/export_disk_ext.sh
```diff
@@ -14,31 +14,55 @@
...
+# 2. Round up to the next GB.
+SIZE_OUTPUT_GB=$(awk "BEGIN {print int(((${SIZE_BYTES}-1)/${BYTES_1GB}) + 1)}")
+# 3. Add 5GB of additional space to max size to prevent the corner case that output
+# file is slightly larger than source disk.
+MAX_BUFFER_DISK_SIZE_GB=$(awk "BEGIN {print int(${SIZE_OUTPUT_GB} + 5)}")
 
-echo "GCEExport: Exporting disk of size ${SIZE_GB}GB and format ${FORMAT}."
+# Prepare buffer disk. We have confidence that it won't fail in this step.
+echo "GCEExport: Initializing buffer disk for qemu-img output..."
+sudo mkfs.ext4 /dev/sdc
```
hopkiw: what user wll this script be invoked as? (i.e. is `sudo` necessary?)
what error handling is there for these two..what error will user see if they fail?

daisy_workflows/export/export_disk_ext.sh
hopkiw: discussed irl, recommend replacing with `chmod +x ${DISK_RESIZING_MON_LOCAL_PATH}` after download

dntczdx: modified as discussed

daisy_workflows/export/export_disk_ext.sh
dntczdx: As discussed, to avoid confusion output, added a failure path here.

dntczdx:
1. replaced explicit calling to "bash" by assigning execution right with chmod
2. added error handling for mkfs/mount to prevent confusion output to enduser.

PR #704, opened 2019-03-04
Add random suffix to resource names

iamsubratp:
Without this, the tests from the subsequent test run fails if the resources
from previous tests are not clean'd up.

### DIFF
--- a/osconfig_tests/test_suites/package_management/package_management_test_data.go
+++ b/osconfig_tests/test_suites/package_management/package_management_test_data.go
@@ -53,7 +53,7 @@ func addCreateOsConfigTest(pkgTestSetup []*packageManagementTestSetup, testProje
 	for _, tuple := range tuples {
 		var oc *osconfigpb.OsConfig
 		var image string
-		uuid := utils.RandString(8)
+		uuid := utils.RandString(5)
 
 		switch tuple.platform {
 		case "debian":
@@ -90,7 +90,7 @@ func addPackageInstallTest(pkgTestSetup []*packageManagementTestSetup, testProje
 	for _, tuple := range tuples {
 		var oc *osconfigpb.OsConfig
 		var image, vs string
-		uuid := utils.RandString(8)
+		uuid := utils.RandString(5)
 
 		switch tuple.platform {
 		case "debian":
@@ -140,7 +140,7 @@ func addPackageRemovalTest(pkgTestSetup []*packageManagementTestSetup, testProje
 	for _, tuple := range tuples {
 		var oc *osconfigpb.OsConfig
 		var image, vs string
-		uuid := utils.RandString(8)
+		uuid := utils.RandString(5)
 
 		switch tuple.platform {
 		case "debian":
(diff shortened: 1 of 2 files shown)


osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -52,20 +53,21 @@ func addCreateOsConfigTest(pkgTestSetup []*packageManagementTestSetup, testProje
 	for _, tuple := range tuples {
 		var oc *osconfigpb.OsConfig
 		var image string
+		uuid := utils.RandString(8)
```
ryanwe: How about using a timestamp instead? Going down to millis should avoid duplicates and will make it easier to work backward from leaked resources.

adjackura: That's probably too long, we are limited here on resource name length and having the random bits be the thing that puts us over the limit is undesirable. Use utils.RandString(5) here instead. Also have this generated only once for the entire test suite (and add it to testName) so that we can easily associate test runs and resources (osconfigs and instances)

adjackura: Also I guess it may not be clear, the length issue is for compute resources, but I still feel strongly we should associate all resources created by the test. We could put timestamp in the osconfig description or something

iamsubratp: i initially used uuid(they have time stamp and chances of collision is very very low), but since they are 36 character long, we could have ended up with exceeding 64 characters for resource name.
I can change to use timestamp but then i would like to change the Randstring func instead of using timestamp just here, just want to keep it consistent throughout the application

ryanwe: I'm fine with using a random slug and putting the timestamp in the description. Alternatively, having a timestamp based ID to associate with the entire test run would probably be ideal. Also, if we're using `RandString(5)` we'll want to call it something other than UUID.

iamsubratp: > Use utils.RandString(5) here instead. Also have this generated only once for the entire test suite (and add it to testName) so that we can easily associate test runs and resources (osconfigs and instances)

this totally makes sense.

I am going to change it to use 5 digit random string and use timestamp in the description

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -52,20 +53,21 @@ func addCreateOsConfigTest(pkgTestSetup []*packageManagementTestSetup, testProje
 	for _, tuple := range tuples {
 		var oc *osconfigpb.OsConfig
 		var image string
+		uuid := utils.RandString(5)
```
ryanwe: nit: I'd still recommend calling this `suffix` or just `uid` instead of `uuid` for correctness since these are "unique" ids but not meant to be "universally unique".

iamsubratp: ah.. i am so sorry... i totally forgot this. i will send out a different PR for this.

PR #705, opened 2019-03-04
 osconfig agent: read config options out of instance metadata

adjackura:
Rework task logic, attempt to standardize the surface for each component
Patch watcher reworked to allow for cancellation logic. Right now this is just used for disabling of the patch system.

### DIFF
--- a/cli_tools/google-osconfig-agent/config/config.go
+++ b/cli_tools/google-osconfig-agent/config/config.go
@@ -19,6 +19,8 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"net"
+	"net/url"
 	"runtime"
 	"strconv"
 	"sync"
@@ -130,9 +132,29 @@ func SetConfig() error {
 		aptRepoFilePath:    aptRepoFilePath,
 	}
 
-	md, err := metadata.Get("?recursive=true&alt=json")
-	if err != nil {
-		return err
+	var md string
+	var err error
+	webError := 0
+	for {
+		md, err = metadata.Get("?recursive=true&alt=json")
+		if err == nil {
+			break
+		}
+		// Only log the third web error to avoid transient errors and
+		// not to spam the log on network failures.
+		if webError == 1 {
+			if urlErr, ok := err.(*url.Error); ok {
+				if _, ok := urlErr.Err.(*net.DNSError); ok {
+					return fmt.Errorf("DNS error when requesting metadata, check DNS settings and ensure metadata.internal.google is setup in your hosts file")
+				}
+				if _, ok := urlErr.Err.(*net.OpError); ok {
+					return fmt.Errorf("network error when requesting metadata, make sure your instance has an active network and can reach the metadata server")
+				}
+			}
+			return fmt.Errorf(err.Error())
+		}
+		webError++
+		time.Sleep(5 * time.Second)
 	}
 
 	var metadata metadataJSON
--- a/cli_tools/google-osconfig-agent/config/config.go
+++ b/cli_tools/google-osconfig-agent/config/config.go
@@ -130,6 +130,18 @@ func createConfigFromMetadata(md metadataJSON) *config {
 	return c
 }
 
+func handleError(err error) error {
+	if urlErr, ok := err.(*url.Error); ok {
+		if _, ok := urlErr.Err.(*net.DNSError); ok {
+			return fmt.Errorf("DNS error when requesting metadata, check DNS settings and ensure metadata.internal.google is setup in your hosts file")
+		}
+		if _, ok := urlErr.Err.(*net.OpError); ok {
+			return fmt.Errorf("network error when requesting metadata, make sure your instance has an active network and can reach the metadata server")
+		}
+	}
+	return err
+}
+
 // SetConfig sets the agent config.
 func SetConfig() error {
 	var md string
@@ -143,14 +155,7 @@ func SetConfig() error {
 		// Try up to 3 times to wait for slow network initialization, after
 		// that resort to using defaults and returning the error.
 		if webErrorCount == 2 {
-			if urlErr, ok := webError.(*url.Error); ok {
-				if _, ok := urlErr.Err.(*net.DNSError); ok {
-					return fmt.Errorf("DNS error when requesting metadata, check DNS settings and ensure metadata.internal.google is setup in your hosts file")
-				}
-				if _, ok := urlErr.Err.(*net.OpError); ok {
-					return fmt.Errorf("network error when requesting metadata, make sure your instance has an active network and can reach the metadata server")
-				}
-			}
+			webError = handleError(webError)
 			break
 		}
 		webErrorCount++
--- a/cli_tools/google-osconfig-agent/config/config.go
+++ b/cli_tools/google-osconfig-agent/config/config.go
@@ -16,9 +16,12 @@
 package config
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"runtime"
+	"strconv"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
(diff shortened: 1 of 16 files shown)


cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -27,8 +32,17 @@ import (
 const (
 	// InstanceMetadata is the instance metadata URL.
 	InstanceMetadata = "http://metadata.google.internal/computeMetadata/v1/instance"
-	// ReportURL is where OS configurations are written in guest attributes.
+	// ReportURL is the guest attributes endpoint.
```
ryanwe: Can we rename this to `InventoryReportURL` to avoid confusing it with the patch report URL?

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -37,19 +51,119 @@ var (
...
 
-	version string
-)
+func getAgentConfig() config {
+	agentConfigMx.RLock()
+	defer agentConfigMx.RUnlock()
+	return *agentConfig
+}
+
+func enabled(s string) bool {
```
ryanwe: would this be clearer as `parseBool` since that's what it does?

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -37,19 +51,119 @@ var (
...
+	Attributes attributesJSON
+}
+
+type attributesJSON struct {
+	OSInventoryEnabled string `json:"os-inventory-enabled"`
+	OSPatchEnabled     string `json:"os-patch-enabled"`
+	OSPackageEnabled   string `json:"os-package-enabled"`
+}
+
+func (c *config) setFromMetadata(md metadataJSON) {
```
ryanwe: minor: it might be clearer to change this to `buildFromMetadata()` or `createFromMetadata()` and have it return the new object. The advantage being that it would no longer mutate the state of an object that may also be managed in other places.

That would also allow you to set the defaults in the same place where you check for overrides which would make the logic a lot clearer.

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -37,19 +51,119 @@ var (
...
+	if md.Instance.Attributes.OSPatchEnabled != "" {
+		c.osPatchEnabled = enabled(md.Instance.Attributes.OSPatchEnabled)
+	}
+	if md.Instance.Attributes.OSPackageEnabled != "" {
+		c.osPackageEnabled = enabled(md.Instance.Attributes.OSPackageEnabled)
+	}
+}
+
+// SetConfig sets the agent config.
+func SetConfig() error {
```
ryanwe: What do you think about calling this "LoadConfig" instead? Does that convey its behavior a little better?

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -27,8 +32,17 @@ import (
...
 	ReportURL = InstanceMetadata + "/guest-attributes"
+
+	googetRepoFilePath = "C:/ProgramData/GooGet/repos/google_osconfig.repo"
+	zypperRepoFilePath = "/etc/zypp/repos.d/google_osconfig.repo"
+	yumRepoFilePath    = "/etc/yum/repos.d/google_osconfig.repo"
+	aptRepoFilePath    = "/etc/apt/sources.list.d/google_osconfig.list"
+
+	osInventoryEnabled = false
+	osPackageEnabled   = false
+	osPatchEnabled     = false
```
ryanwe: minor: can you add `default` to the names to differentiate them from other variables with the same name?

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -37,19 +51,119 @@ var (
...
+		webError++
+		time.Sleep(5 * time.Second)
+	}
+
+	var metadata metadataJSON
+	if err := json.Unmarshal([]byte(md), &metadata); err != nil {
+		return err
+	}
+
+	agentConfig.setFromMetadata(metadata)
```
ryanwe: if you change `setFromMetadata()` to `buildFromMetadata()`, you can put the logic around locking the `agentConfig ` pointer here where you set it.

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -37,19 +51,119 @@ var (
...
+	var md string
+	var err error
+	webError := 0
+	for {
+		md, err = metadata.Get("?recursive=true&alt=json")
+		if err == nil {
+			break
+		}
+		// Only log the third web error to avoid transient errors and
+		// not to spam the log on network failures.
```
ryanwe: This comment doesn't match what the code is doing and I'm not sure if the comment is out of date or if it points to a bug. First, the special logic happens on the 2nd web error and not the 3rd. But also, this function doesn't log but exits.

So, should it handle the error on the 2nd or 3rd error? And should the comment be changed to "Only return the third web error..."?

Also, did you chose 2nd (or 3rd) error because usually metadata is only down for <= 15 seconds?

cli_tools/google-osconfig-agent/main.go
```diff
@@ -46,25 +46,28 @@ func (l *logWritter) Write(b []byte) (int, error) {
...
+		}
+
+		// This sets up the patching system to run in the background.
+		ospatch.Configure(ctx)
+
 		if config.OSPackageEnabled() {
-			if err := ospackage.RunWithEnqueue(ctx, res); err != nil {
-				logger.Errorf(err.Error())
-			}
+			ospackage.Run(ctx, res)
```
ryanwe: 👍

cli_tools/google-osconfig-agent/ospatch/patch_run.go
```diff
@@ -68,18 +68,43 @@ var (
...
+func Configure(ctx context.Context) {
+	select {
+	case _, ok := <-cancelC:
+		// Channel closed, enable if needed.
+		if !ok && config.OSPatchEnabled() {
+			initPatch(ctx)
+		}
+		// Channel open, disabled if needed.
+		// This should never happen as nothing should be sending on this
+		// channel.
```
ryanwe: If this should never happen, should we log or panic if it does? It's a little scary if we're adding logic for a case that should never occur. While we're in EAP and Alpha is a good time to log loudly to help us catch edge cases early.

cli_tools/google-osconfig-agent/ospatch/watcher.go
```diff
@@ -47,78 +46,100 @@ func updateEtag(resp *http.Response) bool {
...
+				if webError == 1 {
+					if urlErr, ok := ret.err.(*url.Error); ok {
+						if _, ok := urlErr.Err.(*net.DNSError); ok {
+							logger.Errorf("DNS error when requesting metadata, check DNS settings and ensure metadata.internal.google is setup in your hosts file.")
+						}
+						if _, ok := urlErr.Err.(*net.OpError); ok {
+							logger.Errorf("Network error when requesting metadata, make sure your instance has an active network and can reach the metadata server.")
+						}
 					}
+					logger.Errorf(ret.err.Error())
```
ryanwe: This will lead to double logs. Should there be a `continue` after the previous log statements?

cli_tools/google-osconfig-agent/config/config.go
adjackura: We do use this for patch as well

cli_tools/google-osconfig-agent/ospatch/patch_run.go
adjackura: The comment is more about the fact that the code as written should never get here. If can lead to a panic if this is violated though, so this comment would then come in handy. I added a log line just to be sure we catch it.

cli_tools/google-osconfig-agent/config/config.go
adjackura: Fixed, this was not doing what I wanted

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -130,6 +130,18 @@ func createConfigFromMetadata(md metadataJSON) *config {
 	return c
 }
 
+func handleError(err error) error {
```
ryanwe: nit: WDYT about renaming these `formatError()` or `fmtError` since "handle" is really generic and doesn't give much context?

PR #712, opened 2019-03-08
osconfig agent: use a lock file

adjackura:
Use a lock file to ensure only one osconfig process is running at any one time. This is a simple solution to this issue and should be expanded upon or replaced later with a more robust solution.

### DIFF
--- a/cli_tools/google-osconfig-agent/lock_linux.go
+++ b/cli_tools/google-osconfig-agent/lock_linux.go
@@ -1,4 +1,4 @@
-//  Copyright 2018 Google Inc. All Rights Reserved.
+//  Copyright 2019 Google Inc. All Rights Reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
--- a/cli_tools/google-osconfig-agent/lock_linux.go
+++ b/cli_tools/google-osconfig-agent/lock_linux.go
@@ -43,7 +43,7 @@ func obtainLock() {
 	select {
 	case err := <-c:
 		if err != nil {
-			logger.Fatalf("OSConfig agent lock already held, is the agent already running? Error: %v", err)
+			logger.Fatalf("Cannot obtain agent lock, is the agent already running? Error: %v", err)
 		}
 	case <-time.After(time.Second):
 		logger.Fatalf("OSConfig agent lock already held, is the agent already running?")
--- a/cli_tools/google-osconfig-agent/lock_linux.go
+++ b/cli_tools/google-osconfig-agent/lock_linux.go
@@ -1,0 +1,53 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"syscall"
+	"time"
+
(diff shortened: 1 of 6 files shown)


cli_tools/google-osconfig-agent/main.go
```diff
@@ -83,6 +102,17 @@ func main() {
 	flag.Parse()
 	ctx := context.Background()
 
+	// TODO: move to a different locking system or move lockFile definition to config.
```
ryanwe: Consider moving this to a separate function that makes its purpose clearer, for example `ensureSingletonProcess` or `ensureOsConfigNotAlreadyRunning`.

Also, fwiw, this seems like a brittle way to do it since the process could crash w/out removing the lock file. We may need to put a timestamp in the file that the process with the lock continually updates...

ryanwe: Oh. Does `O_EXCL` release/delete the file if the process terminates?

adjackura: Nope, O_EXCL just makes sure the file doesnt exist already. The point of my TODO was that we should find a better solution later, I'll update it to be more clear on the reasons.

Also i feel naming is totally fine, your examples are crazy verbose :) this is all private to main which has like 3 functions :p

cli_tools/google-osconfig-agent/ospatch/patch_run.go
```diff
@@ -69,7 +69,7 @@ var (
 		completeJobCompleted: 9,
 	}
 
-	cancelC chan struct{}
+	cancelC = make(chan struct{})
```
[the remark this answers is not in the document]
adjackura: This fixed a race i found through the unit test, closing a nil channel leads to a panic

cli_tools/google-osconfig-agent/main.go
ryanwe: Does that mean that if the process crashes and doesn't release the lock, the agent won't be able to restart? If so, I don't think we'd want to use this mechanism even for a little while. Can we use `https://golang.org/pkg/syscall/#Flock` instead?

adjackura: flock only works on linux

adjackura: Implemented for windows and linux

cli_tools/google-osconfig-agent/lock_linux.go
```diff
@@ -0,0 +1,53 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019 :smile:

cli_tools/google-osconfig-agent/lock_linux.go
```diff
@@ -0,0 +1,53 @@
...
+	lockFile := "/run/lock/osconfig_agent.lock"
+
+	err := os.MkdirAll(filepath.Dir(lockFile), 0755)
+	if err != nil && !os.IsExist(err) {
+		logger.Fatalf("Cannot obtain agent lock: %v", err)
+	}
+
+	f, err := os.OpenFile(lockFile, os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil && !os.IsExist(err) {
+		logger.Fatalf("Cannot obtain agent lock: %v", err)
```
ryanwe: minor: you might want to differentiate this message from the previous one so you know where it failed if you see it.

cli_tools/google-osconfig-agent/lock_windows.go
```diff
@@ -0,0 +1,88 @@
...
+	ret, _, _ := procUnlockFileEx.Call(
+		hFile,
+		0,
+		uintptr(nNumberOfBytesToLockLow),
+		uintptr(nNumberOfBytesToLockHigh),
+		uintptr(unsafe.Pointer(lpOverlapped)),
+	)
+	// If the function succeeds, the return value is nonzero.
+	if ret == 0 {
+		return fmt.Errorf("nonzero return code from UnlockFileEx: %d", ret)
```
ryanwe: Is `0` success or fail? It looks like this will only happen when `ret == 0` which contradicts the error message.

cli_tools/google-osconfig-agent/lock_linux.go
adjackura: The error contains line number and the actual error received.

cli_tools/google-osconfig-agent/lock_windows.go
adjackura: fixed

PR #714, opened 2019-03-13
osconfig-agent: Fix initialization of global values

iamsubratp:
This commit fixes 3 issues
1) 	As per Golang specs, the init functions from different files
	in a single package are called in lexicographical order of
	the file names.
	Because of such order of initialization, some of the values
	are not initialized before they are used.

2)	The agentconfig is not initialized for cases where the
	users want to only execute a single feature

3)	In ospackage feature, the repo directory might not exist.

### DIFF
--- a/cli_tools/google-osconfig-agent/main.go
+++ b/cli_tools/google-osconfig-agent/main.go
@@ -54,7 +54,7 @@ func run(ctx context.Context) {
 	ticker := time.NewTicker(config.SvcPollInterval())
 	for {
 		if err := config.SetConfig(); err != nil {
-			logger.Errorf(err.Error())
+			logger.Fatalf(err.Error())
 		}
 
 		// This sets up the patching system to run in the background.
@@ -82,6 +82,10 @@ func run(ctx context.Context) {
 func main() {
 	flag.Parse()
 	ctx := context.Background()
+
+	if err := config.SetConfig(); err != nil {
+		logger.Errorf(err.Error())
+	}
 
 	if config.Debug() {
 		packages.DebugLogger = log.New(&logWritter{}, "", 0)
--- a/cli_tools/google-osconfig-agent/main.go
+++ b/cli_tools/google-osconfig-agent/main.go
@@ -54,7 +54,7 @@ func run(ctx context.Context) {
 	ticker := time.NewTicker(config.SvcPollInterval())
 	for {
 		if err := config.SetConfig(); err != nil {
-			logger.Fatalf(err.Error())
+			logger.Errorf(err.Error())
 		}
 
 		// This sets up the patching system to run in the background.
--- a/cli_tools/google-osconfig-agent/ospackage/ospackage.go
+++ b/cli_tools/google-osconfig-agent/ospackage/ospackage.go
@@ -24,6 +24,7 @@ import (
 	"hash"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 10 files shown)


cli_tools/google-osconfig-agent/main.go
```diff
@@ -83,6 +83,10 @@ func main() {
 	flag.Parse()
 	ctx := context.Background()
 
+	if err := config.SetConfig(); err != nil {
```
adjackura: We don't want to do this here, only during service run

go/packages/packages.go
```diff
@@ -94,6 +96,28 @@ var run runFunc = func(cmd *exec.Cmd) ([]byte, error) {
 	return out, nil
 }
 
+func init() {
```
adjackura: Just move the *Exists checks to each file associated with that package manager.

cli_tools/google-osconfig-agent/ospackage/ospackage.go
```diff
@@ -101,8 +101,14 @@ func lookupConfigs(ctx context.Context, client *osconfig.Client, resource string
 
 func setConfig(res *osconfigpb.LookupConfigsResponse) error {
 	var errs []string
-
 	if res.Goo != nil && packages.GooGetExists {
+		if _, err := os.Stat(config.GoogetRepoFilePath()); os.IsNotExist(err) {
+			logger.Debugf("Repo file does not exist, will create one...")
+			if err := os.MkdirAll(config.GoogetRepoFilePath(), 07550); err != nil {
```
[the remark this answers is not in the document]
iamsubratp: i just realized, when i ran my e2e with these bug fixes.

will change it

go/packages/packages.go
iamsubratp: i think, we will end up in the same situation if we use "init" to initialize any of these values. initializing them as variables also, would not work since the variables are the first one that gets initialized during the initialization process, and then the init functions are called inside the package.

iamsubratp: actually, i did not want to keep these initialization in package.go. Instead, t thought that since the order of execution of init function are so specific, i would make sense to gather all into a different file "_inits.go". this way, we ensure that all the values are initialized at first and splitting them out in a separate file would also bring out this information about compilation in lexicographical ordering( it gave me real hard time yesterday while debugging the issue :( )
However, that initialization did not work at all. For some reason, the init method was not called at all.
In the interest of time, i decided to  club them together where it made most sense

cli_tools/google-osconfig-agent/main.go
iamsubratp: as per offline discussion, leaving it as it is

go/packages/packages.go
iamsubratp: discussed offline. moved the *exists to respective package manager

cli_tools/google-osconfig-agent/main.go
```diff
@@ -54,7 +54,7 @@ func run(ctx context.Context) {
 	ticker := time.NewTicker(config.SvcPollInterval())
 	for {
 		if err := config.SetConfig(); err != nil {
-			logger.Errorf(err.Error())
+			logger.Fatalf(err.Error())
```
adjackura: This isn't a fatal error, we don't want to crash just because we can't set configs, we have sane defaults set

iamsubratp: done

PR #716, opened 2019-03-15
Add integration tests for installation from new repo

iamsubratp:
- As a osconfig user, user should be able to add a new repository
and install packages from the new one. For our case, we have create
new rapture repository 'osconfig-agent-test-repository'.
I have only added test for debian based system. Tests for yum based systems
are yet to follow.
- Removed the panics since it does not make sense to crash the test runner as
these configurations are owned by the test runner itself. There will be no situation
where the a package-manager is missing unknowingly

### DIFF
--- a/osconfig_tests/test_suites/package_management/package_management.go
+++ b/osconfig_tests/test_suites/package_management/package_management.go
@@ -357,6 +357,89 @@ func runPackageInstallTest(ctx context.Context, testCase *junitxml.TestCase, tes
 	}
 }
 
+func runPackageInstallFromNewRepoTest(ctx context.Context, testCase *junitxml.TestCase, testSetup *packageManagementTestSetup, logger *log.Logger, testProjectConfig *testconfig.Project) {
+	parent := fmt.Sprintf("projects/%s", testProjectConfig.TestProjectID)
+	oc, err := osconfigserver.CreateOsConfig(ctx, testSetup.osconfig, parent)
+	if err != nil {
+		testCase.WriteFailure("error while creating osconfig: \n%s\n", utils.GetStatusFromError(err))
+		return
+	}
+	defer cleanupOsConfig(ctx, testCase, oc, testProjectConfig)
+
+	assign, err := osconfigserver.CreateAssignment(ctx, testSetup.assignment, parent)
+	if err != nil {
+		testCase.WriteFailure("error while creating assignment: \n%s\n", utils.GetStatusFromError(err))
+		return
+	}
+	defer cleanupAssignment(ctx, testCase, assign, testProjectConfig)
+
+	client, err := daisyCompute.NewClient(ctx)
+	if err != nil {
+		testCase.WriteFailure("error creating client: %v", err)
+		return
+	}
+
+	testCase.Logf("Creating instance with image %q", testSetup.image)
+	i := &api.Instance{
+		Name:        testSetup.name,
+		MachineType: fmt.Sprintf("projects/%s/zones/%s/machineTypes/n1-standard-1", testProjectConfig.TestProjectID, testProjectConfig.TestZone),
+		NetworkInterfaces: []*api.NetworkInterface{
+			&api.NetworkInterface{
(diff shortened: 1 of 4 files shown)


osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -357,19 +357,104 @@ func runPackageInstallTest(ctx context.Context, testCase *junitxml.TestCase, tes
...
+					},
+				},
+			},
+		},
+		Metadata: &api.Metadata{
+			Items: []*api.MetadataItems{
+				testSetup.startup,
+				&api.MetadataItems{
+					Key:   "os-package-enabled",
+					Value: func() *string { v := "true"; return &v }(),
```
ryanwe: Is this a bug in the `api.MetadataItems`? It seems really odd that we have to do this.

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -236,11 +242,54 @@ func addPackageInstallRemovalTest(pkgTestSetup []*packageManagementTestSetup, te
 	return pkgTestSetup
 }
 
+func addPackageInstallFromNewRepoTest(pkgTestSetup []*packageManagementTestSetup, testProjectConfig *testconfig.Project) []*packageManagementTestSetup {
+	testName := "packageinstallfromnewrepotest"
+	desc := "test package installation from new package"
+	packageName := "osconfig-agent-test"
+	for _, tuple := range tuples {
```
ryanwe: Not new to this review but isn't `tuples` naming the variable according to its type? Maybe `platformPkgManagers` or `distros` or `systems` would be clearer. I had to double check the previous source to remember what these things were.

osconfig_tests/test_suites/package_management/package_management_utils.go
```diff
@@ -159,7 +160,53 @@ func getPackageInstallRemovalStartupScript(pkgManager, packageName string) strin
...
+}
+
+func getPackageInstallFromNewRepoTestStartupScript(pkgManager, packageName string) string {
+	var ss string
+
+	switch pkgManager {
+
+	case "apt":
+		ss = "%s\n" +
+			"sleep 10;\n" +
```
ryanwe: Why sleep for 10 seconds? Is it trying to wait for something to happen that could, in some circumstances, take longer than 10 seconds?

osconfig_tests/test_suites/package_management/package_management.go
adjackura: Valuse is a *string, this is the easiest way I know of to inline that type of value

ryanwe: Yeah. I was wondering, since it's hard to assign a string literal to a `*string`, and a `string` is just a reference anyway, why is this a `*string` instead of a `string`? And is that a bug in the metadata api, not our own?

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: changed

osconfig_tests/test_suites/package_management/package_management_utils.go
iamsubratp: while testing it manually,  i noticed that the agent needed a restart. Did not have to wait before restart. to be on the safe side, i added the wait.
I am actually curious to know why the agent needs a restart, ideally it should have worked on the first run itself.
I am thinking of adding debug flag through instancemetadata so the agent can run in debug mode.

ryanwe: Sounds good. Can you add a comment to the code explaining that? @adjackura is this related to agents hanging on startup? If so, the fact that a restart fixes the problem may be a new clue.

iamsubratp: added a comment explaining the test runner behavior

PR #719, opened 2019-03-17
Allow increase log verbosity through instance metadata

iamsubratp:
Currently, there is no way to increase verbosity of agent logging. Until now, for debugging purpose I had to restart the agent on the test vm with the debug flag. This was already a painful process, and with introduction of agent locking, even this option was no more feasible.

### DIFF
--- a/cli_tools/google-osconfig-agent/config/config.go
+++ b/cli_tools/google-osconfig-agent/config/config.go
@@ -43,6 +43,7 @@ const (
 	osInventoryEnabledDefault = false
 	osPackageEnabledDefault   = false
 	osPatchEnabledDefault     = false
+	debugEnabledDefault       = false
 )
 
 var (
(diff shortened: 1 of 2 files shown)


cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -93,13 +94,15 @@ type attributesJSON struct {
 	OSInventoryEnabled string `json:"os-inventory-enabled"`
 	OSPatchEnabled     string `json:"os-patch-enabled"`
 	OSPackageEnabled   string `json:"os-package-enabled"`
+	DebugEnabled       string `json:"debug-enabled"`
```
adjackura: This is a shared namespace, use something like 'os-config-debug'

iamsubratp: changed... made it consistent throughout with other osconfig attributes.

PR #720, opened 2019-03-18
Export e2e: add 2T/4T tests as weekly

dntczdx:
2T is the max of MBR partition.
4T is what we can support stable by testing.

6T/8T failed due to gsutil because of super long copying (> 40hours), so excluded.


dntczdx:
Ping. Can someone help to review this?
@zoran15 @adjackura @zmarano @hopkiw

PR #736, opened 2019-04-03
osconfig agent: actually enable ospatch when set in metadata

adjackura:
Created stuff files for future unit tests.


cli_tools/google-osconfig-agent/ospatch/updates_test_stub.go
```diff
@@ -0,0 +1,33 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019

PR #738, opened 2019-04-04
[osconfig-agent] add windows end to end tests

iamsubratp:
* created the rapture test repository
* Added a cowsay package
* Since windows updates takes a long time, i had to increase the assertion timeout for test case failure.

### DIFF
--- a/osconfig_tests/test_suites/package_management/package_management.go
+++ b/osconfig_tests/test_suites/package_management/package_management.go
@@ -56,6 +56,7 @@ type packageManagementTestSetup struct {
 	assignment *osconfigpb.Assignment
 	startup    *api.MetadataItems
 	vstring    string
+	assertTimeout time.Duration
 	vf         func(*compute.Instance, string, int64, time.Duration, time.Duration) error
 }
 
@@ -133,7 +134,7 @@ func runPackageRemovalTest(ctx context.Context, testCase *junitxml.TestCase, tes
 	//TODO: move instance definition to a common method
 	i := &api.Instance{
 		Name:        testSetup.name,
-		MachineType: fmt.Sprintf("projects/%s/zones/%s/machineTypes/n1-standard-1", testProjectConfig.TestProjectID, testProjectConfig.TestZone),
+		MachineType: fmt.Sprintf("projects/%s/zones/%s/machineTypes/n1-standard-4", testProjectConfig.TestProjectID, testProjectConfig.TestZone),
 		NetworkInterfaces: []*api.NetworkInterface{
 			&api.NetworkInterface{
 				Network: "global/networks/default",
@@ -188,7 +189,7 @@ func runPackageRemovalTest(ctx context.Context, testCase *junitxml.TestCase, tes
 	}
 
 	// read the serial console once
-	if err = testSetup.vf(inst, testSetup.vstring, 1, 10*time.Second, 600*time.Second); err != nil {
+	if err = testSetup.vf(inst, testSetup.vstring, 1, 10*time.Second, testSetup.assertTimeout); err != nil {
 		testCase.WriteFailure("error while asserting: %v", err)
 		return
 	}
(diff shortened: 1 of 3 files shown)


osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -41,6 +41,7 @@ const (
 	debianImage   = "projects/debian-cloud/global/images/family/debian-9"
 	centosImage   = "projects/centos-cloud/global/images/family/centos-7"
 	rhelImage     = "projects/rhel-cloud/global/images/family/rhel-7"
+	windowsImage  = "projects/windows-cloud/global/images/family/windows-2016"
```
ryanwe: What's the plan for extending this to all the versions we want to support?

adjackura: I would suggest making all these []string and then ranging over them for creating tests

dntczdx:
I would suggest having a more specific title for this PR: windows tests for some project? Since there are multiple ongoing projects under /compute-image-tools. It helped people to decide whether to look into it easier. Thanks!

osconfig_tests/test_suites/package_management/package_management.go
adjackura: this will get complicated with your current setup, you may want to use some sort of config struct like we have for inventory tests

ryanwe: +1. But let's do that as a separate PR.

iamsubratp: i actually looked into it. with the current setup, looks like, it is going to get complicated. will send out a separate PR

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -352,7 +361,7 @@ func runPackageInstallTest(ctx context.Context, testCase *junitxml.TestCase, tes
 	testCase.Logf("Agent installed successfully")
 
 	// read the serial console once
-	if err = testSetup.vf(inst, testSetup.vstring, 1, 10*time.Second, 60*time.Second); err != nil {
+	if err = testSetup.vf(inst, testSetup.vstring, 1, 10*time.Second, 1200*time.Second); err != nil {
```
ryanwe: This is a long time to wait and I'm concerned that it may mean that failing tests take 20 times longer to actually fail. Can this be scoped just to Windows, if that's where the extra time is needed? Can we use bigger test instances to make this faster? Are there failure signals that can be checked to ensure that the test is waiting on a success and not waiting to time out 20 minutes after a failure was reported?

iamsubratp: > This is a long time to wait and I'm concerned that it may mean that failing tests take 20 times longer to actually fail. Can we use bigger test instances to make this faster?
windows startup is terribly long. Digging more into it, I found that the time is majorly spent at 2 different stages. 1) windows boot and startup script installation; 2) fetching windows updates.
I also looked at the resource utilization of the test instances and found that with the smaller instances (1 cpu/4gb ram - standard/1), the CPU utilization is almost always 100%. Moving to bigger instances should definitely help in improving the boot time(I have just kicked off a test with bigger instance(4 cpu/8gigs - standard/4); would post the results shortly)
Surprisingly, it also helped in faster fetching of windows updates. [logs](https://pantheon.corp.google.com/logs/viewer?resource=gce_instance%2Finstance_id%2F6364734612471780596&project=subratp-playground&minLogLevel=0&expandAll=false&timestamp=2019-04-10T17:53:57.345000000Z&customFacets=&limitCustomFacetWidth=true&interval=PT1H&scrollTimestamp=2019-04-10T17:50:43.159815000Z&dateRangeStart=2019-04-10T16:49:28.312Z&dateRangeUnbound=forwardInTime)

I also tested with even bigger instance(8 cpu), i dont see an improvement over standard/4. 
I will use standard/4 as test instances.

> Can this be scoped just to Windows, if that's where the extra time is needed?
this can done. I can add a new member to packagemanagementestsetup, so while adding a testsetup, a custom timeout for verification string can be provided for individual tests.

> Are there failure signals that can be checked to ensure that the test is waiting on a success and not waiting to time out 20 minutes after a failure was reported?
Currently, there is no strong failure signal that we get from the tests and i did not want it on purpose for the reason that I wanted to assert on success signals instead of failure signal. while performing all the manual tests, I realized that it was hard to define strong failure signals because there were multiple reasons for a test to fail(these were reasons that i could figure out, there might be others as well). So, instead i chose to have a strong success signal.

PR #741, opened 2019-04-10
[OsConfig Agent] Add other supported images for different platforms

iamsubratp:
This change is built on top of PR:https://github.com/GoogleCloudPlatform/compute-image-tools/pull/738

Out of the 2 commits in this PR, the other one is already being reviewed in the above mentioned PR.

### DIFF
--- a/osconfig_tests/test_suites/package_management/package_management.go
+++ b/osconfig_tests/test_suites/package_management/package_management.go
@@ -36,12 +36,12 @@ import (
 	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/test_config"
 )
 
-const (
+var (
 	testSuiteName = "PackageManagementTests"
-	debianImage   = "projects/debian-cloud/global/images/family/debian-9"
-	centosImage   = "projects/centos-cloud/global/images/family/centos-7"
-	rhelImage     = "projects/rhel-cloud/global/images/family/rhel-7"
-	windowsImage  = "projects/windows-cloud/global/images/family/windows-2016"
+	debianImages  = []string{"projects/ubuntu-os-cloud/global/images/family/ubuntu-1604-lts", "projects/ubuntu-os-cloud/global/images/family/ubuntu-1804-lts", "projects/debian-cloud/global/images/family/debian-9"}
+	centosImages  = []string{"projects/centos-cloud/global/images/family/centos-6", "projects/centos-cloud/global/images/family/centos-7"}
+	rhelImages    = []string{"projects/rhel-cloud/global/images/family/rhel-6", "projects/rhel-cloud/global/images/family/rhel-7"}
+	windowsImages = []string{"projects/windows-cloud/global/images/family/windows-2016"}
 )
 
 var (
@@ -58,6 +58,19 @@ type packageManagementTestSetup struct {
 	vstring       string
 	assertTimeout time.Duration
 	vf            func(*compute.Instance, string, int64, time.Duration, time.Duration) error
+}
+
+func NewPackageManagementTestSetup(image, name, fname, vs string, oc *osconfigpb.OsConfig, assignment *osconfigpb.Assignment, startup *api.MetadataItems, vf func(*compute.Instance, string, int64, time.Duration, time.Duration) error) *packageManagementTestSetup {
+	return &packageManagementTestSetup{
+		image:      image,
+		name:       name,
+		osconfig:   oc,
+		assignment: assignment,
+		fname:      fname,
+		vf:         vf,
+		vstring: 		vs,
+		startup: startup,
+	}
 }
 
 // TestSuite is a PackageManagementTests test suite.
--- a/osconfig_tests/test_suites/package_management/package_management.go
+++ b/osconfig_tests/test_suites/package_management/package_management.go
@@ -60,16 +60,17 @@ type packageManagementTestSetup struct {
 	vf            func(*compute.Instance, string, int64, time.Duration, time.Duration) error
 }
 
-func newPackageManagementTestSetup(setup **packageManagementTestSetup, image, name, fname, vs string, oc *osconfigpb.OsConfig, assignment *osconfigpb.Assignment, startup *api.MetadataItems, vf func(*compute.Instance, string, int64, time.Duration, time.Duration) error) {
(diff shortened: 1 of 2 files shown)


osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -58,6 +59,19 @@ type packageManagementTestSetup struct {
...
 
+func NewPackageManagementTestSetup(image, name, fname, vs string, oc *osconfigpb.OsConfig, assignment *osconfigpb.Assignment, startup *api.MetadataItems, vf func(*compute.Instance, string, int64, time.Duration, time.Duration) error) *packageManagementTestSetup {
+	return &packageManagementTestSetup{
+		image:      image,
+		name:       name,
+		osconfig:   oc,
+		assignment: assignment,
+		fname:      fname,
+		vf:         vf,
+		vstring: 		vs,
```
ryanwe: minor: looks like the spacing is off here.

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -58,39 +58,41 @@ func addCreateOsConfigTest(pkgTestSetup []*packageManagementTestSetup, testProje
...
 		uniqueSuffix := utils.RandString(5)
 
 		switch tuple.platform {
 		case "debian":
-			image = debianImage
-			pkgs := []*osconfigpb.Package{osconfigserver.BuildPackage(packageName)}
-			oc = osconfigserver.BuildOsConfig(fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), desc, osconfigserver.BuildAptPackageConfig(pkgs, nil, nil), nil, nil, nil, nil)
+			for _, image := range debianImages {
+				pkgs := []*osconfigpb.Package{osconfigserver.BuildPackage(packageName)}
+				oc = osconfigserver.BuildOsConfig(fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), desc, osconfigserver.BuildAptPackageConfig(pkgs, nil, nil), nil, nil, nil, nil)
```
ryanwe: minor: consider breaking out the `fmt.Sprintf()` and `BuildAptPackageConfig()` to separate variables. It'll make the line shorter and the code easier to read.

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -58,39 +58,41 @@ func addCreateOsConfigTest(pkgTestSetup []*packageManagementTestSetup, testProje
...
+			}
 		case "windows":
-			image = windowsImage
-			pkgs := []*osconfigpb.Package{osconfigserver.BuildPackage(packageName)}
-			oc = osconfigserver.BuildOsConfig(fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), desc, nil, nil, osconfigserver.BuildGooPackageConfig(pkgs, nil, nil), nil, nil)
+			for _, image := range windowsImages {
+				pkgs := []*osconfigpb.Package{osconfigserver.BuildPackage(packageName)}
+				oc = osconfigserver.BuildOsConfig(fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), desc, nil, nil, osconfigserver.BuildGooPackageConfig(pkgs, nil, nil), nil, nil)
+				setup := NewPackageManagementTestSetup(image, fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), testName, "", oc, nil, nil, vf)
+				pkgTestSetup = append(pkgTestSetup, setup)
```
ryanwe: it looks like the `setup :=` and `pkgTestSetup :=` lines are duplicated for each case. Can they be moved after the case statements? Alternatively, you could factor our these cases into separate functions if you think that would be cleaner.

osconfig_tests/test_suites/package_management/package_management_test_data.go
```diff
@@ -100,49 +102,58 @@ func addPackageInstallTest(pkgTestSetup []*packageManagementTestSetup, testProje
...
-			vs = fmt.Sprintf(packageInstalledString)
+			for _, image := range windowsImages {
+				pkgs := []*osconfigpb.Package{osconfigserver.BuildPackage(packageName)}
+				oc = osconfigserver.BuildOsConfig(fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), desc, nil, nil, osconfigserver.BuildGooPackageConfig(pkgs, nil, nil), nil, nil)
+				vs = fmt.Sprintf(packageInstalledString)
+				instanceName := fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix)
+				assign := osconfigserver.BuildAssignment(fmt.Sprintf("%s-%s-%s", path.Base(image), testName, uniqueSuffix), desc, osconfigserver.BuildInstanceFilterExpression(instanceName), []string{fmt.Sprintf("projects/%s/osConfigs/%s", testProjectConfig.TestProjectID, oc.Name)})
+				ss := getPackageInstallStartupScript(tuple.pkgManager, packageName)
+				setup := NewPackageManagementTestSetup(image, instanceName, testName, vs, oc, assign, ss, vf)
+				pkgTestSetup = append(pkgTestSetup, setup)
```
ryanwe: Here too, I really like the approach, but there's a bunch of duplicate boiler plate code that could be refactored.

iamsubratp: i really wanted to refactor out the common code, but the only difficulty is BuildOsConfig utility method. This method is tightly coupled with the type of test case and the package manager of the platform.
To refactor out the BuildOsConfig call, i either might have to maintain a static mapping of package-manager and testcase to the function or i am thinking i can use builder pattern to build the osconfig.
Since i haven't really thought through this fully, how about i add this as a TODO task for refactoring, for now?

osconfig_tests/test_suites/package_management/package_management.go

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: done

osconfig_tests/test_suites/package_management/package_management_test_data.go
iamsubratp: moving it to separate function since moving it out of switch would require to create a new slice for each platform and in the end append each slice to the common pkgtestSetup slice(too much work).

Refactored it out to a separate function.

PR #745, opened 2019-04-11
OS Config - Make service poll interval overridable from metadata

ryanwe:
- Only supports instance metadata since we don't want this done very often.
- Handy for testing and demos.

### DIFF
--- a/cli_tools/google-osconfig-agent/config/config.go
+++ b/cli_tools/google-osconfig-agent/config/config.go
@@ -83,15 +83,6 @@ func parseBool(s string) bool {
 	return enabled
 }
 
-func parseInt(s string) int {
-	val, err := strconv.Atoi(s)
-	if err != nil {
-		// Bad entry returns default
-		return 0
-	}
-	return val
-}
-
 type metadataJSON struct {
 	Instance instanceJSON
 	Project  projectJSON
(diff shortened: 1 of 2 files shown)


ryanwe:
> just a general question, do we also want customers to have control to tweak this?

Good question. In general, we don't recommend it. But given that this tool is open source, users can and will tweak this. We have limits to throttle these calls on the back end. And making this easier to change now will help with testing and demo'ing.

cli_tools/google-osconfig-agent/config/config.go
```diff
@@ -100,21 +111,23 @@ type projectJSON struct {
...
-	OSPatchEnabled     string `json:"os-patch-enabled"`
-	OSPackageEnabled   string `json:"os-package-enabled"`
-	OSDebugEnabled     string `json:"os-debug-enabled"`
-	OSConfigEndpoint   string `json:"os-config-endpoint"`
+	OSInventoryEnabled   string `json:"os-inventory-enabled"`
+	OSPatchEnabled       string `json:"os-patch-enabled"`
+	OSPackageEnabled     string `json:"os-package-enabled"`
+	OSDebugEnabled       string `json:"os-debug-enabled"`
+	OSConfigEndpoint     string `json:"os-config-endpoint"`
+	OSConfigPollInterval string `json:"os-config-poll-interval"`
```
adjackura: I think *json.Number is safer

PR #747, opened 2019-04-11
Fix nil pointer failure and add unit test

iamsubratp:
If the osconfigpollinterval is not set in the metadata(like in case of our end to end tests), the code path generates nil pointer exception.

[Fail test](https://prow.k8s.io/view/gcs/compute-image-tools-test/logs/osconfig-tests/1116435195327680512)

### DIFF
--- a/cli_tools/google-osconfig-agent/config/config_test.go
+++ b/cli_tools/google-osconfig-agent/config/config_test.go
@@ -76,3 +76,56 @@ func TestSetConfig(t *testing.T) {
 		t.Errorf("Default poll interval: got(%f) != want(%d)", SvcPollInterval().Minutes(), 3)
 	}
 }
+
+func TestSetConfigDefaultValues(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, `{"instance":{"id":12345,"name":"name","zone":"zone","attributes":{"os-config-endpoint":"SvcEndpoint"}}}`)
+	}))
+	defer ts.Close()
+
+	if err := os.Setenv("GCE_METADATA_HOST", strings.Trim(ts.URL, "http://")); err != nil {
+		t.Fatalf("Error running os.Setenv: %v", err)
+	}
+
+	if err := SetConfig(); err != nil {
+		t.Fatalf("Error running SetConfig: %v", err)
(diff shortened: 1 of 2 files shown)


cli_tools/google-osconfig-agent/config/config_test.go
```diff
@@ -76,3 +76,56 @@ func TestSetConfig(t *testing.T) {
...
+		desc string
+		op   func() string
+		want string
+	}{
+		{"SvcEndpoint", SvcEndpoint, "SvcEndpoint"},
+		{"Instance", Instance, "zone/instances/name"},
+		{"ID", ID, "12345"},
+		{"ProjectID", ProjectID, "projectId"},
+		{"Zone", Zone, "zone"},
+		{"Name", Name, "name"},
```
ryanwe: Is this testing these defaults? It looks like you set them on line 82.

cli_tools/google-osconfig-agent/config/config_test.go
```diff
@@ -76,3 +76,56 @@ func TestSetConfig(t *testing.T) {
...
+	}
+	for _, tt := range testsBool {
+		if tt.op() != tt.want {
+			t.Errorf("%q: got(%t) != want(%t)", tt.desc, tt.op(), tt.want)
+		}
+	}
+
+	if SvcPollInterval().Minutes() != float64(osConfigPollIntervalDefault) {
+		t.Errorf("Default poll interval: got(%f) != want(%d)", SvcPollInterval().Minutes(), osConfigPollIntervalDefault)
+	}
```
ryanwe: Nice tests!

cli_tools/google-osconfig-agent/config/config_test.go
iamsubratp: no.. i only intended to test the metadata attributes added by end end user since those may or may not present These are infra level attributes and i assumed that these will be already during events like project setup, instance spin up

ryanwe: Should they be removed from this test then? It looks like they're already tested above.

iamsubratp: i actually updated our tests to test for default svcendpoint

ryanwe: Oh. I see. LGTM 👍

PR #749, opened 2019-04-15
[OsConfig Agent] Add daisy workflow for integration test

iamsubratp:
With the introduction of dependency on metadata attributes, executing osconfig-test binary is no longer an option. This daisy workflow will
- spin up a vm
- pull the latest code from repo & branch specified by user
- run the osconfig-agent-test binary
- publish the logs in gcs bucket

P.S.: To test the latest changes with the integration test framework, make sure you have pushed the changes to a remote branch and use the remote branch as input to the daisy workflow

Future TODO items:
- currently, this does not support testing changes done to agent itself, through e2e tests. This is because we have a hard dependency e2e-test-runner on [rapture](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/osconfig_tests/utils/utils.go#L40).
Will address this TODO item in a future PR

### DIFF
--- a/osconfig_tests/daisy_workflow/build_osconfig_agent_test.sh
+++ b/osconfig_tests/daisy_workflow/build_osconfig_agent_test.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+# Copyright 2019 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


adjackura:
Is this still required as is with the new updates?

iamsubratp:
> Is this still required as is with the new updates?

i would still like to keep it. executing the test-runner from GCE vm would be closest to how our e2e tests would run.

PR #755, opened 2019-04-22
[Osconfig Agent] Preserve test run logs

iamsubratp:
The logs are helpful in debugging end to end tests.

Also, removed the dependency from config library that is only supposed
to be used by the agent.


osconfig_tests/config/config.go
```diff
@@ -0,0 +1,52 @@
...
+
+	TestConfig *config
+)
+
+type config struct {
+	svcEndpoint, oauthPath, logsBucket, logsPath string
+}
+
+// SetConfig sets the test configs
+func SetConfig() error {
```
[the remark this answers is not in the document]
iamsubratp: i had the idea of adding an interface to provide custom location for agent downloads. but i think you are right, that can also be a flag. i do not see any use case where we need to be passed to test runner through any value that is not static.

will change it.

osconfig_tests/test_suites/inventory/inventory.go
```diff
@@ -451,6 +459,7 @@ func runPackagesTest(ga *apiBeta.GuestAttributes, testSetup *inventoryTestSetup,
 func inventoryTestCase(ctx context.Context, testSetup *inventoryTestSetup, tests chan *junitxml.TestCase, wg *sync.WaitGroup, logger *log.Logger, regex *regexp.Regexp, testProjectConfig *testconfig.Project) {
 	defer wg.Done()
 
+	var logwg sync.WaitGroup
```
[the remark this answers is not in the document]
iamsubratp: good catch. added. thanks

osconfig_tests/utils/utils.go
```diff
@@ -111,3 +121,50 @@ func GetStatusFromError(err error) string {
 	}
 	return fmt.Sprintf("%v", err)
 }
+
+// StreamSerialOutput stores the serial output of an instance to GCS bucket
+func StreamSerialOutput(ctx context.Context, i *compute.Instance, storageClient *storage.Client, logsPath, bucket string, wg *sync.WaitGroup, port int64, interval time.Duration) {
+	wg.Add(1)
```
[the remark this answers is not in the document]
iamsubratp: what is the rationale behind adding the wait from outside of the goroutine?

osconfig_tests/compute/instance.go
```diff
@@ -27,13 +27,13 @@ import (
 // Instance is a compute instance.
 type Instance struct {
 	*api.Instance
-	client        daisyCompute.Client
+	Client        daisyCompute.Client
```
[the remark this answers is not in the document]
iamsubratp: good suggestion, will move

osconfig_tests/compute/instance.go
```diff
@@ -82,6 +88,52 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
...
+				// Instance is stopped or stopping.
+				status, sErr := i.client.InstanceStatus(path.Base(i.Project), path.Base(i.Zone), i.Name)
+				switch status {
+				case "TERMINATED", "STOPPED", "STOPPING":
+					if sErr == nil {
+						break Loop
+					}
+				}
+				logger.Errorf("Instance %q: error getting serial port: %v", i.Name, err)
+				break Loop
```
ryanwe: I usually consider this a a really good indicator that the code should be refactored, often by factoring out a function. However, in this case, could you change the logic to something like this and remove the label?
```go
if !isTerminal(status) {
  logger.Errorf(...)
}
continue;
```

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -0,0 +1,32 @@
...
+	if storageClient != nil {
+		return storageClient, nil
+	}
+
+	return storage.NewClient(ctx, option.WithCredentialsFile(config.OauthPath()))
+}
+
+// GetOsConfigClient returns a singleton GCP client for osconfig tests
+func GetOsConfigClient(ctx context.Context) (*osconfig.Client, error) {
+	if osconfigClient != nil {
```
ryanwe: These aren't thread safe. Will that ever be an issue? Will we ever run our tests in parallel from a single process?

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -0,0 +1,32 @@
+package gcpclients
```
ryanwe: Missing copyright header? Here and elsewhere.

osconfig_tests/utils/utils.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 var (
+
```
ryanwe: unnecessary change?

osconfig_tests/compute/instance.go
```diff
@@ -82,6 +88,52 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
 	}
 }
 
+// StreamSerialOutput stores the serial output of an instance to GCS bucket
+func (i *Instance) StreamSerialOutput(ctx context.Context, storageClient *storage.Client, logsPath, bucket string, wg *sync.WaitGroup, port int64, interval time.Duration) {
+	defer wg.Done()
```
ryanwe: minor: consider renaming `wg` to `logwg`. That will make it easier to map these to the ones where you call `logwg.Add(1)` and `logwg.Wait()`.

iamsubratp: done

osconfig_tests/compute/instance.go
iamsubratp: removed the label.

osconfig_tests/gcp_clients/gcp_clients.go
iamsubratp: done

osconfig_tests/utils/utils.go
iamsubratp: removed.

osconfig_tests/utils/utils.go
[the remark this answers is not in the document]
adjackura: It's so you avoid race conditions, if inside the goroutine the calling function could pass Wait() before any goroutine actually calls Add

osconfig_tests/test_suites/inventory/inventory.go
```diff
@@ -21,6 +21,8 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/config"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/gcp_clients"
```
adjackura: You imports are out of order

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -17,7 +17,10 @@ package packagemanagement
 import (
 	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/config"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/gcp_clients"
```
adjackura: fix imports

osconfig_tests/osconfig_server/osconfig_wrapper.go
```diff
@@ -18,8 +18,9 @@ package osconfigserver
 import (
 	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/gcp_clients"
```
adjackura: fix imports

osconfig_tests/osconfig_server/assignment_wrapper.go
```diff
@@ -18,6 +18,7 @@ package osconfigserver
 import (
 	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/gcp_clients"
```
adjackura: fix imports

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -12,28 +12,35 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-package osconfigserver
+package gcpclients
 
 import (
+	"cloud.google.com/go/storage"
```
adjackura: fix imports

osconfig_tests/compute/instance.go
```diff
@@ -16,8 +16,14 @@
 package compute
 
 import (
+	"bytes"
+	"cloud.google.com/go/storage"
+	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/logger"
```
adjackura: fix imports

osconfig_tests/compute/instance.go
```diff
@@ -82,6 +88,54 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
...
+	defer logwg.Done()
+
+	logsObj := path.Join(logsPath, fmt.Sprintf("%s-serial-port%d.log", i.Name, port))
+	logger.Infof("Streaming instance %q serial port %d output to https: //storage.cloud.google.com/%s/%s", i.Name, port, bucket, logsObj)
+	var start int64
+	var buf bytes.Buffer
+	var gcsErr, isTerminalStatus bool
+	tick := time.Tick(interval)
+
+
```
adjackura: extra line here, gofmt should fix

osconfig_tests/compute/instance.go
```diff
@@ -82,6 +88,54 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
...
+		select {
+		case <-tick:
+			resp, err := i.client.GetSerialPortOutput(path.Base(i.Project), path.Base(i.Zone), i.Name, port, start)
+			if err != nil {
+				// Instance is stopped or stopping.
+				status,_ := i.client.InstanceStatus(path.Base(i.Project), path.Base(i.Zone), i.Name)
+				if !isTerminal(status) {
+						logger.Errorf("Instance %q: error getting serial port: %v", i.Name, err)
+				}
+				isTerminalStatus = true
```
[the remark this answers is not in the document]
adjackura: It'd be cleaner to not have the extra function and just do this:
if status == "TERMINATED" || status == "STOPPED" || status == "STOPPING" {
return
}
logger.Errorf("Instance %q: error getting serial port: %v", i.Name, err)
continue

osconfig_tests/compute/instance.go
```diff
@@ -82,6 +88,54 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
...
+			wc.ContentType = "text/plain"
+			if _, err := wc.Write(buf.Bytes()); err != nil && !gcsErr {
+				gcsErr = true
+				logger.Errorf("Instance %q: error writing log to GCS: %v", i.Name, err)
+				continue
+			}
+			if err := wc.Close(); err != nil && !gcsErr {
+				gcsErr = true
+				logger.Errorf("Instance %q: error saving log to GCS: %v", i.Name, err)
+				continue
```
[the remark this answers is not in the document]
adjackura: 2 is better as you don't want to write potentially incorrect data, can you also fix the daisy function that does the same?
Just put the error check before the log line. For this code (not daisy), you can just remove the the gcsErr check all together as this is a test and not user tooling and log spam is not so much of an issue (we may in fact prefer it)
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/step_create_instances.go#L67

osconfig_tests/gcp_clients/gcp_clients.go
iamsubratp: even today, the test suites are run on independent goroutines and each test case in a test suite is again run on independent goroutines.

ryanwe: Then I guess the questions are: Should this be thread safe, it seems weird that it wouldn't be. What happens if two goroutines create different clients? Can the clients be created in init() or is double-check-locking or channels better?

osconfig_tests/compute/instance.go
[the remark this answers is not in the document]
iamsubratp: i just noticed, the extra function can be used in other places. we also check for terminal state in `WaitForSerialOutput()`. I would like to keep the function and change the logic to 


if isTerminal(status) {
return
}
logger.Errorf("Instance %q: error getting serial port: %v", i.Name, err)
continue

osconfig_tests/gcp_clients/gcp_clients.go
iamsubratp: at first i thought this about the client being thread safe. Later, reading through multi-threading paradigms, i realized, it is about the initialization of the the client being thread safe.

yeah, you are right, the client initialization of the singleton client is not thread safe. Double check locking or channels are great way to handle singleton initialization. However, I figured a more idiomatic go way to handle this. The go `sync` library provides a a type called `Once`.
More details [sync.Once](https://golang.org/pkg/sync/#Once)

osconfig_tests/compute/instance.go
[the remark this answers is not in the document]
iamsubratp: as suggested in option 2, i have changed the logic to close the writer only when there is a write. i am getting rid of the logic of already logged logs, altogether

changed the logic for daisy as well.

osconfig_tests/test_suites/inventory/inventory.go
iamsubratp: fixed

osconfig_tests/test_suites/package_management/package_management.go
iamsubratp: my intellij does not do autoimport, need to configure it properly. my vim does that though

osconfig_tests/osconfig_server/assignment_wrapper.go
iamsubratp: fixed

osconfig_tests/osconfig_server/osconfig_wrapper.go
iamsubratp: done

osconfig_tests/config/config.go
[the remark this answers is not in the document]
iamsubratp: ok.. i am adding this back. I am using "logspath" which has the time at which the tests are run. using config struct to store the logspath. 
i dont think there is a harm in keeping the config

osconfig_tests/compute/instance.go
```diff
@@ -82,6 +89,50 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
...
 
+// StreamSerialOutput stores the serial output of an instance to GCS bucket
+func (i *Instance) StreamSerialOutput(ctx context.Context, storageClient *storage.Client, logsPath, bucket string, logwg *sync.WaitGroup, port int64, interval time.Duration) {
+	defer logwg.Done()
+
+	logsObj := path.Join(logsPath, fmt.Sprintf("%s-serial-port%d.log", i.Name, port))
+	logger.Infof("Streaming instance %q serial port %d output to https://storage.cloud.google.com/%s/%s", i.Name, port, bucket, logsObj)
+	var start int64
+	var buf bytes.Buffer
+	var isTerminalStatus bool
```
ryanwe: remove `isTerminalStatus` since it's not used anymore.

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -0,0 +1,80 @@
...
+func createStorageClient(ctx context.Context) {
+	fmt.Printf("creating instance")
+	storageClient, _ = storage.NewClient(ctx, option.WithCredentialsFile(config.OauthPath()))
+	fmt.Printf("accessing storage client: %s\n", func(client *storage.Client) string {
+		if client == nil {
+			return "false"
+		} else {
+			return "true"
+		}
+	}(storageClient))
```
ryanwe: Why use this immediately invoked function? Also, would this be clearer as

```go
if storageClient != nil {
  fmt.Println("Created storage client")
} else {
  fmt.Println("No storage client")
}
```

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -0,0 +1,80 @@
...
+func createOsConfigClient(ctx context.Context) {
+	fmt.Printf("creating instance")
+	osconfigClient, _ = osconfig.NewClient(ctx, option.WithCredentialsFile(config.OauthPath()), option.WithEndpoint(config.SvcEndpoint()))
+	fmt.Printf("accessing storage client: %s\n", func(client *osconfig.Client) string {
+		if client == nil {
+			return "false"
+		} else {
+			return "true"
+		}
+	}(osconfigClient))
```
ryanwe: Same here.

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -0,0 +1,80 @@
...
+
+var (
+	storageClient  *storage.Client
+	osconfigClient *osconfig.Client
+
+	populateClientOnce sync.Once
+)
+
+func PopulateClients(ctx context.Context) {
+	populateClientOnce.Do(func() {
```
ryanwe: Nice! How about calling this function from `create*Client()` instead of forcing consumers to call this first?

osconfig_tests/gcp_clients/gcp_clients.go
iamsubratp: oops... i added these lines while i ran into an issue with client initialization. should have removed them. since these were useful lines for debugging, i will log them with debug level

osconfig_tests/gcp_clients/gcp_clients.go
iamsubratp: > Nice! How about calling this function from `create*Client()` instead of forcing consumers to call this first?
you mean `Get*Client()` ? 
when i relate this to OOP, this should be a part of object initialization which essentially mean dependency injection.
Dependency injection in functional programming language  has been bothering me for a while, now. You are right, consumer should not have to call a method to initialize dependency objects.
I feel `init() ` is Go idiomatic way to perform such tasks. but having faced the issue with initialization sequence of "init methods`, i am reluctant to use it anymore.

calling the populate method from `Get*Client()` is ok but then the initialization of the clients might not be consistent since initializtion would depend on which Get*client method was invoked first. This might be a bit annoying while debugging an issue.
if you agree, i would prefer to solve this the correct way which is to use DI, in a different PR

ryanwe: SGTM. Addressing `SetConfig` would be good too. I've been bit by that before :snake:.

osconfig_tests/compute/instance.go
iamsubratp: done

osconfig_tests/gcp_clients/gcp_clients.go
iamsubratp: done

osconfig_tests/config/config.go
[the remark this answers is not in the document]
adjackura: Just make that a var as well, the issue is your running this initialization right at main, so really there's no point in the struct since its only setup once at start, and it introduces the ability for you to make a nil pointer dereference.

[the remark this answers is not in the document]
iamsubratp: i am removing the struct. but to address the problem of nil pointer dereference, i think DI is the correct way to address it.

osconfig_tests/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+	oauthDefault           = ""
+	bucketDefault          = "osconfig-agent-end2end-tests"
+	logsPathFormatDefault  = "logs-%s"
+	logPushIntervalDefault = 3 * time.Second
+
+	logsPath string
+)
+
+func init() {
+	logsPath = fmt.Sprintf(logsPathFormatDefault, time.Now().Format("2006-01-02-15:04:05"))
```
adjackura: no need to have this in init, just have this in the var declaration

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -0,0 +1,79 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package gcpclients
+
+import (
+	"context"
+	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/google-osconfig-agent/logger"
```
adjackura: Here and elsewhere, the std imports and third party ones should be separate

iamsubratp: done

osconfig_tests/config/config.go
iamsubratp: done

PR #756, opened 2019-04-22
Add "custom_translate_workflow" flag for gce vm image import cli tool

dntczdx:
Add "custom_translate_workflow" flag for gce vm image import cli tool

Background: gcloud will use gce_vm_image_import cli tool instead of daisy. "custom_translate_workflow" is a flag that we want to support to keep backward compatible.


dntczdx:
Adjusted format as hint

cli_tools/gce_vm_image_import/main.go
```diff
@@ -108,6 +109,10 @@ func validateAndParseFlags() error {
 		return fmt.Errorf("either -data_disk or -os has to be specified, but not both")
 	}
 
+	if *osID == "" && *customTranWorkflow != "" {
+		return fmt.Errorf("-custom_translate_workflow can't be specified when -osID is not specified")
```
adjackura: it's just -os

compute-image-tools.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adjackura: You probably didn't mean to include this file

dntczdx:
Addressed

PR #759, opened 2019-04-23
OVF Import: use OVF osType attribute if --os flag not set.

zoran15:


cli_tools/gce_ovf_import/main_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
```
adjackura: Separate this import from the std imports

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -36,6 +36,36 @@ const (
 	usbController          uint16 = 23
 )
 
+//TODO: add Windows 7 and 10 once BYOL is supported
+//Full list: https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html
+
+// Mapping OVF osType attribute to importer OS ID
+var ovfOSTypeToOSID = map[string]string{
```
adjackura: Just fyi I'm pretty sure these fields in vmware mean close to nothing, so users could be setting incorrect values with really no negative side effects. The issue it could cause with import is that the import will fail later on.

zoran15: These values are exported in OVF by VMWare. Users will probably not set them themselves. 

As you pointed out, having these values set incorrectly would fail the import during translate. Not sure we can do much about it, but I doubt this will be an issue, except if users are modifying OVF descriptor themselves.

PR #761, opened 2019-04-24
fix bucket split logic

dntczdx:
Currently, scratch bucket name can't be processed correctly since regex doesn't match. Fix by changing to correct regex.

### DIFF
--- a/cli_tools/common/utils/storage/storage_client.go
+++ b/cli_tools/common/utils/storage/storage_client.go
@@ -31,8 +31,9 @@ import (
 )
 
 var (
-	bucketRegex = regexp.MustCompile(`^gs://([a-z0-9][-_.a-z0-9]*)/?$`)
-	gsRegex = regexp.MustCompile(`^gs://([a-z0-9][-_.a-z0-9]*)/(.+)$`)
+	bucket = `([a-z0-9][-_.a-z0-9]*)`
+	bucketRegex = regexp.MustCompile(fmt.Sprintf(`^gs://%s/?$`, bucket))
+	gsRegex = regexp.MustCompile(fmt.Sprintf(`^gs://%s/(.+)$`, bucket))
 )
 
 // StorageClient implements domain.StorageClientInterface. It implements main Storage functions
@@ -187,14 +188,14 @@ func SplitGCSPath(p string) (string, string, error) {
 	return "", "", fmt.Errorf("%q is not a valid GCS path", p)
 }
 
-// SplitBucketGCSPath splits GCS path to get bucket name
-func SplitBucketGCSPath(p string) (string, error) {
+// GetBucketNameFromGCSPath splits GCS path to get bucket name
+func GetBucketNameFromGCSPath(p string) (string, error) {
 	matches := bucketRegex.FindStringSubmatch(p)
 	if matches != nil {
 		return matches[1], nil
 	}
 
-	return "", fmt.Errorf("%q is not a valid GCS bucket", p)
+	return "", fmt.Errorf("%q is not a valid GCS bucket path", p)
 }
 
 // HTTPClient implements domain.HTTPClientInterface which abstracts HTTP functionality used by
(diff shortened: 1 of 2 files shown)


cli_tools/common/utils/storage/storage_client.go
```diff
@@ -31,6 +31,7 @@ import (
 )
 
 var (
+	bucketRegex = regexp.MustCompile(`^gs://([a-z0-9][-_.a-z0-9]*)/?$`)
```
zoran15: Extract bucket regex string and reuse it in both regex's.

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -186,6 +187,16 @@ func SplitGCSPath(p string) (string, string, error) {
 	return "", "", fmt.Errorf("%q is not a valid GCS path", p)
 }
 
+// SplitBucketGCSPath splits GCS path to get bucket name
+func SplitBucketGCSPath(p string) (string, error) {
```
zoran15: Rename to GetBucketNameFromGCSPath(gcsPath string) or similar current name is not appropriate for what this func does.

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -186,6 +187,16 @@ func SplitGCSPath(p string) (string, string, error) {
...
+	matches := bucketRegex.FindStringSubmatch(p)
+	if matches != nil {
+		return matches[1], nil
+	}
+
+	return "", fmt.Errorf("%q is not a valid GCS bucket", p)
```
zoran15: ...is not a valid GCS bucket path

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -31,7 +31,9 @@ import (
 )
 
 var (
-	gsRegex = regexp.MustCompile(`^gs://([a-z0-9][-_.a-z0-9]*)/(.+)$`)
+	bucket = `([a-z0-9][-_.a-z0-9]*)`
+	bucketRegex = regexp.MustCompile(fmt.Sprintf(`^gs://%s/?$`, bucket))
+	gsRegex = regexp.MustCompile(fmt.Sprintf(`^gs://%s/(.+)$`, bucket))
```
iamsubratp: the variable names are not entirely clear.
"bucket" is essentially the regex that follows gcs naming pattern, right? if so, it should be changed to bucketregex. and then bucketregex, gsregex should be changed to something more specific to what pattern the regex is supposed to match.

dntczdx: Sounds good, however I followed pattern just like in daisy/storage.go to keep the style. We may have a bigger refactoring to resolve this naming confusion.

zoran15: We can do it step by step. Readability trumps consistency IMO.

dntczdx:
Regex renamed as Subrat suggested

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -31,7 +31,9 @@ import (
 )
 
 var (
-	gsRegex = regexp.MustCompile(`^gs://([a-z0-9][-_.a-z0-9]*)/(.+)$`)
+	bucket      = `([a-z0-9][-_.a-z0-9]*)`
+	bucketRegex = regexp.MustCompile(fmt.Sprintf(`^gs://%s/(.*)$`, bucket))
```
adjackura: Won't you need to make the '/' optional?

dntczdx: ouch, I saw this comment after merging...
I didn't make '/' optional because upstream (gcloud) always pass in '/'. If I simply end it with "/?(.*)", then it may accept wrong bucket name. For example, that will pass "gs://bucket-name_mass".
We can make a more complex regex to be compatible with "gs://bucket-name" but so far I don't see the usage

PR #762, opened 2019-04-24
Import: Windows Server 2019

rofuentes:
- Import (translate) workflow
- Daisy integration test
- Test import image (no drivers/GCE binaries) exists in test project

### DIFF
--- a/daisy_integration_tests/2019_translate.wf.json
+++ b/daisy_integration_tests/2019_translate.wf.json
@@ -1,0 +1,89 @@
+{
+  "Name": "windows-2019",
+  "Sources": {
+    "post_translate_test.ps1": "./scripts/post_translate_test.ps1"
+  },
+  "Vars": {
+    "image_name": {
+      "Value": "windows-2019-translate-test-${ID}"
+    },
+    "source_image": {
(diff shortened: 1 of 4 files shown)


daisy_workflows/image_import/windows/translate_windows_81_byol.wf.json
```diff
@@ -0,0 +1,74 @@
...
+        }
+      },
+      "Timeout": "60m"
+    },
+    "create-image": {
+      "CreateImages": [
+        {
+          "Name": "${image_name}",
+          "SourceDisk": "${source_disk}",
+          "Licenses": ["projects/windows-cloud/global/licenses/windows-server-2012-r2-byol"],
```
adjackura: Do we not have a windows-8-1-enterprise-byol license?

rofuentes: This was probably an oversight, I will update and create the licence if it does not exist.

rofuentes: I am moving the Windows 8.1 import workflow into a separate PR.

PR #763, opened 2019-04-25
OVF Import: instance creation in sole tenant nodes

zoran15:
OVF Import: added -node-affinity-label flag to support creating instances in sole tenant nodes (needed for BYOL)

### DIFF
--- a/cli_tools/common/utils/compute/node_affinity_label_parser.go
+++ b/cli_tools/common/utils/compute/node_affinity_label_parser.go
@@ -28,9 +28,6 @@ import (
 func ParseNodeAffinityLabels(labels []string) ([]*compute.SchedulingNodeAffinity, error) {
 	var nodeAffinities []*compute.SchedulingNodeAffinity
 
-	if labels == nil || len(labels) == 0 {
-		return nodeAffinities, nil
-	}
 	for _, label := range labels {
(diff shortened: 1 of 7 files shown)


cli_tools/common/utils/compute/node_affinity_label_parser.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// ParseNodeAffinityLabels parses sole tenant affinities
+// labels - array of strings with node affinity label info. Each label is of the following format:
+// <key>,<operator>,<value>,<value2>... where <operator> can be one of: IN, NOT.
+// For example: workload,IN,prod,test is a label with key 'workload' and values 'prod' and 'test'
+func ParseNodeAffinityLabels(labels []string) ([]*compute.SchedulingNodeAffinity, error) {
+	var nodeAffinities []*compute.SchedulingNodeAffinity
+
+	if labels == nil || len(labels) == 0 {
```
adjackura: This is a bit redundant, just have len(labels) == 0 or remove this check completly as you can range over a nil slice

zoran15: Done. Also added more tests to cover empty/nil labels flags

PR #767, opened 2019-04-25
Untranslated image: remove from workflow

dntczdx:
Untranslated image is reserved for user when trnaslation fail. So, it shouldn't be automatically cleaned up.


dntczdx:
> Update this PR with details on why this change is made.

It's there: "Untranslated image is reserved for user when trnaslation fail. So, it shouldn't be automatically cleaned up."

dntczdx:
Accidentally closed... reopen

dntczdx:
After a broader scope of discussion, we decide to remove this step for now due to lack of usage and high perf gain.

PR #769, opened 2019-04-26
[Osconfig Agent] Fix e2e tests of yum based platforms

iamsubratp:


osconfig_tests/test_suites/package_management/package_management_utils.go
```diff
@@ -21,11 +21,19 @@ import (
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/utils"
 	"github.com/google/logger"
-
 	api "google.golang.org/api/compute/v1"
 )
 
-func getPackageInstallStartupScript(pkgManager, packageName string) *api.MetadataItems {
+var (
+	yumImageToStartupScriptMap = map[string]string {
```
ryanwe: minor: remove "Map" from the name. I always like to think of how the name will impact the usage. Below, you call `yumImageToStartupScriptMap[image]`. Maybe `yumStartupScripts[image]` would be clearer.

iamsubratp: done

PR #772, opened 2019-05-01
Image import locality support for development purposes

zoran15:
This change adds support for image locality property by utilizing compute alpha API.

This change is not expected to be merged into master in full once compute prod API is updated to have image locality property (only resource_labeler and gce_vm_image_import/main will need to be updated for prod).


adjackura:
We can't do this with the Alpha API, its whitelist only (and not backward compatible) so this will break all workflows done in projects without that whitelist.

zoran15:
> We can't do this with the Alpha API, its whitelist only (and not backward compatible) so this will break all workflows done in projects without that whitelist.

This change will not go to master as is. It's only to enable internal development against an API that is in alpha. For GA, we'll only use prod API.

PR #777, opened 2019-05-06
[gce_vm_image_import] Detect compression image source file (tar.gz and more)

dntczdx:
tar.gz suppose to be a exported image file and thus doesn't need image import. Instead, it should be used for image create.

We use gzip lib to detect whether it's a compression file (not just tar.gz but also other gzip format, since any compressed file can't be a valid image import source). If yes, we hint user to change to image create.

Tested by running gcloud with customized docker image.


cli_tools/gce_vm_image_import/main.go
```diff
@@ -141,10 +148,32 @@ func validateAndParseFlags() error {
...
+	rc, err := storageClient.GetObjectReader(*sourceBucketName, *sourceObjectName)
+	if err != nil {
+		return fmt.Errorf("readFile: unable to open file from bucket %q, file %q: %v", *sourceBucketName, *sourceObjectName, err)
+	}
+	defer rc.Close()
+
+	// Try to extract compression file header of the source file.
+	_, err = gzip.NewReader(rc)
+	if err == nil {
+		return fmt.Errorf("source file is a compression file. You may try 'image create' instead of 'image import'")
```
ericdand:
```suggestion
		return fmt.Errorf("source file is a compressed file. Please use 'images create' instead of 'images import'")
```

dntczdx: Yes, any error thrown from gce_vm_image_import (or any other cli tools) will be natually adopted by gcloud.

dntczdx:
addressed

cli_tools/gce_vm_image_import/main.go
```diff
@@ -141,10 +148,32 @@ func validateAndParseFlags() error {
...
+func validateSourceFile(storageClient commondomain.StorageClientInterface) error {
+	if sourceFile == nil {
+		return nil
+	}
+
+	rc, err := storageClient.GetObjectReader(*sourceBucketName, *sourceObjectName)
+	if err != nil {
+		return fmt.Errorf("readFile: unable to open file from bucket %q, file %q: %v", *sourceBucketName, *sourceObjectName, err)
+	}
+	defer rc.Close()
```
zoran15: This will never be executed in case of an error. Defer calls a method at the end of this method and should be moved before if statement above

cli_tools/gce_vm_image_import/main.go
```diff
@@ -141,10 +148,32 @@ func validateAndParseFlags() error {
...
+
+	// Try to extract compression file header of the source file.
+	_, err = gzip.NewReader(rc)
+	if err == nil {
+		return fmt.Errorf("source file is a compressed file. Please use 'images create' instead of 'images import'")
```
zoran15: Create image is not an alternative to image import in most cases (can't read most formats, can't do translate), so it's not a good idea to suggest it as such. The error message should only describe what is the problem and how to resolve it. Something like:

"cannot import an image from a compressed file. Please provide a path to an uncompressed image file"

cli_tools/gce_vm_image_import/main.go
```diff
@@ -346,10 +375,17 @@ func main() {
...
 	if err != nil {
 		log.Fatalf(err.Error())
 	}
 
+	err = validateSourceFile(storageClient)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+
+
```
zoran15: remove extra blank line

cli_tools/gce_vm_image_import/main.go
adjackura: You dont need to close if GetObjectReader returns an error, you should be able to expect it to not create the client

cli_tools/gce_vm_image_import/main.go
dntczdx: Sounds good. Still want to let people know that it's possible that they used wrong command, because that is why the bug is filed - they should have using 'image create'.
Let's say: "cannot import an image from a compressed file. Please provide a path to an uncompressed image file. If the compressed file is a exported image, please use 'image icreate' instead."

dntczdx:
Refined accordingly

cli_tools/gce_vm_image_import/main.go
zoran15: ... "If the compressed file is an image exported from Google Compute Engine, please use 'images create' instead."

dntczdx: Great!

dntczdx:
Added "If the compressed file is an image exported from Google Compute Engine, please use 'images create' instead"

dntczdx:
Typo resolved :P

dntczdx:
Ping

cli_tools/gce_vm_image_import/main.go
```diff
@@ -128,10 +132,13 @@ func validateAndParseFlags() error {
 	}
 
 	if *sourceFile != "" {
-		_, _, err := storageutils.SplitGCSPath(*sourceFile)
+		bucketName, objectName, err := storageutils.SplitGCSPath(*sourceFile)
```
zoran15: if declare err var first, you can avoid having these two temporary variables. You probably also don't need to have sourceBucketName/sourceObjectName be pointers, they can be plain strings

cli_tools/gce_vm_image_import/main.go
```diff
@@ -346,6 +375,12 @@ func main() {
 	metadataGCEHolder := computeutils.MetadataGCE{}
 	storageClient, err := storageutils.NewStorageClient(
 		ctx, createStorageClient(ctx), logging.NewLogger("[image-import]"))
+	defer storageClient.Close()
```
zoran15: I think Andrew correctly pointed out storageClient might be nil if there was an error creating it. Make sure this is handles. I assume reverting to having this call after the error check should work (if there was an error probably no need to call close on storageClient)

cli_tools/gce_vm_image_import/main.go
dntczdx: Having args to be populated as pointers is how other args in this program did, so I just followed the pattern. I don't see a obvious pro/con on pointers v.s. objects here.

I can replace these 2 temp var with sourceBucketName/sourceObjectName, but I chose not to, because it's meaningless to change their value when SplitGCSPath fails. It's an unexpected side effect, even though it's harmless.

dntczdx:
Added unit test and refined

cli_tools/gce_vm_image_import/main.go
zoran15: Other flags are pointers as the are returned from flag methods. This is not a flag but a regular variable so there's no reason to be a pointer. 

It's possible that some other regular variables were declared as pointers but there's no reason for it.

cli_tools/gce_vm_image_import/main_test.go
```diff
@@ -15,10 +15,12 @@
 package main
 
 import (
+	"cloud.google.com/go/storage"
```
zoran15: this is also considered external and should go to the second group (even though it's google)

zoran15:
> yes

What for?

cli_tools/gce_vm_image_import/main.go
dntczdx: I read the code carefully and got the hidden reason why those existing vars need to be a pointer: they need to call "testutils.SetStringP" to set/recover the value. So it's because of testability.
Due that my tests don't need to set them, I changed them to objects. Thanks!

PR #780, opened 2019-05-07
RHEL 8 image build.

zmarano:
- Cleanup old builds which are EOL for RHEL for SAP Apps/HANA.
- Cleanup old unused synopsis logs.
- Normalize EL google-compute-engine repos.


daisy_workflows/image_build/enterprise_linux/ks_helpers.py
```diff
@@ -276,6 +264,17 @@ def BuildKsConfig(release, google_cloud_repo, byol, sap, sap_hana, sap_apps,
     custom_post = '\n'.join([ol_post, el_post])
     cleanup = FetchConfigPart('el7-cleanup.cfg')
     repo_version = 'el7'
+  elif release == "rhel8":
+    logging.info('Building RHEL 8 image.')
+    ks_packages = FetchConfigPart('el8-packages.cfg')
+    ks_options = FetchConfigPart('el8-options.cfg')
+    if uefi:
+      ks_options = FetchConfigPart('el8-uefi-options.cfg')
+    rhel_post = FetchConfigPart('rhel8-post.cfg')
```
hopkiw: is it intentional to have `el8-post` AND `rhel8-post` files? can they be more descriptively named?

zmarano: It is- that's because EL8 will apply to CentOS as well, where as rhel8 only applies to RHEL.

PR #785, opened 2019-05-10
OVF import e2e tests

zoran15:
Notes to make the review easier:
- this change includes refactoring gce_ovf_import/main.go by extracting most of the logic to ovf_importer.go and ovf_import_params_validator.go (including tests). Most of this code was already reviewed. This was necessary to make e2e tests easier to write.
- Code in gce_ovf_import_tests/ is based on osconfig_tests/
- gce_ovf_import_tests/scripts/ovf_import_test_ubuntu_3_disks.sh is based on daisy_integration_tests/scripts/post_translate_test.sh with extra check for mounted disks added

Docker image build was not tested yet. The test did run successfully when executed directly on dev box.

The test suite currently includes only one (fairly complex) scenario. Once we verify Prow is configured correctly to run this test, more test cases will follow.

### DIFF
--- a/daisy_workflows/ovf_import/import_ovf.wf.json
+++ b/daisy_workflows/ovf_import/import_ovf.wf.json
@@ -24,7 +24,6 @@
     },
     "translation_disk_name": "temp-translation-disk-${ID}",
     "boot_image_name": "boot-image-${ID}",
-    "boot_image_name_untranslated": "boot-image-untranslated-${ID}",
     "machine_type": "n1-standard-1",
     "network": {
       "Value": "global/networks/default",
@@ -54,13 +53,6 @@
           "import_subnet": "${subnet}"
         }
       }
-    },
-    "take-image-of-untranslated-disk": {
-      "CreateImages": [{
-        "Name": "${boot_image_name_untranslated}",
-        "Description": "Imported (untranslated) image from ${boot_disk_file}",
-        "SourceDisk": "${translation_disk_name}"
-      }]
     },
     "translate": {
       "IncludeWorkflow": {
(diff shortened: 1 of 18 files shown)


gce_ovf_import_tests/test_config/project_config.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// GetProject creates a Project object for given project ID and zone
+func GetProject(projectID, testZone string) *Project {
+	return &Project{
+		TestProjectID:       projectID,
+		TestZone:            testZone,
+		ServiceAccountEmail: "default",
+		ServiceAccountScopes: []string{
+			"https://www.googleapis.com/auth/cloud-platform",
```
adjackura: This is a very broad scoping, is this needed? Can we get away with a smaller subset?

daisy_workflows/ovf_import/import_ovf.wf.json
```diff
@@ -55,13 +54,6 @@
         }
       }
     },
-    "take-image-of-untranslated-disk": {
-      "CreateImages": [{
```
dntczdx: get 17% of improve as well? :P

zoran15: Same thing as for image import.

gce_ovf_import_tests/test_config/project_config.go
zoran15: This configuration is based on OS Config e2e tests which use the same service account scopes.

Public docs state the best practice is to grant 'cloud-platform' scope to an instance then limit it's permissions using IAM: https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam

Since this is internal code running e2e tests in a test project, what are your concerns in having broad set of permissions on a particular instance?

PR #788, opened 2019-05-13
OS Config - run tests across multiple zones

ryanwe:
- Avoids sporadic stock out issues.
- Test runner flags also supports JSON map of zones to quota.
- Increases frequency of E2e test runs.
- Updates prow test runner to run tests in multiple zones.
- Simply chooses a random zone that still has quota to run the test.

Additional, more complex ways of breaking these tests up will be added when needed.

### DIFF
--- a/osconfig_tests/main.go
+++ b/osconfig_tests/main.go
@@ -17,11 +17,13 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"encoding/xml"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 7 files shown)


osconfig_tests/main.go
```diff
@@ -62,13 +65,24 @@ func main() {
...
+		}
+	}
+
+	if len(zones) == 0 {
+		fmt.Println("Error, no zones specified")
 		os.Exit(1)
 	}
 
-	pr := testconfig.GetProject(*testProjectID, *testZone)
+	pr := testconfig.GetProject(*testProjectID, &zones)
```
adjackura: A pointer to a map seems a bit odd and unnecessary

osconfig_tests/test_config/project_config.go
```diff
@@ -14,23 +14,64 @@
...
 }
+
+// GetZone gets a random zone that still has capacity.
+func (p *Project) GetZone() string {
+	p.mux.Lock()
+	defer p.mux.Unlock()
+
+	zc := len(p.zoneIndices)
+	if zc == 0 {
+		fmt.Println("Not enough zone quota sepcified. Specify additional quota in `test_zones`.")
```
adjackura: This should be returned as an error. To make this make more sense CreateComputeInstance should probably just take testProjectConfig and handle all the logic in that function. Also the way this is currently setup makes it very hard to add logic to just wait for an open slot. 

Since this would all take a large refactor a TODO would probably be ok for now.

PR #799, opened 2019-05-16
Update translation workflows to use debian-9 worker image.

zmarano:
- Will make these faster and more reliable.
- Also use newer RHEL compute engine repo.


dntczdx:
Cool, I saw it had been applied to other wf

PR #801, opened 2019-05-20
[OsConfig Agent Test] Fix package management test setup to test case mapping

iamsubratp:
In order to make test run logic generic enough, we abstracted out the test setup. This added a hard coupling between testsetup and testcase for package management.
With the change in package management test case name, the mapping between the testsetup and the testcase broke.


osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -354,38 +363,47 @@ func runPackageInstallFromNewRepoTest(ctx context.Context, testCase *junitxml.Te
...
+	case packageRemovalFunction:
+		tc = junitxml.NewTestCase(testSuiteName, fmt.Sprintf("[Package removal] [%s]", path.Base(testSetup.image)))
+		f = runPackageRemovalTest
+	case packageInstallRemovalFunction:
+		tc = junitxml.NewTestCase(testSuiteName, fmt.Sprintf("[Package no change] [%s]", path.Base(testSetup.image)))
+		f = runPackageInstallRemovalTest
+	case packageInstallFromNewRepoFunction:
+		tc = junitxml.NewTestCase(testSuiteName, fmt.Sprintf("[Add a new package from new repository] [%s]", path.Base(testSetup.image)))
+		f = runPackageInstallFromNewRepoTest
+	}
```
ryanwe: Can we have a default case that errors early? I'd imagine it's a bug to not specify any tests.

iamsubratp: yeah... that makes sense since the tests are specified in a different location. 
i will add a default case. also, in that case, i would still prefer to just log  a warning and not bail out the whole execution of the test runner.

PR #804, opened 2019-05-21
Support "buffer_disk_type" param for export

dntczdx:
Free-trial user has limited pd-ssd quota (only 100GB right now), however, 200GB is required by export workflow. We allow "buffer_disk_type" to be set to "pd-standard" to workaround this problem.

Added corresponding e2e tests.


daisy_workflows/export/disk_export.wf.json
```diff
@@ -21,6 +21,10 @@
       "Value": "200",
       "Description": "size of the export instances disk, this disk is unused for the export but a larger size increase PD read speed"
     },
+    "buffer_disk_type": {
```
adjackura: Use export_instance_disk_type to match the previous variables

dntczdx: changed accordingly

dntczdx:
Changed accordingly

PR #808, opened 2019-05-23
osconfig tests: add new patch tests

adjackura:
Fixes failing WIndows tests
Adds tests for reboot (expected reboot and no-reboot)
Adds tests for Apt and Yum settings

### DIFF
--- a/osconfig_tests/compute/instance.go
+++ b/osconfig_tests/compute/instance.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net/http"
 	"path"
 	"strings"
 	"sync"
@@ -28,12 +29,14 @@ import (
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 
 	daisyCompute "github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
-	api "google.golang.org/api/compute/v1"
+	computeBeta "google.golang.org/api/compute/v0.beta"
+	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
 )
 
 // Instance is a compute instance.
 type Instance struct {
-	*api.Instance
+	*compute.Instance
 	client        daisyCompute.Client
 	Project, Zone string
 }
@@ -89,6 +92,46 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
 	}
 }
 
+// WaitForGuestAttribute waits to a guest attribute to appear.
+func (i *Instance) WaitForGuestAttribute(queryPath, variableKey string, interval, timeout time.Duration) (string, error) {
+	tick := time.Tick(interval)
+	timedout := time.Tick(timeout)
+	for {
+		select {
+		case <-timedout:
+			return "", fmt.Errorf("timed out waiting for guest attribute %q", path.Join(queryPath, variableKey))
+		case <-tick:
+			attr, err := i.GetGuestAttribute(queryPath, variableKey)
+			if err != nil {
+				apiErr, ok := err.(*googleapi.Error)
+				if ok && apiErr.Code == http.StatusNotFound {
+					continue
+				}
+				return "", err
+			}
+			return attr, nil
+		}
+	}
+}
+
+// GetGuestAttributes gets guest attributes for an instance.
+func (i *Instance) GetGuestAttributes(queryPath, variableKey string) (*computeBeta.GuestAttributes, error) {
+	return i.client.GetGuestAttributes(i.Project, i.Zone, i.Name, queryPath, variableKey)
+}
+
+// GetGuestAttribute returns a single guest attribute VariableValue.
+func (i *Instance) GetGuestAttribute(queryPath, variableKey string) (string, error) {
+	resp, err := i.GetGuestAttributes(queryPath, variableKey)
+	if err != nil {
+		return "", err
+	}
+	if resp.QueryValue == nil || resp.QueryValue.Items[0] == nil {
+		return "", nil
+	}
+
+	return resp.QueryValue.Items[0].Value, nil
+}
+
 // StreamSerialOutput stores the serial output of an instance to GCS bucket
 func (i *Instance) StreamSerialOutput(ctx context.Context, storageClient *storage.Client, logsPath, bucket string, logwg *sync.WaitGroup, port int64, interval time.Duration) {
 	defer logwg.Done()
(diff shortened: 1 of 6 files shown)


osconfig_tests/test_suites/patch/patch.go
```diff
@@ -98,32 +176,31 @@ func runExecutePatchJobTest(ctx context.Context, testCase *junitxml.TestCase, te
...
-	}
-	logwg.Add(1)
-	go inst.StreamSerialOutput(ctx, storageClient, path.Join(testSuiteName, config.LogsPath()), config.LogBucket(), logwg, 1, config.LogPushInterval())
+	/*
+		storageClient, err := gcpclients.GetStorageClient(ctx)
+		if err != nil {
+			testCase.WriteFailure("Error getting storage client: %v", err)
+		}
+		go inst.StreamSerialOutput(ctx, storageClient, path.Join(testSuiteName, config.LogsPath()), config.LogBucket(), logwg, 1, config.LogPushInterval())
+	*/
```
ryanwe: :hocho:

osconfig_tests/test_suites/patch/patch.go
```diff
@@ -98,32 +176,31 @@ func runExecutePatchJobTest(ctx context.Context, testCase *junitxml.TestCase, te
...
-	// create patch job
 	parent := fmt.Sprintf("projects/%s", testProjectConfig.TestProjectID)
 	osconfigClient, err := gcpclients.GetOsConfigClient(ctx)
 
-	assertTimeout := testSetup.assertTimeout
-
 	req := &osconfigpb.ExecutePatchJobRequest{
 		Parent:      parent,
-		Description: "testing default patch job run",
+		Description: "testing reboot patch job run",
```
ryanwe: Isn't this the execute patch test and the reboot one is below?

osconfig_tests/test_suites/patch/patch.go
```diff
@@ -79,7 +123,41 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
 	logger.Printf("Finished TestSuite %q", testSuite.Name)
 }
 
-func runExecutePatchJobTest(ctx context.Context, testCase *junitxml.TestCase, testSetup *patchTestSetup, logger *log.Logger, logwg *sync.WaitGroup, testProjectConfig *testconfig.Project) {
+func waitPatchJob(ctx context.Context, job *osconfigpb.PatchJob, timeout time.Duration) error {
```
ryanwe: nit: consider "awaitPatchJob". It makes the code read a little smoother and only adds one char.

osconfig_tests/compute/instance.go
```diff
@@ -89,6 +92,46 @@ func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeo
 	}
 }
 
+// WaitForGuestAttribute waits to a guest attribute to appear.
+func (i *Instance) WaitForGuestAttribute(queryPath, variableKey string, interval, timeout time.Duration) (string, error) {
```
ryanwe: Since `interval` and `timeout` are always the same, should they be consts instead?

osconfig_tests/test_suites/patch/patch.go
```diff
@@ -133,39 +210,74 @@ func runExecutePatchJobTest(ctx context.Context, testCase *junitxml.TestCase, te
...
-			}
-		}
+	testCase.Logf("Agent installed successfully")
+
+	parent := fmt.Sprintf("projects/%s", testProjectConfig.TestProjectID)
+	osconfigClient, err := gcpclients.GetOsConfigClient(ctx)
+
+	req := &osconfigpb.ExecutePatchJobRequest{
+		Parent:      parent,
+		Description: "testing reboot patch job run",
```
ryanwe: nit: can this be reworded to "testing patch job reboot"? That should make it clearer in the logs.

osconfig_tests/test_suites/patch/patch.go
```diff
@@ -133,39 +210,74 @@ func runExecutePatchJobTest(ctx context.Context, testCase *junitxml.TestCase, te
...
+	if err != nil {
+		testCase.WriteFailure("Error retrieving boot count: %v", err)
+		return
+	}
+	num, err := strconv.Atoi(attr)
+	if err != nil {
+		testCase.WriteFailure("Error parsing boot count: %v", err)
+		return
+	}
+	if num < minBootCount || num > maxBootCount {
```
ryanwe: Instead of verifying bootcount is within a range, how about grabbing the value before running the patch job and verifying that it's higher afterwards? That may be more robust if you're concerned about spurious reboots before and during the patch job.

osconfig_tests/test_suites/patch/test_setup.go
```diff
@@ -29,58 +29,108 @@ type patchTestSetup struct {
...
-	machineType: "n1-standard-2",
-}
+	linuxRecordBoot = `
+uri=http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/osconfig_tests/boot_count
+old=$(curl $uri -H "Metadata-Flavor: Google" -f)
+new=$(($old + 1))
+curl -X PUT --data "${new}" $uri -H "Metadata-Flavor: Google"`
+	aptStartup = utils.InstallOSConfigDeb + linuxRecordBoot
+	el6Startup = utils.InstallOSConfigYumEL6 + linuxRecordBoot
+	el7Startup = utils.InstallOSConfigYumEL7 + linuxRecordBoot
```
ryanwe: Is `gcloud` installed by default on these machines and would using that make this a little cleaner?

osconfig_tests/test_config/project_config.go
```diff
@@ -59,8 +58,9 @@ func (p *Project) GetZone() string {
 	zc := len(p.zoneIndices)
 	if zc == 0 {
 		// TODO: return an error instead of stopping the process.
-		fmt.Println("Not enough zone quota sepcified. Specify additional quota in `test_zones`.")
-		os.Exit(1)
+		msg := "Not enough zone quota sepcified. Specify additional quota in `test_zones`."
+		fmt.Println(msg)
+		return msg
```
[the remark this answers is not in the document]
adjackura: The existing logic does fail fast but blows up the entire test infrastructure on a lack of 'quota' for a single test, this change makes only tests that can't get 'quota' fail (in a rather obtuse way but the error will be clear).

This is still a placeholder for a refactor of this code to have much better error handling and maybe even a test queue.

osconfig_tests/test_suites/patch/patch.go
adjackura: This function has two consumers, the first is that we did reboot (min matters, max does not)
the other tests we didnt reboot (max matters min does not)

osconfig_tests/test_suites/patch/test_setup.go
adjackura: Set guest attribute is not a gcloud command as it can only be run inside the guest

osconfig_tests/compute/instance.go
adjackura: This function is meant to be generic, the callers are all for osconfig_tests/install_done right now but in different packages so there isn't really a good place to put a constant

PR #816, opened 2019-05-24
 osconfig_tests: deflake inventory tests, update to use new metadata keys

adjackura:


adjackura:
FYI the associated agent update is being worked on here:
https://github.com/GoogleCloudPlatform/osconfig/compare/master...adjackura:master

PR #824, opened 2019-05-31
add cleaning up done msg for workflow

dntczdx:
add cleaning up done msg for workflow


daisy/workflow.go
```diff
@@ -259,6 +259,7 @@ func (w *Workflow) cleanup() {
 			w.LogWorkflowInfo("Error returned from cleanup hook: %s", err)
 		}
 	}
+	w.LogWorkflowInfo("Workflow %q finished clean up.", w.Name)
```
zoran15: minor: to be consistent with other logs, either use "cleaning up" (preferred) or "cleanup".

PR #826, opened 2019-06-05
[import/export cli] add more image import/export cli tool e2e tests for rich params

dntczdx:
Add more image import/export cli tool e2e tests for rich params.
Some params are hard to test and is not critical to our main scenario, so we skipped them for now:
-oauth, -compute_endpoint_override, and -scratch_bucket_gcs_path,
and all KMS-related params.

### DIFF
--- a/gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
+++ b/gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
@@ -18,12 +18,13 @@ package importtestsuites
 import (
 	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/gce_image_import_export_tests/compute"
 	"log"
 	"regexp"
+	"strings"
 	"sync"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
-	"github.com/GoogleCloudPlatform/compute-image-tools/gce_image_import_export_tests/compute"
 	"github.com/GoogleCloudPlatform/compute-image-tools/gce_image_import_export_tests/test_suites"
 	"github.com/GoogleCloudPlatform/compute-image-tools/go/e2e_test_utils/junitxml"
 	"github.com/GoogleCloudPlatform/compute-image-tools/go/e2e_test_utils/test_config"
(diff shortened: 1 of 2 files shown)


gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
```diff
@@ -99,6 +102,25 @@ func runImageImportOSFromImageTest(
...
+func runImageImportWithRichParamsTest(
+		ctx context.Context, testCase *junitxml.TestCase,
+		logger *log.Logger, testProjectConfig *testconfig.Project) {
+
+	suffix := pathutils.RandString(5)
+	imageName := "e2e-test-image-import-data-disk-" + suffix
+	cmd := "gce_vm_image_import"
+	args := []string{"-client_id=e2e", fmt.Sprintf("-project=%v", testProjectConfig.TestProjectID),
+		fmt.Sprintf("-image_name=%s", imageName), "-data_disk", fmt.Sprintf("-source_file=gs://%v-test-image/image-file-10g-vmdk", testProjectConfig.TestProjectID),
+		"-no_guest_environment", "-family=test-family", "-description=test-description",
```
zoran15: Should be easy to verify: family, description and labels, as those are image properties.

dntczdx:
Added some aprams validation as suggested

gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
```diff
@@ -18,12 +18,13 @@ package importtestsuites
 import (
 	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/compute-image-tools/gce_image_import_export_tests/compute"
```
zoran15: This should go to the external imports group below

dntczdx: I hate this - always moved by IDE automatically

dntczdx:
> Make all lines fit in github page without scrolling as it's tedious to review code that requires it.

I agree. I tested, github won't wrap the text at a fixed width; it wraps when browser is not wide enough. So it's hard to tell what's the best size

zoran15:
> > Make all lines fit in github page without scrolling as it's tedious to review code that requires it.
> 
> I agree. I tested, github won't wrap the text at a fixed width; it wraps when browser is not wide enough. So it's hard to tell what's the best size

We should use common monitors most of us use at 100% zoom in Chrome. With these settings, line overflows after 119 characters on my screen, so we can use 115 as cut off.

PR #832, opened 2019-06-10
osconfig_tests: attempt to deflake Windows patch tests

adjackura:
Move setting of WSUS server to sysprep startup script so we don't get hung up on the reboot.

### DIFF
--- a/osconfig_tests/test_suites/patch/test_setup.go
+++ b/osconfig_tests/test_suites/patch/test_setup.go
@@ -17,14 +17,15 @@ package patch
 import (
 	"time"
 
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/compute"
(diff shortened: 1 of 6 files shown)


osconfig_tests/test_suites/patch/test_setup.go
```diff
@@ -17,14 +17,15 @@ package patch
 import (
 	"time"
 
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/compute"
 	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/utils"
-	"google.golang.org/api/compute/v1"
+	api "google.golang.org/api/compute/v1"
```
ryanwe: consider the alias `compute` in case this ever references other apis.

adjackura: Yeah, I went ahead and fixed all the import naming

PR #836, opened 2019-06-17
[import wrapper] fix e2e test "ImageImportWithRichParams"

dntczdx:
Current logic of verifying "labels" is incorrect. Changed the logic.
Improved implementation of verifying import/export result as well.

### DIFF
--- a/gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
+++ b/gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
@@ -122,7 +122,7 @@ func runImageImportWithRichParamsTest(
 
 	family := "test-family"
 	description := "test-description"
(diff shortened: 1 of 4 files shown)


gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
```diff
@@ -194,5 +187,20 @@ func verifyImportedImageWithParams(ctx context.Context, testCase *junitxml.TestC
 	} else {
 		logger.Printf("Image '%v' cleaned up.", imageName)
 	}
+}
 
+func contains(arr []string, subarr []string) bool {
```
zoran15: This code is fine for this use case (array lookup where n<10), just make sure we don't use this for large amount of data as it's O(n^2) complexity

PR #838, opened 2019-06-18
[gce_export] Export raw: support exporting to local disk

dntczdx:
A new param "local_path" is added. Now it's a either-or between "local_path" and "gcs_path".
The reason to add this is because we plan to change the way of exporting raw. This PR is the 1st half of the change. The 2nd half would be: change export workflow to use "local_path" and then do "gsutil cp" to gcs. Between 1st half and 2nd half, we need to release a new debian-9-worker image to adopt the new gce_export.

The purpose is to avoid long TCP connection with GCS. Currently, when directly writing .tar.gz file to GCS, a super long TCP connection to write the whole file (by io.CopyN) has to be established. However, it's not as stable as before (tcp connection reset happens a lot), and it's anyways unstable while writing a large image.

Per my testing, the performance of new implementation is the same as the current one.
The cost is higher due to more disk consumed, but no more than $0.1 for a 200GB exporting.

### DIFF
--- a/cli_tools/gce_export/main.go
+++ b/cli_tools/gce_export/main.go
@@ -37,8 +37,8 @@ import (
 )
 
 var (
-	disk      = flag.String("disk", "", "disk to export, on linux this would be something like '/dev/sda', and on Windows '\\\\.\\PhysicalDrive0'")
-	localPath = flag.String("local_path", "", "local path to store the image file on disk, on linux this would be something like '/dev/sdb/buffer', and on Windows '\\\\.\\PhysicalDrive1\buffer'. ")
+	disk      = flag.String("disk", "", "disk to export, on linux this would be something like '/dev/sda', and on Windows '\\\\.\\PhysicalDrive1'")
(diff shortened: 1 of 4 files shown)


cli_tools/gce_export/main.go
```diff
@@ -94,15 +95,42 @@ func main() {
...
+	writer, targetPath := createWriter()
+	createGzipFile(file, size, writer, targetPath)
+}
+
+func createWriter() (io.WriteCloser, string) {
+	var w io.WriteCloser
+	var p string
+	if *gcsPath != "" {
+		bkt, obj, err := storageutils.SplitGCSPath(*gcsPath)
+		if err != nil {
```
zoran15: It's not a good practice to perform fatal on an app in a method outside main. Return error and let main decide if it can recover or not. The same applies for other instances of log.fatal in sub-methods.

dntczdx:
Moved log out from sub func

dntczdx:
> Part of README that describes how no local file is created is outdated now

You are right, I just modified accordingly.

PR #840, opened 2019-06-21
[daisy export] New export workflow for raw format

dntczdx:
New export workflow for raw format.
Old: export directly to GCS
New: export to local disk, and then copy to GCS.

Currently, long TCP connection has a high chance to be reset by GCS. It caused ~40% failure for a 10GB disk exporting. Let's avoid that.

It's quite similar to the existing "export ext" workflow now.

### DIFF
--- a/daisy_workflows/export/disk_export.wf.json
+++ b/daisy_workflows/export/disk_export.wf.json
@@ -35,16 +35,22 @@
     }
   },
   "Sources": {
-    "${NAME}_export_disk.sh": "./export_disk.sh"
+    "${NAME}_export_disk.sh": "./export_disk.sh",
+    "${NAME}_disk_resizing_mon.sh": "./disk_resizing_mon.sh"
   },
   "Steps": {
     "setup-disks": {
       "CreateDisks": [
         {
-          "Name": "disk-${NAME}",
+          "Name": "disk-${NAME}-os",
           "SourceImage": "${export_instance_disk_image}",
-          "SizeGb": "${export_instance_disk_size}",
           "Type": "${export_instance_disk_type}"
+        },
+        {
+          "Name": "disk-${NAME}-buffer-${ID}",
+          "SizeGb": "${export_instance_disk_size}",
+          "Type": "${export_instance_disk_type}",
+          "ExactName": true
         }
       ]
     },
@@ -52,12 +58,17 @@
       "CreateInstances": [
         {
           "Name": "inst-${NAME}",
-          "Disks": [{"Source": "disk-${NAME}"}, {"Source": "${source_disk}", "Mode": "READ_ONLY"}],
+          "Disks": [
+            {"Source": "disk-${NAME}-os"},
+            {"Source": "${source_disk}", "Mode": "READ_ONLY"},
+            {"Source": "disk-${NAME}-buffer-${ID}"}
+          ],
           "MachineType": "n1-highcpu-4",
           "Metadata": {
             "block-project-ssh-keys": "true",
             "gcs-path": "${OUTSPATH}/${NAME}.tar.gz",
-            "licenses": "${licenses}"
+            "licenses": "${licenses}",
+            "resizing-script-name": "${NAME}_disk_resizing_mon.sh"
           },
           "networkInterfaces": [
             {
--- a/daisy_workflows/export/disk_export.wf.json
+++ b/daisy_workflows/export/disk_export.wf.json
@@ -76,7 +76,7 @@
               "subnetwork": "${export_subnet}"
             }
           ],
-          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control"],
+          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control","https://www.googleapis.com/auth/compute"],
           "StartupScript": "${NAME}_export_disk.sh"
         }
       ]
--- a/daisy_workflows/export/disk_export.wf.json
+++ b/daisy_workflows/export/disk_export.wf.json
@@ -68,6 +68,7 @@
             "block-project-ssh-keys": "true",
             "gcs-path": "${OUTSPATH}/${NAME}.tar.gz",
             "licenses": "${licenses}",
+	    "buffer-disk": "disk-${NAME}-buffer-${ID}",
             "resizing-script-name": "${NAME}_disk_resizing_mon.sh"
           },
           "networkInterfaces": [
--- a/daisy_workflows/export/disk_export.wf.json
+++ b/daisy_workflows/export/disk_export.wf.json
@@ -76,7 +76,7 @@
               "subnetwork": "${export_subnet}"
             }
           ],
-          "Scopes": ["https://www.googleapis.com/auth/devstorage.read_write"],
+          "Scopes": ["https://www.googleapis.com/auth/devstorage.full_control"],
           "StartupScript": "${NAME}_export_disk.sh"
         }
       ]
--- a/daisy_workflows/export/export_disk.sh
+++ b/daisy_workflows/export/export_disk.sh
@@ -56,7 +56,7 @@ echo "GCEExport: Launching disk size monitor in background..."
 chmod +x ${DISK_RESIZING_MON_LOCAL_PATH}
 ${DISK_RESIZING_MON_LOCAL_PATH} ${MAX_BUFFER_DISK_SIZE_GB} &
 
-echo "GCEExport: Exporting image to local by gce_export tool..."
+echo "GCEExport: Exporting image to local disk by gce_export tool..."
 if [[ -n $LICENSES ]]; then
   gce_export -local_path "/gs/${IMAGE_OUTPUT_PATH}" -disk /dev/sdb -licenses "$LICENSES" -y
 else
@@ -73,8 +73,6 @@ if ! out=$(gsutil cp "/gs/${IMAGE_OUTPUT_PATH}" "${GCS_PATH}" 2>&1); then
   exit 1
 fi
 echo ${out}
-
-"$GCS_PATH"
 
 echo "ExportSuccess"
 sync
--- a/daisy_workflows/export/export_disk.sh
+++ b/daisy_workflows/export/export_disk.sh
@@ -58,12 +58,12 @@ ${DISK_RESIZING_MON_LOCAL_PATH} ${MAX_BUFFER_DISK_SIZE_GB} &
 
 echo "GCEExport: Exporting image to local by gce_export tool..."
 if [[ -n $LICENSES ]]; then
-  gce_export -local_path "$IMAGE_OUTPUT_PATH" -disk /dev/sdb -licenses "$LICENSES" -y
+  gce_export -local_path "/gs/${IMAGE_OUTPUT_PATH}" -disk /dev/sdb -licenses "$LICENSES" -y
 else
-  gce_export -local_path "$IMAGE_OUTPUT_PATH" -disk /dev/sdb -y
+  gce_export -local_path "/gs/${IMAGE_OUTPUT_PATH}" -disk /dev/sdb -y
 fi
 if [ $? -ne 0 ]; then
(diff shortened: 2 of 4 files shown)


daisy_workflows/export/disk_export.wf.json
```diff
@@ -35,29 +35,40 @@
     }
   },
   "Sources": {
-    "${NAME}_export_disk.sh": "./export_disk.sh"
+    "${NAME}_export_disk.sh": "./export_disk.sh",
+    "${NAME}_disk_resizing_mon.sh": "./disk_resizing_mon.sh"
```
zoran15: Where is this script coming from?

daisy_workflows/export/export_disk.sh
```diff
@@ -13,20 +13,68 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+BYTES_1GB=1073741824
 URL="http://metadata/computeMetadata/v1/instance/attributes"
 GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/gcs-path)
 LICENSES=$(curl -f -H Metadata-Flavor:Google ${URL}/licenses)
+DISK_RESIZING_MON=$(curl -f -H Metadata-Flavor:Google ${URL}/resizing-script-name)
 
-echo "GCEExport: Running export tool."
```
zoran15: Why remove this?

daisy_workflows/export/export_disk.sh
```diff
@@ -13,20 +13,68 @@
...
+  echo "ExportFailed: Failed to copy disk size monitor script. Error: ${out}"
+  exit 1
+fi
+echo ${out}
+
+echo "GCEExport: Launching disk size monitor in background..."
+chmod +x ${DISK_RESIZING_MON_LOCAL_PATH}
+${DISK_RESIZING_MON_LOCAL_PATH} ${MAX_BUFFER_DISK_SIZE_GB} &
+
+echo "GCEExport: Exporting image to local by gce_export tool..."
```
zoran15: Maybe: ..to local disk...?

daisy_workflows/export/export_disk.sh
```diff
@@ -13,20 +13,68 @@
...
 fi
 if [ $? -ne 0 ]; then
-  echo "ExportFailed: Failed to export disk source to ${GCS_PATH}."
+  echo "ExportFailed: Failed to export disk source to /gs/${IMAGE_OUTPUT_PATH}."
+  exit 1
+fi
+
+echo "GCEExport: Copying output image to target GCS path..."
+if ! out=$(gsutil cp "/gs/${IMAGE_OUTPUT_PATH}" "${GCS_PATH}" 2>&1); then
+  echo "ExportFailed: Failed to copy output image to ${GCS_PATH}, error: ${out}"
```
zoran15: Since connectivity from export instance to GCS is flaky with gcsfuse, should we worry about gsutil being flaky as well? If so, we should consider a limited number of retries on fail here (like 1 or 2 since the files are large)

daisy_workflows/export/export_disk.sh
```diff
@@ -13,20 +13,68 @@
...
+
+# Fetch disk size monitor script from GCS1
+DISK_RESIZING_MON_GCS_PATH=gs://${OUTS_PATH%/*}/sources/${DISK_RESIZING_MON}
+DISK_RESIZING_MON_LOCAL_PATH=/gs/${DISK_RESIZING_MON}
+echo "GCEExport: Copying disk size monitor script..."
+if ! out=$(gsutil cp "${DISK_RESIZING_MON_GCS_PATH}" "${DISK_RESIZING_MON_LOCAL_PATH}" 2>&1); then
+  echo "ExportFailed: Failed to copy disk size monitor script. Error: ${out}"
+  exit 1
+fi
+echo ${out}
```
zoran15: Where is this var coming from?

daisy_workflows/export/export_disk.sh
```diff
@@ -13,20 +13,68 @@
...
+fi
+
+echo "GCEExport: Copying output image to target GCS path..."
+if ! out=$(gsutil cp "/gs/${IMAGE_OUTPUT_PATH}" "${GCS_PATH}" 2>&1); then
+  echo "ExportFailed: Failed to copy output image to ${GCS_PATH}, error: ${out}"
   exit 1
 fi
+echo ${out}
+
+"$GCS_PATH"
```
zoran15: What does this do?

daisy_workflows/export/disk_export.wf.json
dntczdx: it's the existing one, named "disk_resizing_mon.sh" under "daisy_workflows_export"

daisy_workflows/export/export_disk.sh
dntczdx: line 49

daisy_workflows/export/export_disk.sh
dntczdx: done

daisy_workflows/export/export_disk.sh
dntczdx: It's replaced by multiple other messages because there are more detailed steps now. The most similar one is "GCEExport: Exporting image to local by gce_export tool..." at line 59

daisy_workflows/export/export_disk.sh
dntczdx: By testing, it's much more reliable. It's production quality, while gcsfuse owner claimed "beta-quality" clearly.
gcsfuse has retry mechanism naturally.

daisy_workflows/export/export_disk.sh
dntczdx: Ouch, this is useless by accident. Removed.

dntczdx:
Addressed 2 and answered 4 from Zoran's comment

daisy_workflows/export/export_disk.sh
zoran15: Having an intro message that export is starting is still useful. Exporting to local is a single step of the export process and not equivalent of the whole process

daisy_workflows/export/export_disk.sh
zoran15: "gcsfuse has retry mechanism naturally."

You mean gsutil?

dntczdx: Oh yes I meant gsutil :P

daisy_workflows/export/export_disk.sh
dntczdx: The thing is, it's not started at this point of time. The 1st message to see would be:
"GCEExport: Initializing buffer disk for gce_export output..."
I understand you are talking about a message for the whole exporting thing. That exists as "step disk-export is running..." in daisy log.

dntczdx:
> This is a huge step backwards in speed and efficiency and will have a huge impact on our image builds. Also if this is all your doing this is way overkill and we had a much cleaner solution before we built gce_export. The entire reason we made gce_export was speed and efficency. Instead of hacking this in the script can't we just update gce_export to handle reconnects? Also since this is a regresion why not put it on GCS to fix?

Let me try to describe everything about this change. It's more than just a bug fix. We can discuss more details offline if it's not clear enough.

1. May I know the data that you had when evaluating the speed/efficiency improvement? Cause my current testing doesn't show a speed backward overall. My current data is (only calculate the critical exporting step, test 3 times for each situation):
10G disk with 2G content (classic official image):
        | 10g disk, 2g content | 10g | 210g
old 1 | 0:01:40 | 0:04:40 | 1:29:14
old 2 | 0:01:30 | 0:05:30 | 1:28:58
old 3 | 0:01:40 | 0:04:40 | 1:29:05
new 1 | 0:01:50 | 0:04:10 | 1:43:20
new 2 | 0:01:50 | 0:04:30 | 1:41:20
new 3 | 0:01:40 | 0:04:10 | 1:41:01
So it's actually slightly a bit slower for super small image, and faster for larger ones.

2. This change is a preparation for another improvement as well - once we use gsutil, we can do parallel copying, which speed up a lot. Per my testing, it can improve >50% for >200GB image, and 10~30% for 2GB~10GB small images. The reason is that bottleneck is network transforming. Preparing data on local and copying in parallel speed it up.

3. This change makes exporting raw almost the same as exporting other formats. We can consolidate them later (both bash script and workflow).

4. About reconnect: Unless we handle the protocol ourselves, I don't find a way to reconnect. Current way of gce_export is using io.CopyN to transform data to GCS, which is not resumable per my understanding. Having GCS fix their stability is a good idea, but client side always need to handle unstable situation for a long connection.

Summarize: gsutil naturally has retry (reconnect) and parallel uploading, which provided stability and better speed/efficiency.

(parallel uploading hasn't been added in this PR, I will do it later).

adjackura:
> > This is a huge step backwards in speed and efficiency and will have a huge impact on our image builds. Also if this is all your doing this is way overkill and we had a much cleaner solution before we built gce_export. The entire reason we made gce_export was speed and efficency. Instead of hacking this in the script can't we just update gce_export to handle reconnects? Also since this is a regresion why not put it on GCS to fix?
> 
> Let me try to describe everything about this change. It's more than just a bug fix. We can discuss more details offline if it's not clear enough.
> 
> 1. May I know the data that you had when evaluating the speed/efficiency improvement? Cause my current testing doesn't show a speed backward overall. My current data is (only calculate the critical exporting step, test 3 times for each situation):
>    10G disk with 2G content (classic official image):
>            | 10g disk, 2g content | 10g | 210g
>    old 1 | 0:01:40 | 0:04:40 | 1:43:20
>    old 2 | 0:01:30 | 0:05:30 | 1:41:20
>    old 3 | 0:01:40 | 0:04:40 | 1:41:01
>    new 1 | 0:01:50 | 0:04:10 |TBD
>    new 2 | 0:01:50 | 0:04:30 |TBD
>    new 3 | 0:01:40 | 0:04:10 | TBD
>    So it's actually slightly a bit slower for super small image, and faster for larger ones.
> 2. This change is a preparation for another improvement as well - once we use gsutil, we can do parallel copying, which speed up a lot. Per my testing, it can improve >50% for >200GB image, and 10~30% for 2GB~10GB small images. The reason is that bottleneck is network transforming. Preparing data on local and copying in parallel speed it up.
> 3. This change makes exporting raw almost the same as exporting other formats. We can consolidate them later (both bash script and workflow).
> 4. About reconnect: Unless we handle the protocol ourselves, I don't find a way to reconnect. Current way of gce_export is using io.CopyN to transform data to GCS, which is not resumable per my understanding. Having GCS fix their stability is a good idea, but client side always need to handle unstable situation for a long connection.
> 
> Summarize: gsutil naturally has retry (reconnect) and parallel uploading, which provided stability and better speed/efficiency.
> 
> (parallel uploading hasn't been added in this PR, I will do it later).

The issues were seeing is not because of a lack of retries, I was unable to repo any issues and the existing code has just as much retry logic as gsutil if not more. Also I don't want to put any unnecessary dependencies on gsutil when the tool can perform better anyway. If we really want to move to a more expensive local cache solution we can just build that into the export tool. Ill do some testing today to see if what the time differences would be but we still have the much bigger issue of space utilization.

dntczdx:
> The issues were seeing is not because of a lack of retries, I was unable to repo any issues and the existing code has just as much retry logic as gsutil if not more. Also I don't want to put any unnecessary dependencies on gsutil when the tool can perform better anyway. If we really want to move to a more expensive local cache solution we can just build that into the export tool. Ill do some testing today to see if what the time differences would be but we still have the much bigger issue of space utilization.

Cool! I am also doing more testing. Let's get a more complete picture on this topic.
I tested 12 times yesterday night and this morning. While exporting 210GB image, 50% of existing code failed by tcp reset, while all new code by "gsutil cp" succeeded. So I believe their retry are different.

As I described above, gsutil provided another benefit that it can upload file in parallel, which is 4 times faster. I will provide testing data later today.

dntczdx:
Closing it due that we finally chose another implementation

PR #841, opened 2019-06-21
[daisy] a new param "print_perf" to show perf statistics on console

dntczdx:
Example as below:

Perf Profile:
- setup-disks: 0:0:12
- image-export-export-disk.setup-disks: 0:0:33
- image-export-export-disk.run-image-export-export-disk: 0:0:18
- image-export-export-disk.wait-for-inst-image-export-export-disk: 0:1:40
- image-export-export-disk.copy-image-object: 0:0:0
- image-export-export-disk.delete-inst: 0:1:4
- image-export-export-disk: 0:3:35
- workflow cleanup: 0:0:2
Total time: 0:3:50

It helped perf testing and debugging.


cli_tools/daisy/main.go
```diff
@@ -38,6 +39,7 @@ var (
 	zone               = flag.String("zone", "", "zone to run in, overrides what is set in workflow")
 	variables          = flag.String("variables", "", "comma separated list of variables, in the form 'key=value'")
 	print              = flag.Bool("print", false, "print out the parsed workflow for debugging")
+	printPerf          = flag.Bool("print_perf", false, "print out the perf profile for debugging")
```
zoran15: Use "performance" in descriptions/help texts

Also, is "for debugging" actually correct? We can use this for workflow optimization etc. not only debugging. I would remove that part.

daisy/step.go
```diff
@@ -250,7 +250,14 @@ func (s *Step) populate(ctx context.Context) dErr {
 	return err
 }
 
+func (s *Step) recordStepTime(startTime time.Time) {
+	endTime := time.Now()
```
zoran15: Can be inlined

cli_tools/daisy/main.go
```diff
@@ -282,3 +285,33 @@ func main() {
 		}
 	}
 }
+
+func printPerfProfile(workflow *daisy.Workflow) {
```
zoran15: Consider moving this to a separate source file/class to avoid clutter

cli_tools/daisy/main.go
```diff
@@ -282,3 +285,33 @@ func main() {
...
+		}
+		if wfEndTime.Before(r.EndTime) {
+			wfEndTime = r.EndTime
+		}
+		fmt.Printf("- %v: %v\n", r.Name, formatDuration(r.EndTime.Sub(r.StartTime)))
+	}
+	fmt.Printf("Total time: %v\n\n", formatDuration(wfEndTime.Sub(wfStartTime)))
+}
+
+func formatDuration(d time.Duration) string {
```
zoran15: move this as well to a separate file

cli_tools/daisy/main.go
```diff
@@ -282,3 +285,33 @@ func main() {
...
+	s := int(d.Seconds())
+	return fmt.Sprintf("%v:%v:%v", s/3600, s/60%60, s%60)
```
zoran15: Should add units, e.g. %vh:%vm:%vs

cli_tools/daisy/main.go
```diff
@@ -253,6 +255,7 @@ func main() {
 		wg.Add(1)
 		go func(w *daisy.Workflow) {
 			defer wg.Done()
+			defer printPerfProfile(w)
```
zoran15: if !*printPerf {
defer...
}

then extract printPerfProfile to it's own file

cli_tools/daisy/main.go
```diff
@@ -282,3 +285,33 @@ func main() {
...
+	timeRecords := workflow.GetStepTimeRecords()
+	if len(timeRecords) == 0 {
+		return
+	}
+
+	wfStartTime := time.Now()
+	wfEndTime, _ := time.Parse(time.RFC3339, "0001-01-01T00:00:00+00:00")
+	fmt.Println("\nPerf Profile:")
+	for _, r := range timeRecords {
+		if wfStartTime.After(r.StartTime) {
```
zoran15: How can start time of a record be before workflow started?

daisy/workflow.go
```diff
@@ -247,7 +257,23 @@ func (w *Workflow) RunWithModifiers(
 	return nil
 }
 
+func (w *Workflow) recordStepTime(stepName string, startTime time.Time, endTime time.Time) {
+	if w.parent == nil {
+		w.recordTimeMx.Lock()
+		w.stepTimeRecords = append(w.stepTimeRecords, TimeRecord{stepName, startTime, endTime})
+		w.recordTimeMx.Unlock()
+	} else {
+		w.parent.recordStepTime(fmt.Sprintf("%s.%s", w.Name, stepName), startTime, endTime)
```
zoran15: Can we  have nested workflows with more than one level of nesting? E.g. w.parent.parent != nil. If so, should we recurse  up the graph until we find the root instead of just doing one step up?

cli_tools/daisy/main.go
```diff
@@ -282,3 +285,33 @@ func main() {
...
+		return
+	}
+
+	timeRecords := workflow.GetStepTimeRecords()
+	if len(timeRecords) == 0 {
+		return
+	}
+
+	wfStartTime := time.Now()
+	wfEndTime, _ := time.Parse(time.RFC3339, "0001-01-01T00:00:00+00:00")
```
adjackura: Whats the point of this? Why not just 'var wfEndTime time.Time' ?

cli_tools/daisy/main.go
dntczdx: Sounds good

daisy/step.go
dntczdx: I want to keep it this way to make it easier to read. Otherwise, a pair of (startTime, time.Now()) doesn't have a obvious meaning.

daisy/workflow.go
dntczdx: This is a recursive call: w.parent.recordStepTime

cli_tools/daisy/main.go
dntczdx: This trying to find the minimum step start time, and use it as workflow start time.

cli_tools/daisy/main.go
dntczdx: Since I didn't know that is the default value of time... changed to time.Time{} accordingly

cli_tools/daisy/main.go
dntczdx: 0:10:32 is a nice format as well. I can directly copy it to sheet or excel for calculation. 0h:10m:32s is not accepted by those sheets program.

zoran15: The problem here is it's not possible to know what the values mean without looking at the code. At least add units before/after the value so it's obvious to anyone reading the perf report

dntczdx: Sure. I though people have aligned understanding on that format. Changed.

dntczdx:
Added time format hint

cli_tools/daisy/print_utils.go
```diff
@@ -0,0 +1,34 @@
+package main
```
zoran15: Move it to a shared location, maybe cli_tools/common/utils/performance or similar, in case we want to print performance of Daisy based wrappers (import/export/ovf). 

Also, add copyright comment

dntczdx: sounds good

PR #844, opened 2019-06-28
use parallel uploading for exporting vmdk

dntczdx:
Use parallel uploading for exporting vmdk.

### DIFF
--- a/daisy_workflows/export/export_disk_ext.sh
+++ b/daisy_workflows/export/export_disk_ext.sh
@@ -65,7 +65,7 @@ fi
 echo ${out}
 
 echo "GCEExport: Copying output image to target GCS path..."
-if ! out=$(gsutil cp "/gs/${IMAGE_OUTPUT_PATH}" "${GS_PATH}" 2>&1); then
+if ! out=$(gsutil -o GSUtil:parallel_composite_upload_threshold=150M cp "/gs/${IMAGE_OUTPUT_PATH}" "${GS_PATH}" 2>&1); then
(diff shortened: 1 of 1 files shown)


daisy_workflows/export/export_disk_ext.sh
```diff
@@ -65,7 +65,7 @@ fi
 echo ${out}
 
 echo "GCEExport: Copying output image to target GCS path..."
-if ! out=$(gsutil cp "/gs/${IMAGE_OUTPUT_PATH}" "${GS_PATH}" 2>&1); then
+if ! out=$(gsutil -o GSUtil:parallel_composite_upload_threshold=150M cp "/gs/${IMAGE_OUTPUT_PATH}" "${GS_PATH}" 2>&1); then
```
adjackura: Note setting this to a low 150M mostly meaningless for our use case since a single composition can only be done with 32 pieces gsutil will simply divide the size by 32 and only use 150 for small files (less than 32*150M).

You may also want to play with parallel_thread_count and parallel_process_count as we are using only a 4 core machine and the defaults may cause slowdown.

dntczdx: Thanks for all the hints! I tested those scenarios.
1. Even for small disk, it is useful. For example, 10G (the smallest disk) with 2G content, which is, a initial debian OS disk. It reduced 20s.
2. I tested different combinations of parallel_thread_count, parallel_process_count. Setting it to 4 has the exactly same test result as default settings. So I would keep it as default in order that gsutil may tune it better in the future and we can automatically adopt it.

PR #845, opened 2019-06-28
 gce_export: add component uploading with local buffer

adjackura:
Allows for multipart component upload while using a local buffer and includes up to 3 retries per part.


cli_tools/gce_export/main.go
```diff
@@ -17,47 +17,69 @@ package main
...
 func (p *progress) Write(b []byte) (int, error) {
+	p.lock.Lock()
 	p.total += int64(len(b))
+	p.lock.Unlock()
 	return len(b), nil
 }
 
+func randString(n int) string {
+	gen := rand.New(rand.NewSource(time.Now().UnixNano()))
+	letters := "bdghjlmnpqrstvwxyz0123456789"
```
dntczdx: Use constant to avoid dup memory allocation?

cli_tools/gce_export/main.go
```diff
@@ -69,99 +91,247 @@ func splitLicenses(input string) []string {
 	return ls
 }
 
-func main() {
-	flag.Parse()
+type buffer struct {
```
dntczdx: It implements "writer" and behaves as a writer in this program, so I suggest to call it bufferedWriter

cli_tools/gce_export/main.go
```diff
@@ -69,99 +91,247 @@ func splitLicenses(input string) []string {
...
+				file, err := os.Open(in)
+				if err != nil {
+					return err
+				}
+				defer file.Close()
+				defer os.Remove(in)
+
+				tmpObj := path.Join(b.obj, strings.TrimPrefix(in, b.prefix))
+				b.addObj(tmpObj)
+				dst := client.Bucket(b.bkt).Object(tmpObj).NewWriter(b.ctx)
```
dntczdx: per my understanding, if an error is thrown on the half way, there will be an incomplete file left on GCS. I am not sure whether GCS allow us to ignore that file and simply redo uploading. Even if it is allowed, it would be better to resume, while we are able to tracked the progress. If that is hard to implement, we can leave it as a improvement in the future.

cli_tools/gce_export/main.go
```diff
@@ -69,99 +91,247 @@ func splitLicenses(input string) []string {
...
+					}
+				}
+
+				return dst.Close()
+			}()
+			if err != nil {
+				if i > 3 {
+					log.Fatal(err)
+				}
+				time.Sleep(time.Duration(1*i) * time.Second)
```
dntczdx: My experience: 3 times (<8s in total) of retry is not enough, especially when we need to wait for the GCS server to recover from a bad status. We can do 2 extra things:
1. retry more, for example 16 times. The total length of retrying period can be 5min.
2. randomize the waiting seconds. For example, int.rand(duration). The reason is that, very likely all uploading threads (from different clients) were cut at the same time. It would be good to avoid all re-connections happened in the same timing.

Due that reset issue has been auto mitigated in recent 4 days, you can leave it to me as a improvement as well.

adjackura: I bumped it to 16, the 3 was fairly arbitrary and a total guess on my part as I have never been able to repo a connection issue. 

As far as randomizing, we could do that but I'm not to worried about a few connection attempts as we default to num of CPUs in order to not cause any contention.

cli_tools/gce_export/main.go
adjackura: The file isn't written before Close is called, we attempt to reupload this bit up to retry times, if the entire process is canceled then yes you have temporary files left around. We could maybe add logic to try to find old temporary bits and check if they match so we don't re upload but that won't really help our process as this is all done in a workflow that we can't really replicate on additional runs.

cli_tools/gce_export/main.go
dntczdx: "I'm not too worried about a few connection attempts as we default to num of CPUs"
That is true for one instance. However, my assumption is that all export instances connected to that GCS server can be reset at the same time (for example, gcs service got restarted), which can be a lot of connections.

cli_tools/gce_export/main.go
dntczdx: If "The file isn't written before Close is called" is correct, then I totally agree all the statements. I am still not 100% sure about that because I failed to find a official description about this. I will do more test after your PR got release, and improve it if needed.

PR #846, opened 2019-07-02
Support for multiple failure matches in Daisy. 

zoran15:
Support for multiple failure matches in Daisy. Keeping the original single value field for backward compatibility.


daisy/step_wait_for_instances_signal.go
```diff
@@ -86,6 +87,9 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput,
 	if so.FailureMatch != "" {
 		msg += fmt.Sprintf(", FailureMatch: %q", so.FailureMatch)
 	}
+	if len(so.FailureMatches) > 0 {
```
dntczdx: I prefer to let FailureMatch and FailureMatches be mutually exclusive

zoran15: Could be either, I don't see a strong reason to go one way or the other. This was simpler to implement, simpler validation and easier on the user in case they mess up and specify both. I doubt anyone will use both but if they do, it will still work.

daisy/step_wait_for_instances_signal.go
```diff
@@ -33,13 +33,14 @@ type WaitForInstancesSignal []*InstanceSignal
...
+// cause the step to fail.
 type SerialOutput struct {
-	Port         int64  `json:",omitempty"`
-	SuccessMatch string `json:",omitempty"`
-	FailureMatch string `json:",omitempty"`
-	StatusMatch  string `json:",omitempty"`
+	Port           int64    `json:",omitempty"`
+	SuccessMatch   string   `json:",omitempty"`
+	FailureMatch   string   `json:",omitempty"`
+	FailureMatches []string `json:",omitempty"`
```
adjackura: If you write a custom unmarshaler you can have the existing FailureMatch take []string, see Var as an example: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/workflow.go#L77

daisy/step_wait_for_instances_signal.go
dntczdx: To me it's confusion to have both accepted. But you can decide based on your best judgement.

daisy/step_wait_for_instances_signal.go
zoran15: This is now deprecated after the refactoring to support both arrays and single strings in the old (FailureMatch) field.

daisy/step_wait_for_instances_signal.go
zoran15: This is now deprecated after the refactoring to support both arrays and single strings in the old (FailureMatch) field.

daisy/step_wait_for_instances_signal.go
```diff
@@ -83,8 +104,8 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput,
 	if so.SuccessMatch != "" {
 		msg += fmt.Sprintf(", SuccessMatch: %q", so.SuccessMatch)
 	}
-	if so.FailureMatch != "" {
-		msg += fmt.Sprintf(", FailureMatch: %q", so.FailureMatch)
+	if len(so.FailureMatch) > 0 {
+		msg += fmt.Sprintf(", FailureMatch: %v", so.FailureMatch)
```
adjackura: Keep this to %q, it'll quote each element in the list

daisy/step_wait_for_instances_signal.go
```diff
@@ -29,17 +30,37 @@ const (
...
+	return nil
+}
+
 // SerialOutput describes text signal strings that will be written to the serial
 // port.
 // A StatusMatch will print out the matching line from the StatusMatch onward.
 // This step will not complete until a line in the serial output matches
-// SuccessMatch or FailureMatch. A match with FailureMatch will cause the step
-// to fail.
+// SuccessMatch, FailureMatch or FailureMatches. A match with FailureMatch or FailureMatches will
```
adjackura: Update the comment.

daisy/step_wait_for_instances_signal.go
zoran15: Done

daisy/step_wait_for_instances_signal.go
zoran15: Done. In other places as well.

PR #848, opened 2019-07-02
Daisy: removed the check for network being in the same project

zoran15:
Daisy: removed the check for network being in the same project to support shared VPC use-case.


dntczdx:
If it's not a hard failure reason then I am fine to remove it

PR #853, opened 2019-07-10
osconfig_tests: fix OSPackage tests

adjackura:
Cleanup client instantiation
Fix issues with the previous cleanup PR

### DIFF
--- a/osconfig_tests/gcp_clients/gcp_clients.go
+++ b/osconfig_tests/gcp_clients/gcp_clients.go
@@ -20,8 +20,8 @@ import (
 	"sync"
 
 	"cloud.google.com/go/storage"
-	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/config"
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
+	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/config"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 	osconfig "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/cloud.google.com/go/osconfig/apiv1alpha1"
 	"google.golang.org/api/option"
@@ -47,14 +47,14 @@ func populateClients(ctx context.Context) error {
 
 func createComputeClient(ctx context.Context) error {
 	var err error
-	computeClient, err = compute.NewClient(ctx,  option.WithCredentialsFile(config.OauthPath()))
+	computeClient, err = compute.NewClient(ctx, option.WithCredentialsFile(config.OauthPath()))
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
-func createStorageClient(ctx context.Context) error{
+func createStorageClient(ctx context.Context) error {
 	logger.Debugf("creating storage client\n")
 	var err error
 	storageClient, err = storage.NewClient(ctx, option.WithCredentialsFile(config.OauthPath()))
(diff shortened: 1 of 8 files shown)


osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -28,37 +29,57 @@ import (
...
 }
 
-func createStorageClient(ctx context.Context) {
+func createComputeClient(ctx context.Context) error {
+	var err error
+	computeClient, err = compute.NewClient(ctx, option.WithCredentialsFile(config.OauthPath()))
+	if err != nil {
+		return err
+	}
+	return nil
```
ryanwe: You could remove the previous 4 lines and just `return err` after creating the client. Here and a few places below.

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -28,37 +29,57 @@ import (
...
-// PopulateClients creates singleton clients for various services
-func PopulateClients(ctx context.Context) {
+func populateClients(ctx context.Context) error {
+	var err error
 	populateClientOnce.Do(func() {
-		createOsConfigClient(ctx)
-		createStorageClient(ctx)
+		err = createComputeClient(ctx)
+		err = createOsConfigClient(ctx)
+		err = createStorageClient(ctx)
```
ryanwe: since these are executed sequentially, won't each call overwrite `err` from the previous invocation?

osconfig_tests/gcp_clients/gcp_clients.go
```diff
@@ -28,37 +29,57 @@ import (
...
 	})
+	return err
```
ryanwe: I'm not sure how the lifecycle of `err` is managed by go. If `populateClientOnce` has already executed and there was an error, won't err be nil the next time this is called and won't that cause problems for tests that check the err?

osconfig_tests/test_suites/inventory/inventory.go
```diff
@@ -78,10 +77,10 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
 
 func runGatherInventoryTest(ctx context.Context, testSetup *inventoryTestSetup, testCase *junitxml.TestCase, logwg *sync.WaitGroup, testProjectConfig *testconfig.Project) ([]*apiBeta.GuestAttributesEntry, bool) {
 	testCase.Logf("Creating compute client")
-	client, err := daisyCompute.NewClient(ctx)
+
+	computeClient, err := gcpclients.GetComputeClient(ctx)
 	if err != nil {
-		testCase.WriteFailure("Error creating client: %v", err)
-		return nil, false
+		testCase.WriteFailure("Error getting storage client: %v", err)
```
ryanwe: This is for the "compute" client. And shouldn't this early exit on failure?

osconfig_tests/test_suites/package_management/package_management.go
```diff
@@ -131,17 +130,16 @@ func runPackageRemovalTest(ctx context.Context, testCase *junitxml.TestCase, tes
 
 	defer cleanupAssignment(ctx, testCase, assign, testProjectConfig)
 
-	client, err := daisyCompute.NewClient(ctx)
+	computeClient, err := gcpclients.GetComputeClient(ctx)
 	if err != nil {
-		testCase.WriteFailure("Error creating client: %v", err)
-		return
+		testCase.WriteFailure("Error getting storage client: %v", err)
```
ryanwe: here too... And there are more places below.

osconfig_tests/gcp_clients/gcp_clients.go
adjackura: I really didn't like this setup, made this much simpler

PR #854, opened 2019-07-10
[import wrapper] source_image param should accept long format

dntczdx:
Just as what I did for export wrapper in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/833:
we need to be compatible with 2 kind of inputs:
"global/images/anImage"
"projects/myproject/global/images/anImage"
So removed the fixed prefix "global/images" from code. Currently, the fixed prefix is conflict with gcloud


dntczdx:
Withdraw this change due that pantheon has a dependency on it

PR #855, opened 2019-07-11
fix param network/subnet logic for import/export wrapper

dntczdx:
Currently there is a logic lost in import/export wrapper: when subnet is specified and network is not, we should set network to empty string to avoid falling back to default network.

This logic exists in gcloud GA. However, since gcloud is going to call wrapper, this logic should be hosted in this repo. There are 2 reasons to do it:
1. Rich logic should be handled by wrapper layer as much as possible.
2. For a cli tool, there is no way to distinguish network=none v.s. network='', due that param is not nullable.

Added e2e test correspondingly.

This fix is required to fix the behavior of gcloud calling wrapper.

### DIFF
--- a/cli_tools/gce_vm_image_import/main.go
+++ b/cli_tools/gce_vm_image_import/main.go
@@ -199,11 +199,15 @@ func buildDaisyVars(translateWorkflowPath string) map[string]string {
 	}
 	varMap["family"] = *family
 	varMap["description"] = *description
-	if *network != "" {
-		varMap["import_network"] = fmt.Sprintf("global/networks/%v", *network)
-	}
 	if *subnet != "" {
 		varMap["import_subnet"] = fmt.Sprintf("regions/%v/subnetworks/%v", *region, *subnet)
+		// When subnet is set, we need to grant a value to network to avoid fallback to default
+		if *network == "" {
+			varMap["import_network"] = ""
+		}
+	}
+	if *network != "" {
+		varMap["import_network"] = fmt.Sprintf("global/networks/%v", *network)
 	}
 	return varMap
 }
--- a/gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
+++ b/gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
@@ -128,7 +128,7 @@ func runImageImportWithRichParamsTest(
 	labels := []string{"key1=value1", "key2=value2"}
 
 	suffix := pathutils.RandString(5)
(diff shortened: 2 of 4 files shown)


gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
```diff
@@ -125,12 +128,14 @@ func runImageImportWithRichParamsTest(
...
 
 	suffix := pathutils.RandString(5)
-	imageName := "e2e-test-image-import-data-disk-" + suffix
+	imageName := "e2e-test-image-import-rich-param-" + suffix
 	cmd := "gce_vm_image_import"
 	args := []string{"-client_id=e2e", fmt.Sprintf("-project=%v", testProjectConfig.TestProjectID),
 		fmt.Sprintf("-image_name=%s", imageName), "-data_disk", fmt.Sprintf("-source_file=gs://%v-test-image/image-file-10g-vmdk", testProjectConfig.TestProjectID),
 		"-no_guest_environment", fmt.Sprintf("-family=%v", family), fmt.Sprintf("-description=%v", description),
-		"-network=default", "-subnet=default", fmt.Sprintf("-zone=%v", testProjectConfig.TestZone),
+		fmt.Sprintf("-network=%v-vpc-1", testProjectConfig.TestProjectID),
```
zoran15: Do we have these networks in the test project?

cli_tools/gce_vm_image_import/main.go
```diff
@@ -199,11 +199,15 @@ func buildDaisyVars(translateWorkflowPath string) map[string]string {
 	}
 	varMap["family"] = *family
 	varMap["description"] = *description
-	if *network != "" {
-		varMap["import_network"] = fmt.Sprintf("global/networks/%v", *network)
-	}
 	if *subnet != "" {
 		varMap["import_subnet"] = fmt.Sprintf("regions/%v/subnetworks/%v", *region, *subnet)
+		// When subnet is set, we need to grant a value to network to avoid fallback to default
```
zoran15: looks like network/subnet processing is the same for import and export wrappers. Consider refactoring to avoid code duplication.

dntczdx: There will be a major refactoring to CLI tools as preparation of API project (so that API can call it as a lib). Will consider that by then.

gce_image_import_export_tests/test_suites/import/image_import_test_suite.go
dntczdx: yes

PR #858, opened 2019-07-15
Add some useful packages to Debian 10 and move repo configs to common names.

zmarano:


daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -95,17 +95,18 @@ def main():
   image_size = '10G'
   disk_name = 'disk.raw'
   config_space = os.getcwd() + work_dir + '/config_space/'
+  apt_sources_base = 'files/etc/apt/sources.list.d/'
```
hopkiw: is it likely to change? this variable makes it a bit harder to read what exactly is happening where it's used.

zmarano: It isn't likely to ever change no... I kept running into problems with lint and character limits because linter is broken for string concatenation across lines (it does not know which rule to folllow). So to get around that I am simply defining the common path for the cases where its used here.

PR #860, opened 2019-07-16
[export cli tool]Exporter refactoring

dntczdx:
This is a preparation for import/export API. It's a refactoring without any feature changes.
1. It moves logic of export (and unit tests) to "exporter.go" from "main.go". The purpose is to make exporter a library so API can call.
2. It renamed a couple of package name under "cli_tools". The reason is that google3/third_party is incompatible with package names different from its folder name.


cli_tools/common/utils/storage/storage_client.go
```diff
@@ -219,12 +219,12 @@ func (hc *HTTPClient) Get(url string) (resp *http.Response, err error) {
 	return hc.httpClient.Get(url)
 }
 
-// StorageObjectDeleter is responsible for deleting object
-type StorageObjectDeleter struct {
-	sc *StorageClient
+// ObjectDeleter is responsible for deleting object
```
zoran15: By removing the Storage prefix, except by looking at the package name, it's not clear what type of object this is deleting. Update docs to reflect this deals with storage objects. Make sure this is addressed in other files (if any).

PR #864, opened 2019-07-24
Export raw with new gce_export with parallel uploading

dntczdx:
1. Export raw with new gce_export with parallel uploading
Perf improvement:

2. Download gce_export form our official GCS path. It will override pre-installed one in debian-9-worker. The purpose is to make the rollout process of gce_export being align with other cli tools. Otherwise, after a compute-image-tools release, we need to wait for one more debian-9-worker release.

After this code is released, we can remove pre-install step of gce_export from debian-9-worker build process.

### DIFF
--- a/daisy_workflows/export/export_disk.sh
+++ b/daisy_workflows/export/export_disk.sh
@@ -17,6 +17,14 @@ URL="http://metadata/computeMetadata/v1/instance/attributes"
 GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/gcs-path)
 LICENSES=$(curl -f -H Metadata-Flavor:Google ${URL}/licenses)
 
+echo "GCEExport: Downloading gce_export."
+curl --output /usr/bin/gce_export https://storage.googleapis.com/compute-image-tools/release/linux/gce_export
+if [[ $? -ne 0 ]]; then
+  echo "ExportFailed: Could not download gce_export."
+  exit 1
+fi
+chmod +x /usr/bin/gce_export
+
 mkdir ~/upload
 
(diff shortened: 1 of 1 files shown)


daisy_workflows/export/export_disk.sh
```diff
@@ -17,13 +17,23 @@ URL="http://metadata/computeMetadata/v1/instance/attributes"
 GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/gcs-path)
 LICENSES=$(curl -f -H Metadata-Flavor:Google ${URL}/licenses)
 
+echo "GCEExport: Downloading gce_export."
+curl --output /usr/bin/gce_export https://storage.googleapis.com/compute-image-tools/release/linux/gce_export
```
zmarano: We pre-load this on the worker image and control the release that way. I am not comfortable doing this as we lose control over the dependencies. What is the case where we need to re-download gce_export every time?

dntczdx: Hi Zach, the plan is to change the way from pre-load to download. I plan to remove it from debian-9-worker building wf later.

The purpose is to make the rollout process of gce_export being align with other cli tools. Otherwise, after a compute-image-tools release, we need to wait for one more debian-9-worker release. Simplifying the rollout/rollback process helps us ti be align with reliability rules better. Please let me know whether we missed some other advantages of releasing gce_export along with debian-9-worker.

dntczdx: Offline discussed, we don't want to pull binary online. So keep current way for now.

dntczdx:
Offline discussed, we don't want to pull binary online. So keep current way for now.

PR #865, opened 2019-07-24
Refactor import wrapper

dntczdx:
Moved all actual logic into importer.go. It's useful for both logging project and API project.
Moved tests accordingly.
It's similar to the previous refactoring for exporter.

No functionality change at all.


dntczdx:
> How is this useful for the logging project?
> 
> We are in reliability reset with import/export releases currently stopped. We shouldn't be creating a huge backlog of PRs that will get released all at once once we start releasing unless they are related to reliability.

We need to log many info before any actual import step started (including parameters, etc), which should happen even earlier than param check. Thus, we need to move all logic out from main func.

In the reliability reset meeting, I think people talked about this - keep checking in code as usual, just don't release?

dntczdx:
And we didn't release it frequently, in the past it was usually 2/3 month ~ 1 month. I think accumulation is common...

zoran15:
> > How is this useful for the logging project?
> > We are in reliability reset with import/export releases currently stopped. We shouldn't be creating a huge backlog of PRs that will get released all at once once we start releasing unless they are related to reliability.
> 
> We need to log many info before any actual import step started (including parameters, etc), which should happen even earlier than param check. Thus, we need to move all logic out from main func.

I still don't understand why we can't add logging in the main function. What logging are we not able to add without this refactoring?

> In the reliability reset meeting, I think people talked about this - keep checking in code as usual, just don't release?

We are still able to commit code but we should take into account we currently can't release. Until we get a clarification from SREs, we don't know how long this will be the case. 

Even though we can still commit code, I don't think we should be doing major refactorings that are not essential, as that's increasing risks once we actually start releasing. 

We also have enough issues to work on that are reliability related in our backlog, that refactoring to support Service API is not a priority, especially as the service might get removed from Q3 OKRs due to reliability reset.

zoran15:
> And we didn't release it frequently, in the past it was usually 2/3 month ~ 1 month. I think accumulation is common...

We still don't have answers from SREs if we need to comply with the 4 reliability rules. Depending on that, we could be frozen for far longer than a couple of weeks. 

I'm not saying we should stop all development, but refactoring to support service pretty low on the priority list right now.

dntczdx:
> I still don't understand why we can't add logging in the main function. What logging are we not able to add without this refactoring?

We can add it in the main function, but the code won't look good. To get a nice code quality, we will need to make some new helper functions, then we will find it's better to move it to a separate file (and finally the code will look like this PR). And that is just what I did in this PR - move it to a separate file. 
In other words, I split the PR for logging project to 2 PRs, and this is the 1st one of them. It is the safer one which doesn't modify any behaviors.
You can ref to my draft PR for logging project to see how it works, which is based on this refactoring: https://github.com/dntczdx/compute-image-tools/pull/4

> We are still able to commit code but we should take into account we currently can't release. Until we get a clarification from SREs, we don't know how long this will be the case.
> 
> Even though we can still commit code, I don't think we should be doing major refactorings that are not essential, as that's increasing risks once we actually start releasing.
> 
> We also have enough issues to work on that are reliability related in our backlog, that refactoring to support Service API is not a priority, especially as the service might get removed from Q3 OKRs due to reliability reset.

This PR unblocks both projects as said above. Even we don't do service at all, this code change will be done sooner or later for logging.

cli_tools/gce_vm_image_export/exporter/exporter.go
```diff
@@ -133,10 +133,9 @@ func runExportWorkflow(ctx context.Context, exportWorkflowPath string, varMap ma
 }
 
 // Run runs export workflow.
-func Run(clientID string, destinationURI string, sourceImage string, format string,
-	project string, network string, subnet string, zone string, timeout string,
-	scratchBucketGcsPath string, oauth string, ce string, gcsLogsDisabled bool,
-	cloudLogsDisabled bool, stdoutLogsDisabled bool, labels string, currentExecutablePath string) error {
+func Run(clientID, destinationURI, sourceImage, format, project, network, subnet, zone, timeout,
```
zoran15: I find the removal of types except for the last variable in a sequence less readable as I have to read the list until the last variable in the sequence. It can also be potentially more risky when adding new variables as type is implied by the last variable.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -0,0 +1,266 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
zoran15: 2019

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/param"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/storage"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/validation"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// Make file paths mutable
```
zoran15: Why are they mutable now?

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -0,0 +1,266 @@
...
+var (
+	WorkflowDir                = "daisy_workflows/image_import/"
+	ImportWorkflow             = "import_image.wf.json"
+	ImportFromImageWorkflow    = "import_from_image.wf.json"
+	ImportAndTranslateWorkflow = "import_and_translate.wf.json"
+)
+
+// Parameter key shared with external packages
```
zoran15: What external packages are referred to here?

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -0,0 +1,266 @@
...
+	if sourceImage != "" {
+		return path.ToWorkingDir(WorkflowDir+ImportFromImageWorkflow, currentExecutablePath), getTranslateWorkflowPath(customTranWorkflow, osID)
+	}
+	if dataDisk {
+		return path.ToWorkingDir(WorkflowDir+ImportWorkflow, currentExecutablePath), ""
+	}
+	return path.ToWorkingDir(WorkflowDir+ImportAndTranslateWorkflow, currentExecutablePath), getTranslateWorkflowPath(customTranWorkflow, osID)
+}
+
+func getTranslateWorkflowPath(customTranWorkflow, osID string) string {
```
zoran15: customTransWorkflow -> customTranslateWorkflow

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -0,0 +1,266 @@
...
+			}}
+		rl.LabelResources(w)
+		daisyutils.UpdateAllInstanceNoExternalIP(w, noExternalIP)
+	}
+
+	return workflow.RunWithModifiers(ctx, nil, workflowModifier)
+}
+
+// Run runs import workflow.
+func Run(clientID, imageName string, dataDisk bool, osID, customTranWorkflow, sourceFile,
```
zoran15: Again, not using type for each flag makes this very unreadable. The only scenario I could see this being a good practice is if all flags have the same type.

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -0,0 +1,334 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
zoran15: 2019

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+
```
zoran15: extra blank line

cli_tools/gce_vm_image_export/exporter/exporter.go
dntczdx: Good point. Let me rollback this change.

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: This is for testability and flexibility in different env. Sometimes we want to use different workflow file for testing purpose, or the file is located in different path in some environment.

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: package for main function. Unfortunately they can't use the same package name.

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: Agree.

PR #868, opened 2019-07-25
Daisy Workflow perf : Async deletion / cleanup

dntczdx:
Deletions can be triggered proactively (by DeleteResources step) or passively (by default cleanup at the end of the workflow execution). These steps wasted time of import/export, because deletion shouldn't be on on the critical path. No matter whether deletion is success ultimately, the workflow is finished and user has gotten the import/export result.

Due to the above reasons, we provided 3 new mechanism for workflow:
1. Async deletion. A "DeleteResources" step can be set as "async", so the whole workflow will continue without waiting for the result of deletion.
2. Async cleanup. With the new mechanism, cleanup is always async, so it will be finished in no more than 2s. In the past, it can spend up to 2 min, especially when it needs to clean up machine instances.
3. Orphan step. Any step can be set as "orphan", so that it won't be waited by any other steps (except for cleanup step). It sounds like being similar with "async deletion", but it is quite different due to 3 advantages:
 --3.1. It can be used by any step, not just DeleteResources.
 --3.2. Sometimes async deletion actually is depended by cleanup step, because a instance must be deleted before attached disk is cleaned up. We don't want critical steps to wait for instance deletion, but we still want deletion is done before cleanup. In that case, we can use "orphan" mode.
 --3.3. Sometimes a workflow is included by a parent workflow, and some steps in the sub workflow is non-critical. We don't want the parent workflow wait the sub workflow as an atomic step. By setting those non-critical sub steps as "orphan", we can unblock parent workflow.

It drastically improved small size import/export perf, for exmaple, base image import/export.


dntczdx:
This is a draft, not ready for reviewing yet

zoran15:
What are the plans for this PR? It's one year old at this point.

dntczdx:
Will close after the proejct is done. Now It's used to compare the new plan v.s. the old plan

PR #870, opened 2019-07-29
Move export related images out of the main Daisy Cloud Build config

zoran15:
Move export related images out of the main Daisy Cloud Build config due to reliability reset so we can continue building images not related to export.

### DIFF
--- a/daisy_release_cloudbuild.yaml
+++ b/daisy_release_cloudbuild.yaml
@@ -24,16 +24,10 @@ steps:
   args: ['build', '-o=linux/gce_image_publish', 'github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_image_publish']
   env: ['CGO_ENABLED=0']
 - name: 'gcr.io/cloud-builders/go'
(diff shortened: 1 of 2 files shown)


zoran15:
> Looks straight forward. The only problem: is the file name (daisy_release_cloudbuild v.s. export_release_cloudbuild) accurately described the scope? These 2 concepts sounds not like either-or.

daisy_release_cloudbuild should be interpreted as everything daisy based except export. I'd prefer not to rename that file to avoid breaking any dependencies (though the risk is low), plus I'm not sure if there would be a good name (daisy_except_export? :)

dntczdx:
> > Looks straight forward. The only problem: is the file name (daisy_release_cloudbuild v.s. export_release_cloudbuild) accurately described the scope? These 2 concepts sounds not like either-or.
> 
> daisy_release_cloudbuild should be interpreted as everything daisy based except export. I'd prefer not to rename that file to avoid breaking any dependencies (though the risk is low), plus I'm not sure if there would be a good name (daisy_except_export? :)

daisy_release_cloudbuild is not bad. For the other one, how do you fell about "daisy_export_release_cloudbuild"? I think you can go ahead to pick anyone and check in

PR #871, opened 2019-07-30
[import/export wrapper] Change bucket creation logic

dntczdx:
Original bucket creation logic is not aligned with regional design.
New logic:
1. import/export wrapper: return error when failed to create bucket for source/dest file. Originally it will create bucket on default region (US).
2. import/export wrapper: either given source/dest file bucket  doesn't exist or source/dest file is not given, we will use "fallback zone" to create bucket. Originally, the fallback zone is default: multi-region US. Now:
  1. The 1st choice of fallback zone is the input zone.
  2. The 2nd choice is the zone of running cloud build, if there isn't input zone.
  3. If running cloud build also doesn't exist (when wrapper is not running on GCE), we finally fallback to default: multi-region US.

I chose not to use single-region US-central1 as default region, because multi-region US is the default value for API struct "BucketAttrs".


cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	defaultRegion       = "US"
-	defaultStorageClass = "MULTI_REGIONAL"
```
zoran15: The way this const is now used, renaming it to multiregionalStorageClass makes more sense

cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -26,8 +26,9 @@ import (
...
+	defaultRegion        = "US"
```
zoran15: since what is default is determined programatically now, this should be renamed to defaultMultiregion

cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -124,3 +137,14 @@ func (c *ScratchBucketCreator) formatScratchBucketName(project string, location
 	}
 	return strings.ToLower(bucket)
 }
+
+func getRegion(zone string) (string, error) {
```
zoran15: This is a copy paste from param_utils. Reuse that method.

cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -45,38 +46,50 @@ func NewScratchBucketCreator(ctx context.Context, storageClient domain.StorageCl
...
-		// source file not provided or couldn't create bucket based on it, create default scratch bucket
-		bucket = c.formatScratchBucketName(project, defaultRegion)
-		region, err = c.createBucketIfNotExisting(bucket, project, &storage.BucketAttrs{Name: bucket, Location: defaultRegion, StorageClass: defaultStorageClass})
-		if err != nil {
+
+	// if source file is not provided, fallback to input / default region.
+	return c.createBucketOnFallbackZone(project, zone)
+}
+
+func (c *ScratchBucketCreator) createBucketOnFallbackZone(project string, zone string) (string, string, error) {
```
zoran15: Move this method below createBucketMatchFile region as it's not the first to be called from CreateScratchBucket

cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -45,38 +46,50 @@ func NewScratchBucketCreator(ctx context.Context, storageClient domain.StorageCl
...
+	bucket := c.formatScratchBucketName(project, fallbackRegion)
+	region, err := c.createBucketIfNotExisting(bucket, project, &storage.BucketAttrs{Name: bucket, Location: fallbackRegion, StorageClass: storageClass})
+	if err != nil {
+		return "", "", err
 	}
 	return bucket, region, nil
 }
 
-func (c *ScratchBucketCreator) createBucketMatchFileRegion(fileGcsPath string, project string) (string, string, error) {
+func (c *ScratchBucketCreator) createBucketMatchFileRegion(fileGcsPath string, project string, zone string) (string, string, error) {
```
zoran15: The logic for creating determining where to create a bucket can be simplified without introducing another method (and calling it from two places).

We have as inputs:
- source file (optional)
- zone (optional)

We need to figure out:
- storage region
- storage class

Once these two are determined, bucket can be created in a single spot in the code. Error can be returned along the way if necessary.

cli_tools/common/utils/param/param_utils.go
```diff
@@ -54,7 +54,12 @@ func PopulateMissingParameters(project *string, zone *string, region *string,
 
 	scratchBucketRegion := ""
 	if *scratchBucketGcsPath == "" {
-		scratchBucketName, sbr, err := scratchBucketCreator.CreateScratchBucket(file, *project)
+		fallbackZone := *zone
+		if fallbackZone == "" && mgce.OnGCE() {
+			// try to get zone which Cloud Build is running in, ignoring error
+			fallbackZone, _ = mgce.Zone()
```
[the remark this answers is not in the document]
dntczdx: if we can't get zone from mgce, we should fallback to input zone without interrupt anything.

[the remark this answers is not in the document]
zoran15: But then you depend on an assumption that fallbackZone will be "" in that case, which might now always be the case. I think we should check for an error and set fallback zone to "" explicitly.

cli_tools/common/utils/param/param_utils.go
```diff
@@ -66,10 +71,13 @@ func PopulateMissingParameters(project *string, zone *string, region *string,
 		if err != nil {
 			return fmt.Errorf("invalid scratch bucket GCS path %v", scratchBucketGcsPath)
 		}
+
 		scratchBucketAttrs, err := storageClient.GetBucketAttrs(scratchBucketName)
-		if err == nil {
-			scratchBucketRegion = scratchBucketAttrs.Location
+		if err != nil {
+			return fmt.Errorf("invalid scratch bucket name %v", scratchBucketName)
```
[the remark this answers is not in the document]
dntczdx: I thought there won't be an error if the path is valid. Let me change it back.

cli_tools/common/utils/storage/scratch_bucket_creator.go
dntczdx: They are still default. We are still using defaultRegion and defaultStorageClass when we can't determine region by any input.

cli_tools/common/utils/storage/scratch_bucket_creator.go
dntczdx: They are still default. We are still using defaultRegion and defaultStorageClass when we can't determine region by any input.
Other fallback region/zone value is infured from input, so I don't call them "default".
Please let me know whether that sounds correct.

cli_tools/common/utils/storage/scratch_bucket_creator.go
dntczdx: param_utils imports storage_utils, so I need to avoid circled dependency.
As a exposed function, param_utils is the correct place to put "GetRegion". But we need to use it in storage_utils.
It's not worth adding an extra common lib for that.

zoran15: I think adding an extra common lib to avoid code duplication of any decent size (and this one is decent) is worth it.

cli_tools/common/utils/storage/scratch_bucket_creator.go
dntczdx: Yes we can do it that way. It's clearer, but not significantly clearer due to the complexity. Check out my new version.

cli_tools/common/utils/storage/scratch_bucket_creator.go
dntczdx: Sounds good. Let me move it.

dntczdx: Sounds good, resetting the value

cli_tools/common/utils/param/param_utils.go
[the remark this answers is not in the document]
dntczdx: Sounds good. Resetting the value.

cli_tools/common/utils/paramhelper/region.go
```diff
@@ -0,0 +1,18 @@
+package paramhelper
```
zoran15: What about calling this package "parameters"? utils\parameters are utils related to parameters.

cli_tools/common/utils/paramhelper/region.go
```diff
@@ -0,0 +1,18 @@
...
+
```
zoran15: Add copyright docs

cli_tools/common/utils/paramhelper/region.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// GetRegion extracts region from a zones
```
zoran15: ...from a zone

mocks/mock_scratch_bucket_creator.go
```diff
@@ -1,19 +1,5 @@
-//  Copyright 2019 Google Inc. All Rights Reserved.
```
zoran15: Add the copyright back (it gets deleted by mockgen when you generate mock again)

cli_tools/common/utils/paramhelper/region.go
dntczdx: We have already had  "param" under "utils". This one is used by "param" and "storageutils".

dntczdx:
Ping for code review

cli_tools/common/utils/param/param_utils.go
```diff
@@ -52,38 +53,61 @@ func PopulateMissingParameters(project *string, zone *string, region *string,
...
+
+		scratchBucketName, sbr, err := scratchBucketCreator.CreateScratchBucket(file, *project, fallbackZone)
 		scratchBucketRegion = sbr
 		if err != nil {
-			return err
+			return "", err
 		}
 
 		*scratchBucketGcsPath = fmt.Sprintf("gs://%v/", scratchBucketName)
 	} else {
```
zoran15: This else block is actually validation of scratchBucketGcsPath (not population) and can be extracted to it's own func.

Then, the whole if/else could be moved to the calling func (PopulateMissingParameters) and this func (populateBucket) could just contain the if block.

cli_tools/common/utils/param/param_utils.go
```diff
@@ -52,38 +53,61 @@ func PopulateMissingParameters(project *string, zone *string, region *string,
...
+		}
+	}
+
+	if err := PopulateRegion(region, *zone); err != nil {
+		return err
+	}
+	return nil
+}
+
+func populateBucket(scratchBucketGcsPath *string, zone *string, mgce domain.MetadataGCEInterface,
```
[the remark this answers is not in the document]
dntczdx: To make sure the smallest side effect, I decide not to change the structure of this function for now. More optimization can be done later.

cli_tools/common/utils/param/param_utils.go
dntczdx: Moving if/else logic to calling func makes that func more complex. Ideally we can split populate to populateIfMissing + validate, but scratchBucketRegion needs to be determined here as well. Let's consider restructure this code section later since I prefer a conservative fix for now - there are many corner cases.

PR #872, opened 2019-08-01
Debian 10 secure boot support.

zmarano:


daisy_workflows/image_build/debian/fai_config/scripts/10-gce-clean
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# Untar secure boot keys.
+chroot $target tar -C /boot/efi/EFI/Google/gsetup -xzf /boot/efi/EFI/Google/gsetup/secure-boot-keys.tar.gz
```
hopkiw: unless we don't think we have `tar` in the executing environment, we shouldn't use chroot here or in the rm line

secondly, please move this tarball to /boot during the image rather than deep into the gsetup command. it will be more future-proof.

zmarano: Thanks, done.

PR #873, opened 2019-08-02
osconfig_tests: large refactor of GuestPolicy for v1alpha2

adjackura:
Move all package tests to v1alpha2 and rename to guestpolicies
Refactor/simplify guestpolicies tests
Removed unneeded PackageInstallRemoval test
Added 'Package update' and 'package does not update' tests
Added debian-10, maded room for suse support but not adding any suse tests yet


osconfig_tests/test_suites/guestpolicies/guest_policies.go
```diff
@@ -0,0 +1,223 @@
...
+	packageInstallFunction            = "pkginstall"
+	packageRemovalFunction            = "pkgremoval"
+	packageInstallFromNewRepoFunction = "pkgfromnewrepo"
+	packageUpdateFunction             = "pkgupdate"
+	packageNoUpdateFunction           = "pkgnoupdate"
+)
+
+type packageManagementTestSetup struct {
+	image         string
+	name          string
```
ryanwe: Here is where you should consider changing this to `guestPolicyId` and adding `vmInstanceId`. It also threw me for a loop since `name` is also used for the label. But I'm guessing that will be changed as soon as the fix is deployed.

osconfig_tests/utils/utils.go
```diff
@@ -255,6 +262,7 @@ func CreateComputeInstance(metadataitems []*api.MetadataItems, client daisyCompu
 				Scopes: serviceAccountScopes,
 			},
 		},
+		Labels: map[string]string{"name": name},
```
ryanwe: 😮After this is fixed, can we have some e2e tests that use labels while others use instance URIs or prefixes? Though maybe probers are a better place for that kind of testing.

adjackura: I think that would be a better added as a prober, we can discuss where we want various test cases later. One we support osInfo filtering then we probably want to add some more e2e tests?

osconfig_tests/test_suites/guestpolicies/guest_policies_test_data.go
```diff
@@ -0,0 +1,246 @@
...
+func buildPkgInstallTestSetup(name, image, pkgManager, key string) *packageManagementTestSetup {
+	assertTimeout := 60 * time.Second
+	testName := packageInstallFunction
+	packageName := "cowsay"
+	machineType := "n1-standard-2"
+	if pkgManager == "googet" {
+		machineType = "n1-standard-4"
+	}
+
+	instanceName := fmt.Sprintf("%s-%s-%s-%s", path.Base(name), testName, key, utils.RandString(3))
```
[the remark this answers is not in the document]
adjackura: I added guestPolicyId to guestPolicyTestSetup (renamed from packageManagementTestSetup) and changed name to instanceName

osconfig_tests/main.go
```diff
@@ -36,6 +36,8 @@ import (
 	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/test_suites/guestpolicies"
 	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/test_suites/inventory"
 	"github.com/GoogleCloudPlatform/compute-image-tools/osconfig_tests/test_suites/patch"
+
+	_ "google.golang.org/genproto/googleapis/rpc/errdetails"
```
ryanwe: 👍

PR #874, opened 2019-08-02
Support for image storage location using Beta compute API

zoran15:


daisy/step_create_images.go
```diff
@@ -16,63 +16,131 @@ package daisy
...
 	"google.golang.org/api/googleapi"
 )
 
+//UseBetaAPI determines if Beta Compute API should be used for calls that use Beta features.
+var UseBetaAPI = false
+
 // CreateImages is a Daisy CreateImages workflow step.
-type CreateImages []*Image
+type CreateImages struct {
+	Images     []*Image
```
dntczdx: Due to this change, do we need to change our import/export workflows accordingly?

zoran15: No, because custom Unmarshaller will still parse workflows in original JSON format.

daisy/step_create_images.go
```diff
@@ -16,63 +16,124 @@ package daisy
...
+	ImagesBeta []*ImageBeta
+}
+
+// UnmarshalJSON unmarshals Image.
+func (ci *CreateImages) UnmarshalJSON(b []byte) error {
+	imagesBeta := &[]*ImageBeta{}
```
adjackura: Dont take the reference here so you don't need to dereference later, also this is more idiomatic:

var imagesBeta []*ImageBeta
if err := json.Unmarshal(b, &imagesBeta); err != nil {
		return err
	}
ci.ImagesBeta = imagesBeta

daisy/step_create_images.go
```diff
@@ -16,63 +16,124 @@ package daisy
...
 }
 
-func (c *CreateImages) validate(ctx context.Context, s *Step) dErr {
+func (ci *CreateImages) validate(ctx context.Context, s *Step) dErr {
 	var errs dErr
-	for _, i := range *c {
-		errs = addErrs(errs, i.validate(ctx, s))
+
+	if usesBetaFeatures(ci.ImagesBeta) {
+		if ci.ImagesBeta != nil {
```
adjackura: These nil checks right before the range are not needed.

daisy/step_create_images.go
```diff
@@ -16,63 +16,124 @@ package daisy
...
+		if err := ci.create(w.ComputeClient); err != nil {
+			e <- newErr(err)
+			return
+		}
+	}
+
+	if usesBetaFeatures(ci.ImagesBeta) {
+		for _, i := range ci.ImagesBeta {
+			wg.Add(1)
+			go createImage(i, i.OverWrite)
```
adjackura: Passing in overwrite is a bit redundant as you already are passing in ImageInterface

zoran15: I didn't want to add another method to the interface since overwrite is part of ImageBase. That's why I pass ImageBase reference in addition to ImageInterface to certain methods, so I can use ImageBase fields without having to add getters/setters in Image Interface.

daisy/step_create_images.go
zoran15: What do you exactly mean with "dont take the reference here"?

daisy/step_create_images.go
zoran15: Removed

daisy/step_create_images.go
adjackura: Meant to say take the address, when you instantiate imagesBeta you take the address of a slice that you then deference in line 37.

zoran15: Wouldn't that copy images/imagesBeta objects when assigning to ci.Images/ciImagesBeta?

zoran15: Updated code after chatting in person

PR #879, opened 2019-08-08
Fixed GuestOsFeatures not being parsed.

zoran15:


daisy/test_data/test.wf.json
```diff
@@ -77,15 +77,29 @@
...
     "create-image": {
       "createImages": [
         {
           "Name": "image-from-disk",
-          "SourceDisk": "image"
+          "SourceDisk": "image",
+          "Description": "Microsoft, SQL Server 2016 Web, on Windows Server 2019",
+          "Family": "sql-web-2016-win-2019",
+          "Project": "a_project",
+          "NoCleanup": true,
```
dntczdx: Is "NoCleanup" required here? It shouldn't be required if validation is done in the workflow.

zoran15: This is testing workflow parsing is actually reading all the fields. Workflow itself will not be run here.

PR #881, opened 2019-08-12
Fix OVF e2e test

zoran15:
Fix OVF e2e test that is failing due to the introduction of a new machine type family today (N2 - https://cloud.google.com/blog/products/compute/expanding-virtual-machine-types-to-drive-performance-and-efficiency)


dntczdx:
May I know why a new machine type can lead to test failure? Is it because this new machine become to the best match?

zoran15:
> May I know why a new machine type can lead to test failure? Is it because this new machine become to the best match?

Yes. OVF asks for 8GB of RAM which is now perfectly matched with n2-standard-2

PR #884, opened 2019-08-14
Extra error message when disk not bootable during import

zoran15:
Extra info in translate timeout due to disk potentially not being bootable in GCE.

### DIFF
--- a/daisy/step.go
+++ b/daisy/step.go
@@ -38,8 +38,8 @@ type Step struct {
 	name string
 	w    *Workflow
 
-	//Run error description
-	ErrorDescription string `json:",omitempty"`
+	//Timeout description
+	TimeoutDescription string `json:",omitempty"`
 	// Time to wait for this step to complete (default 10m).
 	// Must be parsable by https://golang.org/pkg/time/#ParseDuration.
 	Timeout string `json:",omitempty"`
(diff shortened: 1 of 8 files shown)


daisy_workflows/image_import/debian/translate_debian.wf.json
```diff
@@ -77,7 +77,8 @@
           }
         }
       ],
-      "Timeout": "60m"
+      "Timeout": "60m",
+      "ErrorDescription": "This might indicate the disk is not bootable in Google Compute Engine. Make sure the disk satisfies the requirements documented at https://cloud.google.com/compute/docs/import/importing-virtual-disks."
```
dntczdx: is it possible to catch the bootable issue other than waiting for timeout? I think response of instance.start API should be able to give us some info about the boot.

zoran15: We don't call instance.start, instances are auto started on create. 

Also, instance.create and instance.start only return errors, not detailed status on whether OS has booted. GCE doesn't know if OS has booted or not or if there's an OS at all running on the instance.

daisy/step.go
```diff
@@ -37,6 +38,8 @@ type Step struct {
 	name string
 	w    *Workflow
 
+	//Run error description
+	ErrorDescription string `json:",omitempty"`
```
adjackura: TimeoutErrorDescription, TimeoutErrorText, or just TimeoutDescription/TimeoutText? The comment should be updated to reflect that this is just used for time out and not other errors

zoran15: Renamed to TimeoutDescription. Also, changed the actual description based on Shawnique's suggestion.

PR #885, opened 2019-08-14
new daisy error mechanism to remove privacy info from SafeMessage

dntczdx:
This PR is the step 1 to log import/export info to firelog. It built a mechanism to return a "SafeErrorMessages" in each DaisyError which doesn't contain any privacy info.

1. Change internal error type daisy.dErr to public daisy.DError. Reason: we use it to contain a "SafeErrorMessages", and want to access it from public. Changed those error related functions to public accordingly (errf -> Errf).
2. How to implement "SafeErrorMessages" to exclude privacy info? there are 2 things done:
<1> error from startup scripts (.sh, or .py):
  <1.1> when throwing error from ".sh", we enclose privacy info by a special format: "[Privacy-> xxx <-Privacy]". This content will be replaced when outputing.
  <1.2> when throwing error from ".py": since those .py scripts are translation scripts, and we are not able to predict what will be thrown, we have to hide all error details after "TranslateFailed".
<2> error thrown from daisy/wrapper: most of these errors have a clear background what happened. Thus, we hided privacy details by removing string format args and only log format itself.
3. How to guarantee all errors are covered? We accept daisy.DError at the main func of import/export wrapper. Thus, all error sources have to be modified, which is guaranteed by compiler.

This PR only provided a "SafeErrorMessages" interface so it can be used by log later.

Core logic is changed in:
daisy/error.go
daisy_workflows/export/disk_resizing_mon.sh
daisy_workflows/export/export_disk.sh
daisy_workflows/export/export_disk_ext.sh
daisy_workflows/image_import/import_image.sh

Almost all other files only got changed by 2 reasons:
1. class name (daisy.DError) / function name (daisy.Errf)
2. Modified error msg format to make the root cause more clear.

PS: instance import is not covered so far. We need an additional PR for that.


cli_tools/common/domain/interfaces.go
```diff
@@ -20,6 +20,8 @@ import (
 	"net/http"
 
 	"cloud.google.com/go/storage"
+
```
adjackura: You have an extra space with the new import in a few files

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -91,5 +94,20 @@ func updateAllInstanceAccessConfig(workflow *daisy.Workflow, accessConfigProvide
...
+func RemovePrivacyLogInfo(message string) string {
+	// Since translation scripts are vary and is hard to predict the output, we have to hide the
+	// details and only remain "TranslateFailed"
+	if strings.Contains(message, "TranslateFailed") {
+		return "TranslateFailed"
+	}
+
+	// All import/export bash scripts bracks privacy info inside "[Privacy-> XXX <-Privacy]". Let's
+	// remove it for privacy.
+	var re = regexp.MustCompile(`\[Privacy\->.*?<\-Privacy\]`)
```
adjackura: If using MustCompile make this a global var so the panic happens immediately an not at function run time.

dntczdx:
> This seems like an implementation detail for the wrappers, they should just parse the error text as is, DaisyError shouldn't need a function for it. Also the public functions should not return DaisyError, if you need DaisyError callers can attempt to cast, but I'd prefer to keep it private and run the SafeErrorMessages on err.Error()

dErrImpl hides implementations, while its interface, dErr, can be public so external package can operate it. The reason why I need to expose a public interface is because dErr contains multiple errors, while err.error() only return one message. I can split it with "\n", but it's inaccurate - a single error can also be multi-line. In that case, I may be failed to hide some privacy info.

For example: an error can be as below:

TranslationFailed: failed to fetch repository.\n
due to this error, machine "instance-a" can't translate disk "disk-b".

Then I will fail to hide the 2nd line, which contains privacy info. Thoughts?

dntczdx:
> Then I will fail to hide the 2nd line, which contains privacy info. Thoughts?

I can do it like this: keep current dErr interface; add a new interface, DError, which extends dErr with method SafeErrorMessages(removePrivacyInfo func(string) string) []string
How do you think?

cli_tools/common/utils/daisy/daisy_utils.go
dntczdx: Right, that is by design. We use this regex to detect privacy info. It shouldn't fail due that it's a static expression.

zoran15:
There are two aspects we want to learn about with extra logging and usage data:
1. Details on how import/export is used. E.g. file formats, OS, data-disks etc. 
2. Detailed failure reasons to improve reliability

1. doesn't contain any PII data and we can already collect this data from wrappers. We just need to connect to FireLog.

2. is not usage data but detailed debugging/logging data. We discussed this with the Privacy group and agreed we can store PII data as long as we have short retention policy around it. I thought we were going in this direction?

dntczdx:
> There are two aspects we want to learn about with extra logging and usage data:
> 
> 1. Details on how import/export is used. E.g. file formats, OS, data-disks etc.
> 2. Detailed failure reasons to improve reliability
> 3. doesn't contain any PII data and we can already collect this data from wrappers. We just need to connect to FireLog.
> 4. is not usage data but detailed debugging/logging data. We discussed this with the Privacy group and agreed we can store PII data as long as we have short retention policy around it. I thought we were going in this direction?

We are doing #1,#2,#3 by SafeErrorMessages since reliability and usage is the main purpose.
#4 can be done by simply define one more log source and output DError.error().

zoran15:
> > This seems like an implementation detail for the wrappers, they should just parse the error text as is, DaisyError shouldn't need a function for it. Also the public functions should not return DaisyError, if you need DaisyError callers can attempt to cast, but I'd prefer to keep it private and run the SafeErrorMessages on err.Error()
> 
> dErrImpl hides implementations, while its interface, dErr, can be public so external package can operate it. The reason why I need to expose a public interface is because dErr contains multiple errors, while err.error() only return one message. I can split it with "\n", but it's inaccurate - a single error can also be multi-line. In that case, I may be failed to hide some privacy info.
> 
> For example: an error can be as below:
> 
> TranslationFailed: failed to fetch repository.\n
> due to this error, machine "instance-a" can't translate disk "disk-b".
> 
> Then I will fail to hide the 2nd line, which contains privacy info. Thoughts?

zoran15:
> I can split it with "\n", but it's inaccurate - a single error can also be multi-line. In that case, I may be failed to hide some privacy info.

Could you use the same pattern with embedding privacy safe data by using [Privacy-> xxx <-Privacy] tags in error message to avoid having to change the whole error handling API?

This way, wrappers could extract PII and public data and log them appropriately.

dntczdx:
> > I can split it with "\n", but it's inaccurate - a single error can also be multi-line. In that case, I may be failed to hide some privacy info.
> 
> Could you use the same pattern with embedding privacy safe data by using [Privacy-> xxx <-Privacy] tags in error message to avoid having to change the whole error handling API?
> 
> This way, wrappers could extract PII and public data and log them appropriately.

That is a way to implement. The reason why I didn't chose that way:
[1] The format "[Privacy-> xxx <-Privacy]" needs to be processed by wrapper layer. So daisy layer and wrapper needs to cooperate to make it work. Daisy layer can't handle this tag individually.
[2] It still can't handle the multi-line issue. We always need to expose a interface to get multiple errors, thus it has to be public.
The only benefit of that way is to avoid changing a class name and a function name, which is not a big deal from my perspective.

cli_tools/common/utils/daisy/daisy_utils.go
adjackura: But it should be a global var so failure happens immediately on binary execution

daisy/disk.go
```diff
@@ -145,44 +145,44 @@ func (dr *diskRegistry) init() {
...
-func (dr *diskRegistry) deleteFn(res *Resource) dErr {
+func (dr *diskRegistry) deleteFn(res *Resource) DError {
 	m := namedSubexp(diskURLRgx, res.link)
 	err := dr.w.ComputeClient.DeleteDisk(m["project"], m["zone"], m["disk"])
 	if gErr, ok := err.(*googleapi.Error); ok && gErr.Code == http.StatusNotFound {
-		return typedErr(resourceDNEError, err)
+		return typedErr(resourceDNEError, "failed to delete disk", err)
 	}
-	return newErr(err)
+	return newErr("failed to delete disk", err)
```
zoran15: Is it fair to say that here, "failed to delete disk" is practically error type, an enumeration, or error type? This would be the case if safe error messages are always the same when thrown from the same piece of code.

If so, we should consider representing them in a different way than plain strings. This would allow us to remove safeErrMsgs. For example, we could have different error implementation (type) for each type of error. We could use class name as error type.

adjackura:
Not sure I understand the need for all this still, whats the difference to you between []error and []string? If you add a method to get []error from DError then the caller can handle all the SafeErrorMessages and not have the method in DError. Also any returned error should be able to be cast as DError, so please don't have any public functions return DError.

zoran15:
> dErrImpl hides implementations, while its interface, dErr, can be public so external package can operate it.

I guess what Andrew wanted to say is that clients (wrappers) can still use DErr methods even if error is returned from public methods by casting error to DErr (or some other class like different class for different type of error as suggested in one of my comments that would inherit from DErr).

dntczdx:
> Not sure I understand the need for all this still, whats the difference to you between []error and []string? If you add a method to get []error from DError then the caller can handle all the SafeErrorMessages and not have the method in DError. Also any returned error should be able to be cast as DError, so please don't have any public functions return DError.

That []string contains "safe" error messages - which hides privacy info. Example:
derr := daisy.Errf("Failed to create disk for project %v", my-project)
then derr.safeErrorMessage contains "Failed to create disk for project %v",
while derr.err contains "Failed to create disk for project my-project"

Let's discuss offline about this design to better understand why I decide to do it like this and whether your idea is feasible.

zoran15:
> Not sure I understand the need for all this still, whats the difference to you between []error and []string? If you add a method to get []error from DError then the caller can handle all the SafeErrorMessages and not have the method in DError. Also any returned error should be able to be cast as DError, so please don't have any public functions return DError.

Expanding on this, instead of having SafeErrorMessage, could the code that returns an error return two errors: one SafeError and FullError. That way, we can have existing code operate only on FullErrors while client code (wrappers) could filter out SafeErrors. This way, we're not forcing errors to be two things at the same time (safe error and full error) but instead aggregating them, which error handling code already supports. We can even have different classes for different safe errors, enum like.

dntczdx:
> I guess what Andrew wanted to say is that clients (wrappers) can still use DErr methods even if error is returned from public methods by casting error to DErr (or some other class like different class for different type of error as suggested in one of my comments that would inherit from DErr).

That sounds great. I didn't know golang can cast in that way. Let me try whether that works...

daisy/disk.go
dntczdx: Basically yes, even a int number can represent this info (just maintain a int->string table somewhere). However, we removed safeErrMsgs but added errorCode (or a new class) - we anyways need something new to represent it.

My implementation has 3 extra advantages:
1. Easier to read from log DB.
2. Flexibility - we actually don't want to hide details in all cases. A good example: daisy.Errf("Failed to parse workflow step %v", step.Name). In this case, we actually don't want to hide the step name, because it's not privacy related and it's very useful to identify/categorize the error. The correct code here should be:
daisy.Errf(fmt.Sprintf("Failed to parse workflow step %v", step.Name))
This design gives us the flexibility.
3. The easiest way to add new error type. We don't need to track the largest error code, we don't need to create a new class, we can modify existing error message anytime without side effect.

dntczdx:
Offline discussed and changed accordingly.

cli_tools/common/utils/validation/validation_utils.go
```diff
@@ -16,18 +16,19 @@ package validation
...
 // ValidateStringFlagNotEmpty returns error with error message stating field must be provided if
 // value is empty string. Returns nil otherwise.
 func ValidateStringFlagNotEmpty(flagValue string, flagKey string) error {
-	return ValidateStringNotEmpty(flagValue, flagKey, "The flag -%v must be provided")
+	return validateStringNotEmpty(flagValue, flagKey, "The flag -%v must be provided")
 }
 
-// ValidateStringNotEmpty returns error with specific error message if string is empty. Returns nil otherwise.
-func ValidateStringNotEmpty(value string, key string, errorMessage string) error {
+func validateStringNotEmpty(value string, key string, errorMessage string) daisy.DError {
```
adjackura: If no other code uses this just move the code into ValidateStringFlagNotEmpty

cli_tools/daisy/main.go
```diff
@@ -88,23 +88,23 @@ Loop:
 				continue Loop
 			}
 		}
-		return nil, fmt.Errorf("unknown workflow Var %q passed to Workflow %q", k, w.Name)
+		return nil, daisy.Errf("unknown workflow Var %q passed to Workflow %q", k, w.Name)
```
adjackura: No need to use daisy.Errf here as this is package main and has no callers.

daisy/error.go
```diff
@@ -65,39 +67,57 @@ func addErrs(e dErr, errs ...error) dErr {
...
-func newErr(e error) dErr {
+func newErr(replacementMsg string, e error) DError {
 	if e == nil {
 		return nil
 	}
 	if dE, ok := e.(*dErrImpl); ok {
 		return dE
 	}
-	return &dErrImpl{errs: []error{e}}
+	return &dErrImpl{errs: []error{e}, errFormats: []string{replacementMsg}}
```
zoran15: errFormat and replacementMsg naming is not in sync. Both are also not very descriptive of what they actually hold (anonymized error message). Consider renaming both to something like anonymizedErrs or similar. It's already called safeErrMsg in some pieces of code. That one is better although it's not clear what the message is safe from.  anonymized described exactly what these messages are

daisy/error.go
```diff
@@ -30,30 +30,32 @@ const (
...
 	error
 
-	// add shouldn't be called directly, instead call addErrs(dErr, error).
-	// This assists with nil dErrs. addErrs(nil, e) will return a new dErr.
+	// add shouldn't be called directly, instead call addErrs(DError, error).
+	// This assists with nil dErrs. addErrs(nil, e) will return a new DError.
 	add(error)
-	Type() string
+	etype() string
+	ErrorFormats() []string
```
zoran15: should have docs. Also consider renaming as mentioned in a comment below

daisy/error.go
```diff
@@ -65,39 +67,57 @@ func addErrs(e dErr, errs ...error) dErr {
...
-	return typedErr(errType, fmt.Errorf(format, a...))
+func typedErrf(errType, format string, a ...interface{}) DError {
+	return typedErr(errType, format, fmt.Errorf(format, a...))
 }
 
 type dErrImpl struct {
-	errs    []error
-	errType string
+	errs       []error
+	errFormats []string
```
zoran15: As mentioned in other comments, rename this to e.g. anonymizedErrs, as this can hold other values except formats (when newErr is called directly)

cli_tools/gce_vm_image_import/main.go
```diff
@@ -65,6 +67,18 @@ func main() {
 		*stdoutLogsDisabled, *kmsKey, *kmsKeyring, *kmsLocation, *kmsProject, *noExternalIP,
 		*labels, currentExecutablePath, *storageLocation); err != nil {
 
+		defer func() {
+			// TODO: log safe messages to firelog
+			derr := daisy.ToDError(err)
+			if derr == nil {
+				return
+			}
+			safeMsgs := []string{}
```
zoran15: anonymizedErrs or similar

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -45,15 +47,16 @@ var (
...
 
 // ValidateOS validates that osID is supported by Daisy image import
 func ValidateOS(osID string) error {
 	if osID == "" {
-		return fmt.Errorf("osID is empty")
+		return daisy.Errf("osID is empty")
 	}
 	if _, osValid := osChoices[osID]; !osValid {
-		return fmt.Errorf("os `%v` is invalid. Allowed values: %v", osID, reflect.ValueOf(osChoices).MapKeys())
+		return daisy.Errf(fmt.Sprintf("os `%v` is invalid. Allowed values: %v", osID, reflect.ValueOf(osChoices).MapKeys()))
```
zoran15: formatting here is, I assume, so that error gets logged to both personalized and anonymized log? 

Consider either documenting this trick or providing a separate method to log errors that are inherently anonymized?

cli_tools/gce_vm_image_export/main.go
```diff
@@ -50,6 +52,18 @@ func main() {
 		*network, *subnet, *zone, *timeout, *scratchBucketGcsPath, *oauth, *ce, *gcsLogsDisabled,
 		*cloudLogsDisabled, *stdoutLogsDisabled, *labels, currentExecutablePath); err != nil {
 
+		defer func() {
```
zoran15: Why is this deferred? Can't it be run directly here before log.Fatal?

daisy/instance.go
```diff
@@ -182,7 +182,7 @@ func (i *Instance) populateMetadata(w *Workflow) dErr {
 	i.Metadata["daisy-outs-path"] = "gs://" + path.Join(w.bucket, w.outsPath)
 	if i.StartupScript != "" {
 		if !w.sourceExists(i.StartupScript) {
-			return errf("bad value for StartupScript, source not found: %s", i.StartupScript)
+			return Errf("bad value for StartupScript, source not found: %s", i.StartupScript)
```
zoran15: StartupScript is not private so no need to hide it here

daisy/error.go
dntczdx: Renamed  to anonymizedErrs/anonymizedErrMsg as suggested.

cli_tools/gce_vm_image_export/main.go
dntczdx: This func() is useless so far, so I used defer to avoid its execution. log.Fatal() will call to os.Exit(), which will skip deferred func.

daisy/instance.go
dntczdx: It's a URL, which exposes some info, e.g. bucket name, which is PII

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -91,5 +94,19 @@ func updateAllInstanceAccessConfig(workflow *daisy.Workflow, accessConfigProvide
 			}
 		}
 	}
+}
+
+// RemovePrivacyLogInfo removes privacy log information.
+func RemovePrivacyLogInfo(message string) string {
+	// Since translation scripts are vary and is hard to predict the output, we have to hide the
+	// details and only remain "TranslateFailed"
+	if strings.Contains(message, "TranslateFailed") {
```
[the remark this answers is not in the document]
dntczdx: No unfortunately, since it's just a string. We have to do pattern match. Similarly "StatusMatch", "SuccessMatch" and "FailrueMatch" have the same problem.

dntczdx:
Improved details as suggested.

cli_tools/common/utils/daisy/daisy_utils.go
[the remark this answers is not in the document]
zoran15: At least make it configurable and extract failure prefixes to constants

PR #890, opened 2019-08-20
Daisy bugs in non-JSON workflow creation paths.

hopkiw:
WaitForInstancesSignal failure match list should be public
IncludeWorkflow needs to set parent during populate

### DIFF
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -492,9 +492,7 @@ func (w *Workflow) AddDependency(dependent *Step, dependencies ...*Step) error {
 	return nil
 }
 
-// NewIncludedWorkflow instantiates a new workflow with the same resources as the parent.
-func (w *Workflow) NewIncludedWorkflow() *Workflow {
-	iw := New()
+func (w *Workflow) includeWorkflow(iw *Workflow) {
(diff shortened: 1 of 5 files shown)


daisy/workflow.go
```diff
@@ -492,9 +492,7 @@ func (w *Workflow) AddDependency(dependent *Step, dependencies ...*Step) error {
 	return nil
 }
 
-// NewIncludedWorkflow instantiates a new workflow with the same resources as the parent.
-func (w *Workflow) NewIncludedWorkflow() *Workflow {
-	iw := New()
+func (w *Workflow) includeWorkflow(iw *Workflow) {
```
dntczdx: To me, the original way looks more natural. Why do we change to the new way?

hopkiw: the reason for this change is that the 'parenting' this function does is necessary for included workflows to function properly. if you are building workflows from JSON, the NewIncludedWorkflow function was called (from NewIncludedWorkflowFromFile). But if you've crafted a workflow in code, it wasn't ever called. So essentially you could not create a working IncludeWorkflow step in code. This PR splits the parenting from the workflow creation, so that it can be used in both cases.

PR #895, opened 2019-08-21
Fix OVF import failure when location restrictions are on

zoran15:
Fix OVF import failure when location restrictions are on due to images being created in multi-regions. Since this change is internal to the importer, added releaseTrack flag to indicate which Compute API release track is allowed to be used. Image location is only in Beta and we shouldn't use Beta API in GA release.

### DIFF
--- a/cli_tools/common/utils/daisy/resource_labeler_test.go
+++ b/cli_tools/common/utils/daisy/resource_labeler_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
@@ -128,6 +129,30 @@ func TestUpdateWorkflowImagesLabelled(t *testing.T) {
 						},
 					},
 				},
+				ImagesBeta: []*daisy.ImageBeta{
+					{
+						Image: computeBeta.Image{
+							Name:   "final-image-1",
+							Labels: map[string]string{"labelKey": "labelValue"},
+						},
+					},
+					{
+						Image: computeBeta.Image{
+							Name: "final-image-2",
+						},
+					},
+					{
+						Image: computeBeta.Image{
+							Name:   "untranslated-image-1",
+							Labels: map[string]string{"labelKey": "labelValue"},
+						},
+					},
+					{
+						Image: computeBeta.Image{
+							Name: "untranslated-image-2",
+						},
+					},
+				},
 			},
 		},
 	}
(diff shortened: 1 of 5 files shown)


cli_tools/common/utils/daisy/resource_labeler_test.go
```diff
@@ -139,11 +164,51 @@ func TestUpdateWorkflowImagesLabelled(t *testing.T) {
...
+
+	rl := createTestResourceLabeler(buildID, userLabels)
+	rl.ImageLocation = "europe-west5"
+
+	rl.LabelResources(w)
+
+	validateLabels(t, &(*w.Steps["cimg"].CreateImages).ImagesBeta[0].Image.Labels, "gce-image-import",
+		buildID, &existingLabels)
+
+	assert.Equal(t, "europe-west5", (*w.Steps["cimg"].CreateImages).ImagesBeta[0].Image.StorageLocations[0])
```
EricEdens: Should the test fail if there are other labels in the list? Does the test not care?

cli_tools/gce_ovf_import/main.go
```diff
@@ -59,6 +60,7 @@ var (
 	gcsLogsDisabled             = flag.Bool("disable-gcs-logging", false, "do not stream logs to GCS")
 	cloudLogsDisabled           = flag.Bool("disable-cloud-logging", false, "do not stream logs to Cloud Logging")
 	stdoutLogsDisabled          = flag.Bool("disable-stdout-logging", false, "do not display individual workflow logs on stdout")
+	releaseTrack                = flag.String("release-track", ovfimporter.GA, fmt.Sprintf("Release track of OVF import. One of: %s, %s or %s. Impacts which compute API release track is used by the import tool.", ovfimporter.Alpha, ovfimporter.Beta, ovfimporter.GA))
```
EricEdens: Is there an idiom for validating that the user actually passed ga, alpha, or beta?

cli_tools/gce_ovf_import/ovf_importer/ovf_importer_test.go
```diff
@@ -182,6 +183,50 @@ func TestSetUpWorkflowHappyPathFromOVAExistingScratchBucketProjectZoneAsFlags(t
 	assert.Equal(t, "gs://bucket/folder/ovf-import-build123/ovf/", oi.gcsPathToClean)
 }
 
+func TestSetUpWorkflowUsesImageLocationForBetaReleaseTrack(t *testing.T) {
```
EricEdens: How about for GA? Do you want to assert that the location is not set?

cli_tools/common/utils/daisy/resource_labeler_test.go
zoran15: That's already checked using existingLabels

EricEdens: The 'ImagesBeta[0].Image.StorageLocations' list, I mean. For example, to ensure that a duplicate doesn't sneak in there or something like that. A quick check would be to assert that the list has length one, I'd think.

zoran15: Added length==1 assertion

cli_tools/gce_ovf_import/main.go
zoran15: Not sure it's an idiom, but I added a validation function for release-track flag.

cli_tools/gce_ovf_import/ovf_importer/ovf_importer_test.go
zoran15: Added a test case for GA

PR #896, opened 2019-08-21
Remove gcsfuse as a dependency.

EricEdens:
gcsfuse has been fickle, and we've seen problems with importing images that were alleviated by copying the image to a local file.

This change remove gcsfuse by replacing it with a persistent disk that I'm calling a 'scratch' disk. When the workflow starts, we copy the image to the scratch disk, and then proceed as we did before.

## Testing

For testing, I ran a daisy integ test locally:

```bash
daisy -zone=us-central1-a -project=compute-image-tools-test daisy_integration_tests/ubuntu_1604_import_and_translate.wf.json
```


daisy_workflows/image_import/import_image.sh
```diff
@@ -48,12 +51,37 @@ function resizeDisk() {
...
 
-# Mount GCS bucket containing the disk image.
-mkdir -p /gcs/${SOURCEBUCKET}
-gcsfuse --implicit-dirs ${SOURCEBUCKET} /gcs/${SOURCEBUCKET}
+function copyImageToStagingDisk() {
+  local souceSizeGigabytes="$((1 + SOURCESIZE_BYTES / 1024 / 1024 / 1024))"
+  # We allocate double capacity here since the source file
+  # could be an OVA, which would require making an additional
+  # copy of its enclosed VMDK.
+  local stagingDiskSizeGigabytes="$((souceSizeGigabytes * 2))"
```
zoran15: I would guess that percentage of customers importing OVA as an image is very low. Additionally, we have instance import that is specifically designed for importing OVA/OVF files. 

We can still keep support for OVA in image import not to break backward compatibility. However, doubling scratch disk size for all regular/non-OVA imports (which are the bulk of imports) is a waste of resources. Can you resize the disk to double the capacity only if OVA is actually being imported?

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,12 +51,37 @@ function resizeDisk() {
...
 
-# Atrocious OVA hack.
-SOURCEFILE_TYPE="${$SOURCE_DISK_FILE##*.}"
+  # This disk is initially created with 10GB of space.
+  # Enlarge it if that's insufficient to hold the input image.
+  if [[ ${stagingDiskSizeGigabytes} -gt 10 ]]; then
+    resizeDisk "${STAGING_DISKNAME}" "${stagingDiskSizeGigabytes}" "${ZONE}"
+  fi
+
+  local stagingPath="/gcs/$(dirname ${SOURCEPATH})/"
```
zoran15: /gcs/ is not an appropriate directory name anymore as it's not linked to gcs. It can lead someone to assume it's linked to GCS bucket as before when it's not.

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,12 +51,37 @@ function resizeDisk() {
...
+    echo "ImportFailed: Failed to prepare staging disk."
+  fi
+  gsutil cp "${SOURCEURL}" "${stagingPath}"
+  echo "Import: Copied image from ${SOURCEURL} to ${stagingPath}"
+}
+
+# First, copy the disk image from cloud storage to a PD.
+copyImageToStagingDisk
+
+# Second, if the image is an OVA, then copy out its VMDK.
```
zoran15: "Atrocious OVA hack." sounded more fun :)

daisy_workflows/image_import/import_image.sh
```diff
@@ -62,6 +90,7 @@ if [[ "${SOURCEFILE_TYPE}" == "ova" ]]; then
   echo "Import: New source file is ${VMDK}"
 fi
 
+# Third, ensure the output disk has sufficient space to accept the disk image.
```
zoran15: I would not enumerate steps (first, second, third) as they could easily get out of date if we ever add more steps in between.

daisy_workflows/image_import/import_disk.wf.json
```diff
@@ -43,18 +43,28 @@
           "Type": "pd-ssd",
           "ExactName": true,
           "NoCleanup": true
+        },
+        {
+          "Name": "disk-${NAME}-staging-${ID}",
```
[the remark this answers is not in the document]
EricEdens: It is! Thanks for catching this. Switching to 'scratch'

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,12 +51,37 @@ function resizeDisk() {
   fi
 }
 
-# Mount GCS bucket containing the disk image.
-mkdir -p /gcs/${SOURCEBUCKET}
-gcsfuse --implicit-dirs ${SOURCEBUCKET} /gcs/${SOURCEBUCKET}
+function copyImageToStagingDisk() {
+  local souceSizeGigabytes="$((1 + SOURCESIZE_BYTES / 1024 / 1024 / 1024))"
```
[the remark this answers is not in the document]
EricEdens: good catch

daisy_workflows/image_import/import_image.sh
EricEdens: makes sense -- I'll add a check for the filetype and only double it if it's ova.

daisy_workflows/image_import/import_image.sh
EricEdens: ahh, that's a good point. switched to `/daisy-scratch/`

daisy_workflows/image_import/import_image.sh
EricEdens: :)

daisy_workflows/image_import/import_image.sh
EricEdens: I find it helpful for waypointing, but you have a good point about maintenance. So, yeah, I'm good with removing them.

daisy_workflows/image_import/import_image.sh
```diff
@@ -21,9 +21,10 @@ SOURCE_DISK_FILE="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/source_d
 SOURCEURL=${SOURCE_DISK_FILE}
 SOURCEBUCKET="$(echo ${SOURCEURL} | awk -F/ '{print $3}')"
 SOURCEPATH="${SOURCEURL#"gs://"}"
-SOURCESIZE_BYTES="$(gsutil du ${SOURCEURL} | grep -o '^[0-9]\+')"
+SOURCE_FILE_TYPE="${SOURCEURL##*.}"
+SOURCE_SIZE_BYTES="$(gsutil du ${SOURCEURL} | grep -o '^[0-9]\+')"
 DISKNAME="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/disk_name)"
-STAGING_DISKNAME="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/staging_disk_name)"
+SCRATCH_DISKNAME="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/scratch_disk_name)"
```
zoran15: variable naming is not consistent: we have old variables without underscores and new that are a mix of underscores and no underscores. I have no preference for either but it should be consistent across the file

daisy_workflows/image_import/import_image.sh
```diff
@@ -112,10 +114,10 @@ if ! out=$(gcloud -q compute instances attach-disk ${ME} --disk=${DISKNAME} --zo
 fi
 echo ${out}
 
-# Fourth, decompress the image and write it to the disk referenced by $DISKNAME.
+# Decompress the image and write it to the disk referenced by $DISKNAME.
```
zoran15: Inflate? Convert?

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,22 +53,48 @@ function resizeDisk() {
   fi
 }
 
-# Mount GCS bucket containing the disk image.
-mkdir -p /gcs/${SOURCEBUCKET}
-gcsfuse --implicit-dirs ${SOURCEBUCKET} /gcs/${SOURCEBUCKET}
+function copyImageToScratchDisk() {
+  local sourceSizeGigabytes="$((1 + SOURCE_SIZE_BYTES / 1024 / 1024 / 1024))"
```
dntczdx: ref to export_disk_ext.sh about how to calculate size_gb. I think this line always grant one extra GB if source disk is a size of exact X GB.

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,22 +53,48 @@ function resizeDisk() {
...
   SOURCEPATH="${DAISY_SOURCE_URL#"gs://"}/${VMDK}"
   echo "Import: New source file is ${VMDK}"
 fi
 
+# Ensure the output disk has sufficient space to accept the disk image.
 # Disk image size info.
-SIZE_BYTES=$(qemu-img info --output "json" /gcs/${SOURCEPATH} | grep -m1 "virtual-size" | grep -o '[0-9]\+')
+SIZE_BYTES=$(qemu-img info --output "json" /daisy-scratch/${SOURCEPATH} | grep -m1 "virtual-size" | grep -o '[0-9]\+')
  # Round up to the next GB.
 SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
```
dntczdx: ref to export_disk_ext.sh about how to calculate size_gb
We may fix this by the way? Now it almost always grant extra spaces because 1GB=1073741824>1000000000

EricEdens: done!

daisy_workflows/image_import/import_image.sh
EricEdens: done!

daisy_workflows/image_import/import_image.sh
EricEdens: changed to convert

daisy_workflows/image_import/import_image.sh
EricEdens: Sounds good -- all of the variables that I added now use underscores

daisy_workflows/image_import/import_image.sh
```diff
@@ -15,13 +15,17 @@
...
 URL="http://metadata/computeMetadata/v1/instance"
 DAISY_SOURCE_URL="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/daisy-sources-path)"
 SOURCE_DISK_FILE="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/source_disk_file)"
 SOURCEURL=${SOURCE_DISK_FILE}
 SOURCEBUCKET="$(echo ${SOURCEURL} | awk -F/ '{print $3}')"
 SOURCEPATH="${SOURCEURL#"gs://"}"
+SOURCE_FILE_TYPE="${SOURCEURL##*.}"
+SOURCE_SIZE_BYTES="$(gsutil du ${SOURCEURL} | grep -o '^[0-9]\+')"
 DISKNAME="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/disk_name)"
+SCRATCH_DISK_NAME="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/scratch_disk_name)"
```
hopkiw: should all metadata variables be at the top, together?

daisy_workflows/image_import/import_image.sh
```diff
@@ -15,13 +15,17 @@
 
 set -x
 
+BYTES_1GB=1073741824
 URL="http://metadata/computeMetadata/v1/instance"
 DAISY_SOURCE_URL="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/daisy-sources-path)"
 SOURCE_DISK_FILE="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/source_disk_file)"
 SOURCEURL=${SOURCE_DISK_FILE}
```
hopkiw: makes SOURCE_DISK_FILE unused; consider merging with above line so that SOURCEURL=$(curl...)

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,24 +54,50 @@ function resizeDisk() {
...
+  mkfs.ext4 /dev/sdb
+  mount /dev/sdb "${scratchPath}"
+  if [[ $? -ne 0 ]]; then
+    echo "ImportFailed: Failed to prepare scratch disk."
+  fi
+  gsutil cp "${SOURCEURL}" "${scratchPath}"
+  echo "Import: Copied image from ${SOURCEURL} to ${scratchPath}"
+}
+
+copyImageToScratchDisk
```
hopkiw: why define a function only to immediately call it once?

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,24 +54,50 @@ function resizeDisk() {
...
   echo "Import: New source file is ${VMDK}"
 fi
 
+# Ensure the output disk has sufficient space to accept the disk image.
 # Disk image size info.
-SIZE_BYTES=$(qemu-img info --output "json" /gcs/${SOURCEPATH} | grep -m1 "virtual-size" | grep -o '[0-9]\+')
+SIZE_BYTES=$(qemu-img info --output "json" /daisy-scratch/${SOURCEPATH} | grep -m1 "virtual-size" | grep -o '[0-9]\+')
  # Round up to the next GB.
-SIZE_GB=$(awk "BEGIN {print int((${SIZE_BYTES}/1000000000)+ 1)}")
+SIZE_GB="$((1 + (SOURCE_SIZE_BYTES - 1) / BYTES_1GB))"
```
hopkiw: why this change? arithmetic expansion only supports integers, so division cases are risky.

daisy_workflows/image_import/import_image.sh
```diff
@@ -15,13 +15,17 @@
...
 set -x
 
+BYTES_1GB=1073741824
 URL="http://metadata/computeMetadata/v1/instance"
 DAISY_SOURCE_URL="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/daisy-sources-path)"
 SOURCE_DISK_FILE="$(curl -f -H Metadata-Flavor:Google ${URL}/attributes/source_disk_file)"
 SOURCEURL=${SOURCE_DISK_FILE}
 SOURCEBUCKET="$(echo ${SOURCEURL} | awk -F/ '{print $3}')"
 SOURCEPATH="${SOURCEURL#"gs://"}"
+SOURCE_FILE_TYPE="${SOURCEURL##*.}"
```
hopkiw: np: `SOURCE_FILE_EXT`

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,24 +54,50 @@ function resizeDisk() {
   fi
 }
 
-# Mount GCS bucket containing the disk image.
-mkdir -p /gcs/${SOURCEBUCKET}
-gcsfuse --implicit-dirs ${SOURCEBUCKET} /gcs/${SOURCEBUCKET}
+function copyImageToScratchDisk() {
+  local sourceSizeGigabytes="$((1 + (SOURCE_SIZE_BYTES - 1) / BYTES_1GB))"
+  local scratchDiskSizeGigabytes=sourceSizeGigabytes
```
hopkiw: also unnecessary second variable

daisy_workflows/image_import/import_image.sh
```diff
@@ -48,24 +54,50 @@ function resizeDisk() {
...
+     scratchDiskSizeGigabytes=$((sourceSizeGigabytes * 2))
+  fi
+
+  # This disk is initially created with 10GB of space.
+  # Enlarge it if that's insufficient to hold the input image.
+  if [[ ${scratchDiskSizeGigabytes} -gt 10 ]]; then
+    resizeDisk "${SCRATCH_DISK_NAME}" "${scratchDiskSizeGigabytes}" "${ZONE}"
+  fi
+
+  local scratchPath="/daisy-scratch/$(dirname ${SOURCEPATH})/"
```
hopkiw: i don't understand. why do we care about the dir part of the path to source disk file here. why not put the disk file we're touching to a predefined path eg /daisy-scratch/sourcedisk.ext , where we'd know to find it in the tar and convert steps below?

daisy_workflows/image_import/import_image.sh
EricEdens: > why this change?

@dntczdx's asked for it earlier (one billion bytes isn't a gigabyte, and by not subtracting one we will allocate an extra gigabyte to the image if the input image was a multiple of one gigabyte already)

> arithmetic expansion only supports integers, so division cases are risky.

Interesting, risky in what sense? (The previous usage of `int` performs the same rounding that is used here)

daisy_workflows/image_import/import_image.sh
EricEdens: I agree that since the subsequent code is outside of a function this looks a bit awkward. The benefits are limiting the scope of variables inside the function, as well as putting a name around a block of code. I think the benefits are worth it. But I could be convinced otherwise!

If I weren't so averse to refactoring hard-to-test bash scripts, I'd break the subsequent block of code into a couple of functions and then the top-level script would be: `copyImageToScratchDisk && checkAndInflateOva && writeImageToDisk`

daisy_workflows/image_import/import_image.sh
EricEdens: There's subsequent logic that assumes the image will be in this path. That was the motivation I'd say.

daisy_workflows/image_import/import_image.sh
EricEdens: I was hoping to add clarity of the size of the image versus the size of the disk, but now that you point it out, I agree that it'd be clearer to consolidate it!

daisy_workflows/image_import/import_image.sh
EricEdens: sounds good to me!

daisy_workflows/image_import/import_image.sh
EricEdens: sounds good to me!

daisy_workflows/image_import/import_image.sh
EricEdens: sounds good to me!

daisy_workflows/image_import/import_image.sh
hopkiw: we control both ends, though. if there's not a need to use the uploaded filename locally, my suggestion is to simplify

daisy_workflows/image_import/import_image.sh
hopkiw: yes, but the cast to integer happens after the division operation, so it's safe bc it's being applied intentionally and as a step. with bash/posix arithmetic, it simply doesn't support floats, so you get undesired behavior easily. it's maybe just an aversion to bash arithmetic developed over my time as a bash dev

daisy_workflows/image_import/import_image.sh
hopkiw: makes sense. i think you'd be supported in rewriting a script of this length if you wanted, but up to you.

dntczdx: @hopkiw This is a interesting topic actually. I faced similar situations - I want to write a sub script, but there isn't a good way to copy it to the machine (not like startup script). I will have to download it before calling it. Do you have a better idea?

daisy_workflows/image_import/import_image.sh
dntczdx: @hopkiw I am a bit confused. This calculation doesn't rely on float. Oppositely, it intentionally rely on how int is divided. In which case it works unexpected?

daisy_workflows/image_import/import_image.sh
dntczdx: It prevents multiple scripts using duplicate path. The machine can be reused, if we want.

PR #897, opened 2019-08-21
Fix Debian 9 import failing if gnupg is not installed

zoran15:
Fix Debian 9 import failing if gnupg is not installed with "gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation" error.

### DIFF
--- a/daisy_workflows/image_import/debian/translate.py
+++ b/daisy_workflows/image_import/debian/translate.py
@@ -50,7 +50,7 @@ def DistroSpecific(g):
     logging.info('Installing GCE packages.')
 
     g.command(['apt-get', 'update'])
-    g.command(['apt-get', 'install', '-y', 'gnupg'])
+    g.sh('DEBIAN_FRONTEND=noninteractive apt-get install --assume-yes gnupg')
 
     g.command(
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/debian/translate.py
```diff
@@ -48,6 +48,10 @@ def DistroSpecific(g):
 
   if install_gce == 'true':
     logging.info('Installing GCE packages.')
+
+    g.command(['apt-get', 'update'])
+    g.command(['apt-get', 'install', '-y', 'gnupg'])
```
zmarano: g.sh('DEBIAN_FRONTEND=noninteractive apt-get install --assume-yes gnupg)
This is to avoid any user input requests that may come up with different environments.

zoran15: Is g.command(['apt-get', 'update']) still needed?

zmarano: Yes, since this is happening before the next update (and repo additions).

zoran15: Done

PR #901, opened 2019-08-23
Log Cloud Build ID in import/export wrappers to help with debugging

zoran15:


cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -95,8 +95,11 @@ func NewOVFImporter(params *ovfimportparams.OVFImportParams) (*OVFImporter, erro
 	}
 	tarGcsExtractor := storageutils.NewTarGcsExtractor(ctx, storageClient, logger)
 	buildID := os.Getenv("BUILD_ID")
+
 	if buildID == "" {
```
dntczdx: why does ovf importer use a random ID as fallback while others don't?

zoran15: IIRC, it was for testing resource labelling when run from local.

PR #903, opened 2019-08-27
[Logging] return disk/image/file size from import/export script

dntczdx:
We want to log target/source disk/image/file size for import/export for usage/perf analysis.
We need to expose the value from .sh script. Thus, using a string pattern is almost the only way. I defined this pattern: <serial-output key:'(.*)' value:'(.*)'>
key-value pair in this pattern will be catched and recorded in workflow, and thus can be stored to logging.

### DIFF
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -152,7 +152,9 @@ type Workflow struct {
 	targetInstances *targetInstanceRegistry
 	objects         *objectRegistry
 
-	stepTimeRecords []TimeRecord
+	stepTimeRecords             []TimeRecord
+	serialControlOutputValues   map[string]string
+	serialControlOutputValuesMx sync.Mutex
 }
 
 //DisableCloudLogging disables logging to Cloud Logging for this workflow.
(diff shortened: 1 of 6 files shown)


daisy/workflow.go
```diff
@@ -153,6 +153,7 @@ type Workflow struct {
 	objects         *objectRegistry
 
 	stepTimeRecords []TimeRecord
+	outputValues    map[string]string
```
zoran15: A more descriptive name would be serialConsoleOutputValues.

Also, since there could be multiple VMs for which we're capturing serial console output (and could thus return serial console output values), information about VM is lost here. We might not need the info about which step/VM returned this value, but since Daisy is a generic framework, some other client might need this info.

daisy/workflow.go
```diff
@@ -254,6 +263,11 @@ func (w *Workflow) RunWithModifiers(
 		w.LogWorkflowInfo("Error running workflow: %v", err)
 		return err
 	}
+	if w.outputValues != nil {
+		for k, v := range w.outputValues {
+			w.LogWorkflowInfo("Serial-output value -> %v:%v", k, v)
```
zoran15: I assume, in import/export wrappers, we will read w.outputValues directly and log them to sawmill? If so, this output is just for logging purposes so it's visible in main Daisy log, not only VM serial console output?

If we want to read w.outputValues in wrappers, it should be exported or a getting func provided in workflow.go

daisy/workflow.go
```diff
@@ -254,6 +263,11 @@ func (w *Workflow) RunWithModifiers(
 		w.LogWorkflowInfo("Error running workflow: %v", err)
 		return err
 	}
+	if w.outputValues != nil {
```
zoran15: I believe nil check is not needed when range is used.

daisy_workflows/export/export_disk.sh
```diff
@@ -13,12 +13,18 @@
...
 URL="http://metadata/computeMetadata/v1/instance/attributes"
 GCS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/gcs-path)
 LICENSES=$(curl -f -H Metadata-Flavor:Google ${URL}/licenses)
 
 mkdir ~/upload
 
+# Source disk size info.
+SOURCE_SIZE_BYTES=$(lsblk /dev/sdb --output=size -b | sed -n 2p)
+SOURCE_SIZE_GB=$(awk "BEGIN {print int(((${SOURCE_SIZE_BYTES}-1)/${BYTES_1GB}) + 1)}")
+echo "GCEExport: <serial-output key:'source-size-gb' value:'${SOURCE_SIZE_GB}'>"
```
zoran15: Extract <serial-output> formatting to a func to avoid code duplication and reduce introducing errors.

In other scripts as well

PR #908, opened 2019-08-29
Allocate a scratch disk when importing OVFs.

EricEdens:
This fixes an error in the OVF tests:
```
 - "[OVFImportTests] [OVFImport] Windows 2012 R2 two disks": error while performing OVF import: step "wait-for-data-disk-1-signal" run error: WaitForInstancesSignal FailureMatch found for "data-disk-importer-1-import-ovf-33v2j": "ImportFailed: Failed to prepare scratch disk.'"
```

To confirm the fix, I ran:

```
go run gce_ovf_import_tests/main.go -test_project_id edens-test -test_zone us-central1-a
```

### DIFF
--- a/cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
+++ b/cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
@@ -48,7 +48,7 @@ func AddDiskImportSteps(w *daisy.Workflow, dataDiskInfos []ovfutils.DiskInfo) {
 
 		setupDataDiskStepName := fmt.Sprintf("setup-data-disk-%v", dataDiskIndex)
 		diskImporterDiskName := fmt.Sprintf("disk-importer-%v", dataDiskIndex)
-		scratchDiskDiskName := fmt.Sprintf("disk-importer-scratch-%s-%v", w.ID(), dataDiskIndex)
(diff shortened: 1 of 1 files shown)


cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
```diff
@@ -48,6 +48,7 @@ func AddDiskImportSteps(w *daisy.Workflow, dataDiskInfos []ovfutils.DiskInfo) {
 
 		setupDataDiskStepName := fmt.Sprintf("setup-data-disk-%v", dataDiskIndex)
 		diskImporterDiskName := fmt.Sprintf("disk-importer-%v", dataDiskIndex)
+		scratchDiskDiskName := fmt.Sprintf("disk-importer-scratch-%s-%v", w.ID(), dataDiskIndex)
```
zoran15: Not using ExactName: true would remove the need for manually adding workflow ID as it would be added automatically by Daisy. E.g. 'disk-importer-2-import-ovf-7mn7h' was created from diskImporterDiskName above even  though only 'disk-importer-2' was specified. 

ExactName: true should be used for resources that shouldn't include any temp prefix/suffix, such as disks that will be permanently attached to a created instance. Scratch disk is not one of them.

EricEdens: Cool, good feedback!

PR #909, opened 2019-09-03
Add function to gather kubernetes logs

YangLu1031:
For GKE scenario, customers need kubernetes logs, so gather kubelet.log, kube-proxy.log and docker.log under "C:\etc\kubernetes\logs" directory


cli_tools/diagnostics/main_windows.go
```diff
@@ -168,24 +167,48 @@ func gatherProgramLogs(logs chan logFolder, errs chan error) {
...
+	roots := []string{`C:\Windows\System32\winevt\Logs`}
+	filePaths, ers := collectFilePaths(roots)
+	for _, err := range ers {
+		errs <- err
+	}
+	logs <- logFolder{"Event", filePaths}
+}
+
+func gatherKubernetesLogs(logs chan logFolder, errs chan error) {
+	roots := []string{`C:\etc\kubernetes\logs`, `C:\Windows\MEMORY.dump`}
```
dntczdx: extract all these paths as constants?

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,303 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
dntczdx: 2019

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,303 @@
...
+		outputFileName  string
+		cmdProducesFile bool
+	}
+	tests := []struct {
+		name        string
+		fields      fields
+		wantOutPath string
+		wantErr     bool
+	}{
+		// TODO: Add test cases.
```
dntczdx: a bit confused, is this a incomplete PR?

cli_tools/diagnostics/main_windows.go
```diff
@@ -168,24 +167,48 @@ func gatherProgramLogs(logs chan logFolder, errs chan error) {
...
-		dir := dirs[0]
-		dirs = dirs[1:]
-		files, _ := ioutil.ReadDir(dir)
-		for _, f := range files {
-			path := filepath.Join(dir, f.Name())
-			if f.IsDir() {
-				dirs = append(dirs, path)
-			} else {
-				paths = append(paths, path)
+func collectFilePaths(roots []string) ([]string, []error) {
```
pjh: Please describe the method and its arguments and return value. Same for other methods you've added or modified.

cli_tools/diagnostics/main_windows.go
pjh: The memory dump file could quite large - I found some old ones that I still have around and they're 250-870 MB. Will that be a problem for this tool?

If you need to generate a crash dump for testing purposes you can use the NotMyFault tool from Sysinternals: https://docs.microsoft.com/en-us/sysinternals/downloads/notmyfault

YangLu1031: Good to know, will test if the tool works fine on big file dump

cli_tools/diagnostics/main_windows_test.go
YangLu1031: removed the unnecessary test template

cli_tools/diagnostics/main_windows.go
YangLu1031: Did some experiments:
1. For 500MB memory dump, after zipped, got 160MB resulting file, taking 3mins to show up on GCS.
2. Faked a 1G memory dump based on 200MB using (cat MEMORY.dmp >> MEMORY.dump), got 140MB resulting file, taking 2mins to show up.
3. Tried to upload 1G file to GCS directly thru gsutil, taking 20s. 
Looks working good.

YangLu1031:
@pjh @dntczdx Please take another look

pjh:
cc @ning-yang

cli_tools/diagnostics/main_windows.go
```diff
@@ -26,6 +25,12 @@ import (
 	"time"
 )
 
+const (
+	eventLogsRoot = `C:\Windows\System32\winevt\Logs`
+	k8sLogsRoot   = `C:\etc\kubernetes\logs`
+	crashDump     = `C:\Windows\MEMORY.dmp`
```
ning-yang: NIT: user can change the dump path: https://support.microsoft.com/en-us/help/254649/overview-of-memory-dump-file-options-for-windows

it will be safe to read the path from Registry .  But I don't think most people will do that.

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,114 @@
...
+		roots []string
+	}
+	tests := []struct {
+		name  string
+		args  args
+		errOK func(error) bool
+	}{
+		{"Nil roots", args{nil}, nil},
+		{"Empty roots", args{[]string{""}}, pathNonExist},
+		{"Existing roots", args{[]string{k8sLogsRoot, eventLogsRoot}}, nil},
```
pjh: Is this going to fail on a Windows machine where we haven't configured Kubernetes and created C:\etc\kubernetes\logs?

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,114 @@
...
+		{"Non-existing paths", args{[]string{`C:\etc\kubernetes\logs\xxxx`}}, fileNonExist},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, gotErrs := collectFilePaths(tt.args.roots)
```
pjh: Should we have a test that actually checks that the paths returned are the ones we expect?

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,114 @@
...
+func TestGatherEventLogs(t *testing.T) {
+	type args struct {
+		logs chan logFolder
+		errs chan error
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{"GatherEventLogs", args{make(chan logFolder, 2), make(chan error)}},
```
pjh: This test seems to verify that GatherEventLogs does not return errors. That's good, but is it also possible to verify that it does actually gather any present event logs too?

cli_tools/diagnostics/main_windows.go
YangLu1031: Added to TODO

cli_tools/diagnostics/main_windows_test.go
YangLu1031: Created a dummy test folder "collectFilePathsTest" under %TEMP%, it will be deleted after testing.

cli_tools/diagnostics/main_windows_test.go
YangLu1031: Added verification that System event log file exist

YangLu1031:
@dntczdx @pjh please take another look

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,116 @@
...
+			for _, err := range gotErrs {
+				if tt.errOK == nil || !tt.errOK(err) {
+					t.Errorf("collectFilePaths() got unexpected error = %v", gotErrs)
+				}
+			}
+		})
+	}
+}
+
+func include(ss []string, str string) bool {
```
pjh: Either add a comment describing what this function does, or name it more specifically: stringArrayIncludesString ?

ss and str are also not great argument names; maybe "string_array" or "search_array", and "target"?

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,116 @@
...
+func TestGatherEventLogs(t *testing.T) {
+	type args struct {
+		logs chan logFolder
+		errs chan error
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{"GatherEventLogs", args{make(chan logFolder, 2), make(chan error)}},
```
pjh: Since there's only one test case, can you get rid of the for loop over tests? That would make this code much more straightforward.

This test verifies that gatherEventLogs correctly gathers an expected system log file. This is great, but it's not obvious that this is actually what the test does without looking very closely. Can you either add a comment explaining this, or (even better) name this function "TestGatherLogsGathersExpectedSystemLogFile"?

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}{
+		{"Nil roots", args{nil}, []string{}, nil},
+		{"Empty roots", args{[]string{""}}, []string{}, pathNonExist},
+		{"Existing roots", args{[]string{testRoot}}, []string{kubeletlogFilePath}, nil},
+		{"Non-existing paths", args{[]string{nonExistFilePath}}, []string{}, fileNonExist},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotFiles, gotErrs := collectFilePaths(tt.args.roots)
+			if !reflect.DeepEqual(gotFiles, tt.want) {
```
pjh: Thanks for updating this, looks much better!

cli_tools/diagnostics/main_windows_test.go
YangLu1031: Agree. And the golang suggest mixedCap naming convention, so change to stringArray. https://golang.org/doc/effective_go.html

YangLu1031:
@pjh please take another look :)

cli_tools/diagnostics/main_windows_test.go
pjh: Nice, looks much better!

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,118 @@
...
+	test := struct {
+		name string
+		args args
+	}{
+		"GatherEventLogs",
+		args{make(chan logFolder, 2), make(chan error)},
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		go gatherEventLogs(test.args.logs, test.args.errs)
```
pjh: This is much easier to read, but still seems like it could be simpler. Are test.args.logs and test.args.errs really needed? Why not something like:

```
func TestGatherEventLogsGathersExpectedSystemLogFile(t *testing.T) {
    logFolderChannel = make(chan logFolder, 2)
    errorChannel = make(chan error)

	t.Run("TestGatherEventLogsGathersExpectedSystemLogFile", func(t *testing.T) {
		go gatherEventLogs(logFolderChannel, errorChannel)
		select {
		case l := <-logFolderChannel:
			if !stringArrayIncludesString(l.files, systemLogPath) {
				t.Errorf("Expect %s, but it's missing", systemLogPath)
			}
		case e := <-errorChannel:
			t.Errorf(e.Error())
		}
	})
}
```

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -0,0 +1,118 @@
...
+		select {
+		case l := <-test.args.logs:
+			if !stringArrayIncludesString(l.files, systemLogPath) {
+				t.Errorf("Expect %s, but it's missing", systemLogPath)
+			}
+		case e := <-test.args.errs:
+			t.Errorf(e.Error())
+		}
+	})
+}
```
pjh: Should we add a test that invokes the full `gatherLogs` function and checks that it doesn't return errors?

I'm still concerned that we're not specifically testing that with this PR, if we run the gcloud diagnose command on a Windows VM that doesn't have Kubernetes set up then the command will not fail.

cli_tools/diagnostics/main_windows.go
```diff
@@ -237,7 +276,7 @@ func gatherLogs(trace bool) ([]logFolder, error) {
 			break
 		}
 	}
-
+	// Note: errors are swallowed if error count <= gathterxxxLogs func count.
```
pjh: I'm a little wary about this but I'm not reading the code above in detail. If you think this behavior could be improved then please either add a TODO here or file an issue for this repository, describing what would be better to do instead.

cli_tools/diagnostics/main_windows_test.go
YangLu1031: Since we cannot mock the other gatherXXXLogs behavior, it will retrieve the real logs which takes 2 mins to finish, so better not put in UT. 
Will sync with Ning if we can add e2e test somewhere. Also will validate manually on VM without kubernetes setup to gain confidence.

cli_tools/diagnostics/main_windows_test.go
YangLu1031: The resulting test name is like [FunctionName]/[TestCaseName], so "TestGatherEventLogsGathersExpectedSystemLogFile" test case name seems too long.

Will revert function name to "TestGatherEventLogs" and put "Gathers Expected SystemLog File" as test case name. 
It will look like "TestGatherEventLogs/Gathers_Expected_SystemLog_File" in test logs

cli_tools/diagnostics/main_windows_test.go
YangLu1031: Validated manually, it works well for image without kubernetes setup.

YangLu1031:
@dntczdx Can you take another look if it's ok to merge? Thanks!

YangLu1031:
cc @adjackura

PR #910, opened 2019-09-04
Increase scratch disk to account for EXT4 overhead.

EricEdens:
Prior to this change, large image files would exceed the scratch disk, and cause the `gsutil cp` command to fail.

An additional fix is included to ensure that we don't exceed the logging daemon's buffer with the progress meter for `gsutil cp`.

For testing, I ran these integration tests:
 - image_export_vmdk_given_size_1024gb.wf.json
 - image_export_vmdk_given_size_210gb.wf.json

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -61,7 +61,7 @@ function copyImageToScratchDisk() {
   # the overhead is less than 10% using the default number of inodes
   # and a reserved block percentage of 1%. We conserve a little more space
   # by avoiding the reserved blocks via `mkfs.ext4 -m 0`.
-  local scratchDiskSizeGigabytes=$(awk "BEGIN {print int(${SOURCE_SIZE_GB} * 1.1)}")
+  local scratchDiskSizeGigabytes=$(awk "BEGIN {print int((${SOURCE_SIZE_GB} * 1.1) + 1)}")
   # We allocate double capacity for OVA, which would
   # require making an additional copy of its enclosed VMDK.
   if [[ "${IMAGE_PATH}" =~ \.ova$ ]]; then
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -56,7 +56,12 @@ function resizeDisk() {
 }
 
 function copyImageToScratchDisk() {
-  local scratchDiskSizeGigabytes=${SOURCE_SIZE_GB}
+  # We allocate an extra 10% capacity to account for ext4's
+  # filesystem overhead. According to https://petermolnar.net/why-use-btrfs-for-media-storage/ ,
+  # the overhead is less than 10% using the default number of inodes
+  # and a reserved block percentage of 1%. We conserve a little more space
+  # by avoiding the reserved blocks via `mkfs.ext4 -m 0`.
+  local scratchDiskSizeGigabytes=$(awk "BEGIN {print int(${SOURCE_SIZE_GB} * 1.1)}")
   # We allocate double capacity for OVA, which would
   # require making an additional copy of its enclosed VMDK.
   if [[ "${IMAGE_PATH}" =~ \.ova$ ]]; then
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -56,7 +56,12 @@ function resizeDisk() {
 }
 
 function copyImageToScratchDisk() {
-  local scratchDiskSizeGigabytes=${SOURCE_SIZE_GB}
+  # We allocate an extra 10% capacity to account for ext4's
+  # filesystem overhead. According to https://petermolnar.net/why-use-btrfs-for-media-storage/ ,
+  # the overhead is less than 10% using the default number of inodes
+  # and a reserved block percentage of 1%. We conserve a little more space
+  # by avoiding the reserved blocks via `mkfs.ext4 -m 0`.
+  local scratchDiskSizeGigabytes=$(awk "BEGIN {print int(${SOURCE_SIZE_GB} * 1.1)}")
```
dntczdx: do we really need as much as 10%? If so, shall we handle the rounding issue? A couple of things to consider:
1. We can't create a disk less than 10GB. Shall we set 10GB as lower bound?
2. For example, a 18GB source image will cause a 19GB of scratch disk due to rounding, which is significantly less than 10%. Need to round up always.

daisy_workflows/image_import/import_image.sh
```diff
@@ -72,13 +77,24 @@ function copyImageToScratchDisk() {
...
   fi
-  gsutil cp "${SOURCE_URL}" "${IMAGE_PATH}"
-  echo "Import: Copied image from ${SOURCE_URL} to ${IMAGE_PATH}"
+
+  # Standard error for `gsutil cp` contains a progress meter that when written
+  # to the console will exceed the logging daemon's buffer for large files.
+  # The end of the stream may contain useful debugging messages, hence the
+  # usage of `tail` when we echo the error inside this block.
+  if ! out=$(gsutil cp "${SOURCE_URL}" "${IMAGE_PATH}" 2> gsutil.cp.err); then
+    echo "ImportFailed: Failed to download image to scratch [Privacy-> from ${SOURCE_URL} to ${IMAGE_PATH}, error: $(tail gsutil.cp.err) <-Privacy]"
```
dntczdx: There is a problem here, I think: "StatusMatch", "SuccessMatch" and "FailureMatch" only matches single line. "tail" may give multiple lines.

EricEdens: Interesting. In my testing I thought it was working, but looking at it again, it's not. I'll iterate over the lines in the error log and explicitly print the ones that I want to keep.

daisy_workflows/image_import/import_image.sh
EricEdens: Based on https://petermolnar.net/why-use-btrfs-for-media-storage/ , 10% overhead is sufficient. Good point on rounding up -- I'll add that.

EricEdens:
> What about other file system formats? Will the extra of 10% cover all other formats customers might use?

Yeah, good question. This is specifically for the scratch disk which is using the `ext4` file system, and is initialized here: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/a1a55c8f0d3b2117c0c135dbab9b8aa6c3373ef1/daisy_workflows/image_import/import_image.sh#L83

So it's making room for the inode table, journal, and other [book keeping items](https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout) used by ext4. 

The image coming from the customer probably has its own file system, but it's compressed and baked into the image's file, and will later get inflated by qemu-img onto /dev/sdc.

daisy_workflows/image_import/import_image.sh
dntczdx: As said in the last comment, 10GB is the lower bound of google cloud. Would you test what happens if we try create a <10GB disk?

EricEdens: That's already handled here, right? Or were you thinking of something else?

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/dcd98d020cd82f0ef2edb9e90194b5fb08868589/daisy_workflows/image_import/import_image.sh#L73

dntczdx: Cool! I missed it

PR #914, opened 2019-09-09
Refactor: Move ovf programmatic unit parsing to its own file.

EricEdens:
In addition, converts the existing unit test into a table-based test.

This is in preparation for fixing
https://groups.google.com/d/msg/gce-discussion/k0LKjsH3dw8/Z5q_yv4nAAAJ

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units.go
+++ b/cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units.go
@@ -1,0 +1,49 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package ovfutils
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+)
+
+// Returns capacity in GB taking into account allocation units which should be in `bytes * 2^x`
+// format. If capacity and allocationUnits are valid, returns at least 1 even if given capacity
+// is less than 1GB
+func getCapacityInGB(capacity string, allocationUnits string) (int, error) {
+	capacityRaw, err := strconv.Atoi(capacity)
+	if err != nil {
+		return 0, err
+	}
+	allocationUnitPower, err := getAllocationUnitPowerOfTwo(allocationUnits)
+	if err != nil {
+		return 0, err
+	}
+	allocationUnitPowerToGB := float64(allocationUnitPower) - 30.0
+	allocationUnitFactorToGB := math.Pow(2.0, allocationUnitPowerToGB)
+	capacityInGB := float64(capacityRaw) * allocationUnitFactorToGB
+
+	return int(math.Ceil(capacityInGB)), nil
+}
+
+func getAllocationUnitPowerOfTwo(allocationUnits string) (int, error) {
+	allocationUnits = strings.ToLower(allocationUnits)
+	if !strings.HasPrefix(allocationUnits, "byte * 2^") {
+		return 0, fmt.Errorf("can't parse `%v` disk allocation units", allocationUnits)
+	}
+	return strconv.Atoi(strings.TrimSpace(strings.TrimPrefix(allocationUnits, "byte * 2^")))
+}
--- a/cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
+++ b/cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
@@ -1,0 +1,64 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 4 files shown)


cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Parse error due to allocation units not being recognized.
+	{0, "1024", "megabytes",
+		errors.New("can't parse `megabytes` disk allocation units")},
+	{0, "1024", "mb",
+		errors.New("can't parse `mb` disk allocation units")},
+}
+
+func TestGetCapacityInGB(t *testing.T) {
+	for i := range getCapacityInGBTests {
```
dntczdx: can we do 'for i, test := range getCapacityInGBTests'?

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
```diff
@@ -0,0 +1,64 @@
...
+		test := &getCapacityInGBTests[i]
```
dntczdx: AFAIK, in most of the cases, "&" won't make it faster, even for a struct. instead, it is slower. Use it only when we have to share data

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"errors"
+	"reflect"
+	"testing"
+)
+
+type getCapacityInGBTest struct {
+	expected        int
+	capacity        string
+	allocationUnits string
+	err             error
```
dntczdx: expectedErr?

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
```diff
@@ -0,0 +1,64 @@
...
+		errors.New("can't parse `megabytes` disk allocation units")},
+	{0, "1024", "mb",
+		errors.New("can't parse `mb` disk allocation units")},
+}
+
+func TestGetCapacityInGB(t *testing.T) {
+	for i := range getCapacityInGBTests {
+		test := &getCapacityInGBTests[i]
+		capacityInGB, err := getCapacityInGB(test.capacity, test.allocationUnits)
+		if capacityInGB != test.expected || !reflect.DeepEqual(err, test.err) {
```
dntczdx: is DeepEqual required? If so, we may just compare err.Error()

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
EricEdens: sgtm

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
EricEdens: fixed by using your above suggestion.

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
EricEdens: DeepEqual came from an example I followed in go's source:

https://github.com/golang/go/blob/b9596aea50a0703f89c6f11c206cfd2c7dd189fa/src/strconv/atoi_test.go#L483

If I replace it with err != test.expectedErr, then we get false positives when one is non-nil.

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
EricEdens: nice

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
dntczdx: it seems that we are trying to prevent DeepEqual for error: https://github.com/golang/lint/issues/232
Up to you to determine whether this way is fine

dntczdx:
LGTM but ZOran may comment more since he owns it

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
```diff
@@ -0,0 +1,63 @@
...
+	{1, "1", "byte * 2^20", nil},
+	{1, "1024", "byte * 2^20", nil},
+	{5 * 1024, "5242880", "byte * 2^20", nil},
+
+	// TB
+	{1024, "1", "byte * 2^40", nil},
+	{5242880 * 1024, "5242880", "byte * 2^40", nil},
+
+	// Parse error due to allocation units not being recognized.
+	{0, "1024", "megabytes",
```
zoran15: What's the purpose of adding this test when we need to support "megabytes" anyway?

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units.go
```diff
@@ -0,0 +1,49 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
```
zoran15: Why "programmatic_units" for the file name?

As is, the name for this file should be "ovf_unit_parser" or similar.

EricEdens: It's the vernacular used in the OVF 2.x spec:

```
Values for ovf:capacityAllocationUnits shall match the format for
programmatic units defined in DSP0004 with the restriction that
the base unit shall be "byte".
```


https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf

cli_tools/gce_ovf_import/ovf_utils/ovf_programmatic_units_test.go
EricEdens: It's a habit for working in unfamiliar code: introduce additional test coverage to help with my understanding and to increase overall coverage. The subsequent change should break this test, which is good, since it's exactly where we'll be modifying the code.

PR #921, opened 2019-09-17
OVF import E2E tests: override Build ID

zoran15:
OVF import E2E tests: override Build ID (which is used for temp GCS path) as it appears OVF importer has been picking up "BUILD_ID" env variable (which works in Cloud Build) from Kubernetes environment resulting in shared temporary GCS path for extracted OVFs and overwritten OVF resources between different tests.

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
+++ b/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
@@ -74,10 +74,12 @@ type OVFImporter struct {
 	zoneValidator         domain.ZoneValidatorInterface
 	gcsPathToClean        string
 	workflowPath          string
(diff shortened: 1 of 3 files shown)


dntczdx:
It sounds not correct to use BUILD_ID to determine gcs path. Even for a real user, the tool can be executed outside of cloud build env, so the conflict can happen to them

zoran15:
> It sounds not correct to use BUILD_ID to determine gcs path. Even for a real user, the tool can be executed outside of cloud build env, so the conflict can happen to them

We use a random generated string to substitute BUILD_ID if it's not present. The issue that is fixed here is that BUILD_ID is present and we run imports in parallel resulting in shared temp path and files being overwritten.

PR #923, opened 2019-09-17
gce_image_publish: set optional ObsoleteDate

adjackura:


cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -60,7 +60,9 @@ type Publish struct {
 	// 24h*7*4 = ~1 month
 	// 24h*365 = ~1 year
 	DeleteAfter string `json:",omitempty"`
-	expiryDate  *time.Time
+	// Optional DeprecationStatus.Obsolete entry for the image (RFC 3339).
+	ObsoleteDate *time.Time `json:",omitempty"`
+	expiryDate   *time.Time
```
rofuentes: Since we will be setting this per image version, lets move this to the Image struct rather than keeping it here.

PR #924, opened 2019-09-17
Add debug logs prior to using dynamically-sized disks.

EricEdens:
We've seen intermittent disk-full failures when using /dev/sdb and /dev/sdc in import_image.sh. These debug logs will help to diagnose the issue.

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -86,6 +86,12 @@ function copyImageToScratchDisk() {
     echo "ImportFailed: Failed to prepare scratch disk."
   fi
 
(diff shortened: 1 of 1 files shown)


dntczdx:
Can we repro? It seems not ideal to publish this to prod just for debugging

EricEdens:
> Can we repro? It seems not ideal to publish this to prod just for debugging

I agree that it's not ideal.

I can only repro the issue without debug logs (and I've been trying for two weeks). So, either these calls are mysteriously fixing the problem, or the issue will happen and we'll get some more debug info.

In any case, this is a first step, and I'd still like to root cause and fix the issue.

PR #927, opened 2019-09-18
test-infra: make sure artifact directory is set in environmnet

adjackura:


dntczdx:
Lack of background here - what happens without this setting?

adjackura:
> Lack of background here - what happens without this setting?

Check the tests, the large gap yesterday were because of another change to fix unittetss, all other tests were only working because everything happened to be setup the same. This change ensures that artifacts exist where they are expected to be.

PR #928, opened 2019-09-18
Support VirtualBox ovfs

EricEdens:
Two changes were required here, and each has its own commit:

1. Update the govmomi schema to support `StorageItem`, which is the node where virtualbox defines to which controller a disk is attached. (vmware puts that info in an `Item` node)
2. Update `ovf_utils` to use the new node.

Testing:
- `go test ./...`
- `gofmt -d $(find . -type f -name "*.go")`
- `go run gce_ovf_import_tests/main.go -test_project_id edens-test -test_zone us-central1-a`

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
+++ b/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
@@ -276,11 +276,11 @@ func getDiskControllersPrioritized(virtualHardware *ovf.VirtualHardwareSection) 
 }
 
 func filterItemsByResourceTypes(virtualHardware *ovf.VirtualHardwareSection, resourceTypes ...uint16) []ovf.ResourceAllocationSettingData {
-	all := make([]ovf.ResourceAllocationSettingData, len(virtualHardware.Item))
-	copy(all, virtualHardware.Item)
+	allItems := make([]ovf.ResourceAllocationSettingData, len(virtualHardware.Item))
+	copy(allItems, virtualHardware.Item)
 
 	for _, storageItem := range virtualHardware.StorageItem {
-		all = append(all, ovf.ResourceAllocationSettingData{
+		allItems = append(allItems, ovf.ResourceAllocationSettingData{
 			CIMResourceAllocationSettingData: ovf.CIMResourceAllocationSettingData{
 				AddressOnParent: storageItem.AddressOnParent,
 				Caption:         storageItem.Caption,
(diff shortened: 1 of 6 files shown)


vendor/github.com/vmware/govmomi/ovf/cim.go
```diff
@@ -16,11 +16,14 @@ limitations under the License.
 
 package ovf
 
+import (
+	"github.com/vmware/govmomi/vim25/types"
```
zoran15: This is the same package we are vendoring here. This folder should be copied here (to the vendored package). I don't see it being added in this PR

Also, what exact types are used here? If you need just a struct or two, just copy it.

vendor/github.com/vmware/govmomi/ovf/cim.go
```diff
@@ -16,11 +16,14 @@ limitations under the License.
...
+)
+
 /*
 Source: http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2.24.0/CIM_VirtualSystemSettingData.xsd
 */
 
-// CIMVirtualSystemSettingData holds virtual system settings
```
zoran15: Why is public doc removed? This might break prechecks

vendor/github.com/vmware/govmomi/ovf/env.go
```diff
@@ -18,16 +18,12 @@ package ovf
 
 import (
 	"bytes"
-	"encoding/xml"
 	"fmt"
+
+	"github.com/vmware/govmomi/vim25/xml"
```
zoran15: Again, importing outside vendored package.

vendor/github.com/vmware/govmomi/ovf/env.go
```diff
@@ -47,39 +43,26 @@ const (
 	ovfEnvFooter         = `</Environment>`
 )
 
-// A Name represents an XML name (Local) annotated
-// with a name space identifier (Space).
-// In tokens returned by Decoder.Token, the Space identifier
-// is given as a canonical URL, not the short prefix used
-// in the document being parsed.
-type Name struct {
```
zoran15: These were added here so we don't have to import vim25 which contains 100s of thousands of limes. It's a bit hackish but reduces this vendored package significantly.

gce_ovf_import_tests/test_suites/ovf_import_test_suite.go
```diff
@@ -184,6 +184,25 @@ func TestSuite(
 			expectedMachineType:   "n1-standard-4",
 			expectedStartupOutput: "All tests passed!",
 		},
+		{
+			importParams: &ovfimportparams.OVFImportParams{
```
dntczdx: @zoran15 an irrelative topic to this PR: can we call the ovf import cli tool instead of lib in gce_ovf_import_tests? That is closer to the real user experience. I am not sure whether current implementation has its specific purpose.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -276,8 +276,25 @@ func getDiskControllersPrioritized(virtualHardware *ovf.VirtualHardwareSection)
 }
 
 func filterItemsByResourceTypes(virtualHardware *ovf.VirtualHardwareSection, resourceTypes ...uint16) []ovf.ResourceAllocationSettingData {
+	all := make([]ovf.ResourceAllocationSettingData, len(virtualHardware.Item))
```
zoran15: Minor: allItems or similar

gce_ovf_import_tests/test_suites/ovf_import_test_suite.go
zoran15: I think it would be a good idea to refactor these tests to use gcloud using latest import wrapper. That would be the closest to what customers can expect.

PR #930, opened 2019-09-19
Disable startup script e2e test validation for Debian 9

zoran15:
Disable startup script e2e test validation for Debian 9 for the time being. This image has SSH disabled which might be causing the issue. Added support for no startup script in the test runner.

### DIFF
--- a/gce_ovf_import_tests/test_suites/ovf_import_test_suite.go
+++ b/gce_ovf_import_tests/test_suites/ovf_import_test_suite.go
@@ -321,34 +321,35 @@ func runOvfImportTest(
 		return
 	}
 
-	if testSetup.startup != nil {
-		logger.Printf("[%v] Setting instance metadata with test startup script", testSetup.name)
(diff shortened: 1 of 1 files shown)


gce_ovf_import_tests/test_suites/ovf_import_test_suite.go
```diff
@@ -322,30 +321,34 @@ func runOvfImportTest(
 		return
 	}
 
-	logger.Printf("[%v] Setting instance metadata with test startup script", testSetup.name)
-	err = client.SetInstanceMetadata(testProjectConfig.TestProjectID, testProjectConfig.TestZone,
-		instanceName, &api.Metadata{Items: []*api.MetadataItems{testSetup.startup},
-			Fingerprint: instance.Metadata.Fingerprint})
+	if testSetup.startup != nil {
```
dntczdx: Minor: to be align with below coding style (if .... return), you may consider "if  testSetup.startup==nil ... Printf ... return"

zoran15: Changed

PR #934, opened 2019-09-20
[export wrapper] support simple image name

dntczdx:
[export wrapper] support simple image name.
Before: only support format like "global/images/anImage"
Now: support "anImage" as well, which will be extended to "global/images/anImage" automatically

### DIFF
--- a/cli_tools/gce_vm_image_export/exporter/exporter.go
+++ b/cli_tools/gce_vm_image_export/exporter/exporter.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/compute"
 	daisyutils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/daisy"
@@ -83,7 +84,7 @@ func buildDaisyVars(destinationURI string, sourceImage string, format string, ne
 
 	varMap["destination"] = destinationURI
 
-	varMap["source_image"] = sourceImage
+	varMap["source_image"] = getFullImagePath(sourceImage)
 
 	if format != "" {
 		varMap["format"] = format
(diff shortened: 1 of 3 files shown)


cli_tools/gce_vm_image_export/exporter/exporter.go
```diff
@@ -101,6 +102,14 @@ func buildDaisyVars(destinationURI string, sourceImage string, format string, ne
 	return varMap
 }
 
+func getFullImagePath(imageName string) string {
+	if !strings.Contains(imageName, "/") {
+		// Extend simple image name to full image name
+		return fmt.Sprintf("global/images/%v", imageName)
```
zoran15: Is this going to work with project specific images?

dntczdx: It's "current project image". User can specify any project as current project by --project=aProject.
If the image is from a different project from running project, user has to specify full image path - "projects/another-project/global/images/animage"

dntczdx:
I have 

> Can you please add e2e test for export wrapper to make sure this actually works as intended?

I have added unit test for it. It's too expensive to run a e2e for such a minor change in long term in my opinion. I have ran the wrapper manually to confirm it works.

zoran15:
> I have
> 
> > Can you please add e2e test for export wrapper to make sure this actually works as intended?
> 
> I have added unit test for it. It's too expensive to run a e2e for such a minor change in long term in my opinion. I have ran the wrapper manually to confirm it works.

What about modifying one of the existing tests to use short name?

dntczdx:
> > I have
> > > Can you please add e2e test for export wrapper to make sure this actually works as intended?
> > 
> > 
> > I have added unit test for it. It's too expensive to run a e2e for such a minor change in long term in my opinion. I have ran the wrapper manually to confirm it works.
> 
> What about modifying one of the existing tests to use short name?

Good idea, I used it in rich_param test due that it covers all unusual params

dntczdx:
Ping @zoran15

PR #935, opened 2019-09-20
Enable GuestOsFeatures for intermediate Windows disks

EricEdens:
We've seen some Windows images that aren't able to boot during the translate step. This change surfaces a parameter to CreateDisk that results in the disk getting the WINDOWS GuestOsFeature flag.

I've updated the import wrapper program to include this flag if the word 'windows' is in the translate script's name (I'm definitely open to other heuristics as well!).

Testing:

```
 go run compute-image-tools/cli_tools/gce_vm_image_import/main.go \
  -client_id=edens-test \
  -source_image=projects/ajackura-dev/global/images/customer-srv203-issue886 \
  -os=windows-2008r2 \
  -image_name=guest-os-test-7 \
  -project=edens-test
```

I'm currently running `go run gce_ovf_import_tests/main.go -test_project_id edens-test -test_zone us-central1-a`, and I'll post the results when they're finished.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -150,12 +150,17 @@ func buildDaisyVars(translateWorkflowPath, imageName, sourceFile, sourceImage, f
 	if translateWorkflowPath != "" {
 		varMap["translate_workflow"] = translateWorkflowPath
 		varMap["install_gce_packages"] = strconv.FormatBool(!noGuestEnvironment)
+		varMap["is_windows"] = strconv.FormatBool(strings.Contains(translateWorkflowPath, "windows"))
 	}
 	if sourceFile != "" {
 		varMap["source_disk_file"] = sourceFile
 	}
 	if sourceImage != "" {
-		varMap["source_image"] = fmt.Sprintf("global/images/%v", sourceImage)
+		if strings.Contains(sourceImage, "global/images") {
+			varMap["source_image"] = sourceImage
+		} else {
+			varMap["source_image"] = fmt.Sprintf("global/images/%v", sourceImage)
+		}
 	}
 	varMap["family"] = family
 	varMap["description"] = description
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -150,7 +150,7 @@ func buildDaisyVars(translateWorkflowPath, imageName, sourceFile, sourceImage, f
 	if translateWorkflowPath != "" {
 		varMap["translate_workflow"] = translateWorkflowPath
 		varMap["install_gce_packages"] = strconv.FormatBool(!noGuestEnvironment)
-		varMap["is_windows"] = strconv.FormatBool(strings.Contains(translateWorkflowPath, "windows"))
+		varMap["IsWindows"] = strconv.FormatBool(strings.Contains(translateWorkflowPath, "windows"))
 	}
 	if sourceFile != "" {
 		varMap["source_disk_file"] = sourceFile
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -150,7 +150,7 @@ func buildDaisyVars(translateWorkflowPath, imageName, sourceFile, sourceImage, f
 	if translateWorkflowPath != "" {
 		varMap["translate_workflow"] = translateWorkflowPath
 		varMap["install_gce_packages"] = strconv.FormatBool(!noGuestEnvironment)
-		varMap["IsWindows"] = strconv.FormatBool(strings.Contains(translateWorkflowPath, "windows"))
+		varMap["isWindows"] = strconv.FormatBool(strings.Contains(translateWorkflowPath, "windows"))
 	}
 	if sourceFile != "" {
 		varMap["source_disk_file"] = sourceFile
--- a/cli_tools/gce_vm_image_import/importer/importer_test.go
+++ b/cli_tools/gce_vm_image_import/importer/importer_test.go
@@ -247,56 +247,109 @@ func TestFlagsInvalidOS(t *testing.T) {
 	}
 }
 
-func TestBuildDaisyVarsFromDisk(t *testing.T) {
+func TestBuildDaisyVars(t *testing.T) {
 	resetArgs()
-	imageName = "image-a"
-	noGuestEnvironment = true
-	sourceFile = "source-file-path"
-	sourceImage = ""
-	family = "a-family"
-	description = "a-description"
-	network = "a-network"
-	subnet = "a-subnet"
-	region := "a-region"
 
-	got := buildDaisyVars("translate/workflow/path", imageName, sourceFile,
-		sourceImage, family, description, region, subnet, network, noGuestEnvironment)
+	testCases := []struct {
+		imageName             string
+		translateWorkflowPath string
+		noGuestEnvironment    bool
+		sourceFile            string
+		sourceImage           string
+		family                string
+		description           string
+		network               string
+		subnet                string
+		region                string
+		want                  map[string]string
+	}{
+		{
+			imageName:             "image-a",
+			translateWorkflowPath: "translate/workflow/path",
+			noGuestEnvironment:    true,
+			sourceFile:            "",
+			sourceImage:           "source-image",
+			family:                "a-family",
+			description:           "a-description",
+			network:               "a-network",
+			subnet:                "a-subnet",
+			region:                "a-region",
+			want: map[string]string{
+				"image_name":           "image-a",
+				"translate_workflow":   "translate/workflow/path",
+				"install_gce_packages": "false",
+				"source_image":         "global/images/source-image",
+				"family":               "a-family",
+				"description":          "a-description",
+				"import_network":       "global/networks/a-network",
+				"import_subnet":        "regions/a-region/subnetworks/a-subnet",
+				"is_windows":           "false",
+			},
+		},
+		{
+			imageName:             "image-a",
+			translateWorkflowPath: "windows/translate_windows_2008_r2.wf.json",
+			noGuestEnvironment:    true,
+			sourceFile:            "source-file-path",
+			sourceImage:           "",
+			family:                "a-family",
+			description:           "a-description",
+			network:               "a-network",
+			subnet:                "a-subnet",
+			region:                "a-region",
+			want: map[string]string{
+				"image_name":           "image-a",
+				"translate_workflow":   "windows/translate_windows_2008_r2.wf.json",
+				"install_gce_packages": "false",
+				"source_disk_file":     "source-file-path",
+				"family":               "a-family",
+				"description":          "a-description",
+				"import_network":       "global/networks/a-network",
+				"import_subnet":        "regions/a-region/subnetworks/a-subnet",
+				"is_windows":           "true",
+			},
+		},
+		{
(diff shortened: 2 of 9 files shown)


daisy/step_create_disks.go
```diff
@@ -76,3 +84,28 @@ func (c *CreateDisks) run(ctx context.Context, s *Step) DError {
 		return nil
 	}
 }
+
+// Merges two slices of features and returns a new slice instance.
+// Duplicates are removed, and the result is sorted lexically.
+func combineGuestOSFeatures(currentFeatures []*compute.GuestOsFeature,
```
zoran15: current/additional doesn't matter here since you're creating a return value so naming can me features1, features2

daisy/step_create_disks.go
```diff
@@ -76,3 +84,28 @@ func (c *CreateDisks) run(ctx context.Context, s *Step) DError {
...
+	}
+	ret := make([]*compute.GuestOsFeature, 0)
+	for feature := range featureSet {
+		ret = append(ret, &compute.GuestOsFeature{
+			Type: feature,
+		})
+	}
+	// Sort elements by type, lexically. This ensures
+	// stability of output ordering for tests.
+	sort.Slice(ret, func(i, j int) bool {
```
zoran15: We should not have code that is needed only for tests. If tests need sorted values, they should sort them themselves.

daisy_workflows/image_import/import_from_image.wf.json
```diff
@@ -41,7 +45,8 @@
         "Name": "${disk_name}",
         "Type": "pd-ssd",
         "ExactName": true,
-        "SourceImage": "${source_image}"
+        "SourceImage": "${source_image}",
+        "is_windows": "${is_windows}"
```
zoran15: Shouldn't this be: "IsWindows": "${is_windows}"?

daisy/step_create_disks.go
```diff
@@ -47,6 +51,10 @@ func (c *CreateDisks) run(ctx context.Context, s *Step) DError {
 		go func(cd *Disk) {
 			defer wg.Done()
 
+			if res, _ := strconv.ParseBool(cd.IsWindows); res {
+				cd.GuestOsFeatures = combineGuestOSFeatures(cd.GuestOsFeatures, "WINDOWS")
```
zoran15: This should probably be done in populate so it's validated, instead of modifying it directly in run. Otherwise we might run an invalid workflow and not know until we attempt to run CreateDisk

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -247,56 +247,109 @@ func TestFlagsInvalidOS(t *testing.T) {
...
-	sourceImage = "source-image"
-	family = "a-family"
-	description = "a-description"
-	network = "a-network"
-	subnet = "a-subnet"
-	region := "a-region"
-
-	got := buildDaisyVars("translate/workflow/path", imageName, sourceFile,
-		sourceImage, family, description, region, subnet, network, noGuestEnvironment)
+	testCases := []struct {
```
zoran15: Split these tests back. Batching tests in one test method only makes it harder to know what actually failed without looking at unit test code. On the other hand, if we keep the "one test - one method" convention, just looking at the test method name will tell us what failed.

daisy/step_create_disks.go
EricEdens: Hmm. I'm not sure I agree, but it sounds like your opinion is stronger than mine, so I'm fine moving it :)

daisy/step_create_disks.go
EricEdens: sgtm

daisy_workflows/image_import/import_from_image.wf.json
EricEdens: Yeah, I wasn't sure either since both naming formats are used in different JSON definitions. Now that you mention it, I think it makes sense to use `IsWindows` here, though, to at least match the rest of this block

daisy/step_create_disks.go
EricEdens: sgtm

daisy/step_create_disks.go
EricEdens: Actually, looking at it again, I'd say I still don't agree, so let me explain more :)  Right now the resulting feature list is being tested in two places, so we'd have to re-sort it twice in our tests. Any time we do more tests, we'll have to re-sort on the test side. 

So, by having it pre-sorted, that gives a stronger contract that makes testing easier.

zoran15: This discussion is due to the fact we are returning an array when we should be returning a set. By sorting the array you're making the array behave like a set. But then we should use set in the first place. Or, leave it as is, but don't guarantee ordering and let tests verify if two arrays are equal set wise.

daisy_workflows/image_import/import_from_image.wf.json
zoran15: Would it even work with is_windows? The field name is IsWindows. Unless JSON unmarshalling is smart enough to convert between underscore and camel case notation, I don't see how this would work.

EricEdens: Yeah, this is the syntax for specifying how the field appears in your json document:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/896a2ffdf8a18584815433aa58e278deebdfed79/cli_tools/common/utils/logging/service/log_entry.go#L35

daisy/disk.go
```diff
@@ -66,6 +66,13 @@ type Disk struct {
 	compute.Disk
 	Resource
 
+	// If this is enabled, then WINDOWS will be added to the
+	// disk's guestOsFeatures. This is a string since daisy
+	// replaces variables after JSON has been parsed.
+	// (If it were boolean, the JSON marshaller throws
+	// an error when it sees something like `${is_windows}`)
+	IsWindows string `json:"is_windows,omitempty"`
```
zoran15: use lower case camel case, same as sizeGb below

daisy_workflows/image_import/import_from_image.wf.json
zoran15: This was probably an oversight from Tony. The rest of Daisy uses lower case camel case notation, we should not mix different notations for new code.

zoran15: Also, he might have been forced to use underscore notation because these jsons are sent to Firelog and they might have different notation.

For internal Daisy fields, we should keep camel case. You actually don't need to specify json field name, it will default to Go field name. Check other structs in Daisy.

daisy/step_create_disks.go
EricEdens: Updated to keep the sorting, but remove sorting from the contract.

daisy_workflows/image_import/import_from_image.wf.json
EricEdens: Changed to `isWindows`

cli_tools/gce_vm_image_import/importer/importer_test.go
EricEdens: Reverted

daisy/disk.go
EricEdens: done!

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -150,12 +150,17 @@ func buildDaisyVars(translateWorkflowPath, imageName, sourceFile, sourceImage, f
...
 		varMap["translate_workflow"] = translateWorkflowPath
 		varMap["install_gce_packages"] = strconv.FormatBool(!noGuestEnvironment)
+		varMap["isWindows"] = strconv.FormatBool(strings.Contains(translateWorkflowPath, "windows"))
 	}
 	if sourceFile != "" {
 		varMap["source_disk_file"] = sourceFile
 	}
 	if sourceImage != "" {
-		varMap["source_image"] = fmt.Sprintf("global/images/%v", sourceImage)
+		if strings.Contains(sourceImage, "global/images") {
```
dntczdx: I have a similar change to do this. Is this required by what you are doing? It looks like different thing

zoran15: I guess this is due to Eric merging your changes? Although your changes are not yet merged. It's weird

daisy/disk.go
```diff
@@ -66,6 +67,13 @@ type Disk struct {
 	compute.Disk
 	Resource
 
+	// If this is enabled, then WINDOWS will be added to the
+	// disk's guestOsFeatures. This is a string since daisy
+	// replaces variables after JSON has been parsed.
+	// (If it were boolean, the JSON marshaller throws
+	// an error when it sees something like `${is_windows}`)
+	IsWindows string `json:"isWindows,omitempty"`
```
dntczdx: I believe you can do this:
IsWindows bool                `json:"isWindows,string,omitempty"`

dntczdx: It convert the string "true|false" to bool

daisy_workflows/image_import/import_from_image.wf.json
```diff
@@ -17,6 +17,10 @@
       "Value": "true",
       "Description": "Whether to install GCE packages."
     },
+    "is_windows": {
```
dntczdx: I don't 100% get the motivation. You have triggered CombineGuestOSFeatures in disk.go. Do you need to set this flag for "family" as well?

EricEdens: I agree that there's a bit of overlap. Here are the differences:
 - Family is a free-form metadata field for users -- whatever they put in that field shows up in their web UI.
 - is_windows instructs daisy to add WINDOWS to GuestOsFeatures if it's not already there

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: Yeah, it was a bug that I found when running the command below:

```
go run compute-image-tools/cli_tools/gce_vm_image_import/main.go \
  -client_id=edens-test \
  -source_image=projects/ajackura-dev/global/images/customer-srv203-issue886 \
  -os=windows-2008r2 \
  -image_name=guest-os-test-7 \
  -project=edens-test
```

daisy/disk.go
EricEdens: The problem is that daisy populates variables *after* the JSON document has been parsed into a struct. So, the JSON parser will throw something like: `trying to unmarshal "${is_windows}" into bool`

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: It's basically the same thing Tony worked on. Either of you will have to merge the changes.

daisy/disk.go
```diff
@@ -237,3 +256,29 @@ func (dr *diskRegistry) regDetachAll(iName string, s *Step) DError {
...
+	}
+	ret := make([]*compute.GuestOsFeature, 0)
+	for feature := range featureSet {
+		ret = append(ret, &compute.GuestOsFeature{
+			Type: feature,
+		})
+	}
+	// Sort elements by type, lexically. This ensures
+	// stability of output ordering for tests.
+	sort.Slice(ret, func(i, j int) bool {
```
zoran15: I still I don't agree we should be doing anything just to make tests happy (like this sorting) as it's boilerplate code that should belong in the test (client code of this function doesn't require it so it shouldn't be there). If you copy-pasted this code in the tests before asserting, tests will succeed as is now and we won't do unnecessary things in real code. It will also save future developers of this code trying to understand why are we sorting the list.

PR #940, opened 2019-09-26
add OS Config agent e2e tests for pre-/post-patch steps

wj-chen:
Tested images:
-test_case_filter=debian-9
-test_case_filter=debian-10
-test_case_filter=ubuntu-1604-lts
-test_case_filter=centos-7

/woof

### DIFF
--- a/osconfig_tests/test_suites/patch/test_setup.go
+++ b/osconfig_tests/test_suites/patch/test_setup.go
@@ -37,11 +37,11 @@ while ($true) {
   $old = Invoke-RestMethod -Method GET -Uri $uri -Headers @{"Metadata-Flavor" = "Google"}
   $new = $old+1
   try {
-	Invoke-RestMethod -Method PUT -Uri $uri -Headers @{"Metadata-Flavor" = "Google"} -Body $new -ErrorAction Stop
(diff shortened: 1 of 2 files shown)


osconfig_tests/test_suites/patch/test_setup.go
```diff
@@ -61,6 +61,24 @@ if (Test-Connection $wu_server -Count 1 -ErrorAction SilentlyContinue) {
 	Set-ItemProperty -Path $windows_update_path -Name WUStatusServer -Value "http://${wu_server}:8530"
 	Set-ItemProperty -Path $windows_update_au_path -Name UseWUServer -Value 1
 }
+`
+
+	windowsLocalPostPatchScript = `
+New-Item -Path . -Name "windows_local_post_patch_script.cmd" -ItemType "file" -Value "echo OS Config - local patch script - Windows > post_patch_script_out.txt"
+`
+
+	windowsPrePostStepValidation = `
```
adjackura: Can we make the scripts themselves write to guest attributes, that way we don't need to have this validation run at all

wj-chen: Thanks, it made the setup much simpler, I tested the new scripts on the same images as mentioned in the top comment, but not on windows because of bug 141310538

PR #942, opened 2019-09-26
Fix is_windows variable name causing E2E test failures

zoran15:
Fix is_windows variable name causing E2E test failures: `unknown workflow Var "isWindows" passed to Workflow "import-from-image"`


EricEdens:
Thanks for fixing this!

PR #946, opened 2019-09-30
Wait until persistent disks are attached prior to proceeding with import workflow.

EricEdens:
Testing:

Ran `go run gce_ovf_import_tests/main.go -test_project_id edens-test -test_zone us-central1-a`

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -43,37 +43,33 @@ echo "DISKNAME: ${DISKNAME}" 2> /dev/null
 echo "ME: ${ME}" 2> /dev/null
 echo "ZONE: ${ZONE}" 2> /dev/null
 
-function waitUntilDeviceAttached() {
-  local devPath="${1}"
-  local capacityGb="${2}"
-
-  echo "Import: Checking for $devPath ${capacityGb}G"
-  for t in {1..60}; do
-    if [[ -e ${devPath} ]]; then
-      capacity=$(lsblk ${devPath})
-      if [[ "${capacity}" =~ "${capacityGb}G" ]]; then
-        echo "Import: $devPath is attached and ready."
-        return
-      fi
-    fi
-
-    sleep 5
-  done
-  echo "ImportFailed: Failed to attach disk $devPath"
-  exit
-}
-
 
 function resizeDisk() {
   local diskId="${1}"
   local newSizeInGb="${2}"
   local zone="${3}"
+  local deviceId="${4}"
 
   echo "Import: Resizing ${diskId} to ${newSizeInGb}GB in ${zone}."
   if ! out=$(gcloud -q compute disks resize ${diskId} --size=${newSizeInGb}GB --zone=${zone} 2>&1); then
     echo "ImportFailed: Failed to resize disk. [Privacy-> resize disk ${diskId} to ${newSizeInGb}GB in ${zone}, error: ${out} <-Privacy]"
     exit
   fi
+
+  echo "Import: Checking for ${deviceId} ${newSizeInGb}G"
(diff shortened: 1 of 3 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -43,6 +43,27 @@ echo "DISKNAME: ${DISKNAME}" 2> /dev/null
...
 
+function waitUntilDeviceAttached() {
+  local devPath="${1}"
+  local capacityGb="${2}"
+
+  echo "Import: Checking for $devPath ${capacityGb}G"
+  for t in {1..60}; do
+    if [[ -e ${devPath} ]]; then
+      capacity=$(lsblk ${devPath})
+      if [[ "${capacity}" =~ "${capacityGb}G" ]]; then
```
zoran15: Why is capacity check needed? Did you see cases where the device is attached but size is incorrect, like 0?

daisy_workflows/image_import/import_image.sh
```diff
@@ -149,13 +169,11 @@ if ! out=$(gcloud -q compute instances attach-disk ${ME} --disk=${DISKNAME} --zo
...
 echo ${out}
 
-# Output the size of the persistent disks for debugging.
-# The '-i' parameter is for ascii; without it the output is
-# garbled in serial output.
-#
-# No df here since we're interested in /dev/sdc which doesn't
-# have a filesystem yet.
-lsblk -i
+if [[ ${SIZE_GB} -gt 10 ]]; then
```
zoran15: Can ${SIZE_GB} be less than 10 or it's rounded up to 10? If it is, this if statement is redundant and we could just do:
waitUntilDeviceAttached /dev/sdc "${SIZE_GB}"

daisy_workflows/image_import/import_image.sh
```diff
@@ -43,6 +43,27 @@ echo "DISKNAME: ${DISKNAME}" 2> /dev/null
...
+function waitUntilDeviceAttached() {
+  local devPath="${1}"
+  local capacityGb="${2}"
+
+  echo "Import: Checking for $devPath ${capacityGb}G"
+  for t in {1..60}; do
+    if [[ -e ${devPath} ]]; then
+      capacity=$(lsblk ${devPath})
+      if [[ "${capacity}" =~ "${capacityGb}G" ]]; then
+          echo "Import: $devPath is attached and ready."
```
zoran15: Seems indentation is incorrect here (return and echo are in the same block)

daisy_workflows/image_import/import_image.sh
EricEdens: Yeah, I've seen examples where the disk image requires about 30GB, we resize it, and qemu-img fails at about 10GB (which is the normal disk size).

daisy_workflows/image_import/import_image.sh
EricEdens: Yeah, it can be less than 10. The check has been used for awhile: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/dbe6534a73bef27454034d9eb274a31011319f3f/daisy_workflows/image_import/import_image.sh#L61

daisy_workflows/image_import/import_image.sh
EricEdens: good catch!

adjackura:
I think this is much easier solved by just skipping the attach detach steps in the script altogether, we want to avoid as many calls from in the vm as possible. I believe all you need to do is create the instance with the disk attached and then just add a delete step at the end of that section of the workflow

EricEdens:
@adjackura  - Good idea! I removed the manual attach and detach steps.

@zoran15  - One change landed in data_disk_workflow_updater.go

PR #949, opened 2019-10-01
adding software recipe end to end tests for linux

TrevorBoone:
Windows support will be added later. 

This PR is a little longer in length than I would generally like, let me know if it would be best to split it into multiple PRs.


osconfig_tests/test_suites/guestpolicies/guest_policies.go
```diff
@@ -90,6 +98,43 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
 	testSuite := junitxml.NewTestSuite(testSuiteName)
 	defer testSuite.Finish(testSuites)
 
+	storage, err := gcpclients.GetStorageClient()
```
TrevorBoone: I would prefer the storage stuff to be in it's own function. It seemed to me like to do that it would need to return a cleanup function which would then be deffered but I'm on the fence as to whether that would be cleaner or not

osconfig_tests/test_suites/guestpolicies/guest_policies_test_data.go
```diff
@@ -229,6 +229,130 @@ func addPackageInstallFromNewRepoTest(key string) []*guestPolicyTestSetup {
...
+						Id:            "zip-test",
+						Artifact: &osconfigpb.SoftwareRecipe_Artifact_Gcs_{
+							Gcs: &osconfigpb.SoftwareRecipe_Artifact_Gcs{
+								Bucket: tempBucketName,
+								Object: "zip_test.zip",
+							},
+						},
+					},
+				},
+				[]*osconfigpb.SoftwareRecipe_Step{
```
TrevorBoone: These steps so far are fairly simple and all of them just put a file somewhere, which makes verification easy

osconfig_tests/test_suites/guestpolicies/guest_policies.go
```diff
@@ -90,6 +98,43 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
...
+	if err != nil {
+		logger.Printf("error retrieving storage client: %s", err)
+		logger.Printf("skipping test suite %s due to error", testSuiteName)
+		return
+	}
+	bucket := storage.Bucket(tempBucketName)
```
adjackura: Can we use a common bucket for all tests, just a temp 'folder', since buckets are in a shared namespace I don't want this to go crazy. You can leave the create call if it doesnt exist, just don't make it unique and there is no need to clean it up after.

TrevorBoone: The issue with using the same bucket name every time is that bucket names are global so if different people try to run the tests the names will collide.

adjackura: True, can you generate it based on the project the tests run in? that way its just one bucket per project

hopkiw: we can use one bucket so long as the subdir is unique and has collision detection. but generating a bucket based on the project name is pretty standard. should do both imo, so two running in one project won't collide, and different projects will use different buckets

osconfig_tests/test_suites/guestpolicies/testfiles/exec_test
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
hopkiw: i don't quite follow this paradigm. this will get uploaded so it can be used in an exec step test? where do exec_test_windows.* files get uploaded? also, i don't think we should be storing any binary artifacts like tarballs in git

TrevorBoone: Currently it uploads the recipes to a temporary bucket it creates. Generally I don't store these things in source control but these shouldn't change often, would it be better to store them in an open bucket instead?

osconfig_tests/test_suites/guestpolicies/guest_policies.go
hopkiw: i think this is the only outstanding change blocking this PR. for most of the other software we use or maintain that dumps to GCS, it typically assumes creating a bucket in the project you're running in for the application overall (so e.g. osconfig-tests or similar), with file naming based on the test run. e.g. gs://${project}-osconfig-tests/testrun-${date}/ or similar. bucket can remain existing after test completion, but you can remove files within it.

osconfig_tests/test_suites/guestpolicies/testfiles/exec_test
hopkiw: you mean it uploads these files as artifacts for use in a recipe? i'm fine with that, in general i also would not store in git. for such simple files i am just going to say it's fine for now, though.

adjackura:
Moved to https://github.com/GoogleCloudPlatform/osconfig/pull/116

PR #952, opened 2019-10-02
CentOS 6: Wait until network is ready.

EricEdens:
Testing: Ran `go run gce_ovf_import_tests/main.go -test_project_id edens-test -test_zone us-central1-a`

### DIFF
--- a/daisy_integration_tests/scripts/post_translate_test.sh
+++ b/daisy_integration_tests/scripts/post_translate_test.sh
@@ -18,6 +18,25 @@
 FAIL=0
 FAILURES=""
 
+function wait_for_connectivity {
+  if [[ $(cat /etc/centos-release) =~ "CentOS release 6" ]]; then
+    status "CentOS 6: Waiting for network connectivity."
+  else
+    return
+  fi
+
+  for i in {0..60}; do
+    if [[ $(nmcli -t -f STATE nm) == "connected" ]]; then
+      status "Network ready."
+      return
+    fi
+    sleep 5s
+  done
+
+  echo "FAILED: Unable to initialize network connection."
+  exit 1
+}
+
 function status {
   local message="${1}"
   echo "STATUS: ${message}"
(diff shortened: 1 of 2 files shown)


daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -18,6 +18,25 @@
 FAIL=0
 FAILURES=""
 
+function wait_for_connectivity {
```
zoran15: Do we need to do this for Centos 6 translate as well (or is that already fixed)?

daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -18,6 +18,25 @@
...
+  if [[ $(cat /etc/centos-release) =~ "CentOS release 6" ]]; then
```
zmarano: CentOS and RHEL 6 are identical in this realm. If you ran into this in one you can run into it in the other as well.

EricEdens: You're right!

This `wait_for_connectivity` is a surgical fix for the CentOS6 OVF integration test. The test has been failing for awhile, and I want to make sure that we get it passing so that it doesn't hide other errors.

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/c4dd645bf723d7a934a719a865ee8f6f7afce6bb/gce_ovf_import_tests/test_suites/ovf_import_test_suite.go#L113 .

Zooming out, @hopkiw is working on a fix that will address both CentOS6 and RHEL6 in https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/packages/google-compute-engine/src/etc/init/google-startup-scripts.conf

daisy_integration_tests/scripts/post_translate_test.sh
EricEdens: I haven't seen any network errors in the translation script, so I'm fine waiting for the root-cause fix that Liam is working on. But if we see errors, we can rethink that.

daisy_integration_tests/scripts/post_translate_test.sh
zoran15: Centos 6 tests were failing for regular image import, not only OVF import. Your fix will apply to both though so no action needed.

PR #953, opened 2019-10-02
Fix CentOS 8 build.

zmarano:


daisy_workflows/image_build/enterprise_linux/kickstart/el8-options.cfg
```diff
@@ -2,8 +2,7 @@
 
 ### Anaconda installer configuration.
 # Install in text mode.
-text
-install
```
hopkiw: are you sure? `install` has historically been a necessary directive

zmarano: It is deprecated in EL8...https://docs.centos.org/en-US/8-docs/advanced-install/assembly_kickstart-commands-and-options-reference/#deprecated-kickstart-comands-and-options_kickstart-changes

In this case, the subcommand (harddrive) is where the sources come from.

PR #954, opened 2019-10-02
Upping the timeout for cloudbuild due to recent timeouts.

zmarano:


daisy_latest_cloudbuild.yaml
```diff
@@ -1,4 +1,4 @@
-timeout: 600s
+timeout: 700s
```
zoran15: I would go with at least 900s (15min) to have some buffer

zmarano: Done

PR #960, opened 2019-10-08
Resize post-check: Support terabyte-sized disks

EricEdens:
I added this check as part of #946 as a lower-priority check. It's breaking a large test that uses a disk of size 1024 GB; lsblk returns a size of 1T rather than 1024G, and hence the check is failing.

So, I'm reverting this check for now since it would break all imports of size greater than 1T.


EricEdens:
> I believe you can force lsblk output GB. Just like I used "--output=size -b" to force output bytes in export_disk_ext.sh

Good idea! Updated

PR #961, opened 2019-10-09
Tag OVF Windows imports with WINDOWS GuestOSFeature

EricEdens:


cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
```diff
@@ -70,6 +70,7 @@ func AddDiskImportSteps(w *daisy.Workflow, dataDiskInfos []ovfutils.DiskInfo) {
 					ExactName: true,
 					NoCleanup: true,
 				},
+				IsWindows: w.Vars["is_windows"].Value,
```
zoran15: These are data disks. Why do we need to tag them with WINDOWS GuestOsFeature? I would have thought only the boot disk should be tagged.

EricEdens: Right you are! Good find.

```
guestOsFeatures[]
A list of features to enable on the guest operating system. 
Applicable only for bootable images.
```
https://cloud.google.com/compute/docs/reference/rest/v1/regionDisks/get

EricEdens: Updated!

PR #972, opened 2019-10-16
[logging] fix log race condition

dntczdx:
It guards "logParams". The "logParams" has a chance to be updated while another thread is trying to send it to log server.

### DIFF
--- a/cli_tools/common/utils/logging/service/log_entry.go
+++ b/cli_tools/common/utils/logging/service/log_entry.go
@@ -171,15 +171,15 @@ type OutputInfo struct {
 	FailureMessageWithoutPrivacyInfo string `json:"failure_message_without_privacy_info,omitempty"`
 }
 
-func (i *InputParams) updateParams(p *string) {
+func (i *InputParams) updateParams(projectPointer *string) {
 	logParamsMutex.Lock()
 	defer logParamsMutex.Unlock()
 
-	if p == nil {
+	if projectPointer == nil {
 		return
 	}
 
-	project := *p
+	project := *projectPointer
 	obfuscatedProject := Hash(project)
 
 	if i.ImageImportParams != nil {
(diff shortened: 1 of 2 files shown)


cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -172,6 +172,9 @@ type OutputInfo struct {
 }
 
 func (i *InputParams) updateParams(p *string) {
+	logParamsMutex.Lock()
```
zoran15: Why is this called update params when it's updating project info?

Should p called project?

EricEdens:
Can you add some more info about the race condition that you've observed, and which resources you're trying to guard access to?

cli_tools/common/utils/logging/service/log_entry.go
dntczdx: There can be more params updated if necessary. I won't provide multiple functions for that.
Let me adjust the project param name

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -171,12 +171,15 @@ type OutputInfo struct {
 	FailureMessageWithoutPrivacyInfo string `json:"failure_message_without_privacy_info,omitempty"`
 }
 
-func (i *InputParams) updateParams(p *string) {
-	if p == nil {
+func (i *InputParams) updateParams(projectPointer *string) {
```
zoran15: project is enough as it's obvious it's pointer from the type

dntczdx:
> Can you add some more info about the race condition that you've observed, and which resources you're trying to guard access to?

As the mutex's name said, it guards "logParams". The "logParams" has a chance to be updated while another thread is trying to send it to log server.

cli_tools/common/utils/logging/service/log_entry.go
dntczdx: just in order to letting project & obfuscatedProject having the same type...

EricEdens:
> > Can you add some more info about the race condition that you've observed, and which resources you're trying to guard access to?
> 
> As the mutex's name said, it guards "logParams". The "logParams" has a chance to be updated while another thread is trying to send it to log server.

What are `logParams`? An instance of the InputParams struct? If so, why not put your mutex as a field within that struct?

PR #978, opened 2019-10-18
[logging] override 'project' for imports

dntczdx:
I have fixed project override for ImageExport and verified it works as expected. Now let's apply it to ImageImport and InstanceImport.


cli_tools/gce_ovf_import/ovf_import_params/ovf_import_params.go
```diff
@@ -49,7 +49,7 @@ type OVFImportParams struct {
 	BootDiskKmsLocation         string
 	BootDiskKmsProject          string
 	Timeout                     string
-	Project                     string
+	Project                     *string
```
zoran15: why the change to pointer?

cli_tools/gce_ovf_import/main.go
```diff
@@ -83,30 +83,14 @@ func buildImportParams() *ovfimportparams.OVFImportParams {
 		ShieldedIntegrityMonitoring: *shieldedIntegrityMonitoring, ShieldedSecureBoot: *shieldedSecureBoot,
 		ShieldedVtpm: *shieldedVtpm, Tags: *tags, Zone: *zoneFlag, BootDiskKmskey: *bootDiskKmskey,
 		BootDiskKmsKeyring: *bootDiskKmsKeyring, BootDiskKmsLocation: *bootDiskKmsLocation,
-		BootDiskKmsProject: *bootDiskKmsProject, Timeout: *timeout, Project: *project,
+		BootDiskKmsProject: *bootDiskKmsProject, Timeout: *timeout, Project: project,
 		ScratchBucketGcsPath: *scratchBucketGcsPath, Oauth: *oauth, Ce: *ce,
```
EricEdens: Can you explain the history of the bug a bit, and why this fixes it?

cli_tools/gce_ovf_import/main.go
```diff
@@ -153,7 +137,24 @@ func main() {
 		},
 	}
 
-	if err := service.RunWithServerLogging(service.InstanceImportAction, paramLog, project, runImport); err != nil {
+	params := buildImportParams()
```
EricEdens: What's the motivation for making this function an anonymous function? (It's hard for me to trace what changed to fix your bug vs the refactoring)

dntczdx: Because buildImportParams has to be executed before calling RunWithServerLogging. But I reverted this change in the latest commit by removing the ref to param.project in order to simplify the PR, so we can ignore this now.

cli_tools/gce_ovf_import/main.go
dntczdx: "project" is a input param. However, in many cases, user didn't type in this value (so it will be initialized as an empty value). Thus, the 'project' will be populated after the tool is lunched. We need that populated value. A pointer gives us the ability to reflect the updated 'project' in log.

cli_tools/gce_ovf_import/ovf_import_params/ovf_import_params.go
dntczdx: "project" is a input param. However, in many cases, user didn't type in this value (so it will be initialized as an empty value). Thus, the 'project' will be populated after the tool is lunched. We need that populated value. A pointer gives us the ability to reflect the updated 'project' in log.

zoran15: Can we explicitly update params to make this workflow obvious?

dntczdx: "updateParam" is called not very deep, so it's realtively explicit from my perspective.
It's hard to put that in "main" function. The design tried to separate logging service from import/export logic, so it can't be that explicit.

dntczdx:
Replied comment

cli_tools/gce_ovf_import/ovf_import_params/ovf_import_params.go
zoran15: where is project updated after the fact?

dntczdx: log_entry.go -> "updateParam". The reason why it's not clear in this PR is becuase the code is partially landed in the last PR which I only fixed image-export. Link here: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/970

dntczdx:
https://github.com/GoogleCloudPlatform/compute-image-tools/pull/970 for reference to clarify how that pointer takes effect

cli_tools/gce_ovf_import/main.go
zoran15: I'm not a big fan of updating a field via a pointer as it's hard to follow for others since it's implicit. Having an explicit update would be easier to follow and understand for anyone working on this code base in the future.

EricEdens:
+1

cli_tools/gce_ovf_import/main.go
dntczdx: I totally understand. This patch is to resolve our current problem of excluding test project in order to unblock our failure analysis - it's a fast fix. I have verified it works well and won't break.
I will work on a redesign to avoid implicit update soon.

PR #979, opened 2019-10-19
Transfrom Application & System event logs from evtx files to plain text files

YangLu1031:
Evtx file is compressed binary xml format, not readable on Linux/Mac without any tools. 
If transformed all evtx files, it would take around 20mins.
So just transfrom Application,System which are most valuable logs, to plain text files.

### DIFF
--- a/cli_tools/diagnostics/main_windows.go
+++ b/cli_tools/diagnostics/main_windows.go
@@ -34,6 +34,11 @@ const (
 	crashDump = `C:\Windows\MEMORY.dmp`
 )
 
+type winEvt struct {
+	logName      string
+	fromProvider bool
+}
+
 type cmd struct {
 	path           string
 	args           string
@@ -118,8 +123,7 @@ func runAll(commands []runner, errCh chan error) []string {
 	for _, command := range commands {
 		path, err := command.run()
 		if err != nil {
-			log.Printf("Error: %s while running %v", err, command)
-			errCh <- err
+			errCh <- fmt.Errorf("Error: %s while running %v", err, command)
 		} else {
 			paths = append(paths, path)
 		}
@@ -204,6 +208,30 @@ func collectFilePaths(roots []string) ([]string, []error) {
 	return filePaths, errs
 }
 
+// getPlainEventLogs generates plain text event logs thru `Get-WinEvent` powershell cmd,
+// return list file paths and errors(if any).
+func getPlainEventLogs(evts []winEvt, errs chan error) []string {
+	pwshPath, err := exec.LookPath("powershell")
+	if err != nil {
+		errs <- err
+		return []string{}
+	}
+	argStr := `Get-WinEvent %s'%s' | Format-Table -Auto -Wrap`
+	commands := make([]runner, 0)
+	for _, evt := range evts {
+		source := "-LogName "
+		if evt.fromProvider {
+			source = "-ProviderName "
+		}
+		commands = append(commands, cmd{
+			path:            pwshPath,
+			args:            fmt.Sprintf(argStr, source, evt.logName),
+			outputFileName:  evt.logName + ".log",
+			cmdProducesFile: false})
+	}
+	return runAll(commands, errs)
+}
+
 // gatherEventLogs put all the event log file paths in logFolder channel
 // and errors in error channel.
 func gatherEventLogs(logs chan logFolder, errs chan error) {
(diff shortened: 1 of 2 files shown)


pjh:
cc @ning-yang @adjackura

cli_tools/diagnostics/main_windows.go
```diff
@@ -204,6 +208,30 @@ func collectFilePaths(roots []string) ([]string, []error) {
...
+	argStr := `Get-WinEvent %s'%s' | Format-Table -Auto -Wrap`
+	commands := make([]runner, 0)
+	for _, evt := range evts {
+		source := "-LogName "
+		if evt.fromProvider {
+			source = "-ProviderName "
+		}
+		commands = append(commands, cmd{
+			path:            pwshPath,
+			args:            fmt.Sprintf(argStr, source, evt.logName),
```
pjh: The use of argStr is pretty far from its declaration here, making it a bit difficult to understand what's being executed. You could consider writing a small function

`func getCommandToConvertEvt(fromProvider bool) string { ... }`

then this loop would just iterate over `commands = append(... args: getCommandToConvertEvt(evt.fromProvider) ...)`

cli_tools/diagnostics/main_windows.go
```diff
@@ -34,6 +34,11 @@ const (
 	crashDump = `C:\Windows\MEMORY.dmp`
 )
 
+type winEvt struct {
```
pjh: This could use a comment describing what it represents and what the "fromProvider" arg means.

Alternatively, instead of tracking a fromProvider arg, you could write the code below so that it tries the command with `-LogName` first, and if it fails then it tries `-ProviderName` instead. That would eliminate the need for the winEvt struct.

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -44,6 +47,50 @@ func fileNonExist(e error) bool {
...
+
+	tests := []struct {
+		name         string
+		args         []winEvt
+		want         []string
+		expectErrStr string
+	}{
+		{name: "Nil events", args: nil, want: []string{}, expectErrStr: ""},
+		{name: "Empty events", args: []winEvt{}, want: []string{}, expectErrStr: ""},
+		{name: "Existing events logName", args: []winEvt{{"Application", false}}, want: []string{filepath.Join(tmpFolder, applicationTextLogFileName)}, expectErrStr: ""},
```
pjh: I don't know Go style at all, but it seems like some line breaks might make this long line more readable.

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -84,7 +131,7 @@ func TestCollectFilePaths(t *testing.T) {
 
 func stringArrayIncludesString(stringArray []string, target string) bool {
```
pjh: If this is changing then it should have a comment, or perhaps be renamed to `stringArrayIncludesSubstring`.

cli_tools/diagnostics/main_windows_test.go
```diff
@@ -94,13 +141,16 @@ func stringArrayIncludesString(stringArray []string, target string) bool {
 func TestGatherEventLogs(t *testing.T) {
 	logFolderCh := make(chan logFolder, 2)
 	errCh := make(chan error)
+	expectedFiles := []string{systemLogPath, applicationTextLogFileName, systemTextLogFileName}
```
pjh: Testing in this manner is a "change-detector" test, which often isn't very valuable. It's testing the internals of the function, rather than the API / contract.

If these are all the files that **a "user"** should expect gatherEventLogs to return, then these should be defined somewhere (somewhere in the code (e.g. a constant) preferably, or in a comment for gatherEventLogs at least). This makes the set of files part of the API / contract and then it's ok to test for the presence of all of them.

Alternatively, perhaps just choose one "important" file that we expect gatherEventLogs to always return, and test for that. This is the approach that this test took before this PR.

cli_tools/diagnostics/main_windows.go
YangLu1031: Prefer adding comments here. The alternative seems a little complicate, as it might fail for other reasons, then we need to stop retry for those cases. And found that when the powershell cmd fails, we only got general `Exit status 1` error without any details, which seems interesting, need to figure out later if needed.

cli_tools/diagnostics/main_windows_test.go
YangLu1031: Makes sense. Reverted back

YangLu1031:
@pjh Please take a look again :)

cli_tools/diagnostics/main_windows.go
```diff
@@ -204,6 +211,36 @@ func collectFilePaths(roots []string) ([]string, []error) {
 	return filePaths, errs
 }
 
+// getCommandToConvertEvt return `Get-WinEvent (-ProviderName/-LogName) ${logName}` powershell cmd string
+// based on log source: provider or log name.
```
pjh: This comment is a little too detailed. First, there's no need to include the function name in the comment describing the function. Consider something like:

```
Returns a Get-WinEvent command to get the event log for the specified log source.
```

PR #980, opened 2019-10-21
Don't allow empty disk images to be imported.

EricEdens:
We've seen some errors lately where an import was run against an empty file. This will help to catch the error earlier, and to give more helpful debug information.

### DIFF
--- a/cli_tools/daisycommon/io.go
+++ b/cli_tools/daisycommon/io.go
@@ -10,6 +10,7 @@ type ByteCountingReader struct {
 	BytesRead int64
 }
 
+// NewByteCountingReader is a contructor for ByteCountingReader.
 func NewByteCountingReader(r io.Reader) *ByteCountingReader {
 	return &ByteCountingReader{r, 0}
 }
--- a/cli_tools/daisycommon/io.go
+++ b/cli_tools/daisycommon/io.go
@@ -1,0 +1,21 @@
+package daisycommon
+
+import "io"
+
+// ByteCountingReader forwards calls to a delegate reader, keeping
+// track of the number bytes that have been read in `BytesRead`.
+// Errors are propagated unchanged.
+type ByteCountingReader struct {
+	r         io.Reader
+	BytesRead int64
+}
+
+func NewByteCountingReader(r io.Reader) *ByteCountingReader {
+	return &ByteCountingReader{r, 0}
+}
+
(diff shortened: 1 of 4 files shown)


cli_tools/daisycommon/io_test.go
```diff
@@ -0,0 +1,35 @@
+package daisycommon
```
zoran15: Add copyright

cli_tools/daisycommon/io.go
```diff
@@ -0,0 +1,22 @@
+package daisycommon
```
zoran15: Add copyright

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -115,11 +116,19 @@ func validateSourceFile(storageClient domain.StorageClientInterface, sourceBucke
...
 	// Detect whether it's a compressed file by extracting compressed file header
-	if _, err = gzip.NewReader(rc); err == nil {
+	if _, err = gzip.NewReader(byteCountingReader); err == nil {
 		return daisy.Errf("cannot import an image from a compressed file. Please provide a path to an uncompressed image file. If the compressed file is an image exported from Google Compute Engine, please use 'images create' instead")
 	}
 
+	// By calling gzip.NewReader above, a few bytes were read from the Reader in
+	// an attempt to decode the compression header. If the Reader represents
+	// an empty file, then BytesRead will be zero.
+	if byteCountingReader.BytesRead <= 0 {
```
zoran15: Does this work for non-zip files as well?

EricEdens: Very good question -- it does work, since that gzip.NewReader constructor reads bytes from the stream no matter what. In part to figure out whether it's compressed at all.

Good news: Here's a test that confirms this assumption: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/gce_vm_image_import/importer/importer_test.go#L192

cli_tools/daisycommon/io_test.go
EricEdens: fixed

cli_tools/daisycommon/io.go
EricEdens: fixed

dntczdx:
Nice finding, and don't forget to list all these fixes so we can look back which reliability improvements have been done

PR #981, opened 2019-10-21
[error log] add log prefix to pass gcloud filter

dntczdx:
This prefix is required because gcloud filters log based on prefix. Without this prefix, the final error msg won't show up in gcloud console.


cli_tools/gce_vm_image_export/exporter/exporter.go
```diff
@@ -46,6 +47,10 @@ const (
 	SourceImageFlagKey    = "source_image"
 )
 
+const (
+	logPrefix = "[image-export]"
```
zoran15: Prefix should end with a space. Otherwise, log message will be concatenated directly with the "]".

Applies for other prefixes.

dntczdx: I added a space when SetPrefix. For gcs logging, the space is added when writing the actual log, so I won't add the space for both prefix.

PR #983, opened 2019-10-22
[logging] abstracts class for cli tool with logging

dntczdx:
Use CliToolWithLogging as the abstract class of all cli tools with logging.
ImageImport/ImageExport/InstanceImport follows the pattern.
In the future, new tool (e.g. InstanceExport?) can just follow the pattern to support logging by simply implementing CliToolWithLogging.

This new abstracts avoided implicitly update input params by pointer. It forced tool's MainFunc to return a "updatedParams map[string]string", which is adopted by CliToolWithLogging to update log.

"updatedParams" is used to update 'project', and can be used for other params as well in the future.

### DIFF
--- a/cli_tools/common/utils/logging/service/cli_tool_with_log.go
+++ b/cli_tools/common/utils/logging/service/cli_tool_with_log.go
@@ -1,0 +1,44 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package service
+
+import (
+	"flag"
+	"os"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// CliToolWithLogging abstracts the interface of a cli tool with logging
+type CliToolWithLogging interface {
+	// ActionType indicates the action type of the tool
+	ActionType() ActionType
+
+	// InitParamLog initialize tool params in the log
+	InitParamLog() InputParams
+
+	// MainFunc is the main function of the tool
+	MainFunc() (*daisy.Workflow, map[string]string, error)
+}
+
+// RunCliToolWithLogging runs the cli tool with server logging
+func RunCliToolWithLogging(t CliToolWithLogging) {
+	flag.Parse()
+	paramLog := t.InitParamLog()
+	l := NewLoggingServiceLogger(t.ActionType(), paramLog)
+	if _, err := l.runWithServerLogging(t.MainFunc); err != nil {
+		os.Exit(1)
+	}
+}
--- a/cli_tools/common/utils/logging/service/cli_tool_with_log.go
+++ b/cli_tools/common/utils/logging/service/cli_tool_with_log.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"os"
 
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/param"
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
 
@@ -32,7 +33,7 @@ type CliToolWithLogging interface {
 	InitParamLog() InputParams
 
 	// Run is the entry function of the tool
-	Run() (*daisy.Workflow, map[string]string, error)
+	Run() (*daisy.Workflow, *param.UpdatedParams, error)
 }
 
 // RunCliToolWithLogging runs the cli tool with server logging
--- a/cli_tools/common/utils/logging/service/cli_tool_with_log.go
+++ b/cli_tools/common/utils/logging/service/cli_tool_with_log.go
@@ -21,16 +21,17 @@ import (
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
 
-// CliToolWithLogging abstracts the interface of a cli tool with logging
+// CliToolWithLogging abstracts the interface of a cli tool with logging. A tool
+// which implemented this interface adopts log service automatically.
 type CliToolWithLogging interface {
 	// ActionType indicates the action type of the tool
 	ActionType() ActionType
 
-	// InitParamLog initialize tool params in the log
+	// InitParamLog initializes tool's input params in order to be logged
 	InitParamLog() InputParams
 
-	// MainFunc is the main function of the tool
-	MainFunc() (*daisy.Workflow, map[string]string, error)
+	// Run is the entry function of the tool
+	Run() (*daisy.Workflow, map[string]string, error)
 }
 
 // RunCliToolWithLogging runs the cli tool with server logging
@@ -38,7 +39,7 @@ func RunCliToolWithLogging(t CliToolWithLogging) {
 	flag.Parse()
 	paramLog := t.InitParamLog()
 	l := NewLoggingServiceLogger(t.ActionType(), paramLog)
-	if _, err := l.runWithServerLogging(t.MainFunc); err != nil {
+	if _, err := l.runWithServerLogging(t.Run); err != nil {
 		os.Exit(1)
 	}
 }
--- a/cli_tools/common/utils/logging/service/cli_tool_with_log.go
+++ b/cli_tools/common/utils/logging/service/cli_tool_with_log.go
@@ -24,7 +24,8 @@ import (
 // CliToolWithLogging abstracts the interface of a cli tool with logging. A tool
 // which implemented this interface adopts log service automatically.
 type CliToolWithLogging interface {
-	// ActionType indicates the action type of the tool
+	// ActionType indicates the action type of the tool. It represents which tool
+	// function this log line is about. "ImageImport" is an example.
 	ActionType() ActionType
 
 	// InitParamLog initializes tool's input params in order to be logged
--- a/cli_tools/common/utils/logging/service/log_entry.go
+++ b/cli_tools/common/utils/logging/service/log_entry.go
@@ -85,6 +85,19 @@ type InputParams struct {
 	InstanceImportParams *InstanceImportParams `json:"instance_import_input_params,omitempty"`
 }
 
+func (p *InputParams) commonParams() *CommonParams {
+	if p.ImageImportParams != nil {
+		return p.ImageImportParams.CommonParams
+	}
+	if p.ImageExportParams != nil {
+		return p.ImageExportParams.CommonParams
+	}
+	if p.InstanceImportParams != nil {
+		return p.InstanceImportParams.CommonParams
+	}
+	return nil
+}
+
 // ImageImportParams contains all input params for image import
 type ImageImportParams struct {
 	*CommonParams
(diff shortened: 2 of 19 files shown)


cli_tools/gce_ovf_import/main.go
```diff
@@ -93,7 +93,18 @@ func buildImportParams() *ovfimportparams.OVFImportParams {
 	}
 }
 
-func runImport() (*daisy.Workflow, error) {
+// InstanceExportTool is the tool for instance import
+type InstanceExportTool struct {
```
zoran15: InstanceImportTool

cli_tools/gce_ovf_import/main.go
```diff
@@ -93,7 +93,18 @@ func buildImportParams() *ovfimportparams.OVFImportParams {
...
+	service.CliToolWithLogging
+}
+
+// ActionType implements CliToolWithLogging
+func (t *InstanceExportTool) ActionType() service.ActionType {
+	return service.InstanceImportAction
+}
+
+// MainFunc implements CliToolWithLogging
+func (t *InstanceExportTool) MainFunc() (*daisy.Workflow, map[string]string, error) {
```
zoran15: Why not just Main()? Func is redundant as it's obvious it's a func. It shouldn't conflict with main().

Even better, Run() or similar would prevent any confusion

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -242,48 +242,51 @@ func runImport(ctx context.Context, varMap map[string]string, importWorkflowPath
...
-		return nil, err
+		return nil, updatedParams, err
 	}
 	zoneRetriever := storage.NewZoneRetriever(metadataGCE, computeClient)
 
 	region := new(string)
-	err = param.PopulateMissingParameters(project, &zone, region, &scratchBucketGcsPath,
+	err = param.PopulateMissingParameters(&project, &zone, region, &scratchBucketGcsPath,
 		sourceFile, metadataGCE, scratchBucketCreator, zoneRetriever, storageClient)
+	updatedParams["project"] = project
```
zoran15: Extract "project" as a constant to a common place shared across all cli tools. Maybe in service package?

daisy/step_includeworkflow.go
```diff
@@ -65,6 +65,7 @@ func (i *IncludeWorkflow) populate(ctx context.Context, s *Step) DError {
 	i.Workflow.Logger = i.Workflow.parent.Logger
 	i.Workflow.Name = s.name
 	i.Workflow.DefaultTimeout = s.Timeout
+	fmt.Println("incldued_wf name:", i.Workflow.Name, " timeout:", i.Workflow.DefaultTimeout)
```
zoran15: This should be removed? If not, typo in incldued_wf

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,44 @@
...
+package service
+
+import (
+	"flag"
+	"os"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// CliToolWithLogging abstracts the interface of a cli tool with logging
```
zoran15: Add more description what does CLI tool with logging mean

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,44 @@
...
+type CliToolWithLogging interface {
+	// ActionType indicates the action type of the tool
```
zoran15: more docs on what action type is

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,44 @@
...
+	ActionType() ActionType
+
+	// InitParamLog initialize tool params in the log
```
zoran15: What does it mean "initialize tool params in the log"?

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,44 @@
...
+	InitParamLog() InputParams
+
+	// MainFunc is the main function of the tool
+	MainFunc() (*daisy.Workflow, map[string]string, error)
```
zoran15: as commented below, a better name is needed here. Maybe Run()?

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// RunCliToolWithLogging runs the cli tool with server logging
+func RunCliToolWithLogging(t CliToolWithLogging) {
```
zoran15: minor: just Run() might be enough as "CliToolWithLogging" is already in the file name and in parameter type

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -171,27 +184,23 @@ type OutputInfo struct {
...
-func (l *Logger) updateParams(projectPointer *string) {
+func (l *Logger) updateParams(params map[string]string) {
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
 
-	if projectPointer == nil {
+	l.updateProject(params)
+}
+
+func (l *Logger) updateProject(params map[string]string) {
```
zoran15: since params are predefined, a struct is more appropriate instead of a map. This way, if some values in the future are not strings, we will be able to support them

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -251,8 +246,8 @@ func (l *Logger) sendLogToServerWithRetry(logExtension *ComputeImageToolsLogExte
 	for i := 0; i < maxRetry; i++ {
 		// Before sending a new request, wait for a while if server asked to do so
 		if nextRequestWaitMillis > 0 {
-			nextRequestWaitMillis = 0
 			time.Sleep(time.Duration(nextRequestWaitMillis) * time.Millisecond)
+			nextRequestWaitMillis = 0
```
zoran15: What was this change about?

cli_tools/common/utils/logging/service/cli_tool_with_log.go
dntczdx: modified

cli_tools/common/utils/logging/service/cli_tool_with_log.go
dntczdx: But on the caller side, file name and param type don't present. So This name looks more clear for caller.

cli_tools/common/utils/logging/service/logger.go
dntczdx: Fixed a bug that nextRequestWaitMillis is reset before it's consumed. This bug ignored the waiting mechanism of log service.

cli_tools/common/utils/logging/service/log_entry.go
dntczdx: I chose not to use a struct because the implementation (exporter, for example) shouldn't ref to the structs defined specifically for logging. The logging framework only allows main function to interact with logging.service package. The import, export and ovf_importer knows nothing about logging and can be executed separately.

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"flag"
+	"os"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// CliToolWithLogging abstracts the interface of a cli tool with logging
+type CliToolWithLogging interface {
```
zoran15: There's nothing specific to logging in this interface. It should be in a package common to CLI tools in general, not in a logging package. 

You can put UpdatedParams struct there as well (the other comment about not using map[string]string for updated params)

cli_tools/common/utils/logging/service/log_entry.go
zoran15: See my other comment about the location of some of the structs/interfaces.

Also, CLI tools already refer to logging/service so they are aware of being run within logging framework. E.g. they import
"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging/service"

and implement InitParamLog

cli_tools/common/utils/logging/service/cli_tool_with_log.go
dntczdx: It does have specific logic for logging. It defines some methods, which serves the function splitting of logging. RunCliToolWithLogging Executes the main flow of a cli tool with logging.

Correspondingly, core functionality of those tools can also be executed directly without this wrap. Directly call "Run" and no logging will happen.

There are also some other cli tools, e.g. gce_image_publish, who doesn't implement this interface. So, those tools won't have logging feature naturally.

cli_tools/common/utils/logging/service/log_entry.go
dntczdx: Only main function ref to logging/service, but not the core exporter/importer/ovfImproter. Just like we have daisy and daisy cli. That is what I am trying to achieve.

zoran15: Then have this struct in a package common to all importer/exporter or CLI tools in general. 

By using a map, we're not achieving encapsulation or separation of concerns because importer/exporter still have to be aware they need to return some parameters. We only loose type safety with the current implementation.

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -171,27 +188,23 @@ type OutputInfo struct {
...
-	if l.Params.ImageExportParams != nil {
-		l.Params.ImageExportParams.CommonParams.Project = project
-		l.Params.ImageExportParams.CommonParams.ObfuscatedProject = obfuscatedProject
-	}
-	if l.Params.InstanceImportParams != nil {
-		l.Params.InstanceImportParams.CommonParams.Project = project
-		l.Params.InstanceImportParams.CommonParams.ObfuscatedProject = obfuscatedProject
+	if commonParams := l.Params.commonParams(); commonParams != nil {
+		commonParams.Project = project
+		commonParams.ObfuscatedProject = obfuscatedProject
```
EricEdens: Having project mutable, and allowing another object to mutate its field, makes it difficult for me to reason about it. Here are some issues that come up in my mind:
 1. What code is responsible for calling this method?
 2. What happens if it doesn't get called in the the future?
 3. Is there an order of events that need to be maintained?
 4. What code reads this field?

cli_tools/common/utils/logging/service/cli_tool_with_log.go
```diff
@@ -0,0 +1,46 @@
...
+	// ActionType indicates the action type of the tool. It represents which tool
+	// function this log line is about. "ImageImport" is an example.
+	ActionType() ActionType
+
+	// InitParamLog initializes tool's input params in order to be logged
+	InitParamLog() InputParams
+
+	// Run is the entry function of the tool
+	Run() (*daisy.Workflow, map[string]string, error)
```
EricEdens: The usage of map[string]string here and elsewhere feels dangerous for two reasons:
1. There's no contract that the caller or the receiver can agree upon.
2. The map is mutable, so it's unclear:
   a. Who is responsible for setting the fields
   b. Who is allowed to mutate the fields

cli_tools/common/utils/logging/service/log_entry.go
dntczdx: I read again about your comment. My design was: RunCliToolWithLogging ask for a returned map, and consume it in its own way. Thus, we can change the logging implementation safely. So I tried to avoid a fixed struct from service package. Your suggestion sounds reasonable: a struct from common package doesn't force tools to understand how logging is implemented. Let me do it.

Overall, Eric prefer a different pattern of architect. Let me finish my implementation, and you can discuss more which one makes more sense when I am oof...

dntczdx:
@zoran15 @EricEdens 
I have tried to resolve all comments from Zoran.

Overall, Eric prefer a different pattern of architect.
My perspective: the abstraction is straight forward and the easiest to understand.
Eric: This object-oriented abstraction is not easy to understand.

 Let me finish my implementation, and you can discuss more which one makes more sense when I am oof... If none of them are ideal, we can even keep it as-is.

cli_tools/common/utils/logging/service/cli_tool_with_log.go
dntczdx: Agreed, all that I tried to avoid is to prevent reusing strcuts from logging.service package. I have already used a struct from param package to resolve this problem in my latest commit.

dntczdx: Need to work more on "mutable" thing. That may imply the ideal solution.

PR #984, opened 2019-10-23
Save serial logs if there's an import failure.

EricEdens:
Notes:
- I've piggy-backed on Daisy's `DisableCloudLogging` flag; if the user specifies that flag, then we won't keep the serial output.
- Logs are only retained if there's a workflow error.

Testing:
 - In addition to unit tests, I ran OVF's import integration tests.

### DIFF
--- a/cli_tools/common/utils/logging/service/logger.go
+++ b/cli_tools/common/utils/logging/service/logger.go
@@ -181,6 +181,9 @@ func (l *Logger) getOutputInfo(w *daisy.Workflow, err error) *OutputInfo {
 	if err != nil {
 		o.FailureMessage = getFailureReason(err)
 		o.FailureMessageWithoutPrivacyInfo = getAnonymizedFailureReason(err)
+		if w != nil && w.Logger != nil {
+			o.SerialOutputs = w.Logger.ReadSerialPortLogs()
+		}
 	}
 
 	return &o
--- a/daisy/logger.go
+++ b/daisy/logger.go
@@ -171,11 +171,11 @@ func (l *daisyLog) WriteSerialPortLogs(w *Workflow, instance string, buf bytes.B
 }
 
 func (l *daisyLog) ReadSerialPortLogs() []string {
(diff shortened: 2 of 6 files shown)


cli_tools/common/utils/logging/service/logger.go
```diff
@@ -178,6 +178,9 @@ func (l *Logger) getOutputInfo(w *daisy.Workflow, err error) *OutputInfo {
 	if err != nil {
 		o.FailureMessage = getFailureReason(err)
 		o.FailureMessageWithoutPrivacyInfo = getAnonymizedFailureReason(err)
+		if w != nil && w.Logger != nil {
```
zoran15: Due to serial logs being potentially large, it would good if we had a flag/config that would control if serial console logs are stored (it can default to true).

daisy/logger.go
```diff
@@ -159,6 +170,14 @@ func (l *daisyLog) WriteSerialPortLogs(w *Workflow, instance string, buf bytes.B
 	writeLog(str)
 }
 
+func (l *daisyLog) ReadSerialPortLogs() []string {
+	logs := make([]string, 0, len(l.serialLogs))
```
zoran15: We lose the information of which instance has logged a particular serial log (e.g. disk importer vs translator). In most cases we might be able to figure which one is which, but logging instance name in some way would reduce the workload on our side and uniquely identify which instance has logged a particular log file. It could be as simple as adding instance name with a specific prefix as the first line of each serial log.

cli_tools/common/utils/logging/service/logger.go
EricEdens: We don't save logs if the use has specified [DisableCloudLogging](https://github.com/GoogleCloudPlatform/compute-image-tools/search?q=DisableCloudLogging&unscoped_q=DisableCloudLogging). Or were you thinking of something else?

daisy/logger.go
EricEdens: Good idea! Added that.

PR #985, opened 2019-10-23
uefi-compatible flag in image import and OVF import wrappers

zoran15:


cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -84,28 +85,44 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
 			}
 		}
-	}
+	})
+}
+
+// UpdateToUEFICompatible updates workflow resources (disks and images) to be UEFI compatible
+func UpdateToUEFICompatible(workflow *daisy.Workflow) {
+	WorkflowStepIterator(workflow, func(step *daisy.Step) {
+		if step.CreateDisks != nil {
```
EricEdens: My gut feeling is that "daisy_utils.go" isn't the best place for this, given how specific it is to a particular use case.

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -131,3 +148,19 @@ func RemovePrivacyLogTag(message string) string {
 
 	return message
 }
+
+// WorkflowStepIterator iterates over all workflow steps, including included
+// workflow steps, and calls cb callback function
+func WorkflowStepIterator(workflow *daisy.Workflow, cb func(step *daisy.Step)) {
+	if workflow == nil {
```
EricEdens: How about making this a method directly on workflow?

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -131,3 +148,19 @@ func RemovePrivacyLogTag(message string) string {
...
+		return
+	}
+
+	for _, step := range workflow.Steps {
+		if step.IncludeWorkflow != nil {
+			//recurse into included workflow
+			WorkflowStepIterator(step.IncludeWorkflow.Workflow, cb)
+		}
```
EricEdens: Does Daisy allow circular dependencies?

cli_tools/common/utils/string/string_utils.go
```diff
@@ -0,0 +1,40 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package string
+
+import "sort"
+
+// CombineStringSlices merges two slices of Beta Guest OS features and
```
EricEdens: Stale comment ("Beta Guest OS features")

cli_tools/common/utils/daisy/daisy_utils_test.go
```diff
@@ -151,6 +153,28 @@ func testRemovePrivacyTag(t *testing.T, originalMessage string, expectedMessage
 	}
 }
 
+func TestUpdateToUEFICompatible(t *testing.T) {
+	w := createWorkflowWithCreateDiskImageAndIncludeWorkflow()
+
+	UpdateToUEFICompatible(w)
+
```
EricEdens: I think it'd be clearer to put the whole structs into this method: one for before update, and one for what you expect after the update, and then do a single assertion that the result is what you expected. The alternative below makes it difficult for me to track where the differences are expected to be.

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -84,28 +85,44 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
+// UpdateToUEFICompatible updates workflow resources (disks and images) to be UEFI compatible
+func UpdateToUEFICompatible(workflow *daisy.Workflow) {
+	WorkflowStepIterator(workflow, func(step *daisy.Step) {
+		if step.CreateDisks != nil {
+			for _, disk := range *step.CreateDisks {
+				// for the time being, don't run Debian 9 worker in UEFI mode
+				if strings.Contains(disk.SourceImage, "debian-9-worker") {
+					continue
+				}
+				disk.Disk.GuestOsFeatures = daisy.CombineGuestOSFeatures(disk.Disk.GuestOsFeatures, "UEFI_COMPATIBLE")
```
EricEdens: Some more background here would be good -- why are we not doing UEFI for Debian 9? What's blocking us? 

Also, can we encode this exception to Debian 9 elsewhere, outside of code, for example in the template/script that creates the Debian 9 worker? Two advantages: One, this current approach invites future one-off modifications, and two, that template/script will own all customizations regarding Debian 9.

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -84,28 +85,44 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
-					networkInterface.AccessConfigs = accessConfigProvider()
+					networkInterface.AccessConfigs = []*compute.AccessConfig{}
 				}
 			}
 		}
-	}
+	})
+}
+
+// UpdateToUEFICompatible updates workflow resources (disks and images) to be UEFI compatible
```
EricEdens: It'd be nice for the comment to describe what updates are happening. As it stands, it could be modifying bytes in the image, inserting a new bootloader on the disk, who knows!

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: This is daisy_utils in cli_tools: Daisy utilities for CLI tools. This is used by both import wrapper and OVF importer. I think this is the most appropriate location for this piece of code. If you think there's a better place for it, please do suggest.

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: No

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: This is for Debian 9 worker, not for Debian 9 imports. Disks for workers are part of the workflow. I don't want to be changing how workers boot until UEFI is default.

cli_tools/common/utils/daisy/daisy_utils_test.go
EricEdens: What I'm trying to say is that to read the assertions in this test I have to do a lot of parsing and array indexing in my head and I think that having the two structs printed out in front of me would be easier. But this isn't the highest priority thing, and if you like this, then I'm fine.

zoran15: There are no arrays to parse in the assert, only 3 arguments: t, expected, actual.

cli_tools/common/utils/daisy/daisy_utils.go
EricEdens: So how about moving this to a location that is specific to your worker?

In addition to the reasons above, here's another: If a user has the subtstring "debian-9-worker" in their source image's name and tries to use UEFI.

cli_tools/common/utils/daisy/daisy_utils.go
EricEdens: A few places that come to mind:
 - Its own file
 - A file related to guest os features
 - A method on the workflow itself

cli_tools/common/utils/daisy/daisy_utils_test.go
EricEdens: Parsing in my head

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: Moved to workflow.go

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: Those are all valid options. I feel the current location is good enough, considering the temporary nature of this feature. If you feel strongly against it, I will look into moving it

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: Since import workflow only creates Debian 9 instances implicitly (no way for customer to have a CreateInstance with their own image in import workflow) the issue of customer provided image with "debian-9-worker" in it is not really an issue. Just in case, I switched to using the full URI to Debian 9 worker instead of just the portion.

What do you mean by "moving this to a location that is specific to your worker"?

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: Updated the description

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -84,28 +85,46 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
-	})
-}
-
-func updateAllInstanceAccessConfig(workflow *daisy.Workflow, accessConfigProvider func() []*compute.AccessConfig) {
-	for _, step := range workflow.Steps {
-		if step.IncludeWorkflow != nil {
-			//recurse into included workflow
-			updateAllInstanceAccessConfig(step.IncludeWorkflow.Workflow, accessConfigProvider)
-		}
+	daisy.WorkflowStepIterator(workflow, func(step *daisy.Step) {
```
dntczdx: IterateWorkflowSteps is better from my perspective. Iterator sounds more like a object than a operation.

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: Also renamed per Tony's suggestion and made it a member func

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: renamed and made a member func

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -84,28 +85,44 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
+	})
+}
+
+// UpdateToUEFICompatible updates workflow resources (disks and images) to be UEFI compatible
+func UpdateToUEFICompatible(workflow *daisy.Workflow) {
+	WorkflowStepIterator(workflow, func(step *daisy.Step) {
+		if step.CreateDisks != nil {
+			for _, disk := range *step.CreateDisks {
+				// for the time being, don't run Debian 9 worker in UEFI mode
+				if strings.Contains(disk.SourceImage, "debian-9-worker") {
```
dntczdx: This is hacky - relying on a specific image name is fragile. I feel we should have a tag for CreateInstance to prevent creating worker instance with UEFI.
Going further, why is UEFI not working for debian-9-worker if most of our images will be upgraded to UEFI?

cli_tools/common/utils/daisy/daisy_utils.go
EricEdens: Good point about it being temporary. Sounds like you're happy with it, so I'm fine!

cli_tools/common/utils/daisy/daisy_utils.go
zoran15: This is not image name but image family. As Debian 9 worker gets updated, the family name stays the same.

Alternatively, we could make any disks associated with CreateInstance in import workflows not marked as UEFI, only stand-alone ones (basically customer provided)

I'm not sure if debian-9 worker will work with UEFI. Most likely it will, but we have not done the testing and I would leave it booting using BIOS as it doesn't affect customer GPT formatted images. We can remove this once Debian 9 worker is tested with UEFI.

dntczdx: Sounds good, let's remove it later once it's ready due that UEFI is becoming a standard

PR #986, opened 2019-10-25
[logging] output 'imported file format'

dntczdx:
In order to build dashboard for imported file format, we need to output it. The source is from qemu-info.

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -152,8 +152,6 @@ echo "Import: $(serialOutputKeyValuePair "source-size-gb" "${SOURCE_SIZE_GB}")"
 echo "Import: $(serialOutputKeyValuePair "import-file-format" "${IMPORT_FILE_FORMAT}")"
 set -x
 
-echo "ImportFailed: Failed to convert source to raw. [Privacy-> test <-Privacy]" # TODO remove
-
 # Ensure the disk referenced by $DISKNAME is large enough to
 # hold the inflated disk. For the common case, we initialize
 # it to have a capacity of 10 GB, and then resize it if qemu-img
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -139,7 +139,9 @@ fi
 
 # Ensure the output disk has sufficient space to accept the disk image.
 # Disk image size info.
-SIZE_BYTES=$(qemu-img info --output "json" "${IMAGE_PATH}" | grep -m1 "virtual-size" | grep -o '[0-9]\+')
+FILE_INFO=$(qemu-img info --output "json" "${IMAGE_PATH}")
+SIZE_BYTES=$(echo $FILE_INFO | grep -oP  '(?<="virtual-size":[ *])[0-9]+' | head -1)
+IMPORT_FILE_FORMAT=$(echo $FILE_INFO | grep -oP  '(?<="format":[ *]").*?(?=")')
  # Round up to the next GB.
 SIZE_GB=$(awk "BEGIN {print int(((${SIZE_BYTES} - 1)/${BYTES_1GB}) + 1)}")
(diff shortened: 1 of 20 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -139,14 +139,17 @@ fi
 
 # Ensure the output disk has sufficient space to accept the disk image.
 # Disk image size info.
-SIZE_BYTES=$(qemu-img info --output "json" "${IMAGE_PATH}" | grep -m1 "virtual-size" | grep -o '[0-9]\+')
+FILE_INFO=$(qemu-img info --output "json" "${IMAGE_PATH}")
+SIZE_BYTES=$(echo $FILE_INFO | grep -oP  '(?<="virtual-size":[ *])[0-9]+' | head -1)
```
EricEdens: Is [jq](https://stedolan.github.io/jq/) available on the worker? If not, probably a good idea to add it and use it here.

dntczdx: It's not. In order to use it, we need to add it to debian-9-worker. I will ask Zach later whether it's proper to add this one more tool (in other words, dependency). It's enough for now to use regex to deliver this specific feature.

daisy_workflows/image_import/import_image.sh
```diff
@@ -139,14 +139,17 @@ fi
...
+IMPORT_FILE_FORMAT=$(echo $FILE_INFO | grep -oP  '(?<="format":[ *]").*?(?=")')
```
zoran15: Have you tested `qemu-img info` with various formats? Is it reliable? Does it happen that format is not present in the JSON? We should handle that scenario in case it ever happens, so we don't break the import just because format is not present. I would test this grep statement with a JSON without "format" field. If that happens, you can fallback to file extensions of the imported file.

daisy_workflows/image_import/import_image.sh
EricEdens: > It's enough for now to use regex to deliver this specific feature.

If it were enough than I wouldn't have given the feedback before :)

Bash commands are fragile and difficult to debug. Let's get jq added (or something similar); doing that will also address part of Zoran's concerns below.

daisy_workflows/image_import/import_image.sh
dntczdx: yes it is stable.
Even if failed to fetch the format, it won't break anything, just got an empty format.

When matching failed (which won't happen per current testing), I would leave it empty other than file extension because file extension is not reliable and in many cases it can be empty.

daisy_workflows/image_import/import_image.sh
dntczdx: Adding an extra dependency is relatively harder (we need to release a new debian-9-worker before we release this change). I won't add if I can find a as-good solution. So I tried a different way.

In previous PR, I used json just because it's already there. We can actually use regular output, which is super easy and stable to parse. It looks like:

tzz@tzz:~/t$ qemu-img info gce_export_new
image: gce_export_new
file format: raw
virtual size: 17M (17350656 bytes)
disk size: 17M

So I just grep "file format" and get the content after comma. It just matches a fixed and simple single-line format, so it would be safe enough.

daisy_workflows/image_import/import_image.sh
```diff
@@ -140,13 +140,15 @@ fi
 # Ensure the output disk has sufficient space to accept the disk image.
 # Disk image size info.
 SIZE_BYTES=$(qemu-img info --output "json" "${IMAGE_PATH}" | grep -m1 "virtual-size" | grep -o '[0-9]\+')
+IMPORT_FILE_FORMAT=$(qemu-img info "${IMAGE_PATH}" | grep -m1 "file format" | grep -oP '(?<=^file format:[ *]).*')
```
zoran15: Minor (not a regex expert): match end of line after .*

dntczdx: It's not required - grep always output single line; since ".*" matches everything in a line, it is ended at end of line.

PR #987, opened 2019-10-25
Mitigate race condition between createInstance and waitForInstanceSignal

EricEdens:
waitForInstanceSignal is able to kill an instance depending on its serial output, while createInstance is responsible for archiving serial output. Typically these steps run in parallel, leading to situations where an instance is killed prior to archiving the latest batch of logs. 

This commit adds a couple of mitigations:
 1. Always get the latest batch of logs when a workflow is canceled.
 2. Delay the return of createInstance by 20 seconds.

Testing:
 - Ran the OVF import integ tests.

### DIFF
--- a/daisy/step_create_instances.go
+++ b/daisy/step_create_instances.go
@@ -41,8 +41,6 @@ func logSerialOutput(ctx context.Context, s *Step, i *Instance, port int64, inte
 Loop:
 	for {
 		select {
-		case <-w.Cancel:
-			break Loop
 		case <-tick:
 			resp, err := w.ComputeClient.GetSerialPortOutput(path.Base(i.Project), path.Base(i.Zone), i.Name, port, start)
 			if err != nil {
@@ -72,6 +70,10 @@ Loop:
 				gcsErr = true
 				w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error saving log to GCS: %v", i.Name, err)
 				continue
+			}
+
+			if w.isCanceled() {
+				break Loop
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


adjackura:
Please don't add sleep to an unconnected step to solve this, waitForInstanceSignal doesn't take any action other than register the signal as soon as it comes in. I think it would be better to have one watcher registered per serial pot output that is required, then we don't need to have multiple readers. the wait for instance signal step can then just register its frequency and the watcher will write the output to a shared location, then the logger and the wait step can just pull from that.

EricEdens:
I 100% agree that this doesn't solve the problem, but rather it's a *mitigation* for it. 

The issue is that a fraction of serial port logs don't include the tail, and the tail contains the context around why waitForInstanceSignal decided to kill the workflow.

I propose pushing this mitigation now, allowing us to address the immediate problem, and then we can look at re-working log monitoring at a subsequent point.

dntczdx:
Offline discussed with Eric. 20s is too long from perf's perspective and it doesn't guarantee the result. On the other hand, this implementation can unblock the investigation for now. Let's do it as a workaround; meanwhile, open a tracking bug to follow up with the ideal solution suggested by Andrew.

adjackura:
20s is still way to long and this will impact every single workflow that uses WaitForInstanceSignal. if logging errors is what you care about simply add a 4s wait to cleanup for instances, that will make sure we don't delete an instance before we log its output (interval is set to 3s)

EricEdens:
> Add a 4s wait to cleanup if all you care about is failure logs, this would not impact current usage other than failures

Sounds good!

I moved the sleep to `cleanup`. AFAICT, though, it'll be executed for both success and failure. Or were you thinking of something else?

PR #989, opened 2019-10-30
[logging] v2 - abstracts class for cli tool with logging

dntczdx:
This is a different implementation of https://github.com/GoogleCloudPlatform/compute-image-tools/pull/983

Similarly, it tries to resolve the problem that some parameters have to be updated during the tool is executed.
When I try to adopt some feedbacks from Eric and Zoran, the whole solution became closer to the Eric's proposed one - it's like a mixture of my last implementation and Eric's. Since the shape of the design is quite different from 983, I created a new PR so you can compare them.

1. I removed the abstract class in this version due that it's harder to get a common top-level common code, as well as the abstract class shared pretty few common logic.
2. The key is that I added a "UpdatableParam" (maybe MutableParam is a better name), which replaces param pointer that needs to be updated during the tool's executed. 
3. The "UpdatableParam" is registered in InputParams, so logger.Update can consume.
4. Logically, it collects most of the params at the beginning for "Start" log, while it reflects updatable params after the execution and before "Success/Failure" log.
5. "UpdatableParam" is updated by core importer/exporter.

### DIFF
--- a/cli_tools/common/utils/logging/service/log_entry.go
+++ b/cli_tools/common/utils/logging/service/log_entry.go
@@ -14,6 +14,10 @@
 
 package service
 
+import (
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/param"
+)
+
 // logRequest is a server-side pre-defined data structure
 type logRequest struct {
 	ClientInfo    clientInfo `json:"client_info"`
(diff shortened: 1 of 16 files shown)


cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -171,27 +190,22 @@ type OutputInfo struct {
...
-	}
-	if l.Params.ImageExportParams != nil {
-		l.Params.ImageExportParams.CommonParams.Project = project
-		l.Params.ImageExportParams.CommonParams.ObfuscatedProject = obfuscatedProject
-	}
-	if l.Params.InstanceImportParams != nil {
-		l.Params.InstanceImportParams.CommonParams.Project = project
-		l.Params.InstanceImportParams.CommonParams.ObfuscatedProject = obfuscatedProject
+		commonParams.Project = project
+		commonParams.ObfuscatedProject = obfuscatedProject
```
EricEdens: `UpdatableProject` has a bad code smell to me.
1. From its name, it sounds mutable. Is that what you want?
2. Who is responsible for initializing it? Is it really okay that the value is ever nil?

The approach that I would have expected is to have the 'project' value created and validated once, and then made immutable. Are there limitations that make that impossible?

PR #993, opened 2019-10-31
Run start script asynchronously

dntczdx:
StartScript has a limit of execution time - even though I failed to find a doc to clarify that, I realized this behavior in the weekly export test, which exports a 4TB image.
Exporting it self works fine (up to 13 hours); however, the "verify" step always fail after ~20 hours, which imports the huge image from gcs and takes really long.

Looking into the log, there isn't any clue what happened, the instance just got terminated if the start script executes for ~20 hours.

In order to unblock this exporting test, and more importantly, unblock any long import/export operation, I moved the logic of the scripts into a "main" function, and call it asynchronously:

main &

So the main process of start script can finish in 1s, while the core logic is executed in a background process.


hopkiw:
What is the surrounding context? How will success or failure be reported, and what was causing the timeout?

dntczdx:
> What is the surrounding context? How will success or failure be reported, and what was causing the timeout?

It was initially detected by ci-daisy-e2e-weekly:
https://prow.k8s.io/view/gcs/compute-image-tools-test/logs/ci-daisy-e2e-weekly/1187781896604487680

The test should pass with the fix.

It's not a timeout. Importing a 4TB image from GCS needs >20 hours.

PR #994, opened 2019-11-04
Update error message when input image is gzipped.

EricEdens:


cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -124,7 +124,7 @@ func validateSourceFile(storageClient domain.StorageClientInterface, sourceBucke
 	byteCountingReader := daisycommon.NewByteCountingReader(rc)
 	// Detect whether it's a compressed file by extracting compressed file header
 	if _, err = gzip.NewReader(byteCountingReader); err == nil {
-		return daisy.Errf("cannot import an image from a compressed file. Please provide a path to an uncompressed image file. If the compressed file is an image exported from Google Compute Engine, please use 'images create' instead")
+		return daisy.Errf("the input file appears to be a tar.gz, which is not supported by 'gcloud compute images import'. To import a tar.gz that was exported from Google Compute Engine, please use 'gcloud compute images create' instead")
```
zoran15: At this point, we are sure the file is a zipped file. "appears" sounds like we are not sure.

EricEdens: We're sure that it's gzipped, not that it's a tar.gz. I'd be fine with saying "the input file is gzipped, which is not ..."

PR #997, opened 2019-11-06
Import/export wrappers: Remove new lines from log errors

zoran15:
Remove new lines from log errors as only the first line is filtered by gcloud due to subsequent lines not having an appropriate prefix. Alternate solution where each line would have a prefix would be more effort to implement, so this simpler solution was chosen. Not many cases where multi line errors are returned from Daisy.


cli_tools/common/utils/logging/service/logger.go
```diff
@@ -237,6 +237,12 @@ func (l *Logger) runWithServerLogging(function func() (*daisy.Workflow, error),
 	return logExtension, err
 }
 
+func removeNewLinesFromMultilineError(s string) string {
```
EricEdens: unit test?

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -237,6 +237,12 @@ func (l *Logger) runWithServerLogging(function func() (*daisy.Workflow, error),
...
+	// first line in a multi error line is of "Multiple errors" type and doesn't need a separator
```
EricEdens: I'm not familiar with the "multiple errors" type. Is that a golang thing? Daisy?

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -237,6 +237,12 @@ func (l *Logger) runWithServerLogging(function func() (*daisy.Workflow, error),
...
+	firstNewLineRemoved := strings.Replace(s, "\n", " ", 1)
+	return strings.Replace(firstNewLineRemoved, "\n", "; ", -1)
```
EricEdens: minor: Any reason to not use `ReplaceAll`? (I had to read the source code of strings.Replace to figure out what `-1` meant.)

cli_tools/common/utils/logging/service/logger.go
zoran15: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/error.go#L154

PR #998, opened 2019-11-06
Allow instances without GCS access to use startup scripts.

EricEdens:
Prior to this, instances needed read access to GCS in order to access their startup script. Lacking this permission, a booting instance would experience the `Failed to download metadata script` error:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/e1ff1f2bd9cf7df342b4f36a79e91db6bc1f0e49/daisy_workflows/image_import/import_disk.wf.json#L100

The change is best-effort, since only some scripts are inlinable. Specifically, there's a maximum size of 256kb for the script, and there is a limited set of script types that are accepted. More info: https://cloud.google.com/compute/docs/startupscript


EricEdens:
> This is a pretty big change in the default setup, could you potentially just use the existing ${SOURCE:<source_name>} syntax in the workflows you care about?

Well, I'll be -- yeah, I wasn't aware of that, and that looks perfect.

PR #1006, opened 2019-11-12
Image export permission error handling

zoran15:
Image Export: 
* inline startup script
* log GCS read permission with "ExportFailed:" prefix so it gets picked up by the calling code
* don't retry on permission error


cli_tools/gce_export/main.go
```diff
@@ -145,9 +148,13 @@ func (b *bufferedWriter) uploadWorker() {
 			}()
 			if err != nil {
 				fmt.Printf("Failed %v time(s) to upload '%v', error: %v\n", i, in, err)
-				if i > 16 {
-					log.Fatal(err)
+
+				// Don't retry if permission error as it's not recoverable.
+				gAPIErr, isGAPIErr := err.(*googleapi.Error)
+				if i > 16 || (isGAPIErr && gAPIErr.Code == 403 && gcsPermissionErrorRegExp.MatchString(gAPIErr.Message)) {
+					log.Fatalf("ExportFailed: Failed to export disk source to GCS. [Privacy->%v<-Privacy]", err)
```
adjackura: Don't log with this style in this tool. This should be wrapped by something else higher up the stack

zoran15: How?

PR #1008, opened 2019-11-13
osconfig_test: copy autogen proto and gapic code for use with tests

adjackura:


osconfig_tests/utils/utils.go
```diff
@@ -93,6 +93,7 @@ EOM`
 // InstallOSConfigDeb installs the osconfig agent on deb based systems.
 func InstallOSConfigDeb() string {
 	return fmt.Sprintf(`systemctl stop google-osconfig-agent
+dpkg --configure -a
```
iamsubratp: why this?

adjackura: This is a hack to fix the service restart from breaking dpkg, I'm working on a packaging fix to have more garceful shutdowns of the service.

PR #1010, opened 2019-11-14
osconfig_tests: allow for tests to wait until zone quota is available

adjackura:


osconfig_tests/test_config/project_config.go
```diff
@@ -52,24 +52,45 @@ func GetProject(projectID string, testZones map[string]int) *Project {
...
-		return msg
+		p.mux.Unlock()
+		return z
 	}
+}
 
-	zi := rand.Intn(zc)
-	z := p.zoneIndices[zi]
+// ReturnZone returns a zone so other tests can use it.
+func (p *Project) ReturnZone(z string) {
```
ryanwe: GetZone and ReturnZone threw me off at first. Would Aquire/Release (following Go's semaphore semantics) be better?

osconfig_tests/test_config/project_config.go
```diff
@@ -52,24 +52,45 @@ func GetProject(projectID string, testZones map[string]int) *Project {
...
-	p.mux.Lock()
-	defer p.mux.Unlock()
+	for {
+		p.mux.Lock()
+
+		zc := len(p.zoneIndices)
+		if zc == 0 {
+			p.mux.Unlock()
+			time.Sleep(10 * time.Second)
+			continue
```
ryanwe: Can you update the comment that this can block forever if other tests fail to release their zoneCapacity?

osconfig_tests/test_config/project_config.go
```diff
@@ -52,24 +52,45 @@ func GetProject(projectID string, testZones map[string]int) *Project {
...
 func (p *Project) GetZone() string {
-	p.mux.Lock()
-	defer p.mux.Unlock()
+	for {
+		p.mux.Lock()
+
+		zc := len(p.zoneIndices)
+		if zc == 0 {
+			p.mux.Unlock()
+			time.Sleep(10 * time.Second)
```
iamsubratp: should we have a upper bound for the amount of time it should wait before just bailing out the execution?

adjackura: Maybe, we do have a timeout on the test though so I can look into adding this later

adjackura: Actually this was simple to implement

PR #1018, opened 2019-11-18
Support importing of Ubuntu 18.04.

EricEdens:
### DHCP configuration
I removed creating `/etc/network/interfaces` in `translate.py`, in favor of using [cloud-init's automatic DHCP configuration](https://cloudinit.readthedocs.io/en/latest/topics/network-config.html). This was motivated by two factors:
 1.  Ubuntu 18.04 [uses netplan](https://ubuntu.com/blog/ubuntu-bionic-netplan), requiring *some* alternative. 
 2. The instance from Azure mapped eth0 to `enp0s4`, while vmware and AWS mapped eth0 to `ens4`. I spent a few hours trying to find why Azure was different, without luck. However, by using cloud-init's DHCP autoconfig, we don't have to worry about which name to use.

### Improved reliability of integ tests
The integ tests attempt to install a package using `apt`. If an instance is configured to auto-update at boot, then there's a race condition with the auto-update process. To address this, I added retry with exponential backoff.

## Testing
 - Ran all of the new integ tests (one for each vmware, aws, and azure)
 - Ran the existing translate tests for Ubuntu 16.04 and 14.04.


daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -145,17 +145,13 @@ function check_cloud_init {
...
-      fail "apt-get update failed."
-    fi
-
     status "Checking if apt can install a package."
-    apt-get install --reinstall make
-    if [[ $? -ne 0 ]]; then
-      fail "apt-get cannot install make."
-    fi
+    for i in $(seq 1 20) ; do
+      apt-get update && apt-get install --reinstall make && return 0
```
zoran15: why are these two checks merged?

daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -160,6 +159,20 @@ def DistroSpecific(g):
...
+    logging.debug(str(e))
+    logging.warn('Could not uninstall Azure agent. Continuing anyway.')
+
+  try:
+    g.command(['apt-get', 'remove', '-y', '-f', 'waagent'])
+  except Exception as e:
+    logging.debug(str(e))
+    logging.warn('Could not uninstall Azure agent. Continuing anyway.')
+
+
```
zoran15: What's the relationship between waagent and walinuxagent?
Is it expected that both are present at the same time? 

If only one can be present, showing the warning message for the other one might be confusing. Also, showing the same warning twice.

Can we differentiate between the two in warning messages?

daisy_integration_tests/scripts/post_translate_test.sh
EricEdens: The previous code was making two assertions -- one that apt was updatable and one that make could be installed -- and either assertion could fail if the autoupdate script was holding a dpkg lock. By chaining them in this way, we benefit from one set of retry logic that will wrap both.

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: I made an Ubuntu 18.04 instance in Azure, and it had walinuxagent but not waagent. I posted a new commit that differentiates the error messages. 

I'm not aware of any condition that requires one or the other to be present. This just makes it flexible in case one is present and the other is not.

daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -160,6 +159,20 @@ def DistroSpecific(g):
   g.command(['update-grub2'])
 
 
+def remove_azure_agents(g):
```
zoran15: Isn't it most likely that Azure agent won't be present, unless an image is actually from Azure? Will the warning message show in that case as well? If so, we should make it sound less serious.

EricEdens: That's a very good point. We're already logging the apt message at a debug level, which will just say the package wasn't found. So I just removed the warn message.

PR #1021, opened 2019-11-19
Remove osconfig_tests, they moved to the osconfig repo

adjackura:
See https://github.com/GoogleCloudPlatform/osconfig/pull/105


iamsubratp:
might want to remove this from build targets

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/785e15ddb7b6c4e9fb51113ec6e1e551b82a8b33/test-infra/prowjobs/gobuild/main.sh#L41

PR #1022, opened 2019-11-19
Clarify roles that are required during import.

EricEdens:
- Makes the pattern matching more flexible by removing the surrounding text.
- Indicates which role needs to be added.

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -59,8 +59,8 @@ function resizeDisk() {
 
   echo "Import: Resizing ${diskId} to ${requiredSizeInGb}GB in ${zone}."
   if ! out=$(gcloud -q compute disks resize "${diskId}" --size="${requiredSizeInGb}"GB --zone="${zone}" 2>&1 | tr "\n\r" " "); then
-    if echo "$out" | grep -q "Required .compute.disks.resize. permission"; then
-      echo "ImportFailed: Failed to resize disk. The Compute Engine default service account needs the compute.disks.resize permission."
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -59,8 +59,8 @@ function resizeDisk() {
 
   echo "Import: Resizing ${diskId} to ${requiredSizeInGb}GB in ${zone}."
   if ! out=$(gcloud -q compute disks resize "${diskId}" --size="${requiredSizeInGb}"GB --zone="${zone}" 2>&1 | tr "\n\r" " "); then
-    if echo "$out" | grep -q "Required .compute.disks.resize. permission"; then
-      echo "ImportFailed: Failed to resize disk. The Compute Engine default service account needs the compute.disks.resize permission."
+    if echo "$out" | grep -qP "compute\.disks\.resize"; then
+      echo "ImportFailed: Failed to resize disk. The Compute Engine default service account needs the role: roles/compute.storageAdmin"
```
zoran15: Is admin role needed or the specific role is needed? 

I assume you switched to broader role as this is what's documented in our public docs? I wonder if we can be more specific in console logs and link to public docs.

EricEdens: Yeah, I switched to use `roles/compute.storageAdmin` since that phrase is easily findable through web searches, while `compute.disks.resize` was not. Also, the web UI is focused on attaching roles to accounts, rather than permissions.

I thought about adding the link, but my concern is that a broken link in the future would be a bad experience, whereas having the role here would be sufficient for either a web search or for someone to click around in the IAM web console to find it and add it.

PR #1025, opened 2019-11-22
Added -hostname to OVF import to allow custom instance hostnames.

zoran15:


cli_tools/common/utils/validation/validation_utils.go
```diff
@@ -16,10 +16,15 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
 
+var (
+	fqdnRegexp = regexp.MustCompile("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)+([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$")
```
EricEdens: What generally happens with big regexes like this is that they're either too restrictive or not restrictive enough, both of which lead to bad customer experiences. (If they're not restrictive enough then they're going to get the failure later in the import, which I assume is what you're trying to avoid by adding this.)

A couple of concrete examples:
1.  The regex here doesn't match the regex in the error message since it allows uppercases throughout and a leading digit.
2. The regex here doesn't validate the length of labels.

I'd suggest a couple of ways to move forward:

1. Do a quick validation here looking for characters that are not allowed ever. For example, throw an error if there's an underscore. Then, allow gcloud to throw a future error. This approach ensures that you're not too restrictive.
2. Update your match to strictly follow GCE's hostname rules. If you go this route, I suspect you'll find it more maintainable to look at approaches other than one big regex.

Personally, I'd go with (1) since it will give quick feedback for cases that are obviously incorrect. The solution will be simple enough, however, to verify that it won't have false positive matches.

PR #1027, opened 2019-11-27
-no-translate option in OVF import

zoran15:


cli_tools/test_data/test_import_ovf_no_translate.wf.json
```diff
@@ -0,0 +1,93 @@
+{
```
dntczdx: shouldn't test workflow include "daisy_workflows/ovf_import/import_ovf_no_translate.wf.json"? And I don't see the difference between this one and the real one

PR #1028, opened 2019-12-04
OVF Import: data disk steps to use workflow default timeout

zoran15:
OVF Import: data disk steps to use workflow default timeout instead of timeout being hardcoded to 1hr

### DIFF
--- a/daisy/step.go
+++ b/daisy/step.go
@@ -69,8 +69,12 @@ type Step struct {
 	testType stepImpl
 }
 
-// NewStep creates a Step with given name and timeout  with the specified workflow
+// NewStep creates a Step with given name and timeout  with the specified workflow.
+// If timeout is less or equal to zero, defaultTimeout from the workflow will be used
 func NewStep(name string, w *Workflow, timeout time.Duration) *Step {
+	if timeout <= 0 {
+		return &Step{name: name, w: w, Timeout: w.DefaultTimeout}
+	}
 	return &Step{name: name, w: w, timeout: timeout}
 }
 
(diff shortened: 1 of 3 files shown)


daisy/step.go
```diff
@@ -69,8 +69,12 @@ type Step struct {
 	testType stepImpl
 }
 
-// NewStep creates a Step with given name and timeout  with the specified workflow
+// NewStep creates a Step with given name and timeout  with the specified workflow.
```
dntczdx: I would create a new method (NewStepWithDefaultTimeout?) for this purpose because timeout<=0 looks like a hidden logic

EricEdens: +1

daisy/step.go
```diff
@@ -69,11 +69,20 @@ type Step struct {
...
+// If timeout is less or equal to zero, defaultTimeout from the workflow will be used
 func NewStep(name string, w *Workflow, timeout time.Duration) *Step {
+	if timeout <= 0 {
+		return &Step{name: name, w: w, Timeout: w.DefaultTimeout}
+	}
 	return &Step{name: name, w: w, timeout: timeout}
 }
 
+// NewStepDefaultTimeout creates a Step with given name using default timeout from the workflow
+func NewStepDefaultTimeout(name string, w *Workflow, timeout time.Duration) *Step {
```
EricEdens: need to remove the timeout param?

zoran15: Yeah. I think I'll go home get some sleep. :)

PR #1029, opened 2019-12-04
detect QUOTA_EXCEEDED when falling back to pd-standard

dntczdx:


dntczdx:
@googlebot I fixed it.

daisy/step_create_disks.go
```diff
@@ -93,3 +95,8 @@ func (c *CreateDisks) run(ctx context.Context, s *Step) DError {
 		return nil
 	}
 }
+
+func isQuotaExceeded(err error) bool {
+	gAPIErr, isGAPIErr := err.(*googleapi.Error)
+	return isGAPIErr && gAPIErr.Code == 403 && strings.Contains(gAPIErr.Message, "QUOTA_EXCEEDED")
```
EricEdens: Instead of string matching, can we use `error_details.proto` as discussed in https://cloud.google.com/apis/design/errors ?

dntczdx: Looking into https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L65
We will see that the only extra thing we can do is to check whether there is a "QuotaFailure" existing. However, beyond that, we still need a string match towards QuotaFailure.vialations.description. Basically, there isn't a enum field for us to match a fixed code.

So, it's not better than directly matching string towards the message body.

Thoughts?

EricEdens: But you can use type checking rather than depending on the HTTP error code and the string formatting of "QUOTA_EXCEEDED" . For example, from https://cloud.google.com/apis/design/errors , it looks like HTTP code 429 is may also be returned, in which case your check would break, right?  

https://github.com/googleapis/go-genproto/blob/master/googleapis/rpc/errdetails/error_details.pb.go#L724

dntczdx: Hi Eric, basically it's a process of narrow-down to check "code-->type-->description" (or, "code-->message"). It's a general practice to firstly check the code. 403 is wrong, I will change to 429, and I will do more test to ensure the condition is correct.

I chose to verify "code-->message" other than "code-->type-->description" is because they both ultimately rely on a string match in either way. The advantage of verifying "message" is because it contains "QUOTA_EXCEEDED", which behaves more like a code. The QuotaFailure.description contains a natural sentence like "Quota limit 'xxx' exceeded.", which relatively is not as stable.

Above statement is based on documents and my experience of testing quota issue. After the actual e2e testing, the answer may be different. I will choose the more reliable one.

dntczdx: Per my testing, the code is really 403, and it's not a google API error - it's handled by async helper layer, being transformed to an operation error, and then being constructed as a normal error.
Due that it's better not to assume how the helper layer construct the error, simply doing string match is a relatively better way.
https://github.com/dntczdx/compute-image-tools/blob/ad8d70929d00999fc2ca4dc26f6b6c611860aece/daisy/compute/compute.go#L275

EricEdens: The type, though, is `QuotaFailure`, right? So, you just need code-->type? Or am I missing something?

https://github.com/googleapis/go-genproto/blob/master/googleapis/rpc/errdetails/error_details.pb.go#L724

dntczdx: What you reference to belongs to googleAPIError. The caught error is not an googleAPIError - it's  wrapped by mutiple layers of normal error. The top-level wrapper is here:

EricEdens: The code that you referenced is utility method in daisy that makes a user-facing error message out of a complex error object. It intentionally throws aways details in order to simplify the representation of the error.

Now, you're taking the simplified representation and trying to use string matching to determine the aforementioned details. This is the fragility that I'm trying to describe. :)

Take a look at refactoring that method. One option is to have two versions: One that gives the original error, and one that gives a user friendly error. This is probably the easiest and a good place to start.

dntczdx: Hi Eric, the "original" error, without any utility processing, is a "OperationErrorErrors", which is from compute-gen, aka is from GCE official lib. It doesn't inherit golang error interface, so we can't handle it in regular error handling way. There is never a real golang error in the whole context, until the wrapper constructed one.

So, no matter how to do the refactoring, we always need to do:
1. extract error code and message from OperationErrorErrors.
2. Embed it in a golang error.
I don't feel it's worth introducing a new error class just for this specific scenario. Instead, we can improve the string match by using regex to (mostly) ensure the match is not an accident.

Please checkout my new code.

PR #1033, opened 2019-12-06
OVF Import: limit disk names to 63 characters

zoran15:
OVF Import: limit disk names to 63 characters to prevent failures when long instance name is provided.

### DIFF
--- a/cli_tools/common/utils/validation/validation_utils.go
+++ b/cli_tools/common/utils/validation/validation_utils.go
@@ -22,7 +22,9 @@ import (
 )
 
 var (
-	fqdnRegexp = regexp.MustCompile("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)+([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$")
+	rfc1035LabelRegexpStr = "[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9]"
+	rfc1035LabelRegexp    = regexp.MustCompile(rfc1035LabelRegexpStr)
+	fqdnRegexp            = regexp.MustCompile(fmt.Sprintf("^((%v)\\.)+(%v)$", rfc1035LabelRegexpStr, rfc1035LabelRegexpStr))
 )
 
 // ValidateStringFlagNotEmpty returns error with error message stating field must be provided if
@@ -41,6 +43,13 @@ func ValidateFqdn(flagValue string, flagKey string) error {
 		return daisy.Errf(fmt.Sprintf("The flag `%v` must conform to RFC 1035 requirements for valid hostnames. "+
 			"To meet this requirement, the value must contain a series of labels and each label is concatenated with a dot."+
 			"Each label can be 1-63 characters long, and the entire sequence must not exceed 253 characters.", flagKey))
+	}
+	return nil
+}
+
+func ValidateRfc1035Label(value string) error {
+	if len(value) > 63 || !rfc1035LabelRegexp.MatchString(value) {
+		return daisy.Errf(fmt.Sprintf("Value `%v` must conform to RFC 1035 requirements for valid labels.", value))
 	}
 	return nil
 }
--- a/cli_tools/common/utils/validation/validation_utils.go
+++ b/cli_tools/common/utils/validation/validation_utils.go
@@ -47,6 +47,7 @@ func ValidateFqdn(flagValue string, flagKey string) error {
 	return nil
 }
 
+// ValidateRfc1035Label validates a single label per RFC 1035
 func ValidateRfc1035Label(value string) error {
 	if len(value) > 63 || !rfc1035LabelRegexp.MatchString(value) {
 		return daisy.Errf(fmt.Sprintf("Value `%v` must conform to RFC 1035 requirements for valid labels.", value))
--- a/cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
+++ b/cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
@@ -40,13 +40,11 @@ func AddDiskImportSteps(w *daisy.Workflow, dataDiskInfos []ovfutils.DiskInfo) {
 	for i, dataDiskInfo := range dataDiskInfos {
 		dataDiskIndex := i + 1
 		dataDiskFilePath := dataDiskInfo.FilePath
-		diskNames = append(
-			diskNames,
-			fmt.Sprintf("%v-data-disk-%v", w.Vars["instance_name"].Value, dataDiskIndex))
+		diskNames = append(diskNames, generateDataDiskName(w.Vars["instance_name"].Value, dataDiskIndex))
 
 		setupDataDiskStepName := fmt.Sprintf("setup-data-disk-%v", dataDiskIndex)
-		diskImporterDiskName := fmt.Sprintf("disk-importer-%v", dataDiskIndex)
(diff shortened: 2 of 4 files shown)


cli_tools/common/utils/validation/validation_utils.go
```diff
@@ -44,3 +46,10 @@ func ValidateFqdn(flagValue string, flagKey string) error {
 	}
 	return nil
 }
+
+func ValidateRfc1035Label(value string) error {
+	if len(value) > 63 || !rfc1035LabelRegexp.MatchString(value) {
```
dntczdx: I am not sure how is the disk name decided here. Is it provided by user input, or generated by import workflow? If prior, we may explicitly tell them the length exceeds limit.

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater_test.go
```diff
@@ -122,6 +87,29 @@ func TestAddDiskImportSteps(t *testing.T) {
...
+	w := createBaseImportWorkflow("a-very-long-instance-name-that-is-at-the-limit-of-allowed-leng")
+
+	diskInfos := []ovfutils.DiskInfo{
+		{FilePath: "gs://abucket/apath/disk1.vmdk", SizeInGB: 20},
+		{FilePath: "gs://abucket/apath/disk2.vmdk", SizeInGB: 1},
+	}
+	AddDiskImportSteps(w, diskInfos)
+
+	assert.NotNil(t, w)
+	assert.Equal(t, 2+4*len(diskInfos), len(w.Steps))
```
EricEdens: Can you add a comment to explain what `2+4*len(diskInfos)` refers to?

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater_test.go
```diff
@@ -122,6 +87,29 @@ func TestAddDiskImportSteps(t *testing.T) {
 	assert.Equal(t, w.DefaultTimeout, w.Steps["wait-for-data-disk-2-signal"].Timeout)
 	assert.Equal(t, w.DefaultTimeout, w.Steps["delete-data-disk-2-import-instance"].Timeout)
 }
+func TestAddDiskImportStepsLongInstanceName(t *testing.T) {
```
EricEdens: Is this test expecting that 'a-very-long-instance-name-that-is-at-the-limit-of-allowed-leng' gets truncated? It'd be clear if you had two tests -- one where it gets truncated and one where it doesn't. Also, it would help the reader if you added some assertions showing the exact name (prefix plus suffix) that you're expecting.

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
```diff
@@ -40,13 +40,11 @@ func AddDiskImportSteps(w *daisy.Workflow, dataDiskInfos []ovfutils.DiskInfo) {
...
-		diskNames = append(
-			diskNames,
-			fmt.Sprintf("%v-data-disk-%v", w.Vars["instance_name"].Value, dataDiskIndex))
+		diskNames = append(diskNames, generateDataDiskName(w.Vars["instance_name"].Value, dataDiskIndex))
 
 		setupDataDiskStepName := fmt.Sprintf("setup-data-disk-%v", dataDiskIndex)
-		diskImporterDiskName := fmt.Sprintf("disk-importer-%v", dataDiskIndex)
-		scratchDiskDiskName := fmt.Sprintf("disk-importer-scratch-%v-%v", dataDiskIndex, w.Vars["instance_name"].Value)
+		diskImporterDiskName := fmt.Sprintf("disk-importer-%v-%v", dataDiskIndex, w.ID())
+		scratchDiskDiskName := fmt.Sprintf("disk-importer-scratch-%v-%v", dataDiskIndex, w.ID())
```
EricEdens: I'd recommend using generateDataDiskName here as well -- might as well benefit from the safety that it gives to protect yourself from future changes to w.ID().

cli_tools/common/utils/validation/validation_utils.go
zoran15: Customer provides instance name, we decide on disk names. Customer can provide an instance name up to 63 characters, but the problem is disk names have the same limitation while possibly there are multiple disks. So we run into the problem of disk names exceeding 63 characters if we just append a suffix to instance name.

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
zoran15: These two disk names will never exceed 63 characters, since dataDiskIndex will be up to 2-3 characters long once converted to string and w.ID() is 5 characters.

EricEdens: I'm aware that it's fine now, but my point is that you can make your code defensive against future changes to w.ID().

zoran15: I feel this is an overkill. We control what w.ID() is. 

Even if it gets switched under us it to something really long, like UUID, which is 36 characters when formatted as a string, we still have 39 characters of available space disk-importer-scratch and 47 characters for disk-importer.

zoran15: Also, the tests below are checking that all disk names are valid. So, if someone changes w.ID() to something longer than 39 characters (again, unlikely), data_disk_workflow_updater_test.go and disk naming can be fixed at that time. Since, user doesn't control any part of these two disk's names, I don't think we need to do anything now.

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater_test.go
zoran15: The scenario where disk name is not getting truncated is tested in the first test. I can add assertion that it's still valid.

The test doesn't really care what the name is as long as it's unique and valid. I will add uniqueness assertions.

EricEdens: Maybe just rename the test to make it clear that you're expecting the host name to be truncated. That'd make it clearer to me.

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
EricEdens: I'm encouraging defensive coding, and it would also ensure consistency to the reader (hey, we sanitize all disk names!). Is there a downside that I'm not seeing?

zoran15: Just to make it clear: you're asking me to write code that is defensive against itself?

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater_test.go
zoran15: Added a comment

cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater_test.go
zoran15: Renamed the test to include precondition and expected behavior in test method name.

cli_tools/common/utils/validation/validation_utils.go
dntczdx: Is there a hard requirement that disk name has to contain the full instance name? If not, we can just keep a partial instance name + a suffix. This is how daisy workflow handle the similar situation today - for example, when worker instance name + id exceeds 63, the name will be shortened, just kept a partial name + id.

zoran15: That's what generateDataDiskName() does.

dntczdx: perfect!

PR #1034, opened 2019-12-06
Add retry to serial log polling.

EricEdens:
Testing:
 - Ran OVF's integration tests.

### DIFF
--- a/daisy/step_create_instances.go
+++ b/daisy/step_create_instances.go
@@ -36,6 +36,7 @@ func logSerialOutput(ctx context.Context, s *Step, i *Instance, port int64, inte
 	var start int64
 	var buf bytes.Buffer
 	var gcsErr bool
+	var numErr int
 	tick := time.Tick(interval)
 
 Loop:
@@ -44,17 +45,22 @@ Loop:
 		case <-tick:
 			resp, err := w.ComputeClient.GetSerialPortOutput(path.Base(i.Project), path.Base(i.Zone), i.Name, port, start)
 			if err != nil {
-				// Instance is stopped or stopping.
+				numErr++
 				status, sErr := w.ComputeClient.InstanceStatus(path.Base(i.Project), path.Base(i.Zone), i.Name)
 				switch status {
 				case "TERMINATED", "STOPPED", "STOPPING":
+					// Instance is stopped or stopping.
 					if sErr == nil {
 						break Loop
 					}
 				}
-				w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error getting serial port: %v", i.Name, err)
-				break Loop
+				if numErr > 10 {
+					w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error getting serial port: %v", i.Name, err)
+					break Loop
+				}
+				continue
 			}
+			numErr++
 			start = resp.Next
 			buf.WriteString(resp.Contents)
 			wc := w.StorageClient.Bucket(w.bucket).Object(logsObj).NewWriter(ctx)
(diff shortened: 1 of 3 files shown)


daisy/step_create_instances.go
```diff
@@ -44,17 +45,22 @@ Loop:
...
 				}
-				w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error getting serial port: %v", i.Name, err)
-				break Loop
+				if numErr > 10 {
+					w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error getting serial port: %v", i.Name, err)
+					break Loop
+				}
+				continue
 			}
+			numErr++
```
zoran15: Shouldn't this be numErr=0 to reset error counter if no error occured on this tick?

daisy/step_create_instances.go
```diff
@@ -134,7 +140,7 @@ func (c *CreateInstances) run(ctx context.Context, s *Step) DError {
 				eChan <- newErr("failed to create instances", err)
 				return
 			}
-			go logSerialOutput(ctx, s, i, 1, 3*time.Second)
+			go logSerialOutput(ctx, s, i, 1, 5*time.Second)
```
zoran15: Why the change to 5 second refresh interval?

EricEdens: Three felt a bit spammy. On the other hand, it's been working for awhile, so, it's probably higher risk to change it at this point (since when an instance stops we lose access to its serial logs, so that would mean potentially losing two extra seconds of logs)

Fixed back to three.

daisy/step_create_instances.go
EricEdens: Oh wow, yeah, you're completely right. Running the tests again ..

zoran15: What about a unit test?

EricEdens: splendid idea! Added.

dntczdx:
May I know the background of this retry logic? Is t because a accidental polling failure can break the whole workflow?

EricEdens:
@dntczdx  -- It won't kill the workflow. Take a look at where log collection is started. The go routine is started and the results are not monitored.

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/894b5348c9445f123fb925642c4e1b8183beadef/daisy/step_create_instances.go#L143

The consequences, however, are twofold First, when this fails, we lose serial logs. Second, you get errors similar to those below. We've had error reports where the overall import was successful, but the user sees this message and thinks there was a problem. Also, we've had reports where there was a problem, and there's a message explaining the issue, but the user stops reading the logs when they see this issue.

```
CreateInstances: Instance "inst-bootstrap-import-and-translate-translate-translate--43242": error getting serial port: googleapi: Error 503: TIMEOUT, backendError
```

daisy/step_create_instances.go
```diff
@@ -44,17 +45,22 @@ Loop:
...
 				switch status {
 				case "TERMINATED", "STOPPED", "STOPPING":
+					// Instance is stopped or stopping.
 					if sErr == nil {
 						break Loop
 					}
 				}
-				w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error getting serial port: %v", i.Name, err)
-				break Loop
+				if numErr > 10 {
```
zoran15: Is ignoring the error for 30 seconds (since this is called with a 3 second interval) too long? If serial console logs can't be retrieved due to a transient error, this might be too long to wait before showing an error.

EricEdens: Parameter tuning is important for good user experience.

That said, in this case, this code is purely run in the background. So the large number of retries is to increase the probability that we'll gather all serial logs, in spite of any transient errors.

(Compare this with the waitForInstanceSignal, which *is* monitored by the main thread.)

PR #1035, opened 2019-12-06
Fix debug message when copying from GCS fails.

EricEdens:


daisy_workflows/image_import/import_image.sh
```diff
@@ -128,7 +128,7 @@ function copyImageToScratchDisk() {
       echo "Import: ${line}"
     done
     if grep -qP "storage\.objects\.(list|get)" gsutil.cp.err; then
-      echo "ImportFailed: Failed to resize disk. The Compute Engine default service account needs the role: roles/storage.objectViewer"
+      echo "ImportFailed: Failed to download image to scratch. The Compute Engine default service account needs the role: roles/storage.objectViewer"
```
zoran15: "scratch" might be confusing to users. Maybe "Failed to download image to local disk" or similar?

EricEdens: sgtm

EricEdens: switched to 'worker instance' instead of 'scratch'

PR #1037, opened 2019-12-09
Added old variants of --os flag values for Windows client

zoran15:
Added old variants of --os flag values for Windows client for backward compatibility


EricEdens:
I saw errors for 'windows-8-1-x64-byol' as well

@rofuentes any ideas?

PR #1044, opened 2019-12-13
Image import: set storage location automatically 

zoran15:
Image import: set storage location automatically to match source file region/multi-region.

ZoneRetriever was renamed to ResourceLocationRetriever. The new method added is GetLargestStorageLocation.


cli_tools/common/utils/storage/resource_location_retriever.go
```diff
@@ -28,44 +28,45 @@ var (
 	// for each multi region, regions are sorted by cost in ascending order as of the time of writing
 	// this code.
 	multiRegions = map[string][]string{
```
EricEdens: - Which cost is sorted by?
- Can you put a link to the page where someone can find that cost.
- For future proofing, can we switch to an API?
   - For example, on https://cloud.google.com/storage/docs/locations#location-dr , I see `NORTHAMERICA-NORTHEAST1`

zoran15: Which API are you referring to? I couldn't find an API for region to multi-region mapping.

zoran15: Added links to docs and expanded on the description.

I don't believe there's an API for region to multiregion mapping. The best I found is the docs you linked to. 

Regarding NORTHAMERICA-NORTHEAST1, it's in Canada, so I don't believe it should belong to "US" multi-region, although I couldn't find any docs that it doesn't belong there.

cli_tools/common/utils/storage/resource_location_retriever.go
```diff
@@ -0,0 +1,150 @@
...
+// GetZone retrieves GCE zone to run import in based on imported source file location and available
+// zones in the project.
+// If storageRegion is provided and valid, first zone within that region will be used.
+// If no storageRegion is provided, GCE Zone from the running process
+// will be used.
+func (rlr *ResourceLocationRetriever) GetZone(storageLocation string, project string) (string, error) {
+	zone := ""
+	var err error
+	if storageLocation != "" {
+		// pick a zone from the region where data is stored
```
EricEdens: I'm having a hard time figuring out what changed here. Let's revert the renames from this PR to retain the diffs, and then if you want to do the renames, add them as a separate PR.

zoran15: I'm not going to revert diffs just for code review purposes. We can sit at my desk and I can explain you what changed.

zoran15: The only func added was GetLargestStorageLocation and funcs it uses. The rest of the code is the same. Please review GetLargestStorageLocation.

cli_tools/common/utils/storage/resource_location_retriever.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func isZoneUp(zone *compute.Zone) bool {
+	return zone != nil && zone.Status == "UP"
+}
+
+// GetLargestStorageLocation returns the largest storage location that includes provided argument.
+// If argument is a multi-region, the argument is returned. If argument is a region within a multi-region,
+// the multi-region is returned. If argument is a region not within a multi-region, argument is returned.
+func (rlr *ResourceLocationRetriever) GetLargestStorageLocation(storageLocation string) string {
```
dntczdx: The meaning of 'Largest' is not straight forward to me. Is there a better word? Or a longer description.

zoran15: What about 'broadest'?

cli_tools/common/utils/storage/resource_location_retriever.go
zoran15: I don't want to set a precedent where code submitter has to rename files in order to facilitate easier code reviews. This will become messy very quickly.

The problem here is that git doesn't support rename but is using remove+add for renames. On top of that GitHub is using heuristics to determine if something is a rename or a new file. In this case, the changes are big enough that it decided it's a new file. 

I've created a separate commit with the changes to resource_location_retriever.go so you can review them easily:

https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1044/commits/1ba9196e85275a66deb452bb020119b337d26719

PR #1046, opened 2019-12-13
Support import of OpenSUSE 15.

EricEdens:
Testing:
- Built cli_tools/gce_vm_image_import/main.go, and ran an import using `-os=opensuse-15` 
- Ran `daisy_integration_tests/opensuse_15_1.wf.json`

### DIFF
--- a/daisy_workflows/image_import/suse/translate.py
+++ b/daisy_workflows/image_import/suse/translate.py
@@ -72,10 +72,7 @@ def DistroSpecific(g):
   # Update grub config to log to console, remove quiet and timeouts.
   g.command(
       ['sed', '-i',
-       '-e', r'/GRUB_CMDLINE_LINUX/s#"$# console=ttyS0,38400n8'
-       ' scsi_mod.use_blk_mq=Y"#',
-       '-e', r'/GRUB_CMDLINE_LINUX/s#quiet##',
-       '-e', r'/GRUB_[^=]*TIMEOUT=/s#=.*#=0',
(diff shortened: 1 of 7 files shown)


daisy_workflows/image_import/suse/translate.py
```diff
@@ -72,10 +72,7 @@ def DistroSpecific(g):
   # Update grub config to log to console, remove quiet and timeouts.
   g.command(
       ['sed', '-i',
-       '-e', r'/GRUB_CMDLINE_LINUX/s#"$# console=ttyS0,38400n8'
-       ' scsi_mod.use_blk_mq=Y"#',
-       '-e', r'/GRUB_CMDLINE_LINUX/s#quiet##',
-       '-e', r'/GRUB_[^=]*TIMEOUT=/s#=.*#=0',
```
zoran15: Why is this removed?

EricEdens: 1. It didn't work -- sed complained about an extra 's'.
2. Reading through it didn't make sense. `scsi_mod.use_blk_mq=Y` is an optimization that requires kernel support -- newer kernels are turning it on by default, but it doesn't seem safe to blindly add it. If the user already had it present, then nothing changes. Also, it's not included in the other translation scripts.
3. The expression '/GRUB_CMDLINE_LINUX/s#"$# console=ttyS0,38400n8' is broken, since it's throwing away the terminal quote.

So, I'm dubious of the whole block. I copied what we have from the ubuntu translate script to get something minimal that we know is working.

PR #1047, opened 2019-12-17
Don't emit error if we've previously read data from a serial log.

EricEdens:
Quit silently if we were able to read *some* data from the instance,
since there's a race condition where an instance can shut down
fast enough that the call to InstanceStatus will return a 404.

### DIFF
--- a/daisy/step_create_instances.go
+++ b/daisy/step_create_instances.go
@@ -36,6 +36,7 @@ func logSerialOutput(ctx context.Context, s *Step, i *Instance, port int64, inte
 	var start int64
 	var buf bytes.Buffer
 	var gcsErr bool
+	var readFromSerial bool
 	var numErr int
 	tick := time.Tick(interval)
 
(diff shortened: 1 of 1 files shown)


daisy/step_create_instances.go
```diff
@@ -55,11 +56,18 @@ Loop:
 					}
 				}
 				if numErr > 10 {
-					w.LogStepInfo(s.name, "CreateInstances", "Instance %q: error getting serial port: %v", i.Name, err)
+					// Quit silently if we were able to read *some* data from the instance,
+					// since there's a race condition where an instance can shut down
+					// fast enough that the call to InstanceStatus will return a 404.
```
zoran15: Move this below the if. Otherwise, it appears as if this comment applies to the if statement, when in fact it applies to the nop outside.

EricEdens: So, the comment is intended to explain the if block, and the circumstances under which we expect it to be executed ("*if* we were able"). I changed the wording of th comment, however, to focus on logging rather than quitting.

adjackura: Why don't we just add a check for sErr, if we get a 404 then quit the same as if we got "TERMINATED", "STOPPED", "STOPPING"

EricEdens: I've seen 404s returned prior to PROVISIONING, so we'd still need to decide whether the 404 meant to quit or to wait longer.

https://cloud.google.com/compute/docs/instances/instance-life-cycle

adjackura: Ack

PR #1050, opened 2019-12-20
Add support for SUSE and SLES.

EricEdens:
Added import support for:
 - OpenSUSE 15
 - SLES 12 BYOL , SLES 15 BYOL

Testing:
 - sles_12_5_byol.wf.json
 - sles_15_1_byol.wf.json
 - opensuse_15_1.wf.json


daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
...
+    for subscription in distro.subscriptions:
+      try:
+        g.command(['SUSEConnect', '-p', subscription])
+      except Exception as e:
+        raise ValueError(
+            'Command failed: SUSEConnect -p {}: {}'.format(subscription, e))
+
+
+def install_packages(distro, g, install_gce):
+  g.command(['zypper', 'refresh'])
```
zmarano: You may want to catch errors here as well. In the case there is no repo access or there are errors.

daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
...
 
-  # Update grub config to log to console, remove quiet and timeouts.
-  g.command(
-      ['sed', '-i',
-       '-e', r'/GRUB_CMDLINE_LINUX/s#"$# console=ttyS0,38400n8'
-       ' scsi_mod.use_blk_mq=Y"#',
-       '-e', r'/GRUB_CMDLINE_LINUX/s#quiet##',
-       '-e', r'/GRUB_[^=]*TIMEOUT=/s#=.*#=0',
-       '/etc/default/grub'])
+def install_virtio_drivers(g):
```
zmarano: This is interesting. This shouldn't be necessary- did you run into problems here?

daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
...
+gcp_sdk = Package('google-cloud-sdk', gce=True, required=False)
+gce_init = Package('google-compute-engine-init', gce=True, required=True)
+gce_oslogin = Package('google-compute-engine-oslogin', gce=True, required=True)
+
+_distros = [
+    SuseRelease(
+        flavor='opensuse',
+        major='15',
+        minor='1',
+        packages=[cloud_init, gcp_sdk, gce_init, gce_oslogin]
```
zmarano: We shouldn't need cloud-init... in fact it will likely cause conflicts.

daisy_workflows/linux_common/utils/diskutils.py
```diff
@@ -56,6 +56,10 @@ def MountDisk(disk):
   # mounting the filesystems in the correct order.
   mps = g.inspect_get_mountpoints(roots[0])
 
+  g.gcp_image_distro = g.inspect_get_distro(roots[0])
+  g.gcp_image_major = str(g.inspect_get_major_version(roots[0]))
+  g.gcp_image_minor = str(g.inspect_get_minor_version(roots[0]))
```
zoran15: Do all distributions have major/minor versions?

daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
 import utils.diskutils as diskutils
 
 
-network = '''
-BOOTPROTO='dhcp'
-STARTMODE='auto'
-DHCLIENT_SET_HOSTNAME='yes'
-'''
+class Package:
```
zoran15: Add class docs

daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
...
+  gce = False
+  required = False
+
+  def __init__(self, name, gce, required):
+    self.name = name
+    self.gce = gce
+    self.required = required
+
+
+class SuseRelease:
```
zoran15: Add class docs

daisy_workflows/image_import/suse/translate.py
zoran15: packages are initialized to the same list for all 3 distros. Doesn't seem that it should be an argument as it's the same for all 3.

daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
...
+        flavor='sles',
+        major='12',
+        minor='5',
+        packages=[cloud_init, gcp_sdk, gce_init, gce_oslogin],
+        subscriptions=['sle-module-public-cloud/12/x86_64']
+    ),
+]
+
+
+def get_distro(g) -> SuseRelease:
```
zoran15: Add docs to all global methods for easier review and later maintenance.

daisy_workflows/image_import/suse/translate.py
```diff
@@ -26,67 +26,159 @@
...
+  if licensing == 'byol':
+    return
 
-    # Installing google-compute-engine-init and not installing
-    # gce-compute-image-packages as there is no port to SUSE
-    logging.info('Installing GCE packages.')
-    g.command(
-        ['zypper', '-n', 'install', '--no-recommends',
-        'google-compute-engine-init'])
+  # TODO: Implement GCP licensing for SLES.
```
zoran15: How are you tracking this work? Can this script receive 'gcp' licensing as the code is currently set up?

daisy_workflows/image_import/suse/translate.py
EricEdens: Interesting .. what kind of conflicts?

This test motivated me to add it: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/06b4dc2fa750de2c40fd5ec37f96c01c5e297371/daisy_integration_tests/scripts/post_translate_test.sh#L142-L141

daisy_workflows/image_import/suse/translate.py
EricEdens: I ran into a problem with SLES 12.5 image that I created with vmware. SLES's installer didn't include virtio kernel modules in initramfs, and booting the instance without re-running dracut stalled when trying to load the disks. 

(Interestingly, dracut isn't required if the instance is created in vmware using `xfs` intead of `btrfs`. `btrfs` is the default filesystem on SLES 12.5)

daisy_workflows/linux_common/utils/diskutils.py
EricEdens: Zero is returned from `inspect_get_major_version` and `inspect_get_minor_version` if those fields aren't found. For sanity testing, I ran the ovf import tests using this code and there were no errors.

Here's what the python code is calling:
  http://libguestfs.org/guestfs.3.html#guestfs_inspect_get_major_version

daisy_workflows/image_import/suse/translate.py
EricEdens: These fields are for supporting GCP licensing and the SAP variant of SLES, both of which will need some additional packages. Check out the zypper command on this page:

https://www.suse.com/c/create-sles-demand-images-gce/

I considered having a 'base' set of common packages, but in my experience that is confusing for the reader, and creates a large blast radius when someone updates the base set.

zoran15: So this is for future proofing?

daisy_workflows/image_import/suse/translate.py
EricEdens: done

daisy_workflows/image_import/suse/translate.py
EricEdens: done

daisy_workflows/image_import/suse/translate.py
EricEdens: done

daisy_workflows/image_import/suse/translate.py
EricEdens: went ahead and refactored the list of packages out of the field and into a variable.

daisy_workflows/image_import/suse/translate.py
EricEdens: done

daisy_workflows/image_import/suse/translate.py
EricEdens: To simplify this review, I went ahead and removed this method

PR #1052, opened 2019-12-23
Fixes for handling GCS paths

zoran15:
Fixes for handling GCS paths where existing code was rejecting valid paths. E.g. gs://bucket-name/ was being rejected from OVF import as it didn't have object name portion, even though it's a valid path to an OVF package if files are directly in the bucket (no folders). Added GetGCSObjectPathElements func to be used when an actual file path is needed

### DIFF
--- a/cli_tools/common/utils/storage/scratch_bucket_creator.go
+++ b/cli_tools/common/utils/storage/scratch_bucket_creator.go
@@ -81,7 +81,7 @@ func (c *ScratchBucketCreator) getBucketAttrs(fileGcsPath string, project string
 func (c *ScratchBucketCreator) getBucketAttrsFromInputFile(fileGcsPath string, project string,
 	fallbackZone string) (*storage.BucketAttrs, error) {
 
-	fileBucket, _, err := SplitGCSPath(fileGcsPath)
+	fileBucket, err := GetBucketNameFromGCSPath(fileGcsPath)
 	if err != nil || fileBucket == "" {
 		return nil, daisy.Errf("file GCS path `%v` is invalid: %v", fileGcsPath, err)
 	}
--- a/cli_tools/common/utils/storage/storage_client.go
+++ b/cli_tools/common/utils/storage/storage_client.go
@@ -34,7 +34,7 @@ import (
 
 var (
 	bucketNameRegex = `[a-z0-9][-_.a-z0-9]*`
-	gsPathRegex     = regexp.MustCompile(fmt.Sprintf(`^gs://(%s\/?)(.*)$`, bucketNameRegex))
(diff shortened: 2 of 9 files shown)


cli_tools/common/utils/storage/storage_client.go
```diff
@@ -34,8 +34,7 @@ import (
 
 var (
 	bucketNameRegex = `([a-z0-9][-_.a-z0-9]*)`
-	bucketPathRegex = regexp.MustCompile(fmt.Sprintf(`^gs://%s/(.*)$`, bucketNameRegex))
-	gsPathRegex     = regexp.MustCompile(fmt.Sprintf(`^gs://%s/(.+)$`, bucketNameRegex))
+	gsPathRegex     = regexp.MustCompile(fmt.Sprintf(`^gs://%s/?(.*)$`, bucketNameRegex))
```
EricEdens: How does your regular expression handle `gs://bucket||||`? Is it what you expected?

zoran15: Good point. Made the slash between bucket name and object path a part of bucket matching group and removed it (if any) from the resulting string.

zoran15: Actually, this one is still not working as expected. Will fix it tomorrow.

zoran15: Actually handled it now

cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -81,7 +81,7 @@ func (c *ScratchBucketCreator) getBucketAttrs(fileGcsPath string, project string
 func (c *ScratchBucketCreator) getBucketAttrsFromInputFile(fileGcsPath string, project string,
 	fallbackZone string) (*storage.BucketAttrs, error) {
 
-	fileBucket, _, err := SplitGCSPath(fileGcsPath)
+	fileBucket, err := GetBucketNameFromGCSPath(fileGcsPath)
```
dntczdx: hmm this has been merged... I think "getBucketAttrsFromInputFile" rely on SplitGCSPath to check whether it's a valid file path (bucket name + object name). If you need to create bucket by a bucket name, it would be better to make a separate function.

zoran15: I've changed SplitGCSPath to work for both bucket only paths and for bucket+object. I've added GetBucketNameFromGCSPath for this specific use case where only bucket name is needed.

Do you have any concerns with this solution?

dntczdx: Oh I didn't notice the new function behaves the same as the previous function. It's perfect,

PR #1054, opened 2019-12-26
fix typo in ks_helpers

hopkiw:
Turns out google-osconfig-agentgoogle-cloud-sdk is not a real package.
Fixes installed packages test.


hopkiw:
I have to merge this as admin, apparently, as it thinks gobuild and gocheck are expected. They're not expected for this change..

PR #1059, opened 2020-01-02
Make image import translate step use Daisy Workflow default timeout

zoran15:
Make image import translate step use Daisy Workflow default timeout (configurable) instead of it being hardcoded to 1hr. This is to allow for longer image translates as there have been examples where translate run successfully for more than 1hr.

### DIFF
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -394,7 +394,6 @@ func (w *Workflow) PopulateClients(ctx context.Context) error {
 func (w *Workflow) populateStep(ctx context.Context, s *Step) DError {
 	if s.Timeout == "" {
 		s.Timeout = w.DefaultTimeout
-
 	}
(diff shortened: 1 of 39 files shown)


dntczdx:
Is the translation step related to specific ovf image content? Usually it doesn't need >1 hour, in which case we need it?

zoran15:
> Is the translation step related to specific ovf image content? Usually it doesn't need >1 hour, in which case we need it?

Yes. We had a customer with an image that takes around 70 minutes to translate successfully. 

I think if we run our code in a controlled environment (e.g. images we provide, like Debian-9 or Windows workers) we can have timeouts set. However, if we run in uncontrolled environments (like in customer images for translate), even if we run our code, we can't guarantee things will run at certain speeds as we don't know what's running in that environment besides our translate scripts. Thus the removal of hardcoded timeouts.

PR #1061, opened 2020-01-02
Move cleanerupper to use beta osconfig

adjackura:


adjackura:
> Why do we still need references to osconfig in prow jobs if tests are removed? Just to clean up any artifacts from earlier?

This repo still handles all the running of tests (all the prow stuff), just the code was moved to the osconfig repo.

PR #1062, opened 2020-01-02
Python rhel8

hopkiw:
these files are already being used as templates, so might as well declare packages there to be obvious. keep the replacement for packages that vary, but don't vary based on distro.

after removing legacy guest we find that python packages need to be explicitly installed. start doing that now.


daisy_workflows/image_build/enterprise_linux/kickstart/el8-post.cfg
```diff
@@ -37,7 +37,7 @@ rpmkeys --import /etc/pki/rpm-gpg/google-rpm-package-key.gpg
 rpmkeys --import /etc/pki/rpm-gpg/google-key.gpg
 
 # Install GCE guest packages.
-dnf install -y __PACKAGES__ python27
+dnf install -y google-compute-engine google-osconfig-agent google-cloud-sdk gce-disk-expand python27 python36 __PACKAGES__
```
zmarano: We need to use python2 (instead of python27) here. See: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1016/files

PR #1068, opened 2020-01-06
Image import: Support OVAs where enclosed vmdk has spaces.

EricEdens:
Prior to this fix, if a user imported an OVA that contained "The Image Name.vmdk", they'd see the following logs and errors:

```
Import: Unpacking VMDK files from ova.
tar: The: Not found in archive
tar: Image: Not found in archive
tar: Name.vmdk: Not found in archive
tar: Exiting with failure status due to previous errors
Import: New source file is The Image Name.vmdk
...
Could not open '/daisy-scratch/The Image Name.vmdk': No such file or directory
```

## Testing
 - Ran the new integration test

### DIFF
--- a/daisy_integration_tests/daisy_e2e.test.gotmpl
+++ b/daisy_integration_tests/daisy_e2e.test.gotmpl
@@ -137,8 +137,8 @@
     "Image import and translate ova": {
       "Path": "./image_import_and_translate_ova.wf.json"
     },
-    "Image import and translate ova with spaces in VMDK": {
-      "Path": "./image_import_and_translate_ova_with_spaces_in_vmdk.wf.json"
+    "Image import and translate ova with spaces in VMDK filename.": {
+      "Path": "./image_import_and_translate_ova_with_spaces_in_vmdk_filename.wf.json"
     },
     "Image import and translate no extension": {
       "Path": "./image_import_and_translate_no_extension.wf.json"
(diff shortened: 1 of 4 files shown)


daisy_integration_tests/daisy_e2e.test.gotmpl
```diff
@@ -137,6 +137,9 @@
     "Image import and translate ova": {
       "Path": "./image_import_and_translate_ova.wf.json"
     },
+    "Image import and translate ova with spaces in VMDK": {
```
zoran15: ...spaces in VMDK file name"

It's VMDK files that have space, not the file itself. File names and other references in this PR should be updated as well.

daisy_integration_tests/image_import_and_translate_ova_with_spaces_in_vmdk.wf.json
```diff
@@ -0,0 +1,78 @@
...
+        "Images": [
+          "${image}"
+        ],
+        "Instances": [
+          "test-the-image-instance"
+        ]
+      }
+    }
+  },
+  "Dependencies": {
```
zoran15: nit: it's common to have dependencies sorted in a chronological order to make it easier to figure out ordering of steps. E.g. "test-the-image": ["import-image"] would be the first in this list, so it's easy to know that "import-image" happens first.

PR #1069, opened 2020-01-07
[ovf e2e] fix dup ovf test name

dntczdx:
Dup instance name lead to test conflict. Fix it.
It potentially can be the reason of deleted subnet. Will keep digging that.


dntczdx:
duplicate test instance

PR #1073, opened 2020-01-07
[ovf import / daisy] avoid undesired deletion

dntczdx:
A resource should be deleted by cleanup only when it's created by the workflow.
Now, there are 2 bugs which can lead to undesired deletion:
1. ovf import: ovf import set "forceCleanup" to "True". It conflicts with original design of "NoCleanup", which is used by regURL to mark placeholder resource as skipping cleanup. The problem: when ovf import failed, it will try deleting all used resources, including subnet, network, etc. We need to avoid this by distinguish whether the resource is a placeholder resource. The correct way is to check "resource.creator == nil".
2. daisy: for any daisy workflow, the create_resource step (for example, CreateDisks) will delete the created resource when the workflow failed on it. The problem is, when multiple workflows run in parallel to create a resource with the same name, one of them can fail on "resource exists", and delete it ultimately. To fix this issue, we need to set a flag to indicate whether creation succeeded; if not, the deletion should be skipped.

### DIFF
--- a/daisy/image.go
+++ b/daisy/image.go
@@ -116,6 +116,7 @@ type ImageInterface interface {
 	getRawDiskSource() string
 	setRawDiskSource(rawDiskSource string)
 	create(cc daisyCompute.Client) error
+	markCreatedInWorkflow()
 	delete(cc daisyCompute.Client) error
 	appendGuestOsFeatures(featureType string)
 	getGuestOsFeatures() guestOsFeatures
@@ -190,6 +191,10 @@ func (i *Image) create(cc daisyCompute.Client) error {
 	return cc.CreateImage(i.Project, &i.Image)
 }
 
+func (i *Image) markCreatedInWorkflow() {
+	i.createdInWorkflow = true
+}
+
 func (i *Image) delete(cc daisyCompute.Client) error {
 	return cc.DeleteImage(i.Project, i.Name)
 }
@@ -258,6 +263,10 @@ func (i *ImageBeta) setRawDiskSource(rawDiskSource string) {
 
 func (i *ImageBeta) create(cc daisyCompute.Client) error {
 	return cc.CreateImageBeta(i.Project, &i.Image)
+}
+
+func (i *ImageBeta) markCreatedInWorkflow() {
+	i.createdInWorkflow = true
 }
 
 func (i *ImageBeta) delete(cc daisyCompute.Client) error {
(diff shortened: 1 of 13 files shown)


daisy/resource_registry.go
```diff
@@ -40,7 +40,10 @@ func (r *baseResourceRegistry) init() {
 func (r *baseResourceRegistry) cleanup() {
 	var wg sync.WaitGroup
 	for name, res := range r.m {
-		if (res.NoCleanup && !r.w.forceCleanup) || res.deleted {
+		if res.creator == nil || // placeholder resource
+			(res.creator != nil && !res.created) || // resource is'nt created successfully
```
EricEdens: isn't

daisy/step_create_images.go
```diff
@@ -113,6 +113,7 @@ func (ci *CreateImages) run(ctx context.Context, s *Step) DError {
 			e <- newErr("failed to create images", err)
 			return
 		}
+		ci.markCreated()
```
EricEdens: Any reason why you use the method here, but in the other steps you set the field directly?

daisy/image.go
```diff
@@ -116,6 +116,7 @@ type ImageInterface interface {
 	getRawDiskSource() string
 	setRawDiskSource(rawDiskSource string)
 	create(cc daisyCompute.Client) error
+	markCreated()
```
EricEdens: Does this mean that the resource was created in general (perhaps it was found to exist already), or that it was explicitly created by this workflow? If the second, I'd suggest a more descriptive name for the field and method.

Also, where's the field defined?

daisy/step_create_images.go
dntczdx: Image is the only special resource. It used a 'ImageInterface' as a wrapper because it wants to be compatible with 2 versions: Image & ImageBeta. So I have to add a method to the interface to operate the 'created' field.

daisy/image.go
dntczdx: It's the 2nd. Let me name it "createdInWorkflow"

daisy/resource.go
```diff
@@ -44,6 +44,7 @@ type Resource struct {
 	deleteMx *sync.Mutex
 
 	creator, deleter *Step
+	created          bool
```
EricEdens: createdInWorkflow?

PR #1074, opened 2020-01-07
Image import: Add hint for when VMDK decoding fails.

EricEdens:
The message is intended to help with two common failures in VMDK decoding:
1. The user uploads the descriptor file, but not the image itself
2. The user uploads one VMDK from a set of split VMDKs

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -178,9 +178,19 @@ fi
 # /dev/sdc is used since it's the third disk that's attached in import_disk.wf.json.
 if ! out=$(qemu-img convert "${IMAGE_PATH}" -p -O raw -S 512b /dev/sdc 2>&1); then
   if [[ "${IMAGE_PATH}" =~ \.vmdk$ ]]; then
-    hint="Verify that you have a monolithic VMDK with an embedded descriptor file."
+    if file "${IMAGE_PATH}" | grep -qiP ascii; then
+      hint="When importing a VMDK disk image, ensure that you specify the VMDK disk "
+      hint+="image file, rather than its text descriptor file. In some virtual "
+      hint+="machine managers, given a text descriptor called <disk.vmdk>, "
+      hint+="the disk image file would be called <disk-flat.vmdk>."
+    else
+      hint="When preparing a VM import, ensure that you create a monolithic "
+      hint+="image file, where the disk is contained in a single VMDK file, as "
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -177,10 +177,13 @@ fi
 # Convert the image and write it to the disk referenced by $DISKNAME.
 # /dev/sdc is used since it's the third disk that's attached in import_disk.wf.json.
 if ! out=$(qemu-img convert "${IMAGE_PATH}" -p -O raw -S 512b /dev/sdc 2>&1); then
-  echo "ImportFailed: Failed to convert source to raw. [Privacy-> error: ${out} <-Privacy]"
+  if [[ "${IMAGE_PATH}" =~ \.vmdk$ ]]; then
+    hint="Verify that you have a monolithic VMDK with an embedded descriptor file."
```
zoran15: - "Verify that the VMDK image file being imported is a..."
- "monolithic VMDK" might not be something that all users know about. Some extra help text or a link to a website describing the differences would be useful. 
- I've seen users try to import vmdk descriptor in the past. It makes sense to call this potential issue in more detail than just referring to an "embedded descriptor". Ideally, we would detect if a file is a descriptor since it's a text file. More info here: https://www.vmware.com/support/developer/vddk/vmdk_50_technote.pdf

dntczdx: Agreed. I myself don't know too much details about VMDK. It would be better if there is a guide how to verify it - is there a simple command/tool to do that?

EricEdens: Our help guides [don't currently have information on creating VMDKs](https://cloud.google.com/compute/docs/import/importing-virtual-disks), and I wouldn't feel comfortable adding a link to an external site. That said, let's talk adding some guides around exporting and creating VMDKs (and other image files). If we do that, then we can add a link to it here.

In the meantime, how does this sound:

*If we detect that it's an ascii text file using the `file` command:*
> When importing a VMDK disk image, ensure that you specify the VMDK disk image file, rather than its text descriptor file.


*Otherwise:*
> When preparing a VM import, ensure that you create a monolithic image file, where the disk is contained in a single VMDK file, as opposed to a split file, where the disk is spread across multiple files.

dntczdx: Maybe both info? I am not sure whether it's reliable to decide by whether it's an ascii text file.

PR #1075, opened 2020-01-08
prowjobs: reorganize cloudbuild

adjackura:
- Move to use kaniko for container builds
- reorganize dockerfiles to avoid pulling from git unnecessarily
- use go modules for wrapper and cleanerupper


gce_image_import_export_tests.Dockerfile
```diff
@@ -11,27 +11,30 @@
...
-FROM golang:alpine
-
-RUN apk add --no-cache git
+FROM golang
 
 # Build test runner
-WORKDIR /
-RUN go get -d github.com/GoogleCloudPlatform/compute-image-tools/gce_image_import_export_tests
-RUN CGO_ENABLED=0 go build -o /gce_image_import_export_test_runner github.com/GoogleCloudPlatform/compute-image-tools/gce_image_import_export_tests
+WORKDIR /gce_ovf_import_tests
```
zoran15: why gce_ovf_import_tests?

Shouldn't it be gce_image_import_export_tests?

test-infra/prowjobs/cleanerupper/Dockerfile
```diff
@@ -11,13 +11,10 @@
...
-FROM golang:alpine
+FROM golang
 
-RUN apk add --no-cache git
-
-WORKDIR /tmp
-COPY main.go main.go
-RUN go get -d ./...
+WORKDIR /build
+COPY . .
```
zoran15: What does this do?

test-infra/prowjobs/gobuild/Dockerfile
```diff
@@ -11,15 +11,12 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM gcr.io/compute-image-tools-test/wrapper:latest
+FROM gcr.io/$PROJECT_ID/wrapper:latest
 
 FROM golang
 
 ENV GOOGLE_APPLICATION_CREDENTIALS /etc/compute-image-tools-test-service-account/creds.json
 
-RUN apt-get update && apt-get install -y git && \
```
zoran15: git comes pre-installed?

gce_image_import_export_tests.Dockerfile
adjackura: Yep, this was a typo

test-infra/prowjobs/cleanerupper/Dockerfile
adjackura: Copies everything in the current directory to the working directory, its a bit odd here because of how the WORKDIR works but the build starts in a folder (in this case it has a checkout of the git repo in it). I then set the WORKDIR to /build, so all build actions take place in there. I then copy the starting directory to that working directory

test-infra/prowjobs/gobuild/Dockerfile
adjackura: yeah, i went through and just trimmed items that didn't actually need installing

PR #1078, opened 2020-01-10
Added machine image import to OVF importer

zoran15:
Missing from this PR:
- E2E tests
- Logging to firelog


cli_tools/gce_ovf_import/README.md
```diff
@@ -1,7 +1,7 @@
 ## Compute Engine OVF Import
 
 The `gce_ovf_import` tool imports a virtual appliance in OVF format created in VMware environments
-to Google Compute Engine VM. It supports importing OVF and OVA archives.
+to Google Compute Engine VM or machine image. It supports importing OVF and OVA archives.
```
dntczdx: I feel "machine image" is not clear in the context. I know it's GMI, but "machine image" sounds more general. Can we use GMI?

cli_tools/gce_ovf_import/README.md
```diff
@@ -93,33 +96,50 @@ go get github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_ovf_impo
 + `-disable-stdout-logging` do not display individual workflow logs on stdout
 + `-node-affinity-label` Node affinity label used to determine sole tenant node to schedule this instance on. Label is of the format: <key>,<operator>,<value>,<value2>... where <operator> can be one of: IN, NOT. For example: workload,IN,prod,test is a label with key 'workload' and values 'prod' and 'test'. This flag can be specified multiple times for multiple labels.
 + `-release-track` Release track of OVF import. One of: %s, %s or %s. Impacts which compute API release track is used by the import tool.
++ `-hostname` Specify the hostname of the instance to be created. The specified hostname must be RFC1035 compliant.
++ `-machine-image-storage-location` GCS bucket storage location of the machine image being imported (regional or multi-regional). Applicable only if `-machine-image-name` is provided.
```
dntczdx: It's long and not that clear. How about machine-image-output-uri?

zoran15: This is not an URI but storage location, like a region or a multi-region. Equivalent for images is --storage-location. Since using that name would cause confusion in the context of ovf improt (since it imports instances), -machine-image-storage-location accurately describes what this flag is. From gcloud perspective, the flag will be --storage-location since the context will be machine images.

cli_tools/gce_ovf_import/README.md
zoran15: Changed to "Google Compute Engine machine image". We shouldn't use acronyms. Most people external to Google will not know what a GMI is until it becomes widespread, like Amazon's AMIs.

PR #1081, opened 2020-01-13
Better hint for default resource

dntczdx:
Now we got:

step "import-disk" validation error: step "import-virtual-disk" validation error: ResourceDoesNotExist: projects/xxxx/global/networks/default does not exist

But actually, there is no "network" input by user. The resource name is a default resource. To make it clear, we added one more sentence for "default" situation:

step "import-disk" validation error: step "import-virtual-disk" validation error: ResourceDoesNotExist: projects/xxxx/global/networks/default does not exist. please provide a specific resource name.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -292,7 +292,20 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
 		oauth, ce, gcsLogsDisabled, cloudLogsDisabled, stdoutLogsDisabled, kmsKey, kmsKeyring,
 		kmsLocation, kmsProject, noExternalIP, userLabels, storageLocation, uefiCompatible); err != nil {
 
+		postProcessErrorForNetwork(err, network, w)
+
 		return w, err
 	}
 	return w, nil
+}
+
+func postProcessErrorForNetwork(err error, network string, w *daisy.Workflow) {
+	if derr, ok := err.(daisy.DError); ok {
+		if derr.ErrType() == "networkResourceDoesNotExist" && network == "" {
+			w.LogWorkflowInfo("A VPC network is required for running image import," +
+				" and the default VPC network does not exist in your project. You will need to" +
+				" specify a VPC network with the --network flag. For more information about" +
+				" VPC networks, see https://cloud.google.com/vpc.")
+		}
+	}
 }
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -301,7 +301,7 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
 
 func postProcessErrorForNetwork(err error, network string, w *daisy.Workflow) {
 	if derr, ok := err.(daisy.DError); ok {
-		if derr.ErrType() == "networkResourceDoesNotExist" && network == "" {
+		if derr.CausedByErrType("networkResourceDoesNotExist") && network == "" {
 			w.LogWorkflowInfo("A VPC network is required for running image import," +
 				" and the default VPC network does not exist in your project. You will need to" +
 				" specify a VPC network with the --network flag. For more information about" +
--- a/daisy/resource_registry.go
+++ b/daisy/resource_registry.go
@@ -49,7 +49,7 @@ func (r *baseResourceRegistry) cleanup() {
 		wg.Add(1)
 		go func(name string) {
 			defer wg.Done()
-			if err := r.delete(name); err != nil && err.etype() != resourceDNEError {
+			if err := r.delete(name); err != nil && err.ErrType() != resourceDNEError {
 				fmt.Println(err)
 			}
 		}(name)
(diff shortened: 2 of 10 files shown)


daisy/resource_registry.go
```diff
@@ -197,7 +197,11 @@ func (r *baseResourceRegistry) regURL(url string) (*Resource, DError) {
 		if err != nil {
 			return nil, err
 		}
-		return nil, typedErrf(resourceDNEError, "%s does not exist", url)
+		errStr := "%s does not exist"
+		if strings.HasSuffix(url, "default") {
+			errStr += ". please provide a specific resource name."
```
zoran15: While an improvement, this still doesn't tell customer what to do (provide --network flag in this case).

Can we catch this exception in wrappers, detect it's network related and provide a more specific error message?

dntczdx: We can't, unless with a small refactoring. Let me try and propose in the next commit.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -292,7 +292,17 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
...
 		return w, err
 	}
 	return w, nil
 }
+
+func postProcessErrorForNetwork(err error, network string, w *daisy.Workflow) {
+	if derr, ok := err.(daisy.DError); ok {
+		if derr.ErrType() == "networkResourceDoesNotExist" && network == "" {
+			w.LogWorkflowInfo("Please specify --network flag.")
+		}
```
EricEdens: Can you add a couple more sentences? Here's a potential starting point:

"The default VPC has been removed from your GCP project, so you will need to specify one with the --network flag. The import workflow uses the network for [TODO] . For more information about creating networks, see [TODO] ."

Also, will the user need to provide a subnet?

zoran15: If the (non-default) network provided via --network flag is in auto mode, providing subnetwork is not mandatory. If it's a custom network, it is.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -292,7 +292,17 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
...
 
+		postProcessErrorForNetwork(err, network, w)
+
 		return w, err
 	}
 	return w, nil
 }
+
+func postProcessErrorForNetwork(err error, network string, w *daisy.Workflow) {
+	if derr, ok := err.(daisy.DError); ok {
```
zoran15: ok -> isDaisyError. Otherwise, ok sounds like things are ok when they are not (error was thrown)

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: sgtm

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -292,7 +292,17 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
...
+		if derr.ErrType() == "networkResourceDoesNotExist" && network == "" {
```
zoran15: Consider extracting as a constant in error.go.

Is the ErrType actually returning a lower case error name? Other errors (like resourceDNEError from error.go) are upper case.

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: Using `ok` is the idiomatic approach in golang: 


"use the "comma, ok" idiom to test, safely, whether the value is a string": https://golang.org/doc/effective_go.html

zoran15: And we should also call all our for indices "i"? :)

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: This is constructed by resource type name + resourceDNEError. All resource types have their own name.  That's why I didn't list all error names in error.go.
The purpose is to determine whether it's caused by network resource.

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: "xxx, ok" is everywhere including standard lib. I think it's not a problem for reader to understand fluently...

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: "A VPC network is required for running worker instances" - may be too detailed for the user, I'd propose "A VPC network is required for running image import". How do you think?

zoran15:
This change should be implemented in export/OVF import wrappers as well.

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: sounds good to me -- I like that it clarifies that the VPC is only used during import, rather than a permanent decision that is somehow tied to the imported resource.

PR #1092, opened 2020-01-21
[daisy][log] allow nested DError anonymized msg

dntczdx:
Currently, nested DError anonymized error msg is hidden. We got a lot of "Step ^v failed on error: %v"

The actual error should show up - it's also anonymized and is safe to show. Improved the code to allow showing that.

[will check in after the preparing release]


EricEdens:
Are you thinking these errors will be shown in the final error message to users? If so, could you give a few concrete examples of the current error messages, as well as the future error messages?

dntczdx:
> Are you thinking these errors will be shown in the final error message to users? If so, could you give a few concrete examples of the current error messages, as well as the future error messages?

It impacts only "anonymized error msg", which is used by our log. These error will be shown on dashboard for trouble shooting. With this change, many string-pattern match can be avoided to identify the root cause.

Example:
Before:
step %q run error: %s (it covers a ton of possible errors)
After:
step %q run error: ExportFailed: Failed to export disk source to GCS .

daisy/error_test.go
```diff
@@ -124,3 +125,15 @@ func TestDErrImplAdd(t *testing.T) {
...
 	}
 }
+
+func TestNestedAnonymizedDErrorMessage(t *testing.T) {
+	innerDErr1 := Errf("inner error 1: %v %v", "root cause 1", "root cause 2")
+	innerDErr2 := Errf("inner error 2: %v %v", "root cause 3", "root cause 4")
+	innerDErr1.add(innerDErr2)
+	outerDErr := wrapErrf(innerDErr1, "outer error: %v: ", "bad news")
+	got := strings.Join(outerDErr.AnonymizedErrs(), ",")
+	want := "outer error: %v: inner error 1: %v %v; inner error 2: %v %v"
```
zoran15: Why aren't placeholders (%v) replaced with values in the resulting error here ("root cause 1", ..."bad news") etc.?

daisy/error.go
```diff
@@ -77,8 +77,13 @@ func Errf(format string, a ...interface{}) DError {
 }
 
 // wrapErrf returns a DError by keeping errors type and replacing original error message.
-func wrapErrf(e DError, format string, a ...interface{}) DError {
-	return &dErrImpl{errs: []error{fmt.Errorf(format, a...)}, errsType: e.errorsType(), anonymizedErrs: []string{format}}
+func wrapErrf(e DError, formatPrefix string, a ...interface{}) DError {
+	f := formatPrefix + strings.Join(e.AnonymizedErrs(), "; ")
+	return &dErrImpl{
+		errs:           []error{fmt.Errorf(fmt.Sprintf(formatPrefix, a...) + e.Error())},
```
zoran15: What if formatPrefix doesn't include a space? Will the resulting error have e.Error() stuck at the end without any space?

daisy/error_test.go
dntczdx: It's anonymized msg. It's designed to hide the concrete value of placeholder.

daisy/error.go
dntczdx: You mean formatPrefix doesn't include a tailing space? Then it's an ugly format. Functionality won't be impacted.
There isn't necessarily a space there. For example, maybe people want to format it as "!!!SEERIOUS_ERROR>>>%v"

zoran15: I mean, in: 
fmt.Sprintf(formatPrefix, a...) + e.Error()

If format error is "Serious error in value `%v`" to which you concatenate e.Error(), formatting might get ugly, like: "Serious error in value `hostname`InvalidResourceException"

dntczdx: If they format it like that, then yes. Would that be a problem? People can ruin any format...

zoran15: Whoever is calling wrapErrf might not be aware how error will be formatted. It's safe to assume formatPrefix will be used to format a..interface{} values, but it's not clear error will be appended as well. That's why I think we should add a space or any other separator between the formatted formatPrefix and e.Error()

PR #1093, opened 2020-01-22
Add support for importing RHEL8.

EricEdens:
Includes support for GCP-licensed import. BYOL to come in a future PR.

### DIFF
--- a/daisy_integration_tests/rhel_8_translate.wf.json
+++ b/daisy_integration_tests/rhel_8_translate.wf.json
@@ -16,7 +16,7 @@
     }
   },
   "Steps": {
-    "create-disk-from-image": {
+    "create-disk-to-be-translated": {
       "CreateDisks": [
         {
           "name": "translate-me",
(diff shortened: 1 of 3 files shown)


daisy_integration_tests/rhel_8_translate.wf.json
```diff
@@ -0,0 +1,99 @@
...
+            "Port": 1,
+            "SuccessMatch": "PASSED:",
+            "FailureMatch": "FAILED:",
+            "StatusMatch": "STATUS:"
+          }
+        }
+      ]
+    }
+  },
+  "Dependencies": {
```
zoran15: Sort these in chronological order for easier understanding of the full workflow

daisy_workflows/image_import/enterprise_linux/translate_rhel_8_licensed.wf.json
```diff
@@ -0,0 +1,78 @@
...
+          "Name": "disk-translator",
+          "SourceImage": "projects/compute-image-tools/global/images/family/debian-9-worker",
+          "SizeGb": "10",
+          "Type": "pd-ssd",
+          "FallbackToPdStandard": true
+        }
+      ]
+    },
+    "translate-disk": {
+      "Timeout": "2h",
```
zoran15: Remove the timeout so that workflow's default timeout is used. It's unlikely that translate will take more than 2hrs, but we thought the same for 60m timeout which proved to be incorrect. Also, other translate workflows have timeout removed from this step so this workflow would not be consistent with the rest.

daisy_integration_tests/rhel_8_translate.wf.json
EricEdens: sgtm

daisy_workflows/image_import/enterprise_linux/translate_rhel_8_licensed.wf.json
EricEdens: done

PR #1097, opened 2020-01-24
Clarify how to import a disk image that is contained within a gzip compressed file.

EricEdens:
Currently we tell users how to import a file that was exported from GCP. This adds a hint for gzip files that contain disk images from other systems.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -123,11 +123,11 @@ func validateSourceFile(storageClient domain.StorageClientInterface, sourceBucke
 	byteCountingReader := daisycommon.NewByteCountingReader(rc)
 	// Detect whether it's a compressed file by extracting compressed file header
 	if _, err = gzip.NewReader(byteCountingReader); err == nil {
-		return daisy.Errf("the input file is a gzip file, which is not supported by " +
-			"'gcloud compute images import'. To import a file that was exported from " +
-			"Google Compute Engine, please use 'gcloud compute images create'. " +
-			"To import a file that was exported from a different system, decompress it " +
-			"and run 'gcloud compute images import' on the disk image file directly ")
+		return daisy.Errf("the input file is a gzip file, which is not supported by" +
+			"image import. To import a file that was exported from Google Compute " +
+			"Engine, please use image create. To import a file that was exported " +
+			"from a different system, decompress it and run image import on the " +
+			"disk image file directly")
 	}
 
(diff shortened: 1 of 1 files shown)


cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -123,7 +123,11 @@ func validateSourceFile(storageClient domain.StorageClientInterface, sourceBucke
 	byteCountingReader := daisycommon.NewByteCountingReader(rc)
 	// Detect whether it's a compressed file by extracting compressed file header
 	if _, err = gzip.NewReader(byteCountingReader); err == nil {
-		return daisy.Errf("the input file is a gzip file, which is not supported by 'gcloud compute images import'. To import a file that was exported from Google Compute Engine, please use 'gcloud compute images create' instead")
+		return daisy.Errf("the input file is a gzip file, which is not supported by " +
+			"'gcloud compute images import'. To import a file that was exported from " +
+			"Google Compute Engine, please use 'gcloud compute images create'. " +
+			"To import a file that was exported from a different system, decompress it " +
+			"and run 'gcloud compute images import' on the disk image file directly ")
```
zoran15: Extra space at the end.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -123,7 +123,11 @@ func validateSourceFile(storageClient domain.StorageClientInterface, sourceBucke
 	byteCountingReader := daisycommon.NewByteCountingReader(rc)
 	// Detect whether it's a compressed file by extracting compressed file header
 	if _, err = gzip.NewReader(byteCountingReader); err == nil {
-		return daisy.Errf("the input file is a gzip file, which is not supported by 'gcloud compute images import'. To import a file that was exported from Google Compute Engine, please use 'gcloud compute images create' instead")
+		return daisy.Errf("the input file is a gzip file, which is not supported by " +
```
zoran15: Since image import wrapper will now be used via API as well, referring to gcloud commands in errors will be confusing for API users. Can we refer to specific operations in a generic way (image import, image create) instead of referring to commands? We can also provide links where gcloud/API/UI steps are documented.

EricEdens: sgtm!

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: shoot, good find

PR #1099, opened 2020-01-27
Improve stability of import integ tests and enterprise linux installs.

EricEdens:
Two fixes:
 1. Tune the `ping` test to allow longer for the response.
 2. Add retries when installing packages using yum.

Testing:
 1. Ran `daisy_integration_tests/centos_6_qcow2_translate.wf.json`
 2. Ran `daisy_integration_tests/ubuntu_1404_img_translate.wf.json`

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -24,6 +24,7 @@ use_rhel_gce_license: True if GCE RHUI package should be installed
 
 import logging
 import os
+import time
 
 import utils
 import utils.diskutils as diskutils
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -197,14 +198,17 @@ def yum_install(g, *packages):
...
-    g.command(['yum', 'install', '-y'] + list(packages))
-  except Exception as e:
-    logging.debug('Failed to install {}. Details: {}.'.format(packages, e))
-    raise RuntimeError(
-        'Verify that you have specified the correct operating system '
-        'in the `--os` flag.  If you are bringing your own license (BYOL), '
-        'also verify that your subscription is eligible to be run on GCP.')
+  for i in range(6):
+    try:
+      time.sleep(i**2)
```
zoran15: Wouldn't this add delay even when not necessary, since it's before `yum install`? Shouldn't the sleep be in the `except` block?

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -24,6 +24,7 @@
 
 import logging
 import os
+import time
```
zoran15: Since this change is in the actual translate step, the PR description is a bit misleading: this should improve not only test stability but translate stability when used by end users.

EricEdens: updated :)

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Since `i` is zero on the first iteration, there's no delay prior to the first call to yum.

Putting it in the except block incurs a large delay after the last call to yum, if all of the calls fail (assuming there's no extra code to determine if it's the last call).

zoran15: Add a comment that sleep is ignored the first time as it's not obvious

EricEdens: sgtm

PR #1100, opened 2020-01-27
Fall-back to pd-standard for disks used in OVF import if quota exceeded

zoran15:


cli_tools/gce_ovf_import/daisy_utils/data_disk_workflow_updater.go
```diff
@@ -72,7 +74,8 @@ func AddDiskImportSteps(w *daisy.Workflow, dataDiskInfos []ovfutils.DiskInfo) {
 					Name: scratchDiskDiskName,
 					Type: "pd-ssd",
 				},
-				SizeGb: "10",
+				SizeGb:               "10",
```
dntczdx: this align is strange

zoran15: It's auto formatted by `go fmt ./...` to match:
FallbackToPdStandard: true,

Tests would fail otherwise.

PR #1107, opened 2020-01-31
Fixed OVF import e2e tests failing

zoran15:
Fixed OVF import e2e tests failing due to multiple tests sharing single Cloud Build and temporary resources getting cleaned up by shorter running tests thus failing longer running tests. The fix was to provide buildID override that can be utilized by E2E tests.

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
+++ b/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
@@ -100,20 +100,26 @@ func NewOVFImporter(params *ovfimportparams.OVFImportParams) (*OVFImporter, erro
 		return nil, err
 	}
 	tarGcsExtractor := storageutils.NewTarGcsExtractor(ctx, storageClient, logger)
-	buildID := os.Getenv("BUILD_ID")
-
-	if buildID == "" {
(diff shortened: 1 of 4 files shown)


cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -100,22 +100,28 @@ func NewOVFImporter(params *ovfimportparams.OVFImportParams) (*OVFImporter, erro
...
 		mgce:                &computeutils.MetadataGCE{}, bucketIteratorCreator: bic, Logger: logger,
 		zoneValidator: &computeutils.ZoneValidator{ComputeClient: computeClient}, params: params}
 	return ovfImporter, nil
 }
 
+func getBuildID(params *ovfimportparams.OVFImportParams) string {
+	if params != nil && params.BuildID != "" {
+		return params.BuildID
+	}
+	buildID := os.Getenv("BUILD_ID")
```
dntczdx: can we just simply attach a rand string for whatever build id?

zoran15: I guess we could, I just didn't want to change the behavior in case anyone has relied on the fact build ID is used for temporary paths.

PR #1109, opened 2020-02-03
Adopt VMDK api (beta)

dntczdx:
Adopting disk inserting vmdk API.

1. Since the new API feature will be tested in beta, I have to implement a CreateDisksBeta for calling it. There are some dup code. Will remove them once the feature came into GA.

2. Key change 1: new field "SourceStorageObject" for CreateDisks(Beta) API. Accept this param from workflow and send to disk insert API.

3. Key change 2: new workflow / modified workflow to use "SourceStorageObject".

4. Key change 3: importer will use new workflow to call new API. If it failed due to unsupported image format, it will fallback to old workflow.


cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -133,6 +133,20 @@ func UpdateToUEFICompatible(workflow *daisy.Workflow) {
 				disk.Disk.GuestOsFeatures = daisy.CombineGuestOSFeatures(disk.Disk.GuestOsFeatures, "UEFI_COMPATIBLE")
 			}
 		}
+		if step.CreateDisksAlpha != nil {
```
zoran15: Is this a new type of step? In similar changes, for images and instances, we've kept the existing step but added support in it for different release tracks

dntczdx: I chose this way to proactively determine whether alpha-version API needs to be called. Sometimes a single step will try to create 3 disks, but only one of them rely on alpha feature.
Anyways I hope it is migrated to GA when we are ready to release, thus we don;t need these code anymore. At least beta.

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -144,6 +144,20 @@ func UpdateToUEFICompatible(workflow *daisy.Workflow) {
 				disk.Disk.GuestOsFeatures = daisy.CombineGuestOSFeatures(disk.Disk.GuestOsFeatures, "UEFI_COMPATIBLE")
 			}
 		}
+		if step.CreateDisksBeta != nil {
```
dntczdx: This is dup code for beta

zoran15:
Should this be closed?

dntczdx:
Will close after the project is done. Now It's used to compare the new plan v.s. the old plan

XTJX21X (approved):

PR #1110, opened 2020-02-04
Create instance from GMI Daisy step

zoran15:


daisy/step_create_instances.go
```diff
@@ -17,22 +17,43 @@ package daisy
...
+type CreateInstances struct {
+	Instances     []*Instance
+	InstancesBeta []*InstanceBeta
+}
+
+// UnmarshalJSON unmarshals Instance.
+func (ci *CreateInstances) UnmarshalJSON(b []byte) error {
+	var instancesBeta []*InstanceBeta
+	if err := json.Unmarshal(b, &instancesBeta); err != nil {
+		return err
```
dntczdx: A bit confusion to me. It looks like if the json is compatible with both beta and GA, it will be initialized twice, and exist in both ci.InstancesBeta and ci.Instances. Is that the case? If so, will it be created twice?

zoran15: Yes. The same logic is used in CreateImages. validate/run methods in StepCreateInstances used instanceUsesBetaFeatures() to determine which one to use.

daisy/instance.go
```diff
@@ -79,13 +80,41 @@ func instanceExists(client daisyCompute.Client, project, zone, instance string)
...
-// Instance is used to create a GCE instance. Output of serial port 1 will be streamed to the daisy logs directory.
-type Instance struct {
-	compute.Instance
+// MarshalJSON is a workaround to prevent Instance from using compute.Instance's implementation.
+func (i *Instance) MarshalJSON() ([]byte, error) {
+	return json.Marshal(*i)
+}
+
+//InstanceInterface represent abstract Instance across different API stages (Alpha, Beta, API)
+type InstanceInterface interface {
```
dntczdx: Per my understanding, InstanceInterface is for avoiding dup code. However, I still see a lot of dup code except for impl of InstanceInterface (for example, populateDisks). If so, why don't we just dup all code for beta and remove InstanceInterface?

cli_tools/common/utils/compute/node_affinity_label_parser_test.go
```diff
@@ -42,7 +44,7 @@ func TestParseNodeAffinities(t *testing.T) {
 }
 
 func TestParseNodeAffinitiesSingleValue(t *testing.T) {
-	affinities, err := ParseNodeAffinityLabels([]string{"compute.googleapis.com/node-group-name,IN,zoran-playground-node-group"})
+	affinities, _, err := ParseNodeAffinityLabels([]string{"compute.googleapis.com/node-group-name,IN,zoran-playground-node-group"})
```
dntczdx: this pattern indicates affinitiesbeta is not tested

daisy/instance.go
zoran15: Code that modifies nested compute children (such as compute disks and their children) would be difficult to extract to the interface. The combination of duplicate code and interfaced code is what I believed will result in least duplicate code with least effort.

cli_tools/common/utils/compute/node_affinity_label_parser_test.go
zoran15: Base case is tested in TestParseNodeAffinities. I think it's low enough risk to assume other cases will behave the same for Beta

daisy/instance.go
dntczdx: I see. In my PR I didn't do this because saved dup code is not more than dup code for the interface.

zoran15: It's not only about reducing the total amount of code, but avoiding duplicate code due to maintenance risks: if we're performing complex logic that we later update only for GA, but forget for Beta, it could lead to nasty bugs. However, for some functions, avoiding duplication was very costly (mostly around modifying nested objects) so I went with code duplication approach instead.

PR #1112, opened 2020-02-06
Updated CLI tools to support CreateInstances Beta API and some smaller improvements

zoran15:


cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -100,28 +101,22 @@ func NewOVFImporter(params *ovfimportparams.OVFImportParams) (*OVFImporter, erro
 		return nil, err
 	}
 	tarGcsExtractor := storageutils.NewTarGcsExtractor(ctx, storageClient, logger)
+	buildID := os.Getenv("BUILD_ID")
```
dntczdx: why is this change revered?

zoran15: By accident. Thanks for catching this.

PR #1117, opened 2020-02-07
[Windows image import] translate script: add -Force for creating folder

dntczdx:
[Windows image import] translate script: add -Force for creating folder to ignore error if the folder exists.

### DIFF
--- a/daisy_workflows/image_import/windows/translate_bootstrap.ps1
+++ b/daisy_workflows/image_import/windows/translate_bootstrap.ps1
@@ -160,8 +160,8 @@ try {
   }
 
   $driver_dir = 'c:\drivers'
-  New-Item $driver_dir -Force -Type Directory | Out-Null
-  New-Item $script:components_dir -Force -Type Directory | Out-Null
+  New-Item $driver_dir -Type Directory | Out-Null
+  New-Item $script:components_dir -Type Directory | Out-Null
 
   $daisy_sources = Get-MetadataValue -key 'daisy-sources-path'
 
--- a/daisy_workflows/image_import/windows/translate_bootstrap.ps1
+++ b/daisy_workflows/image_import/windows/translate_bootstrap.ps1
@@ -59,7 +59,7 @@ function Get-MetadataValue {
 
 function Setup-ScriptRunner {
   $metadata_scripts = "${script:os_drive}\Program Files\Google\Compute Engine\metadata_scripts"
-  New-Item "${metadata_scripts}\" -ItemType Directory | Out-Null
+  New-Item "${metadata_scripts}\"  -Force -ItemType Directory | Out-Null
   Copy-Item "${script:components_dir}\run_startup_scripts.cmd" "${metadata_scripts}\run_startup_scripts.cmd" -Verbose
   # This file must be unicode with no trailing new line and exactly match the source.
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/windows/translate_bootstrap.ps1
```diff
@@ -160,8 +160,8 @@ try {
   }
 
   $driver_dir = 'c:\drivers'
-  New-Item $driver_dir -Type Directory | Out-Null
-  New-Item $script:components_dir -Type Directory | Out-Null
+  New-Item $driver_dir -Force -Type Directory | Out-Null
```
rofuentes: Force creation of this directory *shouldn't* be necessary, was there an error for a particular image import?

daisy_workflows/image_import/windows/translate_bootstrap.ps1
```diff
@@ -160,8 +160,8 @@ try {
...
+  New-Item $script:components_dir -Force -Type Directory | Out-Null
```
rofuentes: Also, this directory shouldn't exist prior to import.

zoran15: What about images that were exported from GCP and are being re-imported?

rofuentes: This command is running on the import VM, using a GCP image.

daisy_workflows/image_import/windows/translate_bootstrap.ps1
dntczdx: I failed to import a image that is exported from GCE. So I added -Force for all folder creation. I think it's harmless, is it?

daisy_workflows/image_import/windows/translate_bootstrap.ps1
dntczdx: Same as above. If it's harmless, I prefer to add -Force, because a pre-exist folder name shouldn't block the translation in any case.

dntczdx:
> What happens if a directory exists? An empty directory with the same name is created or the existing directory with the existing content is kept?

existing content is kept

daisy_workflows/image_import/windows/translate_bootstrap.ps1
zoran15: My understanding is that it's running on worker VM but is creating directories on the imported image. Is that a correct assumption?

If so, these directories might exist for an image that is being re-imported.

rofuentes: No, your assumption is incorrect. The driver and components directories are created on the worker VM - not the imported image.

I believe these two statements should not contain -Force, but I understand the premise of this PR and not failing when a image in reimported to GCP on more than once.

daisy_workflows/image_import/windows/translate_bootstrap.ps1
rofuentes: If the intent of this PR is to allow for multiple re-imports/exports of an image on GCP, this -Force parameter is NOT necessary, and should be removed.

daisy_workflows/image_import/windows/translate_bootstrap.ps1
dntczdx: removed these 2 lines

PR #1120, opened 2020-02-10
daisy: use operations wait

adjackura:
Using operations wait instead of get with a sleep will significantly decrease qps on many concurrent operations:
https://cloud.google.com/compute/docs/reference/rest/v1/zoneOperations/wait

Cleaned up sleep in cleanup to instead wait for the already existing waitgroup
Cleaned up the instance serial logs recording to use []byte instead of string to significantly increase efficiency (test goes from 1.5s to .1s)


daisy/compute/compute.go
```diff
@@ -273,7 +273,7 @@ type operationGetterFunc func() (*compute.Operation, error)
 
 func (c *client) zoneOperationsWait(project, zone, name string) error {
 	return c.operationsWaitHelper(project, name, func() (op *compute.Operation, err error) {
```
dntczdx: No need to wait for 1s in operationsWaitHelper when we use this wait operation.

adjackura: That is still important as Wait can still return early for a number of reasons (see link in description). So we should treat it the same as Get.

crunk1:
LGTM

PR #1124, opened 2020-02-11
[image_import] Skip BCD rewrite for GPT disks

rofuentes:
Second attemot to detect and refrain from modifying BCD store if the
imported disk is GPT partitioned. Original PR:
https://github.com/GoogleCloudPlatform/compute-image-tools/pull/976

Tested on WS 2019 UEFI/GPT and MBR disks, including booting and sysprepping the resulting image.

### DIFF
--- a/daisy_workflows/image_import/windows/translate_bootstrap.ps1
+++ b/daisy_workflows/image_import/windows/translate_bootstrap.ps1
@@ -119,6 +119,7 @@ try {
   Write-Output 'TranslateBootstrap: Beginning translation bootstrap powershell script.'
   $script:is_x86 = Get-MetadataValue -key 'is_x86'
 
+  $partition_style = Get-Disk 1 | Select-Object -Expand PartitionStyle
   Get-Disk 1 | Get-Partition | ForEach-Object {
     if (-not $_.DriveLetter) {
       # Ensure all available partitions on the import disk are accessible via drive letter.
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/windows/translate_bootstrap.ps1
```diff
@@ -193,7 +194,13 @@ try {
   else {
     Copy-32bitPackages
   }
-  Run-Command bcdboot "${script:os_drive}\Windows" /s $bcd_drive
+  if ($partition_style -eq "MBR") {
+    Write-Output 'MBR partition detected. Resetting bootloader.'
```
zoran15: Should we tag this so it can be picked up by import wrapper so it can modify the workflow and add UEFI_COMPATIBLE? Sync up with Tony on this.

daisy_workflows/image_import/windows/translate_bootstrap.ps1
```diff
@@ -193,7 +194,13 @@ try {
...
+    Run-Command bcdboot "${script:os_drive}\Windows" /s $bcd_drive
+  }
+  else {
+    Write-Output 'GPT partition detected.'
```
zoran15: Here potentially as well.

daisy_workflows/image_import/windows/translate_bootstrap.ps1
dntczdx: discussed with Romain. This one is fine. I will modify the latter one in the next PR. (only return info for UEFI case).

daisy_workflows/image_import/windows/translate_bootstrap.ps1
dntczdx: I will modify this part in the next PR.

PR #1126, opened 2020-02-11
[image_import] Set import images to always use UEFI

rofuentes:
MBR images will continue to boot using SeaBIOS fallback.


rofuentes:
No, we can mark MBR disks as UEFI since they will successfully boot under compatibility mode. I know we also discussed the eventual deprecation of this mode in UEFI Class 3, but you can discuss with @zmarano regarding that.

rofuentes:
OK, following up with @zmarano, it was decided that we are going to detect and apply UEFI_COMPATIBLE programatically. Closing PR.

dntczdx:
Discussed with Zach and Romain again. Ideally we still want to tag only when we detected EFI system. I have found a way to return the detection result and consume, while Romain has helped on detection code for Windows. So let's stick on that plan.

PR #1127, opened 2020-02-11
Dump docker image list

YangLu1031:
Need to dump docker image list for analysis:
https://github.com/kubernetes/kubernetes/issues/88011

### DIFF
--- a/cli_tools/diagnostics/main_windows.go
+++ b/cli_tools/diagnostics/main_windows.go
@@ -31,8 +31,9 @@ const (
 	// TODO: user can change the dump path, so better fetch the path from Registry:
 	// https://support.microsoft.com/en-us/help/254649/overview-of-memory-dump-file-options-for-windows
 	// But it's not likely people will do that.
-	crashDump         = `C:\Windows\MEMORY.dmp`
-	rdpStatusFileName = "rdp_status.txt"
+	crashDump               = `C:\Windows\MEMORY.dmp`
+	rdpStatusFileName       = "rdp_status.txt"
+	dockerImageListFileName = "docker_images.log"
 )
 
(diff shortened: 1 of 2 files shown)


cli_tools/diagnostics/main_windows.go
```diff
@@ -256,6 +256,14 @@ func getPlainEventLogs(evts []winEvt, errs chan error) []string {
 	return runAll(commands, errs)
 }
 
+// getDockerImagesList put docker images list file in logFolder channel and errors in error channel.
+func getDockerImagesList(logs chan logFolder, errs chan error) {
+	var commands = []runner{
+		cmd{`C:\Program Files\Docker\docker.exe`, "image list", "docker_images.log", false},
```
pjh: Remind me, what happens if we invoke this and docker isn't installed on the VM? Does this lead to any failures or error output for the user, or do we just not do anything?

pjh:
LGTM

cli_tools/diagnostics/main_windows.go
YangLu1031: It won't break anything, we will only log a general error message in serial port

pjh: Would it be possible to suppress that error message? (same for other already-existing checks for docker / kubernetes logs). If there's a reasonable way to do this then that would be nice, but not a deal-breaker.

YangLu1031: Yeah, we suppress it like other logs and print one general message like `Errors occured while collecting and zipping some logs...` in serial port.

pjh:
@ning-yang @adjackura PTAL

YangLu1031:
@adjackura thanks for approval, but why it's still not able to merge?

PR #1129, opened 2020-02-12
Improve error messaging when SUSEConnect fails.

EricEdens:
When importing SLES, we need to install the public cloud module to get access to the GCE packages. Prior to this change, if the import failed here, the user would see a failure message such as `Command failed: SUSEConnect -p sle-module-public-cloud/12/x86_64: command: `

This change adds failure detection for three failure modes:
1. The worker instance does not have internet connectivity (SUSEConnect makes an RPC to the SLES central entitlement server.)
2. There is a valid subscription, but the subscription doesn't allow access to the sle-module-public-cloud product.
3. There is not a valid subscription.

### DIFF
--- a/daisy_workflows/image_import/suse/translate.py
+++ b/daisy_workflows/image_import/suse/translate.py
@@ -112,13 +112,11 @@ def _get_distro(g) -> _SuseRelease:
       g.gcp_image_distro, g.gcp_image_major, g.gcp_image_minor))
 
 
-def _disambiguate_suseconnect_product_error(distro, g, product,
-    error) -> Exception:
+def _disambiguate_suseconnect_product_error(g, product, error) -> Exception:
   """Creates a user-debuggable error after failing to add a product
      using SUSEConnect.
 
   Args:
-      distro (_SuseRelease): The distro detected by GuestFS.
       g (GuestFS): Mounted GuestFS instance.
       product (str): The product that failed to be added.
       error (Exception): The error returned from `SUSEConnect -p`.
@@ -148,11 +146,8 @@ def _disambiguate_suseconnect_product_error(distro, g, product,
   # Example 2: SLES 15.1, Active
   # [
   #    {
-  #       "type":"evaluation",
-  #       "starts_at":"2019-12-16 00:00:00 UTC",
   #       "status":"Registered",
   #       "version":"15.1",
-  #       "expires_at":"2020-02-14 00:00:00 UTC",
   #       "arch":"x86_64",
   #       "identifier":"SLES",
   #       "subscription_status":"ACTIVE"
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/suse/translate.py
```diff
@@ -110,19 +112,85 @@ def _get_distro(g) -> _SuseRelease:
...
+  #       "arch":"x86_64",
+  #       "identifier":"sle-module-basesystem"
+  #    }
+  # ]
+
+  for status in statuses:
+    if status.get('identifier') not in ('SLES', 'SLES_SAP'):
+      continue
+
+    if status.get('subscription_status') is 'ACTIVE':
```
zoran15: In which cases an error is returned for an active subscription?

daisy_workflows/image_import/suse/translate.py
```diff
@@ -110,19 +112,85 @@ def _get_distro(g) -> _SuseRelease:
...
+  except Exception as e:
+    return ValueError(
+        'Unable to communicate with SCC. Ensure the import '
+        'is running in a network that allows internet access.', e)
+
+  # `SUSEConnect --status` returns a list of status objects,
+  # where the triple of (identifier, version, arch) uniquely
+  # identifies a product in SCC. Below are two examples.
+  #
+  # Example 1: SLES for SAP 12.2, No subscription
```
zoran15: Try to condense these examples with only relevant info as they make reading the code more difficult due to how much scrolling is needed. E.g. you could list just the relevant fields and list possible values. Or, remove non-relevant fields from examples

daisy_workflows/image_import/suse/translate.py
```diff
@@ -110,19 +112,85 @@ def _get_distro(g) -> _SuseRelease:
...
+      product (str): The product that failed to be added.
+      error (Exception): The error returned from `SUSEConnect -p`.
+  """
+  statuses = []
+  try:
+    statuses = json.loads(g.command(['SUSEConnect', '--status']))
+  except Exception as e:
+    return ValueError(
+        'Unable to communicate with SCC. Ensure the import '
+        'is running in a network that allows internet access.', e)
```
zoran15: https://en.wikipedia.org/wiki/Capitalization_of_Internet

daisy_workflows/image_import/suse/translate.py
EricEdens: Removed the non-relevant fields.

daisy_workflows/image_import/suse/translate.py
EricEdens: It's one of those "should never happen" scenarios, as far as I know, and by that I mean: right now we think it won't happen, but over time, that code will get executed.

PR #1130, opened 2020-02-12
[image_import] Suppress nonessential install errors

rofuentes:
- balloon and pvpanic driver installs fail for WS 2008R2 and W 7
- although Windows does not include default drivers for these devices,
not including them does not substantially impact functionality on GCP
- we are investigating this error, and plan to fix


rofuentes:
[TestRunner] Creating test cases for test suite "daisy-e2e-tests"
  - Creating test case for "Windows Translate: 7 64-bit BYOL"
[TestRunner] Running test case "Windows Translate: 7 64-bit BYOL"
[TestRunner] Test case "Windows Translate: 7 64-bit BYOL" finished
[TestRunner] Creating junit xml file: "junit.xml"
[TestRunner] All test cases completed successfully.

PR #1131, opened 2020-02-13
[export] auto retry CreateInstances when external IP is denied by organization policy

dntczdx:
It's reported by b/149243707 [p1]
Failed to export because org policy denied all external IPs.
There is actually no need to access externals for export. So, just simply retry without external IP when the failure code matches.

This mechanism works for import as well. Will test for import later, and then change the workflow accordingly.

### DIFF
--- a/daisy/common.go
+++ b/daisy/common.go
@@ -280,3 +280,26 @@ func CombineGuestOSFeaturesBeta(features1 []*computeBeta.GuestOsFeature,
 	})
 	return ret
 }
+
+// UpdateInstanceNoExternalIP updates Create Instance step to operate
+// when no external IP access is allowed by the VPC Daisy workflow is running in.
+func UpdateInstanceNoExternalIP(step *Step) {
+	if step.CreateInstances != nil {
+		for _, instance := range step.CreateInstances.Instances {
+			if instance.Instance.NetworkInterfaces == nil {
+				continue
+			}
+			for _, networkInterface := range instance.Instance.NetworkInterfaces {
+				networkInterface.AccessConfigs = []*compute.AccessConfig{}
+			}
+		}
+		for _, instance := range step.CreateInstances.InstancesBeta {
+			if instance.Instance.NetworkInterfaces == nil {
+				continue
+			}
+			for _, networkInterface := range instance.Instance.NetworkInterfaces {
+				networkInterface.AccessConfigs = []*computeBeta.AccessConfig{}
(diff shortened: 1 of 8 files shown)


daisy/common.go
```diff
@@ -280,3 +280,26 @@ func CombineGuestOSFeaturesBeta(features1 []*computeBeta.GuestOsFeature,
...
 	return ret
 }
+
+// UpdateInstanceNoExternalIP updates Create Instance step to operate
+// when no external IP access is allowed by the VPC Daisy workflow is running in.
+func UpdateInstanceNoExternalIP(step *Step) {
+	if step.CreateInstances != nil {
+		for _, instance := range step.CreateInstances.Instances {
+			if instance.Instance.NetworkInterfaces == nil {
+				return
```
zoran15: why return instead of continue?

daisy/common.go
```diff
@@ -280,3 +280,26 @@ func CombineGuestOSFeaturesBeta(features1 []*computeBeta.GuestOsFeature,
...
+			}
+			for _, networkInterface := range instance.Instance.NetworkInterfaces {
+				networkInterface.AccessConfigs = []*compute.AccessConfig{}
+			}
+		}
+		for _, instance := range step.CreateInstances.InstancesBeta {
+			if instance.Instance.NetworkInterfaces == nil {
+				return
```
zoran15: same here

adjackura:
This goes against the whole point of a create instance step. It takes a api create call as described by the creator of the workflow, we should not be modifying it. This also goes for the previous commit for pd-standard when out of quota. A workflow should only ever create what the workflow described with some sane defaults in place.

daisy/common.go
dntczdx: should be continue. I copied from daisy_utils. Changed there as well.

dntczdx:
> This goes against the whole point of a create instance step. It takes a api create call as described by the creator of the workflow, we should not be modifying it. This also goes for the previous commit for pd-standard when out of quota. A workflow should only ever create what the workflow described with some sane defaults in place.

It's specifically hard for this external-ip problem. We have no ability to detect whether it will fail due to org policy before execution. AS discussed offline, let me propose this:
1. Land this fix to unblock customer since it's P1.
2. Later, I will revert this change and replace by a new solution: add a hook for the CreateInstance; in import wrapper, put the retry logic in the hook.

PR #1132, opened 2020-02-13
[image import] tag UEFI_COMPATIBLE correctly

dntczdx:
Tag UEFI_COMPATIBLE correctly for imported image.
This PR will be split to 2 (1. daisy. 2. cli_tools) once it's reviewed due to go-module dependency.

Step 1. Return "<serial-output key:'is-uefi-compatible' value:'true'>" from bootstrap for both windows and linux.
Step 2. Store it in daisy workflow.
Step 3. Provide a hook for daisy steps.
Step 4. Setup the hook in importer.go to update the UEFI_COMPATIBLE tag.

Design preference ref to https://docs.google.com/document/d/1uG9mgiKmTlajESwAyJh4-Smkg1XystZjbLzLW7R0y5I/edit#heading=h.93avqth8c7wc

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -43,7 +43,8 @@ type Importer interface {
 
 // NewImporter constructs an Importer instance.
 func NewImporter(args ImportArguments, computeClient compute.Client, storageClient storage.Client) (Importer, error) {
-	inflater, err := createInflater(args, computeClient, storageClient, imagefile.NewGCSInspector())
+	loggableBuilder := service.NewSingleImageImportLoggableBuilder()
+	inflater, err := newInflater(args, computeClient, storageClient, imagefile.NewGCSInspector(), loggableBuilder)
 	if err != nil {
 		return nil, err
 	}
@@ -62,7 +63,7 @@ func NewImporter(args ImportArguments, computeClient compute.Client, storageClie
 		processorProvider: defaultProcessorProvider{args, computeClient, inspector},
 		traceLogs:         []string{},
 		diskClient:        computeClient,
-		loggableBuilder:   service.NewSingleImageImportLoggableBuilder(),
+		loggableBuilder:   loggableBuilder,
 	}, nil
 }
 
@@ -72,7 +73,7 @@ type importer struct {
 	project, zone     string
 	pd                persistentDisk
 	preValidator      validator
-	inflater          inflater
+	inflater          Inflater
 	processorProvider processorProvider
 	traceLogs         []string
 	diskClient        diskClient
@@ -107,9 +108,9 @@ func (i *importer) Run(ctx context.Context) (loggable service.Loggable, err erro
 func (i *importer) runInflate(ctx context.Context) (err error) {
 	return i.runStep(ctx, func() error {
 		var err error
-		i.pd, _, err = i.inflater.inflate(i.loggableBuilder)
+		i.pd, _, err = i.inflater.Inflate()
 		return err
-	}, i.inflater.cancel, i.inflater.traceLogs)
+	}, i.inflater.Cancel, i.inflater.TraceLogs)
 }
 
 func (i *importer) runProcess(ctx context.Context) error {
--- a/cli_tools/gce_vm_image_import/importer/api_inflater.go
+++ b/cli_tools/gce_vm_image_import/importer/api_inflater.go
@@ -19,7 +19,6 @@ import (
 	"time"
 
 	daisyUtils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/daisy"
-	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging/service"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/storage"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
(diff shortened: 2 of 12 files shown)


cli_tools/go.mod
```diff
@@ -5,26 +5,27 @@ go 1.13
...
-	golang.org/x/tools v0.0.0-20200206152323-64a0f23fc32d // indirect
-	google.golang.org/api v0.16.0
-	google.golang.org/genproto v0.0.0-20200205142000-a86caf926a67 // indirect
-	google.golang.org/grpc v1.27.1 // indirect
+	golang.org/x/tools v0.0.0-20200211205636-11eff242d136 // indirect
+	google.golang.org/api v0.17.0
+	google.golang.org/genproto v0.0.0-20200211111953-2dc5924e3898 // indirect
 )
+
+replace github.com/GoogleCloudPlatform/compute-image-tools/daisy => ../daisy
```
dntczdx: will split to 2 PR and remove this "replace"

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -124,37 +124,64 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
 	})
 }
 
+// SetupUEFIUpdateHook setup pre-run hook for each create-images steps to
+// update "UEFI_COMPATIBLE" tag based on context
+func SetupUEFIUpdateHook(workflow *daisy.Workflow) {
+	workflow.IterateWorkflowSteps(func(s *daisy.Step) {
+		if s.CreateImages != nil {
```
zoran15: We should also check for instance name here. It's unlikely that another instance will output is-uefi-compatible in a workflow, but it's better to be certain the correct instance did output this value

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -124,37 +124,64 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
-				image.Image.GuestOsFeatures = daisy.CombineGuestOSFeatures(image.Image.GuestOsFeatures, "UEFI_COMPATIBLE")
+	}
+}
+
+// UpdateCreateImagesToUEFICompatible marks CreateImages step to be UEFI
+// compatible by adding "UEFI_COMPATIBLE" to GuestOSFeatures.
+func UpdateCreateImagesToUEFICompatible(step *daisy.Step) {
+	if step.CreateImages != nil {
+		for _, image := range step.CreateImages.Images {
+			image.GuestOsFeatures = stringutils.CombineStringSlices(image.GuestOsFeatures, "UEFI_COMPATIBLE")
```
zoran15: Extract "UEFI_COMPATIBLE" to a const

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -124,37 +124,64 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
+			disk.Disk.GuestOsFeatures = daisy.CombineGuestOSFeatures(disk.Disk.GuestOsFeatures, "UEFI_COMPATIBLE")
 		}
-		if step.CreateImages != nil {
-			for _, image := range step.CreateImages.Images {
-				image.GuestOsFeatures = stringutils.CombineStringSlices(image.GuestOsFeatures, "UEFI_COMPATIBLE")
-				image.Image.GuestOsFeatures = daisy.CombineGuestOSFeatures(image.Image.GuestOsFeatures, "UEFI_COMPATIBLE")
+	}
+}
+
+// UpdateCreateImagesToUEFICompatible marks CreateImages step to be UEFI
```
zoran15: It marks image created by CreateImages, not the step itself

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -124,37 +124,64 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
-				if strings.Contains(disk.SourceImage, "projects/windows-cloud/global/images/family/windows-2019-core") && strings.Contains(disk.Name, "disk-bootstrap") {
-					continue
-				}
+		UpdateCreateDisksToUEFICompatible(step)
+		UpdateCreateImagesToUEFICompatible(step)
+	})
+}
 
-				disk.Disk.GuestOsFeatures = daisy.CombineGuestOSFeatures(disk.Disk.GuestOsFeatures, "UEFI_COMPATIBLE")
+// UpdateCreateDisksToUEFICompatible marks CreateDisks step to be UEFI
```
zoran15: It marks disk created by CreateDisks, not the step itself

cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -363,8 +363,13 @@ func (oi *OVFImporter) modifyWorkflowPostValidate(w *daisy.Workflow) {
 	rl.LabelResources(w)
 	daisyutils.UpdateAllInstanceNoExternalIP(w, oi.params.NoExternalIP)
 	if oi.params.UefiCompatible {
```
zoran15: Since --uefi-compatible is not a gcloud flag, we can deprecate it in wrappers if auto UEFI tagging works

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -230,8 +230,13 @@ func runImport(ctx context.Context, varMap map[string]string, importWorkflowPath
 			}}
 		rl.LabelResources(w)
 		daisyutils.UpdateAllInstanceNoExternalIP(w, noExternalIP)
+
 		if uefiCompatible {
```
zoran15: same here

cli_tools/common/utils/daisy/daisy_utils.go
dntczdx: The instance name contains a random suffix.
Even we can get that suffix somehow, I don't feel it's really better: it assumes an instance name which is given by the workflow.
Current PR: assumes no other instances called translate_bootstrap script, which is also defined by workflow.
Both are based on an assumption towards the workflow.

adjackura:
This breaks the contract we have for run(), daisy has three steps for workflow running, populate, validate, and run. The majority of inline changes happen in populate, then validate pre validates the workflow, then run executes it as is, any changes after validate make validation pointless, workflows are meant to be static at runtime.
What you should instead do is run two workflows, the first can export the required variables and then you can create the second one based on the output.

PR #1133, opened 2020-02-14
If an image is undecodable by qemu-img, attempt to repair it.

EricEdens:
This is motivated by the VHDX format, which uses a journal to spool modifications to the underlying image. If the journal is non-empty, then it must be consumed prior to reading the image. VHDX calls this [log replay](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-vhdx/0d588e33-23a6-4c71-b27f-87d97ac3e914). qemu-img's repair isn't limited to VHDK, however [ref](https://manpages.debian.org/testing/qemu-utils/qemu-img.1.en.html).

Testing:
1. Created a truncated VMDK, and observed that import quit after the repair failed.

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -152,6 +152,20 @@ if [[ "${IMAGE_PATH}" =~ \.ova$ ]]; then
   echo "Import: New source file is ${VMDK}"
 fi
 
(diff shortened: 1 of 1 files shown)


dntczdx:
This code won't take effect anymore once we migrated to new PD disk.insert API and they supported VHDX. I will contact PD team to ensure they have the same logic in place.

PR #1134, opened 2020-02-14
Trim import parameters of leading and trailing whitespace before sending to Daisy.

EricEdens:
This was motivated by imports that failed when the subnetwork paramater
reached daisy still having a leading space character.

In addition, this updates the usage of assert.Equal, which expects the first parameter to the be the expected value.


EricEdens:
> This one is probably redundant since it's determined internally by the code

All done!

PR #1140, opened 2020-02-18
[image-import] windows: bring disks online before translate

dntczdx:
This happened when importing a non-uefi windows-2019 image: the secondary disk was offline, so it failed.

Just simply bring all offlined disk online to fix. Manually tested and it passed.


dntczdx:
> Why was the disk offline?

Romain and Andrew can answer it better. So far the theories are:
1. MBR disk may not be online automatically when boot disk (translator) is UEFI.
2. Devices on different bus may not be online automatically.

adjackura:
I had thought this may be because of the different partition tables (mbr vs gpt), but roman did some digging and this may instead be because the attached disk was on a separate bus, so maybe that's why this doesn't fail all the time?

PR #1142, opened 2020-02-20
[daisy] Add "RetryHook" to accept retry logic from external

dntczdx:
Motivation: currently, there are 2 sections of on-demand retry logic which lives in daisy core. Ideally, we should move them out to keep daisy core clean.
(they are: 1. retry CreteDisks with pd-standard when quota exceeded; 2. retry CreateInstances without external IP when it failed on external ip access limit)

Solution: Add a "RetryHook", which is contained by "fallbackRetryableTask". Any steps who need to handle retry just simply add "fallbackRetryableTask" to its corresponding daisy resources class.

To be able to call that hook function, we need to extract the multi-task logic to a common place. "multiTasksStepImpl" is what we add for that purpose.

All above core logic change is mostly in step.go.

Following above patterns, step_create_instances and step_create_disks are changed accordingly. Retry logic was moved to daisy_utils.go so importer, exporter and ovf_importer can call them.

Many test files and workflow.json files got changed to remove "FallbackToPdStandard" and "RetryWhenExternalIPDenied", which is trivial.

TODO in the future: Most of step_xxx.go have multi-task logic in their run(). We can change them by implementing "multiTasksStepImpl" to save a lot of dup code.

### DIFF
--- a/cli_tools/common/utils/daisy/daisy_utils.go
+++ b/cli_tools/common/utils/daisy/daisy_utils.go
@@ -16,6 +16,7 @@ package daisy
 
 import (
 	"fmt"
+	"net/http"
 	"reflect"
 	"regexp"
 	"strings"
@@ -25,6 +26,7 @@ import (
 	daisyCompute "github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
 )
 
 const (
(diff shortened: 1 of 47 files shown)


EricEdens:
As an alternative to this, any thoughts on adding an "or" language construct to Daisy's workflow, such that you can chain together multiple steps. If the first in the "or" chain succeeds, the subsequent steps don't get executed. On the other hand, if the first step fails, the subsequent step is executed. If all of the steps in the "or" chain fail, then the overall workflow fails.

@adjackura -- what do you think?

adjackura:
My main concern with this implementation or a generic "or" construct for tasks is the validation. The validation step needs to be as accurate as possible so we shouldn't be making decisions in 'run' that aren't already validated. For instance making this too open ended and you can create untracked resources that will cause all sorts of issues. Ideally all these errors would be caught in validation so you can then rewrite the workflow and rerun it (this is one of the reasons we run validate at all). Is this really not at all possible with quota or the external IP issue? I have an idea on how to solve the quota problem holistically but it would require a decent amount of work. 

Let's chat in person on this, I also would like to think about the 'or' idea and see if that could be added and still work with validation (maybe we just need strong 'or' validation logic).

dntczdx:
Hi Eric: The "or" grammar doesn't help a lot in this case. To decide which branch to execute, the "or" block needs to detect whether the quota/external ip has any problem. However, the easiest way to detect that is to simply try running CreateDisk/CreateInstance. Directly querying quota or external ip policy needs extra permission, which breaks the existing customers and is unnecessary.

Hi Andrew: As said above, there are APIs to detect quota/external ip policy, but extra permission is required. It breaks the existing customers. It's also strange to ask for those permissions for import/export task. Those permissions are rarely granted to a service account.

dntczdx:
close, not required anymore

PR #1150, opened 2020-02-27
Add retry to linux import package installation.

EricEdens:
Adds retries for:
 - Debian
 - Ubuntu
 - SUSE

And refactors `utils.RetryOnFailure` to allow configurable retry options (initial delay and total elapsed time)

### DIFF
--- a/daisy_integration_tests/daisy_e2e.test.gotmpl
+++ b/daisy_integration_tests/daisy_e2e.test.gotmpl
@@ -47,6 +47,9 @@
     "centos8 translate": {
       "Path": "./centos_8_translate.wf.json"
     },
+    "debian8 translate": {
+      "Path": "./debian_9_translate.wf.json"
+    },
     "Enterprise Linux - Don't use proxy": {
       "Path": "./el_import_dont_use_proxy.wf.json"
     },
--- a/daisy_integration_tests/daisy_e2e.test.gotmpl
+++ b/daisy_integration_tests/daisy_e2e.test.gotmpl
@@ -48,6 +48,9 @@
       "Path": "./centos_8_translate.wf.json"
     },
     "debian8 translate": {
+      "Path": "./debian_8_translate.wf.json"
+    },
+    "debian9 translate": {
       "Path": "./debian_9_translate.wf.json"
     },
     "Enterprise Linux - Don't use proxy": {
--- a/daisy_integration_tests/debian_9_translate.wf.json
+++ b/daisy_integration_tests/debian_9_translate.wf.json
@@ -1,0 +1,95 @@
+{
+  "Name": "debian-9-translate-test",
+  "Sources": {
+    "post_translate_test.sh": "./scripts/post_translate_test.sh"
+  },
+  "Vars": {
+    "image_name": {
+      "Value": "debian-9-translate-test-${ID}"
+    },
+    "source_image": {
+      "Value": "projects/compute-image-tools-test/global/images/debian-9-translate"
+    }
+  },
+  "Steps": {
+    "create-disk-from-image": {
+      "CreateDisks": [
+        {
+          "name": "translate-me",
+          "sourceImage": "${source_image}"
+        }
+      ]
+    },
+    "create-test-disk": {
+      "CreateDisks": [
+        {
+          "name": "disk-import-test",
+          "sourceImage": "${image_name}",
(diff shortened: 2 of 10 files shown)


EricEdens:
SLES 15 and Ubuntu 16.04 tests passed. I'm adding a Debian 9 test, and that's running now.

EricEdens:
The new Debian 9 test passed.

daisy_workflows/image_import/debian/translate.py
```diff
@@ -49,8 +48,8 @@ def DistroSpecific(g):
   if install_gce == 'true':
     logging.info('Installing GCE packages.')
 
-    g.command(['apt-get', 'update'])
-    g.sh('DEBIAN_FRONTEND=noninteractive apt-get install --assume-yes gnupg')
+    utils.update_apt(g)
+    utils.install_apt_package(g, 'gnupg')
```
zoran15: Probably minor: utils.install_apt_package uses --no-install-recommends flag while the replaced call above doesn't.

daisy_workflows/linux_common/utils/common.py
```diff
@@ -721,3 +721,14 @@ def GetForwardingRuleIP(self, name):
         project=self.project, region=self.region, forwardingRule=name)
     response = request.execute()
     return response[u'IPAddress']
+
+
+@RetryOnFailure
```
zoran15: How does RetryOnFailure work? Does it give up after certain number of retries? Is there an exponential back off?

daisy_integration_tests/daisy_e2e.test.gotmpl
```diff
@@ -47,6 +47,9 @@
     "centos8 translate": {
       "Path": "./centos_8_translate.wf.json"
     },
+    "debian8 translate": {
+      "Path": "./debian_9_translate.wf.json"
+    },
```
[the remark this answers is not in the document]
EricEdens: It's 9 (good find!)

daisy_integration_tests/debian_9_translate.wf.json
```diff
@@ -0,0 +1,95 @@
...
+    "delete-image": [
+      "create-test-disk"
+    ],
+    "translate-disk": [
+      "create-disk-from-image"
+    ],
+    "wait-for-test-instance": [
+      "create-test-instance"
+    ]
+  }
```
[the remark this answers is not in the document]
EricEdens: done

daisy_workflows/linux_common/utils/common.py
EricEdens: Here's the decorator. It does linear backoff (1.5 coefficient), stopping after five minutes.

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/aaf4e0b12c637213702aefa762e954eb99a771ac/daisy_workflows/linux_common/utils/common.py#L36

daisy_workflows/image_import/debian/translate.py
EricEdens: I checked and it's being used by apt-key, and apt-key only requires gnupg.

To be confident, I added a test for Debian 8, so we have full coverage of the supported debian versions now.

daisy_workflows/linux_common/utils/common.py
zoran15: It's actually stopping after 15 minutes, as we already discussed in person. 5 minutes sounds more reasonable for this use case.

daisy_workflows/image_import/debian/translate.py
zoran15: My comment was about the --no-install-recommends  being provided now by default in install_apt_package while some of the calls that were replaced by it didn't have the same flag. Will the behavior be the same?

EricEdens: So, what I was trying to say is that the installation of `gnupg` is the only instance where the original code omitted `--no-install-recommends`, as far as I could find.

I checked why we're installing `gnupg`, and it's to allow us to install keys using `apt-key`. `apt-key` only requires `gnupg`, not the recommended packages of `gnupg`.

Lastly, to ensure this was true, I added an integ test for Debian 8, so now we have full integ coverage for the supported versions of Debian.

https://manpages.debian.org/stretch/apt/apt-key.8.en.html

PR #1154, opened 2020-03-05
Windows upgrade tool - minimum version - PR1 - daisy change

dntczdx:
This PR contains necessary daisy changes for windows upgrade (minimum version) which provides the most basic functionality to upgrade windows2008r2 to windows2012r2..

Changes including:
1. Cache cleanup. Cache is global. Need to cleanup before a new workflow started, otherwise validating can't detect new created resources.
2. CreateInstance [used by auto rollback when upgrading failed]
  -2.1 from machine-instance. Some fix applies, for example, don't specify machine type.
  -2.2 with OverWrite, which deletes the old instance before creating new one. In order to delete old disks along with instances, a "SetDiskAutoDelete" method is added.
3. disk: allow detaching without a previous attach step, if the disk has been attached before workflow is executed.
  -3.1 "DeviceName" is sued for detaching other than "DiskName", so changed some validation logic accordingly.
4. resource: added a "startedByWf" besides "stoppedByWf" to allow a resource to be stopped without a previous stop step.
5. Added "WaitForAnyInstancesSignal", which is similar to "WaitForInstancesSignal" but will finish when any instance got the waited signal. It's used to wait on multiple serial ports when upgrading.


adjackura:
The system of loading and modifying json workflows is not the best approach, the only reason we did it for import is because we already had json workflows. The RunWIthModifiers function is a hack that we really should not be leveraging if we don't need to (really we should look to remove all references to it if we can), just generate the workflow the way you want the first time.

daisy_workflows/windows_upgrade/upgrade_script.ps1
```diff
@@ -0,0 +1,55 @@
+#  Copyright 2020 Google Inc. All Rights Reserved.
```
zoran15: Will this script work for other Windows upgrade paths (e.g. 2012->2016)? If not, the name of this script should reflect the specific upgrade path it's implementing.

daisy_workflows/windows_upgrade/reboot.wf.json
```diff
@@ -0,0 +1,25 @@
+{
```
zoran15: reboot_instance.wf.json

daisy/workflow.go
```diff
@@ -342,6 +345,12 @@ func (w *Workflow) cleanup() {
 	w.recordStepTime("workflow cleanup", startTime, time.Now())
 }
 
+func cleanupCache() {
+	cleanupDiskCache()
+	cleanupImageCache()
+	// more
```
zoran15: what does this mean? if it's a TODO, document it as such

daisy/workflow.go
```diff
@@ -312,6 +312,9 @@ func (w *Workflow) GetStepTimeRecords() []TimeRecord {
 }
 
 func (w *Workflow) cleanup() {
+	// cleanup cache so the next workflow can have a clean env
+	defer cleanupCache()
```
zoran15: Why is this needed? What does "next workflow" mean? I thought caches are workflow specific, not global.

daisy/step_wait_for_instances_signal.go
```diff
@@ -224,9 +224,8 @@ func (w *WaitForInstancesSignal) run(ctx context.Context, s *Step) DError {
 			}
 			if is.SerialOutput != nil {
 				go func() {
-					if err := waitForSerialOutput(s, m["project"], m["zone"], m["instance"], is.SerialOutput, is.interval); err != nil {
-						e <- err
-					}
+					// send a signal to end other waiting serial ports, no matter whether an error happens
+					e <- waitForSerialOutput(s, m["project"], m["zone"], m["instance"], is.SerialOutput, is.interval)
```
zoran15: Would sending nil have an effect? Why this change?

daisy/step_detach_disks.go
```diff
@@ -32,15 +31,6 @@ type DetachDisk struct {
...
 func (a *DetachDisks) populate(ctx context.Context, s *Step) DError {
-	for _, dd := range *a {
-		if dd.DeviceName == "" {
-			dd.DeviceName = path.Base(dd.DeviceName)
-		}
-		if diskURLRgx.MatchString(dd.DeviceName) {
-			dd.DeviceName = extendPartialURL(dd.DeviceName, s.w.Project)
-		}
-	}
-
```
zoran15: Why is this code deleted and populate func left empty?

daisy/resource_registry.go
```diff
@@ -236,3 +238,30 @@ func (r *baseResourceRegistry) regUse(name string, s *Step) (*Resource, DError)
...
+	// - s depends on creator of name, if there is a creator.
+	// - name doesn't have a registered deleter yet, usage must occur before deletion.
+	r.mx.Lock()
+	defer r.mx.Unlock()
+	var ok bool
+	var res *Resource
+
+	// don't register use if detached device is not created/attached by an earlier step
+	if res, ok = r.m[name]; !ok {
+		return nil, nil
```
zoran15: No error?

daisy/step_wait_for_instances_signal.go
adjackura: We don't want to do this as a single WaitForInstancesSignal can wait for multiple instances and we do want to actually wait for all of them

daisy/resource_registry.go
```diff
@@ -236,3 +238,30 @@ func (r *baseResourceRegistry) regUse(name string, s *Step) (*Resource, DError)
 	r.m[name].users = append(r.m[name].users, s)
 	return res, nil
 }
+
+// regUseDeviceName registers a Step s as a user of a disk device resource.
+// "DeviceName" is only used by DetachDisks API.
+func (r *baseResourceRegistry) regUseDeviceName(name string, s *Step) (*Resource, DError) {
```
zoran15: What is the need of this func compared to regUse? 

Wouldn't the regUse line:
if r.urlRgx != nil && r.urlRgx.MatchString(name) {

Just be ignored when dd.DeviceName is sent and the two functions would work the same, except for the "return nil, nil" line?

daisy/resource_registry.go
```diff
@@ -89,13 +89,14 @@ func (r *baseResourceRegistry) start(name string) DError {
 		return Errf("cannot start %s %q; does not exist in registry", r.typeName, name)
 	}
 
-	if !res.stopped {
+	if res.started {
```
adjackura: Whats the reason behind this change? What information does this give us that 'stopped' does not?

daisy/disk.go
```diff
@@ -176,23 +182,27 @@ func (dr *diskRegistry) deleteFn(res *Resource) DError {
...
 	if dr.testDetachHelper != nil {
 		return dr.testDetachHelper(dName, iName, s)
 	}
 	pre := fmt.Sprintf("step %q cannot detach disk %q from instance %q", s.name, dName, iName)
+
 	var att *diskAttachment
 
 	if im, _ := dr.attachments[dName]; im == nil {
-		return Errf("%s: not attached", pre)
+		return nil
```
adjackura: I don't think this is what we want, I think we need these errors, what we don't have is registering resources the workflow did not create right? If we register instances and all their attached devices all this logic should still work?

daisy/step_detach_disks.go
```diff
@@ -57,29 +47,30 @@ func (a *DetachDisks) validate(ctx context.Context, s *Step) (errs DError) {
...
-		disk := namedSubexp(diskURLRgx, dr.link)
 		instance := namedSubexp(instanceURLRgx, ir.link)
-		if disk["project"] != instance["project"] {
-			errs = addErrs(errs, Errf("cannot detach disk in project %q from instance in project %q: %q", disk["project"], instance["project"], dd.DeviceName))
-		}
-		if disk["zone"] != instance["zone"] {
-			errs = addErrs(errs, Errf("cannot detach disk in zone %q from instance in zone %q: %q", disk["zone"], instance["zone"], dd.DeviceName))
+
+		// Ensure disk is in the same project and zone if dr exists.
+		if dr != nil {
```
zoran15: What if dr is nil?

daisy_workflows/windows_upgrade/cleanup.wf.json
```diff
@@ -0,0 +1,42 @@
+{
+  "Name": "cleanup",
+  "DefaultTimeout": "5m",
```
zoran15: Might be too short in some cases.

cli_tools/gce_windows_upgrade/upgrader/upgrader.go
```diff
@@ -0,0 +1,234 @@
...
+	daisyCompute "github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
+	"google.golang.org/api/option"
+)
+
+const (
+	logPrefix                      = "[windows-upgrade]"
+	rebootWorkflowPath =             "daisy_workflows/windows_upgrade/reboot.wf.json"
+	upgradePreparationWorkflowPath = "daisy_workflows/windows_upgrade/windows_upgrade_preparation.wf.json"
+	upgradeWorkflowPath            = "daisy_workflows/windows_upgrade/windows_upgrade_2008r2_to_2012r2.wf.json"
+	retryUpgradeWorkflowPath       = "daisy_workflows/windows_upgrade/windows_upgrade_2008r2_to_2012r2_retry.wf.json"
```
zoran15: We should parametrize upgrade paths to easily support future paths (e.g. 2012->2016).

cli_tools/gce_windows_upgrade/main.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging/service"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_vm_image_export/exporter"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_windows_upgrade/upgrader"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// project/zone/network/subnet/labels are not required when the upgrade has to happen in place.
+var (
```
zoran15: How will future upgrade paths (e.g. 2012->2016) be supported? Is there a flag needed or will it be auto detected?

daisy/step_wait_for_instances_signal.go
dntczdx: In my use-case, 'failure' is thrown from port 3 while 'success' is caught from port 1. if "WaitForInstancesSignal" is designed for "all" logic, we need another "WaitForAnyInstancesSignal". Or, we need to ignore waitings for 'failure' when we caught a 'success'. Which one do you think is better? I prefer a new 'WaitForAnyInstancesSignal'

daisy/resource_registry.go
dntczdx: I did this change because the original code doesn't allow a instance to be started without a pre 'stop' step. However, when we operate an instance, it can be either stopped or started.

daisy/step_wait_for_instances_signal.go
dntczdx: Or, to be able to wait on multiple ports for a single "Wait"

daisy/disk.go
dntczdx: You are right, if we want to keep the original check, we need to register all attached devices. I can add that. However, "attach" doesn't have the corresponding check. Do you think we should check whether a "AttachDisk" is not attached to any instance?

daisy/step_wait_for_instances_signal.go
dntczdx: explained below in Andrew's reply. Sending nil wont break a channel.

daisy/resource_registry.go
adjackura: ok, is there a way to simply change this to just use started? just wondering if we can consolidate for readability

daisy/disk.go
adjackura: The point of all the validate steps is to catch any errors before running the workflow, instead of fully checking all attached resources, maybe check on the fly if this is not a daisy created resource? So if a non daisy resource is referenced in an AttachDisk step, lookup attached disks, add them to the registry, then run the normal check. Does that make sense?

dntczdx: "if a non daisy resource is referenced in an AttachDisk step, lookup attached disks, add them to the registry" - that is what AttachDisk do today, by regUse->regURL.

The hard part is DetachDisk, which uses "DeviceName" orther than "Disk" as parameter, but actually ref to a disk. I have already added a "regUseDeviceName" to resolve this problem, collaborating with a deviceNameURLRgx. It did the same thing as you described above - if the referenced resource is not a daisy resource (aka it's a projects/xxx/xx/xx URL), check whether the corresponding instance has a device attached already. Please review it.

daisy/resource_registry.go
dntczdx: no  - if we use only "started", then we are not able to stop a instacne without a pre "start" step. One single bool can only represent 2 scenarios, while there are at least 3 scenarios:
1. instance has been started by a pre step.
2. instance has been stopped by a pre step.
3. instance is not started/stopped by any pre step.

For readability, let me rename the var name to "started_by_wf" and "stopped_by_wf".

daisy/step_wait_for_instances_signal.go
dntczdx: I added "WaitForAnyInstancesSignal", please have a look

daisy/common.go
```diff
@@ -39,6 +39,11 @@ func getUser() string {
 	return "unknown"
 }
 
+// NamedSubexp extracts sub matches in the exp
+func NamedSubexp(re *regexp.Regexp, s string) map[string]string {
```
adjackura: Just make the other function public (youll have more code to update but it beats having two functions)

daisy/step_wait_for_instances_signal.go
```diff
@@ -93,10 +96,10 @@ func waitForInstanceStopped(s *Step, project, zone, name string, interval time.D
 		case <-tick:
 			stopped, err := s.w.ComputeClient.InstanceStopped(project, zone, name)
 			if err != nil {
-				return typedErr(apiError, "failed to check whether instance is stopped", err)
+				return typedErr(apiError, "failed to check whether instance is stoppedByWf", err)
```
adjackura: This shouldn't be renamed

daisy/workflow_test.go
```diff
@@ -290,8 +290,8 @@ func TestNewFromFile(t *testing.T) {
 				},
 			},
 		},
-		"${bootstrap_instance_name}-stopped": {
-			name:                   "${bootstrap_instance_name}-stopped",
+		"${bootstrap_instance_name}-stoppedByWf": {
```
adjackura: these got renamed accidentally

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -192,3 +195,19 @@ func PostProcessDErrorForNetworkFlag(action string, err error, network string, w
...
+
+// RunWorkflowWithCancelSignal runs workflow with accepting cancel signal
+func RunWorkflowWithCancelSignal(ctx context.Context, w *daisy.Workflow, preValidateWorkflowModifier daisy.WorkflowModifier,
+		postValidateWorkflowModifier daisy.WorkflowModifier) error {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func(w *daisy.Workflow) {
+		select {
+		case <-c:
+			w.LogWorkflowInfo("\nCtrl-C caught, sending cancel signal to %q...\n", w.Name)
```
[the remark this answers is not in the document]
dntczdx: I don't see a clue it's not thread safe by looking into the implementation.
It's called in many places, for example, step.run(). Due that steps were run in parallel, so far it performs correctly.

cli_tools/gce_windows_upgrade/main.go
```diff
@@ -0,0 +1,75 @@
...
+				DisableGcsLogging:       *gcsLogsDisabled,
+				DisableCloudLogging:     *cloudLogsDisabled,
+				DisableStdoutLogging:    *stdoutLogsDisabled,
+			},
+			Instance: *instance,
+			SkipBackup: *skipBackup,
+		},
+	}
+
+	if err := service.RunWithServerLogging(service.WindowsUpgrade, paramLog, project, upgradeEntry); err != nil {
```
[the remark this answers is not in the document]
dntczdx: You are right. From engineering's perspective it's not a blocker, we do both in parallel.

daisy/workflow.go
dntczdx: It's global in current implementation. One reason is that a struct "disk" doesn't rely on a workflow to behave.

daisy_workflows/windows_upgrade/cleanup.wf.json
dntczdx: All 3 operations takes only <2s usually. No time consuming step.

daisy_workflows/windows_upgrade/reboot.wf.json
dntczdx: I don't get why that is better since there isn't anything else that can be rebooted

daisy_workflows/windows_upgrade/windows_upgrade_2008r2_to_2012r2.wf.json
```diff
@@ -0,0 +1,66 @@
+{
```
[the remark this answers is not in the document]
dntczdx: There are quite some differences. I feel it's too complex to edit the workflow in code - easy to lost which steps/attributes are modified

daisy/resource.go
```diff
@@ -174,3 +175,25 @@ func resourceNameHelper(name string, w *Workflow, exactName bool) string {
 	}
 	return name
 }
+
+type regionalResourceCache struct {
```
adjackura: I think twoDResourceCache and oneDResourceCache (or simular) would be better as we use regionalResourceCache for zonal resources as well as regional ones

daisy/disk.go
```diff
@@ -176,14 +189,20 @@ func (dr *diskRegistry) deleteFn(res *Resource) DError {
...
 		return dr.testDetachHelper(dName, iName, s)
 	}
 	pre := fmt.Sprintf("step %q cannot detach disk %q from instance %q", s.name, dName, iName)
+
 	var att *diskAttachment
 
+	// if the disk has already been attached before workflow is executed, skip validating its attacher
+	if isAttached {
+		return nil
+	}
```
zoran15: Minor: this can be moved before pre :=...

daisy/image.go
```diff
@@ -80,20 +73,24 @@ func imageExists(client daisyCompute.Client, project, family, name string) (bool
...
+			ila = append(ila, i)
+		}
+		imageCache.exists[project] = ila
 	}
 
 	for _, i := range imageCache.exists[project] {
-		if name == i.Name {
-			if i.Deprecated != nil && (i.Deprecated.State == "OBSOLETE" || i.Deprecated.State == "DELETED") {
-				return true, typedErrf(imageObsoleteDeletedError, "image %q in state %q", name, i.Deprecated.State)
+		if ic, ok := i.(*compute.Image); ok && name == ic.Name {
```
zoran15: What is this cast for? Guard against alpha/beta objects?

dntczdx: The type before this PR was *compute.Image. Since it was changed to interface{}, we have to cast bgefore using its attributes "ic.Name"

PR #1156, opened 2020-03-06
Add support for importing SLES 12.4

EricEdens:
* Adds two new tests: SLES 12.4 and SLES for SAP 12.4
* Includes a new `--os` flag value of `sles-sap-12-byol`

### DIFF
--- a/daisy_workflows/image_import/suse/translate.py
+++ b/daisy_workflows/image_import/suse/translate.py
@@ -105,9 +105,9 @@ def _get_distro(g) -> _SuseRelease:
     ValueError: If there's not a SuseObject for the the OS on the disk.
   """
   for d in _distros:
-    if (re.match(d.flavor, g.gcp_image_distro) and
-        re.match(d.major, g.gcp_image_major) and
-        re.match(d.minor, g.gcp_image_minor)):
+    if re.match(d.flavor, g.gcp_image_distro) \
+        and re.match(d.major, g.gcp_image_major) \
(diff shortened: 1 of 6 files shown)


daisy_workflows/image_import/suse/translate_sles_sap_12_byol.wf.json
```diff
@@ -0,0 +1,116 @@
...
+        "Instances":["inst-translator"]
+      }
+    },
+    "create-image": {
+      "CreateImages": [
+        {
+          "Name": "${image_name}",
+          "SourceDisk": "${source_disk}",
+          "Family": "${family}",
+          "Licenses": ["projects/suse-byos-cloud/global/licenses/sles-sap-12-byos"],
```
EricEdens: @zoran15 -- in terms of this PR, the key difference is the license.

daisy_workflows/image_import/suse/translate.py
```diff
@@ -91,7 +92,7 @@ def __init__(self, flavor, major, minor, products=None):
     _SuseRelease(
         flavor='sles',
         major='12',
-        minor='5',
+        minor='4|5',
```
zoran15: Shouldn't this be a separate _SuseRelease object?

EricEdens: In terms of translation for BYOL, SLES 12.4, SLES for SAP 12.4, and SLES 12.5 are all the same (need the same product enabled and need the same packages). The difference is the license.

PR #1157, opened 2020-03-09
Fixes for dumping RDP settings and Testgrid failure

YangLu1031:
There's test failure on [testgrid](https://testgrid.corp.google.com/windows-image#win-prod), seems due to surfacing the trivial errors, so change it to log.Printf instead of log.Fatalf.

Also Ning found the dumping RDP setting is not working, so figured out that GCE Agent was not running diagnostics.exe inside the `C:\Program Files\Google\Compute Engine\diagnostics\`  folder, so need to get the diagnostics.exe runtime directory path and pass it.

### DIFF
--- a/cli_tools/diagnostics/main_windows.go
+++ b/cli_tools/diagnostics/main_windows.go
@@ -33,6 +33,7 @@ const (
 	// But it's not likely people will do that.
 	crashDump               = `C:\Windows\MEMORY.dmp`
 	rdpStatusFileName       = "rdp_status.txt"
+	rdpScriptFileName       = "rdp_status.ps1"
 	dockerImageListFileName = "docker_images.log"
 )
 
@@ -85,9 +86,12 @@ func (command cmd) run() (outPath string, err error) {
 		c.Stdout = outFile
 		c.Stderr = outFile
 	}
-
 	if command.args != "" {
-		c.Args = append(c.Args, strings.Split(argString, " ")...)
+		args := strings.Split(argString, " ")
+		for _, arg := range args {
+			// Decode the "%20" to space
+			c.Args = append(c.Args, strings.ReplaceAll(arg, "%20", " "))
+		}
 	}
 	err = c.Run()
 	return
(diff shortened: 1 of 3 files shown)


cli_tools/diagnostics/main.go
```diff
@@ -176,6 +176,6 @@ func main() {
 	os.RemoveAll(tmpFolder)
 
 	if nonFatalErrorsPresent {
-		log.Fatal("Errors occured while collecting and zipping some logs.\nUnaffected logs were still packaged and available.")
+		log.Print("Errors occured while collecting and zipping some logs.\nUnaffected logs were still packaged and available.")
```
ning-yang: why we change to print? can we still see it from serial log?

pjh:
LGTM (but please answer Ning's question)

cli_tools/diagnostics/main.go
YangLu1031: The tests are using WinRM to trigger this tool, if some trivial errors happened, Fatal returned a panic. So the tests stopped as they got stderr.
Yes, we can still see it from serial log.

ning-yang: but this is really some kind of error we should capture during our test

pjh: Well the existing code doesn't make much sense in the first place: if _non fatal_ errors present, log _fatal_?

I believe the errors that are being referred to here are due to the tool failing to gather docker/kubernetes logs, which are not guaranteed to be present. We should distinguish between failures like that (which can be logged as just warnings/info) and other failures that should be error/fatal level. Yang, how hard would it be to do that?

YangLu1031: Right, the errors happens if kubernetes logs("C:\etc\kubernetes\logs") or crash dump("C:\Windows\MEMORY.dmp") doesn't exist. 
It's not difficult, we can log warnings/info immediately inside the kubernetes logs collector.

ning-yang: agree. that will be a better approach

YangLu1031: Ok, updated in the PR

ning-yang: thx. it is mush better. please let me know when the fix is released.

pjh:
LGTM

@adjackura PTAL

cli_tools/diagnostics/main_windows.go
```diff
@@ -284,15 +304,39 @@ func gatherEventLogs(logs chan logFolder, errs chan error) {
 // gatherKubernetesLogs put all the kubernetes log file paths in logFolder channel
 // and errors in error channel.
 func gatherKubernetesLogs(logs chan logFolder, errs chan error) {
-	roots := []string{k8sLogsRoot, crashDump}
-	filePaths, ers := collectFilePaths(roots)
-	for _, err := range ers {
-		errs <- err
+	roots := make([]string, 0)
+	var err error
+	_, err = os.Stat(k8sLogsRoot)
```
adjackura: stat can throw another type of error that you would want to catch as well

cli_tools/diagnostics/main_windows.go
```diff
@@ -284,15 +304,39 @@ func gatherEventLogs(logs chan logFolder, errs chan error) {
...
+	if os.IsNotExist(err) {
+		log.Printf("%s doesn't exists, no need to collect inside logs.\n", k8sLogsRoot)
+	} else {
+		roots = append(roots, k8sLogsRoot)
+	}
+	_, err = os.Stat(crashDump)
```
adjackura: same here

cli_tools/diagnostics/main_windows.go
YangLu1031: thanks, fixed

cli_tools/diagnostics/main_windows.go
YangLu1031: done

PR #1158, opened 2020-03-17
Fix the all goroutines asleep - deadlock

YangLu1031:
In the `gatherLogs()` function, it expects each log collector put either logFilePaths in logFolder channel or errors in errCh channel. So shouldn't have done nothing and return, as it caused the goroutines sleeping and failing the [tests](https://g3c.corp.google.com/results/invocations/0f83dcc3-3b68-4893-9b5d-d052b448b92b/targets/%2F%2Fcloud%2Fcluster%2Ftesting%2Fwindows:export_logs_test/tests;group=__main__.ExtractLogsTest;test=testGatherLogsCorrectly_WithOrWithoutKubernetesLog;row=3).

### DIFF
--- a/cli_tools/diagnostics/main_windows.go
+++ b/cli_tools/diagnostics/main_windows.go
@@ -274,15 +274,17 @@ func getPlainEventLogs(evts []winEvt, errs chan error) []string {
 
 // getDockerImagesList put docker images list file in logFolder channel and errors in error channel.
 func getDockerImagesList(logs chan logFolder, errs chan error) {
+	dockerImageFilePath := make([]string, 0, 1)
 	dockerPath, err := exec.LookPath("docker")
 	if err != nil {
 		log.Printf("Docker not installed, couldn't get docker images list.\n")
-		return
+	} else {
+		var commands = []runner{
+			cmd{dockerPath, "image list", dockerImageListFileName, false},
+		}
+		dockerImageFilePath = runAll(commands, errs)
 	}
-	var commands = []runner{
-		cmd{dockerPath, "image list", dockerImageListFileName, false},
-	}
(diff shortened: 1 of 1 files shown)


cli_tools/diagnostics/main_windows.go
```diff
@@ -274,15 +274,17 @@ func getPlainEventLogs(evts []winEvt, errs chan error) []string {
...
-	var commands = []runner{
-		cmd{dockerPath, "image list", dockerImageListFileName, false},
+	} else {
+		var commands = []runner{
+			cmd{dockerPath, "image list", dockerImageListFileName, false},
+		}
+		dockerImageFilePath = runAll(commands, errs)
 	}
-	logs <- logFolder{"Kubernetes", runAll(commands, errs)}
+	logs <- logFolder{"Kubernetes", dockerImageFilePath}
```
pjh: Is this going to work when gatherLogs [returns the list of paths](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/6eb244551f96a457c908fabc11e2d94a39ed9f33/cli_tools/diagnostics/main.go#L151) and then zipFiles [tries to collect them](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/6eb244551f96a457c908fabc11e2d94a39ed9f33/cli_tools/diagnostics/main.go#L157)?

I looked at zipFiles and I think it will work fine if a logFolder entry has an empty files slice, but please run the tool on a Windows VM without docker to verify this code end-to-end.

YangLu1031: Yeah, tested on non-docker and docker Windows VM end-to-end, it works

pjh:
LGTM

@adjackura please take a look, thank.

PR #1164, opened 2020-03-26
Add sysprep_windows param to image import wrapper.

EricEdens:
Testing:

Ran four imports, and confirmed that sysprep ran on (1).
1. Windows 2019, `-sysprep_windows=true`
2. Windows 2019, no `-sysprep_windows` specified
3. Ubuntu 16.04, `-sysprep_windows=true`
4. Ubuntu 16.04, no `-sysprep_windows` specified

### DIFF
--- a/cli_tools/gce_vm_image_import/main.go
+++ b/cli_tools/gce_vm_image_import/main.go
@@ -54,7 +54,7 @@ var (
 	labels               = flag.String("labels", "", "List of label KEY=VALUE pairs to add. Keys must start with a lowercase character and contain only hyphens (-), underscores (_), lowercase characters, and numbers. Values must contain only hyphens (-), underscores (_), lowercase characters, and numbers.")
 	storageLocation      = flag.String("storage_location", "", "Location for the imported image which can be any GCS location. If the location parameter is not included, images are created in the multi-region associated with the source disk, image, snapshot or GCS bucket.")
 	uefiCompatible       = flag.Bool("uefi_compatible", false, "Enables UEFI booting, which is an alternative system boot method. Most public images use the GRUB bootloader as their primary boot method.")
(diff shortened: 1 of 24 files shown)


daisy_workflows/image_import/debian/translate_debian_9.wf.json
```diff
@@ -5,6 +5,10 @@
       "Required": true,
       "Description": "The Debian 9 GCE disk to translate."
     },
+    "sysprep": {
+      "Value": "false",
+      "Description": "If enabled, run Windows Sysprep. Only applicable to Windows instances."
```
zoran15: Minor: consider changing description not to mention Windows but just state it's a sysprep flag that is currently no-op. Otherwise it might be confusing what Windows instances are being referred to here in the context of Linux.

Same for other Linux translate workflow files.

cli_tools/gce_vm_image_import/main.go
```diff
@@ -54,6 +54,7 @@ var (
 	labels               = flag.String("labels", "", "List of label KEY=VALUE pairs to add. Keys must start with a lowercase character and contain only hyphens (-), underscores (_), lowercase characters, and numbers. Values must contain only hyphens (-), underscores (_), lowercase characters, and numbers.")
 	storageLocation      = flag.String("storage_location", "", "Location for the imported image which can be any GCS location. If the location parameter is not included, images are created in the multi-region associated with the source disk, image, snapshot or GCS bucket.")
 	uefiCompatible       = flag.Bool("uefi_compatible", false, "Enables UEFI booting, which is an alternative system boot method. Most public images use the GRUB bootloader as their primary boot method.")
+	sysprepWindows       = flag.Bool("sysprep_windows", false, "Whether to generalize image using Windows Sysprep.")
```
zoran15: Mention that it only applies to Windows. This is kind of implied, but better to be explicit.

PR #1166, opened 2020-03-27
update post translate test for NGE

hopkiw:
if the google guest agent binary exists, modify the test logic for check_google_services


hopkiw:
gobuild and gocheck weren't needed here, nothing changed about go. not sure about unittests. waiting for green status before merging anyway.

PR #1169, opened 2020-03-31
[daisy] Put resource cache inside workflow to avoid race condition

dntczdx:
Put resource cache inside workflow to avoid race condition.

This problem breaks tests which run workflows in parallel.

If somebody run workflow in parallel in one program, it can also happen, even though there isn't a case so far.

### DIFF
--- a/daisy/machinetype.go
+++ b/daisy/machinetype.go
@@ -28,7 +28,7 @@ func (w *Workflow) machineTypeExists(project, zone, machineType string) (bool, D
 		return w.ComputeClient.ListMachineTypes(project, zone)
 	}, project, zone, machineType)
 	if err != nil {
-		return predefinedMachineTypeExists, err
+		return false, err
 	}
 	if predefinedMachineTypeExists {
 		return true, nil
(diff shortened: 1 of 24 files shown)


daisy/machinetype.go
```diff
@@ -18,40 +18,29 @@ import (
...
-		machineTypeCache.exists = map[string]map[string][]interface{}{}
-	}
-	if _, ok := machineTypeCache.exists[project]; !ok {
-		machineTypeCache.exists[project] = map[string][]interface{}{}
+func (w *Workflow) machineTypeExists(project, zone, machineType string) (bool, DError) {
+	predefinedMachineTypeExists, err := w.machineTypeCache.resourceExists(func(project, zone string, opts ...daisyCompute.ListCallOption) (interface{}, error) {
+		return w.ComputeClient.ListMachineTypes(project, zone)
+	}, project, zone, machineType)
+	if err != nil {
+		return predefinedMachineTypeExists, err
```
adjackura, line 31: Convention is to avoid populating other return values on error (this can cause very hard errors to debug):
return false, err

PR #1170, opened 2020-03-31
Don't allow sourceImage to be a GCS path.

EricEdens:
This consolidates the validations for sourceImage and sourceFile and provides a consistent API for dealing with them.

Additionally, this performs syntactic validation of sourceImage.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/importer_test.go
+++ b/cli_tools/gce_vm_image_import/importer/importer_test.go
@@ -16,28 +16,24 @@ package importer
 
 import (
 	"fmt"
-	"io/ioutil"
-	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
-	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/test"
-	"github.com/GoogleCloudPlatform/compute-image-tools/mocks"
-	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 )
 
 var (
-	currentExecutablePath, clientID, imageName, osID, customTranWorkflow, sourceFile, sourceImage,
+	currentExecutablePath, clientID, imageName, osID, customTranWorkflow,
 	family, description, network, subnet, labels string
 	dataDisk, noGuestEnvironment, sysprepWindows bool
+	source                                       resource
 )
 
 func TestGetWorkflowPathsFromImage(t *testing.T) {
 	resetArgs()
-	sourceImage = "image-1"
+	source = imageSource{uri: "uri"}
 	osID = "ubuntu-1404"
-	workflow, translate := getWorkflowPaths(dataDisk, osID, sourceImage, customTranWorkflow, currentExecutablePath)
+	workflow, translate := getWorkflowPaths(source, dataDisk, osID, customTranWorkflow, currentExecutablePath)
 	if workflow != path.ToWorkingDir(WorkflowDir+ImportFromImageWorkflow, currentExecutablePath) || translate != "ubuntu/translate_ubuntu_1404.wf.json" {
 		t.Errorf("%v != %v and/or translate not empty", workflow, WorkflowDir+ImportFromImageWorkflow)
 	}
@@ -47,8 +43,8 @@ func TestGetWorkflowPathsDataDisk(t *testing.T) {
 	resetArgs()
 	dataDisk = true
 	osID = ""
-	sourceImage = ""
-	workflow, translate := getWorkflowPaths(dataDisk, osID, sourceImage, customTranWorkflow, currentExecutablePath)
+	source = fileSource{}
+	workflow, translate := getWorkflowPaths(source, dataDisk, osID, customTranWorkflow, currentExecutablePath)
 	if workflow != path.ToWorkingDir(WorkflowDir+ImportWorkflow, currentExecutablePath) || translate != "" {
 		t.Errorf("%v != %v and/or translate not empty", workflow, WorkflowDir+ImportWorkflow)
 	}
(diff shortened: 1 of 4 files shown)


cli_tools/gce_vm_image_import/importer/source.go
```diff
@@ -0,0 +1,198 @@
...
+	if sourceFile != "" {
+		return newFileSource(sourceFile, storageClient)
+	}
+
+	return newImageSource(sourceImage)
+}
+
+// Whether the resource is a file in GCS.
+func isFile(s resource) bool {
+	_, ok := s.(fileSource)
```
dntczdx, line 63: Minor: my personal preference would be adding a "getSourceType() string" and check source.getSourceType() == SOURCE_TYPE_IMAGE. It's better when there are more source types to handle.

dntczdx:
LGTM overall

cli_tools/gce_vm_image_import/importer/source.go
```diff
@@ -0,0 +1,198 @@
...
+		imageName = s.uri
+	}
+	if imageName == "" || len(imageName) > 63 {
+		return daisy.Errf(
+			"invalid image name %q. Image name must be 1-63 characters long, inclusive", imageName)
+	}
+	if !imageNamePattern.MatchString(imageName) {
+		return daisy.Errf(
+			"invalid image name %q. The first character must be a lowercase letter, and all "+
+				"following characters must be a dash, lowercase letter, or digit, except the last "+
```
dntczdx, line 189: consider merging these logic to validation_utils.go

cli_tools/gce_vm_image_import/importer/source.go
```diff
@@ -0,0 +1,198 @@
...
+//   1. Using a file path instead of an image name.
+//     Examples:
+//        - file://home/user/image.vmdk
+//        - gs://bucket/image.vmdk
+//        - gs://global/images/image
+//        - https://storage.googleapis.com/bucket/image
+//   2. Using a relative file path instead of an image name.
+//     Examples:
+//        - image.vmdk
+//        - path/to/image.vmdk
```
zoran15, line 162: I feel the "offenders" would be better documented in tests instead of in method docs (it would be self documenting). This is really verbose and reduces readability IMO. It will also probably become outdated really fast once we find new offenders. That's why tests are a better documentation since they will have to be updated while docs are easily forgotten.

cli_tools/gce_vm_image_import/importer/source.go
```diff
@@ -0,0 +1,198 @@
...
+//
+// This method does not validate whether the image exist, or whether
+// the calling user has access to it.
+func (s imageSource) validate() error {
```
zoran15, line 166: is the intention to fully validate image URL? E.g. /projects/my-project/global/images/image-name? For that, you would need image URI regex validation.

cli_tools/gce_vm_image_import/importer/source_test.go
```diff
@@ -0,0 +1,133 @@
...
+	cases := []string{"file.vmdk", "gs://bucket", "gs://bucket/"}
+
+	for _, invalidPath := range cases {
+		_, err := initAndValidateSource(invalidPath, "", nil)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "is not a valid Cloud Storage object path")
+	}
+}
+
+func TestEnsureEitherFileOrImageIsPresent(t *testing.T) {
```
zoran15, line 83: These are two different test cases, split them in two functions.

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -150,111 +145,32 @@ func TestFlagsDataDiskAndOSFlagsBothProvided(t *testing.T) {
...
-}
-
-func TestFlagsSourceFileAndSourceImageBothProvided(t *testing.T) {
-	resetArgs()
-	sourceFile = "gs://source_bucket/source_file"
-	dataDisk = false
-	assertErrorOnValidate("Expected error for both source_file and source_image flags set", t)
-}
-
 func TestFlagsSourceFile(t *testing.T) {
```
zoran15, line 148: How is this testing source file when imageName is provided to validate?

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -150,111 +145,32 @@ func TestFlagsDataDiskAndOSFlagsBothProvided(t *testing.T) {
...
-	sourceImage = ""
 	dataDisk = false
 
-	if _, _, _, err := validateAndParseFlags(clientID, imageName, sourceFile, sourceImage, dataDisk,
-		osID, customTranWorkflow, labels); err != nil {
-		t.Errorf("Unexpected error: %v", err)
-	}
-}
-
-func TestFlagSourceFileEmpty(t *testing.T) {
```
zoran15, line 180: These tests are moved to source_test.go which makes sense, but I'm not sure current test setup will ensure that initAndValidateSource is called in Run(). Will tests fail if initAndValidateSource() call is deleted by mistake?

cli_tools/gce_vm_image_import/importer/source.go
```diff
@@ -0,0 +1,198 @@
...
+// sourceImage are specified.
+func initAndValidateSource(sourceFile, sourceImage string,
+	storageClient domain.StorageClientInterface) (resource, error) {
+	sourceFile = strings.TrimSpace(sourceFile)
+	sourceImage = strings.TrimSpace(sourceImage)
+
+	if sourceFile == "" && sourceImage == "" {
+		return nil, daisy.Errf(
+			"either -source_file or -source_image has to be specified")
+	} else if sourceFile != "" && sourceImage != "" {
```
zoran15, line 48: Minor: else if is unnecessary due to the return above

cli_tools/gce_vm_image_import/importer/source.go
EricEdens: My first iteration used this approach, but I figured that if we supported another source in the future it'd be easy to refactor.

Is there another source that you envision? (I couldn't think of any)

cli_tools/gce_vm_image_import/importer/source.go
EricEdens: With potentially-common code, my preference is to wait until it's needed in a second place, and then refactor it. I find that makes a cleaner API, since there are at least two concrete usages, rather than one concrete usage and a future envisioned usage. Also, sometimes that second usage never arises ..

cli_tools/gce_vm_image_import/importer/importer_test.go
zoran15: Which test actually verifies that initAndValidateSource() is called from the importer?

PR #1177, opened 2020-04-03
[daisy] populate daisy input vars early to support non-string types

dntczdx:


daisy/workflow.go
```diff
@@ -794,9 +794,9 @@ func New() *Workflow {
 
 // NewFromFile reads and unmarshals a workflow file.
 // Recursively reads subworkflow steps as well.
-func NewFromFile(file string) (*Workflow, error) {
+func NewFromFile(file string, varMap map[string]string) (*Workflow, error) {
```
zoran15, line 797: since this func is exported, this can break clients who use Daisy as a library.

daisy/step_sub_workflow.go
```diff
@@ -29,7 +29,7 @@ type SubWorkflow struct {
 func (s *SubWorkflow) populate(ctx context.Context, st *Step) DError {
 	if s.Path != "" {
 		var err error
-		if s.Workflow, err = st.w.NewSubWorkflowFromFile(s.Path); err != nil {
+		if s.Workflow, err = st.w.NewSubWorkflowFromFile(s.Path, s.Vars); err != nil {
```
zoran15, line 32: Is the loop over vars and adding them to the subworkflow below in this file needed, since it's already done by NewSubWorkflowFromFile?

daisy/step_includeworkflow.go
```diff
@@ -35,7 +35,7 @@ type IncludeWorkflow struct {
 func (i *IncludeWorkflow) populate(ctx context.Context, s *Step) DError {
 	if i.Path != "" {
 		var err error
-		if i.Workflow, err = s.w.NewIncludedWorkflowFromFile(i.Path); err != nil {
+		if i.Workflow, err = s.w.NewIncludedWorkflowFromFile(i.Path, i.Vars); err != nil {
```
zoran15, line 38: Is the loop over vars and adding them to the subworkflow below in this file needed, since it's already done by NewIncludedWorkflowFromFile (which calls readWorkflow which actually sets variables)?

dntczdx: FromFile is not the only way. The sub/included workflow can actually be directly defined in json inline. Or, when we construct workflow in memory, similarly we can directly set a sub/included workflow.

daisy/step_sub_workflow.go
dntczdx: FromFile is not the only way. The sub/included workflow can actually be directly defined in json inline. Or, when we construct workflow in memory, similarly we can directly set a sub/included workflow.

daisy/workflow.go
dntczdx: Offline discussed with Andrew, it's not a big deal. It's not breaking the core daisy steps semantic, so it won;t break users' workflow or something. The interface change is not a big problem anymore since we introduced go-modules.

daisy/workflow.go
```diff
@@ -841,8 +841,20 @@ func readWorkflow(file string, w *Workflow) DError {
 		return newErr("failed to get absolute path of workflow file", err)
 	}
 
-	if err := json.Unmarshal(data, &w); err != nil {
-		return newErr("failed to unmarshal workflow file", JSONError(file, data, err))
+	// Set input vars and run var substitution before unmarshal.
+	if varMap != nil {
```
adjackura, line 845: I think this can be simplified using bytes.ReplaceAll, strings.ReplaceAll _may_ cost a bit more but i didn't look into the implementation too deeply (use whichever you think is more readable).

```
for k, v := range varMap {
  wfJSON = bytes.RepalceAll(wfJSON, []byte(fmt.Sprintf("${%s}", k)), []byte(v))
}
```

dntczdx: Here we used strings.Replacer other than RepalceAll. Implementation of strings.Replacer is based on bytes array, which write only one pass. It's a O(n) alogorithm.

bytes.RepalceAll in a for-loop is a O(n*m) algorithm, assuming m = loop times.

adjackura: ack, thought Replacer ran in order

badp:
This commit stops older versions of Daisy from compiling.

The Daisy CLI tool depends on HEAD of this repository https://github.com/GoogleCloudPlatform/compute-image-tools/blob/20a7f19be80b4920f8b369384beb161ed580550b/cli_tools/daisy/main.go#L31 and this change broke API compatibility between the CLI tool and the libraries behind it.

EricEdens:
> This commit stops older versions of Daisy from compiling.

Hi @badp -- thanks for the report, and sorry for the breakage! We're discussing this right now and will have an update today.

adjackura:
@badp, can I ask what your build workflow is here? The code you mention (compute-image-tools/cli_tools/daisy) has not existed for a while, the daisy cli has been moved to compute-image-tools/daisy/cli. I would also suggest using go modules for your build so that you don't pull directly from the master branch.

We are still working on fixes to retain backwards compatibility, let me know if you have any more questions.

badp:
Thank you for the prompt revert. I appreciate it. I can imagine that you wouldn't want to be held back by design decisions made in 2018 when reworking your software in 2020. I regret I couldn't get back to you as quickly.

I have limited historical context having just started owning our product, but we have built a little bit of orchestration on top of [a piece of pythonry that originally wrapped daisy](https://github.com/GoogleCloudDataproc/cloud-dataproc/tree/e46cb6086b25a25a533818f037550b2fcb3bef8f) and [this repo](https://github.com/GoogleCloudPlatform/compute-image-tools/tree/20a7f19be80b4920f8b369384beb161ed580550b/). For this reason, we used git submodules for both python and go and pinned both repos to these commits. Things for the most part just worked, and the system quickly was left alone to do its thing. The very start of our CI/CD pipeline checks out the relevant git submodules, dives into [compute-image-tools/cli_tools/daisy/](https://github.com/GoogleCloudPlatform/compute-image-tools/tree/20a7f19be80b4920f8b369384beb161ed580550b/cli_tools/daisy) and runs `go get .`. 

Meanwhile, a number of things happened:

* we made a local fork of the pythonry to add a few features, most of which were contributed upstream. Some of the work we did, however, couldn't or didn't get merged; specifically we were directly manipulating the daisy JSON that the pythonry calculated so we could add GPUs to the virtual machines (turns out the nVidia drivers expect a GPU to be present at setup time), or specify things like `use-internal-proxy`.
* you put couple of years of work went into the cloud-dataproc pythonry. As I had the pleasure of discovering while reconciling our work and yours, the pythonry was first [moved to its own repo](https://github.com/GoogleCloudDataproc/cloud-dataproc/pull/129) (so much for git-imerge 🙃) then [had Daisy removed from its innards quoting a "deprecation"](https://github.com/GoogleCloudDataproc/custom-images/pull/26) in lieu of [automatically generating bash scripts](https://github.com/GoogleCloudDataproc/custom-images/blob/master/custom_image_utils/shell_script_generator.py#L21)
* you also put a couple of years of work into Daisy itself, and to my great surprise today, we've been silently getting most of it silently for free, since the Daisy CLI depended on HEAD of this repository instead of the contents of the repository itself...

I think the shortest path out of this situation for our own use case, and the amount of resources we can allocate to this, is to make one or two forks of this repository, so we can compile the CLI tool at 20a7f19be80b4920f8b369384beb161ed580550b except it's pointing to this repo at tag v20022026; that way we'll be back to a last known good state. We could then perhaps try and adopt v20022026 completely...

PR #1188, opened 2020-04-24
[cli tools] fix dependencies to osconfig

dntczdx:
[cli tools] fix dependencies to osconfig. The packages' path has been changed. Need to fix accordingly.


fejta:
zoink

PR #1191, opened 2020-05-01
Fixed lookup for .ovf descriptor doing a full depth search

zoran15:
Fixed OVF lookup for .ovf descriptor doing afull depth lookup  instead of looking only at the same level as the GCS path provided.


dntczdx:
Overall LGTM, is it ready to review?

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -172,6 +184,27 @@ func (sc *Client) FindGcsFile(gcsDirectoryPath string, fileExtension string) (*s
 		"path %v doesn't contain a file with %v extension", gcsDirectoryPath, fileExtension)
 }
 
+func isDepthValid(lookupDepth int, lookupPath, objectPath string) bool {
```
dntczdx, line 187: This is a bit confusion to me. It looks not like a recursive. Will the it.next() cover all files in any depth of subfolders?

zoran15: Recursive as in going through the full GCS folder system instead of checking only the current level. The recursive part doesn't happen in our code but instead in the results of GCS lookup.

EricEdens: I had the same instinct as Tony (I clarified offline with Zoran), so maybe it's worth adjusting the wording?

EricEdens, line 187: minor: The name isDepthValid took me a bit to understand. An option to consider is refactoring it to have a method such as `distance(gcsDirectoryPath, f string)` that returns the number of directories between gcsDirectoryPath and the file f. Eg:

distance('/tmp', '/tmp/file')  =>  0  (same directory)
distance('/', '/tmp/file')  =>  1  (one sub dir)

This formulation gives you a nice `if` statement at 173, and corresponds directly to your definitions of gcsDirectoryPath and lookupDepth.

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -145,13 +146,22 @@ func (sc *Client) DeleteGcsPath(gcsPath string) error {
...
+// gcsDirectoryPath, lookupDepth should be set as 0. For recursive lookup with
+// no limitations on depth, lookupDepth should be -1
+// File extension can be a file name as well.
+func (sc *Client) FindGcsFileDepthLimited(gcsDirectoryPath string, fileExtension string, lookupDepth int) (*storage.ObjectHandle, error) {
+	bucketName, lookupPath, err := SplitGCSPath(gcsDirectoryPath)
 	if err != nil {
 		return nil, err
 	}
-	it := sc.GetObjects(bucketName, objectPath)
+	it := sc.GetObjects(bucketName, lookupPath)
```
EricEdens, line 164: Since you're depending on the Client object, your test code will need a lot of deep mocking.  An alternative is to express your dependency in terms of the data that you need, which, AFAICT, is an iterator-style source of strings. 

This is just an idea -- either for this PR if you like it, or a future PR where you find yourself doing deep mocking.

cli_tools/common/utils/storage/storage_client.go
dntczdx: @zoran15 if that is the case, it seems that depth check can't reduce the lookup workload. Every single files will be gone through. Per my understanding, depth check is usually for limiting the workload. Here it doesn't help.

zoran15: Updated the description of this PR so "recursive" is not mentioned.

@dntczdx You're right that the lookup code will still look at all the files within a search path. However, this fix is not about reducing the workload or the number of files looked at. It's about correctness: customers provide GCS path to a folder that should contain OVF archive including a descriptor. This lookup should not try to search for OVF archives in any of the subfolders, but only at the level provided by the customer (the flag is currently behaving like a prefix search which is not correct). E.g. if a customer provides "gs://bucket/" as gcs path, we will currently check all subfolders and pick the first that contains an .ovf descriptor. This can lead to unintended OVF archives from being imported.

cli_tools/common/utils/storage/storage_client.go
zoran15: What do you mean "depending on the Client object"? This is a method of storage Client.

Also, what do you mean by "test code will need a lot of deep mocking"?

EricEdens: The ~20 lines of setup code here, for example:

The key insight is that the algorithmic portion of your code doesn't care about GCS or `storage.ObjectAttrs` - its main concern is filepath-like strings.
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/420d6d8c0afb11d1fdba06a50c8759c4ed332524/cli_tools/common/utils/storage/storage_client_test.go#L284

zoran15: isDepthValid is tested on it's own. Then there are some more higher level tests like the one you referenced above that test FindGcsFileDepthLimited() in more detail. 

I don't quite see what is your concern here.

PR #1194, opened 2020-05-02
Decouple daisy from import/export logging.

EricEdens:
We're preparing to break image import into smaller pieces, so we need a way to communicate results without having a concrete `daisy.Workflow` object. 

This change does the following:
* Removes logger's dependency on daisy.Workflow
* Adds a shim so that existing usages still work

I'm vacillating on the interface for `Loggable`, specifically whether to use this generic key/value system, or to have strongly-named and strongly-typed fields, such as `getImportFormat` and `getImportSizesGb`. Thoughts?

### DIFF
--- a/cli_tools/common/utils/logging/service/loggables.go
+++ b/cli_tools/common/utils/logging/service/loggables.go
@@ -1,0 +1,73 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package service
+
+import (
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+	"strconv"
+	"strings"
+)
+
+// WorkflowToLoggable is a shim for a daisy workflow, exposing only those
+// fields that are pertinent to logging.
+func WorkflowToLoggable(wf *daisy.Workflow) Loggable {
+	return workflowLoggable{wf: wf}
+}
+
+type workflowLoggable struct {
+	wf *daisy.Workflow
+}
+
+func (w workflowLoggable) GetKeyValueAsKey(key string) string {
+	return w.wf.GetSerialConsoleOutputValue(key)
+}
+
+func (w workflowLoggable) GetKeyValueAsInt64Slice(key string) []int64 {
+	return getInt64Values(w.wf.GetSerialConsoleOutputValue(key))
+}
+
+func (w workflowLoggable) ReadSerialPortLogs() []string {
+	if w.wf.Logger != nil {
+		logs := w.wf.Logger.ReadSerialPortLogs()
+		view := make([]string, len(logs))
+		copy(view, logs)
+		return view
+	}
+	return nil
+}
+
+type literalLoggable struct {
+	strings map[string]string
+	int64s  map[string][]int64
+	serials []string
+}
+
+func (w literalLoggable) GetKeyValueAsKey(key string) string { return w.strings[key] }
+
+func (w literalLoggable) GetKeyValueAsInt64Slice(key string) []int64 { return w.int64s[key] }
+
+func (w literalLoggable) ReadSerialPortLogs() []string { return w.serials }
+
+func getInt64Values(s string) []int64 {
+	strs := strings.Split(s, ",")
+	var r []int64
+	for _, str := range strs {
+		i, err := strconv.ParseInt(str, 0, 64)
+		if err == nil {
+			r = append(r, i)
+		}
+	}
+	return r
+}
--- a/cli_tools/common/utils/logging/service/loggables.go
+++ b/cli_tools/common/utils/logging/service/loggables.go
@@ -30,7 +30,7 @@ type workflowLoggable struct {
 	wf *daisy.Workflow
 }
 
-func (w workflowLoggable) GetKeyValueAsKey(key string) string {
+func (w workflowLoggable) GetKeyValueAsString(key string) string {
 	return w.wf.GetSerialConsoleOutputValue(key)
 }
 
@@ -54,7 +54,7 @@ type literalLoggable struct {
 	serials []string
 }
 
-func (w literalLoggable) GetKeyValueAsKey(key string) string { return w.strings[key] }
+func (w literalLoggable) GetKeyValueAsString(key string) string { return w.strings[key] }
 
 func (w literalLoggable) GetKeyValueAsInt64Slice(key string) []int64 { return w.int64s[key] }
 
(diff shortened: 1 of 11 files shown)


cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+type literalLoggable struct {
+	strings map[string]string
+	int64s  map[string][]int64
+	serials []string
+}
+
+func (w literalLoggable) GetKeyValueAsKey(key string) string { return w.strings[key] }
```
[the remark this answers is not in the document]
EricEdens: good find, yeah, GetKeyValueAsString is correct. I'll fix that before I push.

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+package service
+
+import (
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+	"strconv"
+	"strings"
+)
+
+// WorkflowToLoggable is a shim for a daisy workflow, exposing only those
+// fields that are pertinent to logging.
```
zoran15, line 24: pertinent -> relevant to improve readability for non-native English speakers in the future

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+	strs := strings.Split(s, ",")
+	var r []int64
+	for _, str := range strs {
+		i, err := strconv.ParseInt(str, 0, 64)
+		if err == nil {
+			r = append(r, i)
+		}
+	}
+	return r
+}
```
zoran15, line 73: Is this something that can be reused in the future? If so, move to utils

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+
+package service
+
+import (
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+	"strconv"
+	"strings"
+)
+
+// WorkflowToLoggable is a shim for a daisy workflow, exposing only those
```
zoran15, line 23: WorkflowToLoggable is actually a function that returns a shim

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+// fields that are pertinent to logging.
+func WorkflowToLoggable(wf *daisy.Workflow) Loggable {
+	return workflowLoggable{wf: wf}
+}
+
+type workflowLoggable struct {
+	wf *daisy.Workflow
+}
+
+func (w workflowLoggable) GetKeyValueAsString(key string) string {
```
zoran15, line 33: GetValue?

The rest is obvious from the func signature (although there would be a conflict with GetKeyValueAsInt64Slice). Also, `KeyValue` could be interpreted as a tuple, which this function doesn't return.

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+	return w.wf.GetSerialConsoleOutputValue(key)
+}
+
+func (w workflowLoggable) GetKeyValueAsInt64Slice(key string) []int64 {
```
zoran15, line 37: Similar, though it conflicts with the above. At least remove `Key` portion of the name

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (w workflowLoggable) ReadSerialPortLogs() []string {
+	if w.wf.Logger != nil {
+		logs := w.wf.Logger.ReadSerialPortLogs()
+		view := make([]string, len(logs))
+		copy(view, logs)
+		return view
+	}
+	return nil
```
zoran15, line 48: Shouldn't this return an error instead of nil?

cli_tools/common/utils/logging/service/loggables.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type literalLoggable struct {
```
zoran15, line 51: How many different types of Loggables do you see we'll have? It might make sense to split them in their own source code files.

cli_tools/common/utils/logging/service/loggables.go
EricEdens: sgtm

cli_tools/common/utils/logging/service/loggables.go
EricEdens: It's moved from logger.go in this same directory. I couldn't think of where we'd need this elsewhere -- did you have something in mind?

In terms of common usage, I'm hesitant mostly over its decision to silently ignore failures. I *think* it's the correct choice here, but it wouldn't be the universal choice.

cli_tools/common/utils/logging/service/loggables.go
EricEdens: good call

cli_tools/common/utils/logging/service/loggables.go
EricEdens: sgtm

cli_tools/common/utils/logging/service/loggables.go
EricEdens: sgtm

cli_tools/common/utils/logging/service/loggables.go
EricEdens: I'm using `nil` since this will get invoked on any workflow failure, and as such, there may not be worker logs. For example: if the user's specified project isn't found, we'd never start a worker.

cli_tools/common/utils/logging/service/loggables.go
EricEdens: While I don't have immediate plans for creating more, I like the idea of splitting the file!

EricEdens:
@zoran15: Good question. Answer below:

> > We're preparing to break image import into smaller pieces, so we need a way to communicate results without having a concrete `daisy.Workflow` object.
> 
> Can you elaborate on this? How will logging facilitate communicating results between workflows? A short doc with a code samples might help what is the intention behind this change.

The current logging code assumes a 1:1 correspondence between a CLI operation and a daisy workflow. The operation runs, then passes its workflow to the logging system. As described in your dynamic daisy doc, however, we're planning on using multiple workflows to implement a single CLI operation. Hence the 1:1 correspondence is no longer valid.

Following this PR, the operation may return a daisy workflow, or it may return those fields that the logging system cares about directly.

In the case of image import, the inflate workflows runs to create the PD. Next, the translation workflow runs. [importer.Run](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/gce_vm_image_import/importer/importer.go) returns a Loggable that contains the combined fields from these two workflows that are relevant to logging.

cli_tools/common/utils/logging/service/loggables.go
zoran15: This func doesn't have anything to do with logging, it's more a string parsing utility. It might be useful in the future for new code and having it in a common util location would make it easier to find.

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -170,39 +169,27 @@ func getAnonymizedFailureReason(err error) string {
...
-		i, err := strconv.ParseInt(str, 0, 64)
-		if err == nil {
-			r = append(r, i)
-		}
-	}
-	return r
-}
-
-func (l *Logger) runWithServerLogging(function func() (*daisy.Workflow, error),
+func (l *Logger) runWithServerLogging(function func() (Loggable, error),
```
zoran15, line 192: minor: can be a good time to rename `function` to something more descriptive. Even 'funcToRun` is better than `function` which almost looks like a reserved word.

PR #1195, opened 2020-05-04
daisy: fix image IgnoreLicenseValidationIfForbidden

### LINKED ISSUE #1187: Unable to embed enable-vmx license in the latest docker release
The latest release image `gcr.io/compute-image-tools/daisy@sha256:efed4cef7eaf128536929ab040538d59478e1ff2ed1c09e705c8b2b46e8fcc34` broke our VM image release which embeds the license: `projects/vm-options/global/licenses/enable-vmx`

The error was:

```
bad license lookup: "projects/vm-options/global/licenses/enable-vmx", error: error listing resource for project "vm-options": googleapi: Error 403: Required 'compute.licenses.list' permission for 'projects/vm-options', forbidden
```

Note that we have set `IgnoreLicenseValidationIfForbidden` to ture and it had been working until April 16.

adjackura:
Fixes #1187

### DIFF
--- a/daisy/test_common_test.go
+++ b/daisy/test_common_test.go
@@ -41,7 +41,10 @@ import (
 	"google.golang.org/grpc"
 )
 
-const DNE = "DNE!"
+const (
+	DNE  = "DNE!"
+	p403 = "p403"
+)
 
 type mockStep struct {
 	populateImpl func(context.Context, *Step) DError
@@ -215,6 +218,9 @@ func newTestGCEClient() (*daisyCompute.TestClient, error) {
 		return []*compute.ForwardingRule{{Name: testForwardingRule}}, nil
 	}
 	c.ListLicensesFn = func(p string, _ ...daisyCompute.ListCallOption) ([]*compute.License, error) {
+		if p == p403 {
+			return nil, &googleapi.Error{Code: 403}
+		}
 		if p != testProject {
 			return nil, errors.New("bad project: " + p)
 		}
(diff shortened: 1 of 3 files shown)


daisy/test_common_test.go
```diff
@@ -215,6 +218,9 @@ func newTestGCEClient() (*daisyCompute.TestClient, error) {
 		return []*compute.ForwardingRule{{Name: testForwardingRule}}, nil
 	}
 	c.ListLicensesFn = func(p string, _ ...daisyCompute.ListCallOption) ([]*compute.License, error) {
+		if p == p403 {
+			return nil, &googleapi.Error{Code: 403}
```
zoran15, line 222: Why is this needed? 

Do we need to handle other error codes in the same way?

adjackura: This is a test for this scenario, were testing whether we handle the 403 correctly.

huangw5: The PR doesn't fix the issue but instead bypass the check. The root cause was https://github.com/GoogleCloudPlatform/compute-image-tools/issues/1187 incorrectly changed the required license permission from "get" to "list". Some GCP projects, e.g. `vm-options`, may allow anyone to "get" the specific license, but they don't allow anyone to "list" all available licenses hosted on their projects.

adjackura: Is this 'get' specific license acl new? This bypass was specifically created to solve the license issue as you can have no access to a license (not get access) but can still use it when creating an image. If there is a specific get acl now we can update the logic to run get if list fails, but that's a separate change. This change just fixes the IgnoreLicenseValidationIfForbidden option.

adjackura: I just tested and this project has no special GET permsions as far as I can tell, I get the same error:
APIError: googleapi: Error 403: Required 'compute.licenses.get' permission for 'projects/vm-options/global/licenses/enable-vmx', forbidden

adjackura: Also moving to the list call vs the get call was because licenses.list was not available when first implemented but it is now. All the cache checks use the list call so this was bringing the license lookup inline. If fallback to get calls is wanted we can implement that across the board.

PR #1198, opened 2020-05-05
Retry apt-get update until a package is successfully installed.

EricEdens:
Over the weekend, our opensuse 15.1 tests [1] started failing. The root cause is that the Debian 10 worker was failing to install a dependency because it couldn't reach security.debian.org on its first attempt.

1. daisy_integration_tests/opensuse_15_1.wf.json

### DIFF
--- a/daisy_workflows/linux_common/utils/common.py
+++ b/daisy_workflows/linux_common/utils/common.py
@@ -88,13 +88,12 @@ YumInstall.first_run = True
 
 @RetryOnFailure()
 def AptGetInstall(package_list, suite=None):
-  if AptGetInstall.first_run:
-    try:
-      Execute(['apt-get', 'update'])
-    except subprocess.CalledProcessError as error:
-      logging.info('Apt update failed, trying again: %s' % error)
-      Execute(['apt-get', 'update'], raise_errors=False)
-    AptGetInstall.first_run = False
+  # When `apt-get update` fails to update a repo, it returns 0.
+  # This check ensures that we retry running update until we've
+  # had one successful install.
+  # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=778357
+  if not AptGetInstall.prior_success:
+    Execute(['apt-get', 'update'])
 
   env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


daisy_workflows/linux_common/utils/common.py
```diff
@@ -88,13 +88,12 @@ def YumInstall(package_list):
...
-    except subprocess.CalledProcessError as error:
-      logging.info('Apt update failed, trying again: %s' % error)
-      Execute(['apt-get', 'update'], raise_errors=False)
-    AptGetInstall.first_run = False
+  # When `apt-get update` fails to update a repo, it returns 0.
+  # This check ensures that we retry running update until we've
+  # had one successful install.
+  # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=778357
+  if not AptGetInstall.prior_success:
+    Execute(['apt-get', 'update'])
```
zoran15, line 96: Should there be some kind of timeout/re-try count limit? Daisy step will timeout eventually but if apt-get is not expected to succeed after 10s of minutes of re-tries anyway, it might make sense to limit them.

EricEdens: Here are the defaults for the retry annotation: 

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/2519227884218345e99c81207db99eaae86dd0e5/daisy_workflows/linux_common/utils/common.py#L36

EricEdens:
> How does the "apt-get retry" actually work here? It seems, with this change, AptGetInstall will throw errors (instead of catching them and retrying) to the calling code. Will the calling code handle cathing these errors and retrying apt-get?

The core issue is that `Execute(['apt-get', 'update'])` doesn't throw when it fails since apt-get update always reports success (see debian#778357). With this change, we assume that it failed, unless the subsequent apt-get install succeeds, in which case we assume that it passed.

In terms of retry, the `@RetryOnFailure` decorator handles that. It throws once the retry limits have been exceeded.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=778357

PR #1206, opened 2020-05-08
If translation fails, alert user of what was detected on disk.

EricEdens:
This adds a new module, `cli_tools/common/distro`, that is used to determine whether the detection result matches the user's input. This will get used in the upcoming OS detection project.

Notably, the module doesn't (yet) support Windows. That's a tactical choice, since currently we only perform detection on Linux, and unlike Linux, in Windows the user-facing version strings are different than the internal version strings [1]. So, I'm leaving implementation of Windows for when it's needed, as that will help define the interface.

Lastly, if the detection result doesn't match the user's input, then we replace the current error with a new one. I like this approach, since it improves the signal-to-noise ratio, and since the Daisy error is logged a few lines before. 

Here are the last few log lines, as an example:

```
[import-from-image]: 2020-05-08T12:16:31-07:00 Error running workflow: step "translate-disk" run error: step "translate-disk" run error: step "wait-for-translator" run error: WaitForInstancesSignal FailureMatch found for "inst-translator-import-from-image-translate-disk-transla-3mzmg": "TranslateFailed: error: sh: /bin/sh: 1: cannot create /etc/resolvconf/resolv.conf.d/base: Directory nonexistent"
[import-from-image]: 2020-05-08T12:16:31-07:00 Serial-output value -> detected_distro:debian
[import-from-image]: 2020-05-08T12:16:31-07:00 Serial-output value -> detected_major_version:9
[import-from-image]: 2020-05-08T12:16:31-07:00 Serial-output value -> detected_minor_version:12
[import-from-image]: 2020-05-08T12:16:31-07:00 Workflow "import-from-image" cleaning up (this may take up to 2 minutes).
[import-from-image]: 2020-05-08T12:17:35-07:00 Workflow "import-from-image" finished cleanup.
[import-image] 2020/05/08 12:17:35 "debian-9" was detected on your disk, but "ubuntu-1604" was specified. Please verify and re-import
```

It wasn't intentional to include the extra lines such as `Serial-output value -> detected_distro:debian`. Those are a side effect of the key/value serial logging mechanism.

https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions

### DIFF
--- a/cli_tools/common/distro/distro.go
+++ b/cli_tools/common/distro/distro.go
@@ -31,7 +31,7 @@ const (
 )
 
 // Release encapsulates product and version information
-// about a Linux or Windows release.
+// about the operating systems we support.
 type Release interface {
 	ImportCompatible(other Release) bool
 	AsGcloudArg() string
--- a/cli_tools/common/distro/distro.go
+++ b/cli_tools/common/distro/distro.go
@@ -1,0 +1,179 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package distro
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// Release encapsulates product and version information
+// about a Linux or Windows release.
+type Release interface {
+	ImportCompatible(other Release) bool
+	AsGcloudArg() string
+}
+
+type release struct {
+	distro  string
+	variant string
+	major   string
+	minor   string
+}
+
+// AsGcloudArg returns a string representation using the format
+// that we use with the `--os` flag.
+func (l release) AsGcloudArg() string {
+	switch l.distro {
+	case "ubuntu":
+		return fmt.Sprintf("%s-%s%s", l.distro, l.major, l.minor)
+	case "sles":
+		if l.variant != "" {
+			return fmt.Sprintf("%s-%s-%s", l.distro, l.variant, l.major)
+		} else {
+			return fmt.Sprintf("%s-%s", l.distro, l.major)
+		}
+	default:
+		return fmt.Sprintf("%s-%s", l.distro, l.major)
+	}
+}
+
+// ImportCompatible uses product-specific information
+// to determine whether two versions are compatible for import.
+func (l release) ImportCompatible(other Release) bool {
+	realOther, ok := other.(release)
+	if !ok {
+		return false
+	}
+
+	if l.distro == "ubuntu" {
+		return l.distro == realOther.distro &&
+			l.major == realOther.major &&
+			l.minor == realOther.minor
+	}
+
+	if l.distro == "sles" {
+		return l.distro == realOther.distro &&
+			l.variant == realOther.variant &&
+			l.major == realOther.major
+	}
+
+	return l.distro == realOther.distro &&
+		l.major == realOther.major
+}
+
+// FromLibguestfs initializes a DistroAndVersion from the fields returned by libguestfs's
+// inspection routines.
+//
+// http://libguestfs.org/guestfs.3.html#guestfs_inspect_get_distro
+func FromLibguestfs(distro string, major string, minor string) (r Release, e error) {
+	distro = strings.ToLower(distro)
+	if distro == "windows" {
+		// We don't currently need windows for this parsing, and punting
+		// since it's not immediately clear how to represent their
+		// version strings, as the user-facing value (Windows 2008r2)
+		// does not match what's used internally by our tools (NT 5.2)
+		// https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
+		return r, errors.New("windows not yet implemented")
+	}
+	if distro == "" || major == "" {
+		return nil, errors.New("distro and major are required")
+	}
+	return release{
+		distro: distro,
+		major:  major,
+		minor:  minor,
+	}, nil
+}
+
+// ParseGcloudOsParam parses the value associated with the `--os` parameter
+// from the `gcloud compute image` tools.
+//
+// https://cloud.google.com/sdk/gcloud/reference/compute/images/import#--os
+func ParseGcloudOsParam(os string) (r Release, e error) {
+	os = strings.ToLower(os)
+	if strings.HasSuffix(os, "-byol") {
+		os = strings.TrimSuffix(os, "-byol")
+	}
+	if strings.HasPrefix(os, "windows") {
+		// We don't currently need windows for this parsing, and punting
+		// since it's not immediately clear how to represent their
+		// version strings, as the user-facing value (Windows 2008r2)
+		// does not match what's used internally by our tools (NT 5.2)
+		// https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
+		return r, errors.New("windows not yet implemented")
+	}
+	parts := strings.Split(os, "-")
+	distro := parts[0]
+
+	// 1. Format is `(debian|centos|rhel|opensuse)-major`.
+	//   Canonical example: debian-8
+	if distro == "debian" ||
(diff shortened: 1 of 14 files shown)


daisy_workflows/linux_common/utils/diskutils.py
```diff
@@ -24,6 +24,18 @@
   AptGetInstall(['python3-guestfs'])
   import guestfs
 
+_STATUS_PREFIX = 'TranslateStatus: '
+
+
+def log_key_value(key, value):
+  """
+  Prints the key and value using the format defined by
+  Daiys's serial output inspector.
```
zoran15, line 33: Daisy's

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -246,6 +248,19 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
...
+		detected, _ := distro.FromLibguestfs(
+			w.GetSerialConsoleOutputValue("detected_distro"),
+			w.GetSerialConsoleOutputValue("detected_major_version"),
+			w.GetSerialConsoleOutputValue("detected_minor_version"))
+		if fromUser != nil && detected != nil && !fromUser.ImportCompatible(detected) {
+			// The error is already logged by Daisy, so skipping re-logging it here.
+			return w, fmt.Errorf("%q was detected on your disk, "+
+				"but %q was specified. Please verify and re-import",
+				detected.AsGcloudArg(), fromUser.AsGcloudArg())
+		}
```
zoran15, line 263: 1. This appears to indiscriminately hides the original error when detected log values are written. What if import returned detected OS version import failed because another reason? 
2. Even if the cause is mismatched OS flag value, for debugging and sanity checks, we should include the original error as well. 
3. This code should be extracted to a func to reduce the size of Run() and improve it's readability. Ideally to a separate util file.

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+package distro
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// Release encapsulates product and version information
+// about a Linux or Windows release.
```
zoran15, line 24: Why only Linux and Windows? If, in the future, we have other OS's, would they be supproted?

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+		} else {
+			return fmt.Sprintf("%s-%s", l.distro, l.major)
+		}
+	default:
+		return fmt.Sprintf("%s-%s", l.distro, l.major)
+	}
+}
+
+// ImportCompatible uses product-specific information
+// to determine whether two versions are compatible for import.
```
zoran15, line 55: What does it mean that two versions are compatible for import?

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+func (l release) ImportCompatible(other Release) bool {
+	realOther, ok := other.(release)
+	if !ok {
+		return false
+	}
+
+	if l.distro == "ubuntu" {
```
zoran15, line 62: extract OS names to constants as they are reused

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+// http://libguestfs.org/guestfs.3.html#guestfs_inspect_get_distro
+func FromLibguestfs(distro string, major string, minor string) (r Release, e error) {
+	distro = strings.ToLower(distro)
+	if distro == "windows" {
+		// We don't currently need windows for this parsing, and punting
+		// since it's not immediately clear how to represent their
+		// version strings, as the user-facing value (Windows 2008r2)
+		// does not match what's used internally by our tools (NT 5.2)
+		// https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
+		return r, errors.New("windows not yet implemented")
```
zoran15, line 90: capitalize Windows

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// FromLibguestfs initializes a DistroAndVersion from the fields returned by libguestfs's
+// inspection routines.
+//
+// http://libguestfs.org/guestfs.3.html#guestfs_inspect_get_distro
+func FromLibguestfs(distro string, major string, minor string) (r Release, e error) {
+	distro = strings.ToLower(distro)
+	if distro == "windows" {
+		// We don't currently need windows for this parsing, and punting
```
zoran15, line 85: Why is Windows not needed?

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+		major:  major,
+		minor:  minor,
+	}, nil
+}
+
+// ParseGcloudOsParam parses the value associated with the `--os` parameter
+// from the `gcloud compute image` tools.
+//
+// https://cloud.google.com/sdk/gcloud/reference/compute/images/import#--os
+func ParseGcloudOsParam(os string) (r Release, e error) {
```
zoran15, line 106: os -> osFlagValue or similar

cli_tools/common/distro/distro_test.go
```diff
@@ -0,0 +1,176 @@
...
+		err string
+	}{
+		{"sles", "unrecognized SLES identifier: \"sles\""},
+		{"kali-12", "unrecognized identifier: \"kali-12\""},
+		{"ubuntu", "expected pattern of `distro-version`"},
+		{"ubuntu-12", "expected version with length four"},
+		{"opensuse-15-leap", "expected pattern of `distro-version`"},
+		{"debian", "expected pattern of `distro-version`"},
+		{"centos7", "unrecognized identifier: \"centos7\""},
+		{"rhel", "expected pattern of `distro-version`"},
```
zoran15, line 100: 1. Empty string
2. non OS ID value, e.g. "notAnOS", "notAnOS-1804"...

cli_tools/common/distro/distro_test.go
```diff
@@ -0,0 +1,176 @@
...
+		safeParse(t, "centos-7").ImportCompatible(
+			safeFromLibguestfs(t, "centos", "7", "1")))
+
+	assert.True(t,
+		safeParse(t, "ubuntu-1404").ImportCompatible(
+			safeFromLibguestfs(t, "ubuntu", "14", "04")))
+
+	assert.False(t,
+		safeParse(t, "ubuntu-1404").ImportCompatible(
+			safeFromLibguestfs(t, "ubuntu", "14", "10")))
```
zoran15, line 127: Add more edge case tests, e.g. differet OS type

daisy_workflows/linux_common/utils/diskutils.py
EricEdens: done

cli_tools/common/distro/distro.go
EricEdens: done

cli_tools/common/distro/distro.go
EricEdens: done

cli_tools/common/distro/distro_test.go
EricEdens: done

cli_tools/common/distro/distro_test.go
EricEdens: done

cli_tools/common/distro/distro.go
EricEdens: talked offline

cli_tools/common/distro/distro.go
EricEdens: talked offline

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: talked offline

cli_tools/common/distro/distro.go
EricEdens: done

cli_tools/common/distro/distro.go
EricEdens: talked offline

cli_tools/common/distro/distro.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		return release{
+			distro: distro,
+			major:  version[:2],
+			minor:  version[2:],
+		}, nil
+	}
+
+	return r, fmt.Errorf("unrecognized identifier: `%s`", os)
```
dntczdx, line 186: Is there any chance to use this error? It seems that the caller swallowed it.

EricEdens: In the current usage, you're right, it gets swallowed. That happens since we've already validated the `--os` argument parameter further up in the method.

PR #1210, opened 2020-05-12
Provide dependencies to importer.Run

EricEdens:
(This PR is working toward splitting `gce_vm_image_import/importer/importer.go` into multiple workflows.)

Prior to this change, importer.Run couldn't be unit tested, since it instantiated concrete dependencies pointing at prod GCE endpoints. By removing dependency construction from importer.Run, this refactoring enables unit tests in the subsequent PRs. 

Additionally, it refactors `common/utils/logging/logger.go` to achieve the following:
1. Export interfaces, not implementations
2. Provide an implementation that uses the standard logger
3. Clarify that the existing logger writes to stdout.

### Testing
- Ran import using gce_vm_image_import/main


cli_tools/common/utils/param/fixer.go
```diff
@@ -0,0 +1,77 @@
...
+//  limitations under the License.
+
+package param
+
+import (
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/domain"
+)
+
+// Fixer standardizes user input, and determines omitted values.
+type Fixer interface {
```
dntczdx, line 22: I rarely see the word "Fixer" used for this purpose. Not sure whether there is a better word.

cli_tools/common/utils/param/fixer_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+	mockMetadataGce := mocks.NewMockMetadataGCEInterface(mockCtrl)
+	mockScratchBucketCreator := mocks.NewMockScratchBucketCreatorInterface(mockCtrl)
+	mockResourceLocationRetriever := mocks.NewMockResourceLocationRetrieverInterface(mockCtrl)
+	mockResourceLocationRetriever.EXPECT().GetZone("us-west2", project).Return("",
+		daisy.Errf("zone not found")).Times(1)
+	mockStorageClient := mocks.NewMockStorageClientInterface(mockCtrl)
+	mockStorageClient.EXPECT().GetBucketAttrs("scratchbucket").Return(&storage.BucketAttrs{Location: "us-west2"}, nil).Times(1)
+
+	err := fixer{
```
dntczdx, line 48: Call NewFixer instead of fixer{} to test one more layer?

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -240,7 +223,7 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
 		description, *region, subnet, network, noGuestEnvironment, sysprepWindows)
 
 	var w *daisy.Workflow
-	if w, err = runImport(ctx, varMap, importWorkflowPath, zone, timeout, *project, scratchBucketGcsPath,
+	if w, err = runImport(context.Background(), varMap, importWorkflowPath, zone, timeout, *project, scratchBucketGcsPath,
```
dntczdx, line 226: if object ctx is not shared across functions, it's not necessary to pass it into runImport. Can call context.Background() inside runImport

cli_tools/common/utils/param/fixer_test.go
EricEdens: Nice idea! Done.

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: sgtm, done

cli_tools/common/utils/param/fixer.go
EricEdens: I struggled as well. Consumers outside this package see `param.Fixer`, which seems reasonable.

cli_tools/common/utils/logging/logger.go
```diff
@@ -33,22 +34,31 @@ func (e *LogEntry) String() string {
...
-// NewLogger creates a new logger which uses prefix for all messages logged
-func NewLogger(prefix string) *Logger {
-	return &Logger{Prefix: prefix}
+// NewStdoutLogger creates a new logger which uses prefix for all messages logged.
+// All messages are sent to stdout.
+func NewStdoutLogger(prefix string) LoggerInterface {
+	return stdoutLogger{prefix: prefix}
 }
 
+type stdoutLogger struct{ prefix string }
```
zoran15, line 48: This could be made more flexible by parametrizing the logger with a file to write to. Stdout is just a file in GoLang so default implementation will use os.Stdout

cli_tools/common/utils/param/fixer.go
zoran15: +1 on this. Fixed implies something is broken (like parameters having incorrect values, when actually parameters don't have values set at all).

cli_tools/common/utils/logging/logger.go
EricEdens: My intuition is that this is beyond the scope of this review -- this PR just renames the method to make it clear what it was already doing.

EricEdens:
> +1 on this. Fixer implies something is broken (like parameters having incorrect values, when actually parameters don't have values set at all).

@dntczdx , @zoran15 :  `param.Fixer` was the best I could think of. You have some ideas on alternatives?

cli_tools/common/utils/param/fixer.go
EricEdens: `param.Fixer` was the best I could think of. You have some ideas on alternatives?

zoran15: Populator?

PR #1215, opened 2020-05-13
Only create workflowLoggable if workflow is non-nil.

EricEdens:
This was prompted by a failed OVF integration test. The import failed while copying a VMDK to GCS. A workflow was never created, and a subsequent call to workflowLoggable.wf caused a seg fault.

### DIFF
--- a/cli_tools/common/utils/logging/service/daisy_loggable.go
+++ b/cli_tools/common/utils/logging/service/daisy_loggable.go
@@ -15,13 +15,17 @@
 package service
 
 import (
-	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 	"strconv"
 	"strings"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
 
-// WorkflowToLoggable provides a Loggable from a daisy workflow.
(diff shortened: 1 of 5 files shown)


cli_tools/common/utils/logging/service/daisy_loggable.go
```diff
@@ -15,13 +15,17 @@
...
 	"strconv"
 	"strings"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
 
-// WorkflowToLoggable provides a Loggable from a daisy workflow.
-func WorkflowToLoggable(wf *daisy.Workflow) Loggable {
+// NewWorkflowToLoggable provides a Loggable from a daisy workflow.
+func NewWorkflowToLoggable(wf *daisy.Workflow) Loggable {
```
zoran15, line 25: I'm a bit confused by the name of this func: NewWorkflowToLoggable sounds like a new workflow is created? Or a loggable is created from a new workflow?

Maybe call it NewLoggable or similar?

EricEdens: Good point: `NewWorkflowToLoggable` is a funky name. Switched to `NewLoggableFromWorkflow`

PR #1218, opened 2020-05-19
remove presubmits

hopkiw:


dntczdx:
Is it replaced by something else?

hopkiw:
> Is it replaced by something else?

yes, this is migration to OSS cluster. see GoogleCloudPlatform/oss-test-infra#363

PR #1222, opened 2020-05-22
Image import: Show detection results only when libguestfs's results can be verified.

EricEdens:
After sampling the results of the previous implementation, we found that our assumptions were incorrect in the following:

- The minor version of Ubuntu is modeled as a string, but it does not contain the leading zero for the April release. This caused us to generate gcloud examples such as "ubuntu-144" instead of "ubuntu-1404"
- Derivative distros (eg: Kali vs Debian) can appears with the following differences:
  - Kali appeared as 'debian-0'
  - Sangoma Linux 7 appeared as 'redhat-based-0'
  - Amazon Linux 2 appeared as 'unknown-0'

This PR changes the following:
 - Ensure Ubuntu has a four-digit version number in its gcloud representation
 - Disable distro suggestions unless we receive a known distro identifier, and a version number greater than zero.

A good place to start the review is cli_tools/gce_vm_image_import/importer/errors_test.go, as this shows the user-facing messages.

Lastly, this adds a new package called `verify`, inspired by:  https://www.martinfowler.com/ieeeSoftware/failFast.pdf

### DIFF
--- a/cli_tools/common/assert/doc.go
+++ b/cli_tools/common/assert/doc.go
@@ -12,25 +12,8 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-package verify
-
-import "fmt"
-
-// GreaterThanOrEqualTo verifies that value is greater than or equal to limit,
-// and panics if the verification fails.
-func GreaterThanOrEqualTo(value int, limit int) {
-	if value < limit {
-		panic(fmt.Sprintf("Expected %d >= %d", value, limit))
-	}
-}
-
-// Contains verifies that element is a member of arr, and panics if the verification fails.
-func Contains(element string, arr []string) {
-	for _, e := range arr {
-		if e == element {
-			return
-
-		}
-	}
-	panic(fmt.Sprintf("%s is not a member of %v", element, arr))
-}
+// Package assert provides convenience functions that help a method or constructor check whether
+// it was invoked correctly (whether its preconditions have been met).
+//
+// When the assertion fails, a panic occurs, allowing bugs to be found in a fail-fast manner.
+package assert
--- a/cli_tools/common/distro/distro.go
+++ b/cli_tools/common/distro/distro.go
@@ -17,7 +17,10 @@ package distro
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/verify"
 )
 
 const (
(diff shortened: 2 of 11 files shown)


cli_tools/common/verify/verify.go
```diff
@@ -0,0 +1,36 @@
...
+
+package verify
+
+import "fmt"
+
+// GreaterThanOrEqualTo verifies that value is greater than or equal to limit,
+// and panics if the verification fails.
+func GreaterThanOrEqualTo(value int, limit int) {
+	if value < limit {
+		panic(fmt.Sprintf("Expected %d >= %d", value, limit))
```
zoran15, line 23: panic is supposed to be used for runtime errors that a program can't recover from, errors it doesn't know how to handle or API errors (made by programmers): https://gobyexample.com/panic

> Mostly we use it to fail fast on errors that shouldn’t occur during normal operation, or that we aren’t prepared to handle gracefully.

I think for generic validation it's better to explicitely return an error. These functions might be used for validating user data in the future where panic is not appropriate. Right now, the use of it in new*Release() makes sense as it's guarding against programmer errors, but having it in a generic verify package opens up these funcs to be used inappropriately.

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
-// inspection routines.
+// FromComponents interprets the (distro, major, minor) tuple and returns
+// a Release if the arguments are syntactically correct, and represent a
+// release that we *may* support. The caller is responsible for verifying
+// whether a translator is available elsewhere in the system.
 //
-// http://libguestfs.org/guestfs.3.html#guestfs_inspect_get_distro
-func FromLibguestfs(distro string, major string, minor string) (r Release, e error) {
+// Prefer this method for generating a new Release object. It is lenient, and
+// returns nice errors. The distro-specific constructors are strict and may panic.
```
zoran15, line 57: What does it mean "nice errors"?

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
-// FromLibguestfs initializes a DistroAndVersion from the fields returned by libguestfs's
-// inspection routines.
+// FromComponents interprets the (distro, major, minor) tuple and returns
+// a Release if the arguments are syntactically correct, and represent a
+// release that we *may* support. The caller is responsible for verifying
+// whether a translator is available elsewhere in the system.
 //
-// http://libguestfs.org/guestfs.3.html#guestfs_inspect_get_distro
-func FromLibguestfs(distro string, major string, minor string) (r Release, e error) {
+// Prefer this method for generating a new Release object. It is lenient, and
```
zoran15, line 56: Prefer to what?

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
+		minorInt, e = strconv.Atoi(minor)
+		if e != nil || minorInt < 0 {
+			return nil, errors.New(
+				"minor version required to be an integer greater than or equal to zero. Received: " + minor)
+		}
+	}
+	switch distro {
+	case "":
+		return nil, errors.New("distro name required")
+	case windows:
```
zoran15, line 78: Is this error visible to users?

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
+	if distro == sles || strings.HasPrefix(distro, sles+"-") {
+		return newSLESRelease(distro, majorInt, minorInt)
 	}
-	return release{
-		distro: distro,
-		major:  major,
-		minor:  minor,
-	}, nil
+
+	return nil, fmt.Errorf("distro `%s` is not importable", distro)
```
zoran15, line 102: Is `distro` a common term used by customers or would `distribution` be more appropriate?

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
 	}
-	if strings.HasPrefix(os, windows) {
-		// We don't currently need Windows for this parsing, and punting
-		// since it's not immediately clear how to represent their
-		// version strings, as the user-facing value (Windows 2008r2)
-		// does not match what's used internally by our tools (NT 5.2)
-		// https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
-		return r, errors.New("Windows not yet implemented")
+	hyphen := strings.LastIndex(os, "-")
+	if hyphen < 0 || hyphen == len(os) {
```
zoran15, line 120: hyphen -> hyphenPosition/hyphenIndex or similar

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
+}
+
+// The caller is responsible for verifying the syntax of the arguments.
+// Verify the following before calling:
+//   - distro is one of {centos, debian, opensuse, rhel}
+//   - major is >= 1 and minor is >= 0
+func newCommonLinuxRelease(distro string, major, minor int) (Release, error) {
+	verify.GreaterThanOrEqualTo(major, 1)
+	verify.GreaterThanOrEqualTo(minor, 0)
+	verify.Contains(distro, []string{centos, debian, opensuse, rhel})
```
zoran15, line 164: Consider extracting []string{centos, debian, opensuse, rhel} to a constant as it would be reusable in the future

cli_tools/common/distro/distro.go
```diff
@@ -33,155 +36,216 @@ const (
...
+type slesRelease struct {
+	variant string
+	major   int
+	minor   int
+}
+
+// The caller is responsible for verifying the syntax of the arguments.
+// Verify the following before calling:
+//   - distroAndVariant starts with "sles" or "sles-"
+//   - major is >= 1 and minor is >= 0
```
zoran15, line 183: These two types of errors are handled in this func in different ways:
- major/minor via panic
- "sles" prefix via errors

cli_tools/common/distro/distro.go
EricEdens: It's not surfaced. (In general, I'm not a fan of exception messages being shown to users, *especially* when they're generated by common code such as this, since that makes it incredibly difficult to tweak error messages in the future.)

cli_tools/common/distro/distro.go
EricEdens: The panics are used when a precondition is violated (major/minor, and requiring a prefix of `sles-`, while an error only occurs when the variant name is not recognized. 

The distinction is the burden that we place on the caller; while we expect the caller to give us well-formed version numbers and version strings, we don't expect them to pre-verify the variant name.

cli_tools/common/distro/distro.go
EricEdens: Similar to the question above, this isn't shown to users at the moment, and it isn't written with that in mind.

cli_tools/common/distro/distro.go
EricEdens: Oh, I just meant a nice error in contrast to a panic.

cli_tools/common/distro/distro.go
EricEdens: Hmm, good point. Originally I intended it to warn against using the distro-specific constructors, but those aren't exported, so that point is moot. Maybe just remove these last two lines?

cli_tools/common/verify/verify.go
EricEdens: Yeah, like you mentioned, the `verify` package is intended to verify preconditions and assumptions, allowing  *programming* errors to be surfaced. It's not intended to be used with unsanitized user input. 

`These functions might be used for validating user data in the future`

Gosh, I hope not! My intention was to create a small, well-defined public API that allows the internal details (such as struct members) to evolve over time. If you have some specific usecases that are not covered by this public API, I'd be happy to brainstorm what they'd look like.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -229,7 +229,10 @@ func Run(clientID string, imageName string, dataDisk bool, osID string, customTr
 
 		daisyutils.PostProcessDErrorForNetworkFlag("image import", err, network, w)
 
-		return customizeErrorToDetectionResults(osID, w, err)
+		return w, customizeErrorToDetectionResults(osID,
```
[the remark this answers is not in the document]
EricEdens: That's a good idea. I recall looking into it, and hitting a bump differentiating whether the OS was provided by the user, or whether it came from the descriptor, at the point where the message would be created. So it ended up being a bigger task than the `image import` case.

cli_tools/common/distro/distro.go
EricEdens: ... and in retrospect, these two lines leak details about the package's implementation, so I went ahead and removed them.

cli_tools/common/distro/distro.go
EricEdens: removed

cli_tools/common/distro/distro.go
EricEdens: sgtm

cli_tools/common/distro/distro.go
EricEdens: sgtm. (I put them behind a method to avoid shared mutable state)

cli_tools/common/verify/verify.go
zoran15: If this package will only hold code to protect against programming errors, then `assert` would be a better name than `verify`. `verify` can apply to both user and programming errors, while `assert` is clear it's about programming errors.

cli_tools/gce_vm_image_import/importer/importer.go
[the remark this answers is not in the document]
zoran15: Do we have a tracking bug for that work?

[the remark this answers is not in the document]
EricEdens: Added b/157516599

cli_tools/common/verify/verify.go
EricEdens: sgtm (done)

cli_tools/common/assert/doc.go
```diff
@@ -0,0 +1,19 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package assert provides convenience functions that help a method or constructor check whether
+// it was invoked correctly (whether its preconditions have been met).
+//
+// When the assertion fails, a panic occurs, allowing bugs to be found in a fail-fast manner.
+package assert
```
dntczdx, line 19: What's the purpose of this file?

EricEdens: It's just for documentation: https://blog.golang.org/godoc

PR #1227, opened 2020-05-28
Split image import into two phases.

EricEdens:
This is the final chunk of refactoring to enable native VMDK import, UEFI tagging, and OS detection.


cli_tools/gce_vm_image_import/importer/finisher.go
```diff
@@ -0,0 +1,47 @@
...
+type finisherProvider interface {
+	provide(pd pd) (finisher, error)
+}
+
+type defaultFinisherProvider struct {
+	ImportArguments
+	imageClient imageClient
+}
+
+func (d defaultFinisherProvider) provide(pd pd) (finisher, error) {
```
zoran15, line 40: `pd pd`, even though allowed, is confusing. Consider using persistentDisk for the type (or similar)

cli_tools/common/utils/logging/service/literal_loggable.go
```diff
@@ -25,3 +25,17 @@ func (w literalLoggable) GetValue(key string) string { return w.strings[key] }
 func (w literalLoggable) GetValueAsInt64Slice(key string) []int64 { return w.int64s[key] }
 
 func (w literalLoggable) ReadSerialPortLogs() []string { return w.serials }
+
+// SingleImageImportLoggable returns a Loggable that is pre-initialized with the metadata
+// fields that are relevant when importing a single image file.
+func SingleImageImportLoggable(fileFormat string, sourceSize, resultSize int64, serials []string) Loggable {
+	loggable := literalLoggable{
```
zoran15, line 32: minor: can be in-lined in return

cli_tools/gce_vm_image_import/importer/finisher.go
```diff
@@ -0,0 +1,47 @@
...
+package importer
+
+import (
+	"context"
+)
+
+// finisher represents the second (and final) phase of import. For bootable
+// disks, this means translation and publishing the final image. For data
+// disks, this means publishing the image.
+type finisher interface {
```
zoran15, line 24: `finisher` sounds like it's just responsible for wrapping things up (which, in case of data disks, is true, kind of), but for bootable disks, there's more to in the `finishing` stage than in the actual disk import. 

Can you think of a more appropriate name to reflect the fact this process can take time and is very involved? `processor`, `transformer`, `adapter`?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
+)
+
+const (
+	inflateFilePath  = "import_disk.wf.json"
+	inflateImagePath = "inflate_image.wf.json"
+)
+
+type inflater interface {
```
zoran15, line 32: Docs are useful for types even if not exported.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+	inflate(ctx context.Context) (pd, error)
+	serials() []string
+}
+
+type daisyInflater struct {
+	wf         *daisy.Workflow
+	uri        string
```
zoran15, line 39: it's not clear to what resource this uri points to. I assume disk? Consider renaming

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+	"google.golang.org/api/compute/v1"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
+)
+
+const (
+	inflateFilePath  = "import_disk.wf.json"
+	inflateImagePath = "inflate_image.wf.json"
```
zoran15, line 29: Workflow file names are inconsistent with these constants. If we are to call import -> inflate (just the byte transfer), we should use it consistently.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+	serials() []string
+}
+
+type daisyInflater struct {
+	wf         *daisy.Workflow
+	uri        string
+	serialLogs []string
+}
+
+func (d daisyInflater) inflate(ctx context.Context) (pd, error) {
```
zoran15, line 43: `d` -> `di` as it can be confusing (disk, daisy etc.)

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type pd struct {
+	uri        string
+	sizeGb     int64
+	sourceGb   int64
+	sourceType string
+}
+
+func createDaisyInflater(t ImportArguments, workflowDirectory string) (inflater, error) {
```
zoran15, line 63: why `t` for ImportArguments instead of `args` or similar?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+		addFeatureToDisk(wf, "WINDOWS", inflationDiskIndex)
+	}
+
+	return daisyInflater{
+		wf:  wf,
+		uri: fmt.Sprintf("zones/%s/disks/%s", t.Zone, diskName),
+	}, nil
+}
+
+func (d daisyInflater) serials() []string {
```
zoran15, line 107: `serials` means `serialLogs`? Consider renaming

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+	return d.serialLogs
+}
+
+// If the inflated disk hold Windows, then it needs the WINDOWS GuestOSFeature
+// in order to boot during the subsequent translate step.
```
zoran15, line 112: Nothing about this func that restricts it to setting only `WINDOWS` guest OS feature. The doc should be more generic.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -0,0 +1,128 @@
...
+	disk := getDisk(workflow, diskIndex)
+	disk.GuestOsFeatures = append(disk.GuestOsFeatures, &compute.GuestOsFeature{
+		Type: feature,
+	})
+}
+
+func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
+	for _, step := range workflow.Steps {
+		if step.CreateDisks != nil {
+			return (*step.CreateDisks)[diskIndex]
```
zoran15, line 123: need to check for length here and throw panic if diskIndex will cause index out of bounds

cli_tools/gce_vm_image_import/importer/finisher.go
EricEdens: sgtm (changed the type to persistentDisk)

cli_tools/common/utils/logging/service/literal_loggable.go
EricEdens: cool, done

cli_tools/gce_vm_image_import/importer/finisher.go
EricEdens: Cool, changed to `processor`.

More info:
I struggled with naming, since there are two different paths to represent:

* dataDisk -> publish
* bootableDisk -> transform, publish

I think your idea of `processor` is more descriptive.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: good call; added comments here and to the other new types.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: Yeah, it's the disk. Fixed that.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: changed to inflate_disk.wf.json

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: done

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: It was a stale holdover from when the args were split into translateSpec, imageSpec, and daisySpec. Thanks for spotting. fixed.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: Updated to 'traceLogs' here and elsewhere, since serial logs are a specific instance of trace logs.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: you're right, thanks for spotting that. done.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: sgtm, done

cli_tools/gce_vm_image_import/importer/bootable_processor.go
```diff
@@ -0,0 +1,131 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+
+	daisy_utils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/daisy"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
+)
+
+type bootableProcessor struct {
```
dntczdx, line 30: you may have discussed this topic since I didn't read old commits - would "bootable_disk_process" be a bteer name since another one is named "data_disk_processor"?

EricEdens: sounds good to me! Changed bootableProcessor to bootableDiskProcessor (type names and file names)

PR #1233, opened 2020-05-29
break main.go for PR testing purposes

hopkiw:
testing the PR presubmits to validate that this change triggers golang presubmits and that they fail (presumably I've made a breaking change here)

### DIFF
--- a/cli_tools/gce_windows_upgrade/main.go
+++ b/cli_tools/gce_windows_upgrade/main.go
@@ -26,8 +26,8 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


hopkiw:
well, nothing got triggered. i see the webhook was sent and the size plugin was triggered, but not blunderbuss or any automatic running of presubmits...

hopkiw:
ok, webhook config was wrong. looking good now..

hopkiw:
ok presubmits correctly failed. going to set this one as on hold so it doesn't merge when i fix that
/hold

hopkiw:
OK auto-cc doesn't work because there's no OWNERS file. Correcting that in a real PR, closing this test PR.

PR #1234, opened 2020-05-29
add OWNERS file

hopkiw:
Seeded initial OWNERS file with everyone with a stake in this repo. This is our only remaining "mega repo" with multiple projects, we can use granular OWNERS files if that seems warranted. Starting with this for now to enable prow based review assignments and approvals.


hopkiw:
> Add Zach?
> 
> Do we need a copyright on this file?

i don't think so. none of the others in other repos do.

PR #1236, opened 2020-05-29
leave empty machine-id file for autogen

hopkiw:
debian 10 images don't have /etc/machine-id populated on boot. create an empty file to trigger its population at boot time.

after this change, the following lines are in the serial console:
```
[    5.101277] systemd[1]: Initializing machine ID from KVM UUID.
[    5.103367] systemd[1]: Installed transient /etc/machine-id file.
```

### DIFF
--- a/daisy_workflows/image_build/debian/fai_config/scripts/10-gce-clean
+++ b/daisy_workflows/image_build/debian/fai_config/scripts/10-gce-clean
@@ -23,6 +23,9 @@ rm -f $target/etc/resolv.conf \
       $target/var/log/dpkg.log \
       $target/var/log/install_packages.list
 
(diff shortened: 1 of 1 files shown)


hopkiw:
for the record, current debian 10 image has the following in serial console output:

```
[    2.449791] systemd[1]: System cannot boot: Missing /etc/machine-id and /etc is mounted read-only.
[    2.451092] systemd[1]: Booting up is supported only when:
[    2.452014] systemd[1]: 1) /etc/machine-id exists and is populated.
[    2.453054] systemd[1]: 2) /etc/machine-id exists and is empty.
[    2.454033] systemd[1]: 3) /etc/machine-id is missing and /etc is writable.
```

PR #1239, opened 2020-06-03
Added custom cancel reason to Daisy workflow. 

zoran15:
Added custom cancel reason to Daisy workflow. This will allow us to provide different error messages when workflows are canceled for different reasons. For example, timed out vs. canceled by users.

### DIFF
--- a/daisy/step.go
+++ b/daisy/step.go
@@ -307,8 +307,7 @@ func (s *Step) run(ctx context.Context) DError {
 	select {
 	case <-s.w.Cancel:
 		// return an error to indicate a canceled workflow is not 'success'
-		s.w.LogWorkflowInfo("Step %q (%s) is canceled.", s.name, st)
-		return Errf("Step %q (%s) is canceled.", s.name, st)
+		return s.w.onStepCancel(s, st)
 	default:
 		s.w.LogWorkflowInfo("Step %q (%s) successfully finished.", s.name, st)
 	}
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -179,6 +179,8 @@ type Workflow struct {
 	ForceCleanupOnError bool
 	// forceCleanup is set to true when resources should be forced clean, even when NoCleanup is set to true
 	forceCleanup bool
+	// CancelReason provides custom reason when workflow is canceled. It replaces "is canceled" in the default error message.
+	CancelReason string `json:"-"`
 }
 
 //DisableCloudLogging disables logging to Cloud Logging for this workflow.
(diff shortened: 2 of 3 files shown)


daisy/step.go
```diff
@@ -307,8 +307,7 @@ func (s *Step) run(ctx context.Context) DError {
 	select {
 	case <-s.w.Cancel:
 		// return an error to indicate a canceled workflow is not 'success'
-		s.w.LogWorkflowInfo("Step %q (%s) is canceled.", s.name, st)
-		return Errf("Step %q (%s) is canceled.", s.name, st)
+		return s.w.onStepCancel(s, st)
```
dntczdx, line 310: I remember you do this in order to distinguish timeout caused cancel. However, it looks like it didn't explicitly tell whether the cancel is triggered by timeout.

zoran15: That portion will be in wrappers. I have another PR for OVF importer as the first implementation of global timeout in our wrappers.

dntczdx: Would you introduce more about the global timeout? Is it because the whole wf is split?

zoran15: Currently, we don't have global timeout on the wrapper or daisy side. -timeout flag currently only sets DefaultTimeout for each step, but the total duration of whole workflows is not limited. I'm planning on adding global timeout on the wrapper side. It's not specifically related to wf split, we would need to do it even with monolithic workflows; but, with the split coming up, implementing global timeout on Daisy side has limited value.

EricEdens: From the OVF side, then, is it going to look something like this?
```
when timeout occurs:
  wf.CancelReason = "timed out"
  close(wf.Cancel)
```

zoran15: Yes. I thought of creating a Timeout channel but that would require changes in all steps which was not worth it IMO. If you have any other idea how to make it more elegant without major changes, let me now.

zoran15: This doesn't appear to work. E.g. on timeout, I do

`w. Cancel <- "timed-out"`

Then, in step.go:

`case cancelReason := <-s.w.Cancel`

cancelReason is empty, whether I do close(w.Cancel) or not after `w. Cancel <- "timed-out"`

dntczdx: I don't see anything bad by closing cancel channel - the caller proactively triggers "cancel" when they think it's timeout, which sounds correct. If we add a "timeout" channel, it feels like "timeout" will trigger something different from cancel, but not the case.

zoran15: After chatting offline with Eric, I've introduced CancelWithReason func as a single entry point for cancelling when a custom reason needs to be provided.

PR #1240, opened 2020-06-04
Refactor buffered writer

fiona-liu:
- Refactor the buffered writer from gce_export into common/utils


cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// ObjectHandleCreator is responsible for creating GCS Object Handle
+type ObjectHandleCreator struct {
+	ctx context.Context
+	sc  *storage.Client
+}
+
+// CreateObjectHandle creates GCS Object Handle
+func (ohc *ObjectHandleCreator) CreateObjectHandle(
```
dntczdx, line 32: ref to bucket-relatednaming style, we can call it "GetObject". "CreateObjectHandle" is a bit confusion for callers

cli_tools/gce_export/main.go
```diff
@@ -320,15 +100,39 @@ func writeGzipProgress(start time.Time, size int64, rp, wp *progress) {
 		if readTotal == totalSize {
 			return
 		}
-		fmt.Printf("GCEExport: Read %s of %s (%s/sec),", readTotal, totalSize, diskSpd)
-		fmt.Printf(" total written size: %s (%s/sec)\n", uploadTotal, upldSpd)
+		log.Println("Read %s of %s (%s/sec),", readTotal, totalSize, diskSpd)
+		log.Println(" Total written size: %s (%s/sec)\n", uploadTotal, upldSpd)
```
dntczdx, line 104: it's continuing the last line, so should be lower case

cli_tools/gce_export/main.go
```diff
@@ -320,15 +100,39 @@ func writeGzipProgress(start time.Time, size int64, rp, wp *progress) {
...
 		oldUpload = wpTotal
 		oldRead = rpTotal
 		oldSince = since
 		time.Sleep(30 * time.Second)
 	}
 }
 
+func gcsClient(ctx context.Context, oauth string) (domain.StorageClientInterface, error) {
+	//return storage.NewClient(ctx)
```
dntczdx, line 113: remove old code

cli_tools/mocks/mock_storage_client.go
```diff
@@ -1,17 +1,3 @@
-//  Copyright 2019 Google Inc. All Rights Reserved.
```
zoran15, line 1: generating mocks removes copyright notice by default so you have to add it manually back. Add it also for the new mocks you generated.

cli_tools/mocks/mock_storage_client.go
```diff
@@ -1,4 +1,4 @@
...
+//  Copyright 2020 Google Inc. All Rights Reserved.
```
dntczdx, line 1: this is a existing file so let's keep 2019

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+func (oh *ObjectHandle) RunComposer(srcs ...domain.StorageObjectInterface) (*storage.ObjectAttrs, error) {
+	var objs []*storage.ObjectHandle
+	for _, obj := range srcs {
+		objs = append(objs, obj.GetObjectHandle())
+	}
+	return oh.ob.ComposerFrom(objs...).Run(oh.ctx)
+}
+
+// RunCopier performs the copy operation.
+func (oh *ObjectHandle) RunCopier(src domain.StorageObjectInterface) (*storage.ObjectAttrs, error) {
```
zoran15, line 79: It's not clear from where and to where this func copies and what. One needs to look at the code to understand what this func is doing. Docs should be extended to make this clearer.

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+	return oh.ob.NewWriter(oh.ctx)
+}
+
+// ObjectName returns the name of the object.
+func (oh *ObjectHandle) ObjectName() string {
+	return oh.ob.ObjectName()
+}
+
+// RunComposer performs the compose operation.
+func (oh *ObjectHandle) RunComposer(srcs ...domain.StorageObjectInterface) (*storage.ObjectAttrs, error) {
```
zoran15, line 70: It's unclear what this func does. What is composed here?

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+
+// GetObject gets GCS object interface.
+func (ohc *ObjectHandleCreator) GetObject(
+	bucket string, object string) domain.StorageObjectInterface {
+	return &ObjectHandle{
+		ob: ohc.sc.Bucket(bucket).Object(object), ctx: ohc.ctx}
+}
+
+// ObjectHandle is a wrapper around storage.ObjectHandle. Implements StorageObjectInterface.
+type ObjectHandle struct {
```
zoran15, line 39: Why not name this struct `StorageObject`? 

Several issues having it named `ObjectHandle` 
- it overlaps with storage.ObjectHandle so it's not clear on first look which one is being referred to.
- ObjectHandle sounds like it's a simple reference without much logic, while StorageObjectInterface provides a lot of functionality on top of the handle.
- having confusing calls such as `objectHandle.GetObjectHandle()`

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+	ob  *storage.ObjectHandle
```
zoran15, line 40: ob -> oh

cli_tools/common/utils/storage/storage_client.go
```diff
@@ -41,29 +41,25 @@ var (
 // Client implements domain.StorageClientInterface. It implements main Storage functions
 // used by image import features.
 type Client struct {
-	ObjectDeleter domain.StorageObjectDeleterInterface
 	StorageClient *storage.Client
 	Logger        logging.LoggerInterface
 	Ctx           context.Context
 	Oic           domain.ObjectIteratorCreatorInterface
+	Ohc           domain.StorageObjectCreatorInterface
```
zoran15, line 48: soc? storageObjectCreator? Longer names are more descriptive IMO, but even the short ones should somehow related to the underlying class, which this one doesn't.

This comment might be redundant depending on whether this interface is needed or not.

cli_tools/common/utils/storage/storage_client_test.go
```diff
@@ -39,12 +39,17 @@ func TestDeleteGcsPath(t *testing.T) {
 	mockObjectIteratorCreator := mocks.NewMockObjectIteratorCreatorInterface(mockCtrl)
 	mockObjectIteratorCreator.EXPECT().CreateObjectIterator("sourcebucket", "sourcepath/furtherpath").Return(mockObjectIterator)
 
-	mockStorageObjectDeleter := mocks.NewMockStorageObjectDeleterInterface(mockCtrl)
-	firstDeletion := mockStorageObjectDeleter.EXPECT().DeleteObject("sourcebucket", "sourcepath/furtherpath/afile1.txt")
-	secondDeletion := mockStorageObjectDeleter.EXPECT().DeleteObject("sourcebucket", "sourcepath/furtherpath/afile2.txt")
-	gomock.InOrder(firstDeletion, secondDeletion)
-
-	sc := Client{Oic: mockObjectIteratorCreator, ObjectDeleter: mockStorageObjectDeleter,
+	mockObjecthandle := mocks.NewMockStorageObjectInterface(mockCtrl)
```
zoran15, line 42: mockObjecthandle -> mockStorageObject. In other places as well

cli_tools/gce_vm_image_export/exporter/exporter.go
```diff
@@ -162,7 +163,7 @@ func Run(clientID string, destinationURI string, sourceImage string, format stri
 	ctx := context.Background()
 	metadataGCE := &compute.MetadataGCE{}
 	storageClient, err := storage.NewStorageClient(
-		ctx, logging.NewStdoutLogger(logPrefix), oauth)
+		ctx, logging.NewStdoutLogger(logPrefix), option.WithCredentialsFile(oauth))
```
zoran15, line 166: Was this a bug earlier?

cli_tools/common/domain/interfaces.go
```diff
@@ -66,9 +65,20 @@ type TarGcsExtractorInterface interface {
 	ExtractTarToGcs(tarGcsPath string, destinationGcsPath string) error
 }
 
-// StorageObjectDeleterInterface represents an object that is responsible for deleting GCS objects
-type StorageObjectDeleterInterface interface {
-	DeleteObject(bucket string, objectPath string) error
+// StorageObjectCreatorInterface represents GCS object creator
+type StorageObjectCreatorInterface interface {
+	GetObject(bucket string, objectPath string) StorageObjectInterface
+}
```
zoran15, line 71: What is the need for this interface? Can it be integrated into StorageClient (so that StorageClient.GetObject does the reference creation)?

Also, this only creates a reference to a GCS object, doesn't actually create anything in GCS so the naming doesn't seem appropriate.

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"context"
+	"io"
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/domain"
+)
+
+// ObjectHandleCreator is responsible for creating GCS Object Handle.
+type ObjectHandleCreator struct {
```
zoran15, line 26: It's actually returning `StorageObjectInterface` and is implementing `StorageObjectCreatorInterface ` so having this named `ObjectHandleCreator` is confusing. `StorageObjectCreator` or similar is more appropriate.

I have another comment on whether this interface and implementation are needed at all, so this comment might be deprecated depending on the resolution of that other comment.

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+	ob  *storage.ObjectHandle
+	ctx context.Context
+}
+
+// GetObjectHandle gets the storage object
+func (oh *ObjectHandle) GetObjectHandle() *storage.ObjectHandle {
+	return oh.ob
+}
+
+// Delete deletes GCS object in given bucket and object path.
```
zoran15, line 49:
```suggestion
// Delete deletes GCS object.
```

cli_tools/common/utils/storage/object_handle.go
```diff
@@ -0,0 +1,81 @@
...
+		ob: ohc.sc.Bucket(bucket).Object(object), ctx: ohc.ctx}
+}
+
+// ObjectHandle is a wrapper around storage.ObjectHandle. Implements StorageObjectInterface.
+type ObjectHandle struct {
+	ob  *storage.ObjectHandle
+	ctx context.Context
+}
+
+// GetObjectHandle gets the storage object
```
zoran15, line 44:
```suggestion
// GetObjectHandle gets the storage object handle
```

cli_tools/common/utils/storage/object_handle.go
fiona-liu: That makes sense, I will change the name to StorageObject.

cli_tools/gce_vm_image_export/exporter/exporter.go
fiona-liu: The buffered writer has other options to pass in so I changed the param to options instead of passing in the oauth string.

cli_tools/common/domain/interfaces.go
fiona-liu: I think this is used so that we can mock the object in storage client tests.

fiona-liu: I think the interface is needed to that we can mock this in storage client tests.

cli_tools/common/utils/storage/object_handle.go
fiona-liu: this function composes the source objects into one object.

cli_tools/common/utils/storage/object_handle.go
fiona-liu: will add more comments, thanks!

cli_tools/common/domain/interfaces.go
dntczdx: StorageObjectInterface is better

cli_tools/common/utils/storage/buffered_writer.go
```diff
@@ -0,0 +1,236 @@
...
+
+	sync.Mutex
+	sync.WaitGroup
+	bytes int64
+	part  int
+	file  *os.File
+}
+
+// NewBuffer creates a BufferedWriter
+func NewBuffer(ctx context.Context, size, workers int64, client gcsClient, oauth, prefix, bkt, obj string) *BufferedWriter {
```
EricEdens, line 63: When I see 'NewBuffer', I think of this from the stdlib: https://golang.org/pkg/bytes/#example_Buffer ; is that your intent? Any reason to not use 'NewBufferedWriter'?

EricEdens, line 63: Do clients need a whole BufferedWriter object? Any thoughts on returning a WriterCloser instead?

https://golang.org/pkg/io/#WriteCloser

cli_tools/common/domain/interfaces.go
EricEdens: > What is the need for this interface? 

It's typically an anti-pattern to expose your implementation details (unless there's an overwhelmeing reason for it). Some good reading: https://golang.org/doc/effective_go.html#generality

cli_tools/common/domain/interfaces.go
```diff
@@ -66,9 +65,20 @@ type TarGcsExtractorInterface interface {
...
+// StorageObjectInterface represents GCS Object
+type StorageObjectInterface interface {
+	Delete() error
+	GetObjectHandle() *storage.ObjectHandle
+	NewReader() (io.ReadCloser, error)
+	NewWriter() io.WriteCloser
+	ObjectName() string
+	RunComposer(src ...StorageObjectInterface) (*storage.ObjectAttrs, error)
+	RunCopier(src StorageObjectInterface) (*storage.ObjectAttrs, error)
 }
```
EricEdens: Any thoughts on making this smaller? For example, an interface for each method?

StorageObjectDeleter
StorageObjectGetter
StorageObjectReader
...

The *huge* benefit is that it clarifies  which dependencies are actually used in a particular context. Also, it avoids the necessity of a big mocking framework, since a particular test can implement a single method on a struct. (I'm not saying that mocking frameworks are bad, but having atomic interfaces gives flexibility)

fiona-liu: The StorageObjectCreatorInterface is so that we're able to pass in mock objects for testing. I was following an existing pattern for this, but if there is another way to test without having the interface, I'd be happy to remove it. 

I think having the a separate interface for each method is not very user-friendly to the caller. The caller will have to pass in the bucket and object path for each of the call. If caller wants to perform multiple operations on the same object, it can feel redundant.

cli_tools/common/utils/storage/buffered_writer.go
fiona-liu: I reused the method name when I refactored. I think NewBufferedWriter makes sense, I will change it. Thanks :)

cli_tools/common/domain/interfaces.go
```diff
@@ -66,9 +65,20 @@ type TarGcsExtractorInterface interface {
...
-// StorageObjectDeleterInterface represents an object that is responsible for deleting GCS objects
-type StorageObjectDeleterInterface interface {
-	DeleteObject(bucket string, objectPath string) error
+// StorageObjectInterface represents GCS object interface
+type StorageObjectInterface interface {
+	GetObject(bucket string, objectPath string) StorageObject
+}
+
+// StorageObject represents GCS Object
+type StorageObject interface {
```
zoran15, line 74: Having `StorageObjectInterface` and `StorageObject` interfaces is confusing. `StorageObjectInterface` should be named something like `StorageObjectCreatorInterface` or `StorageObjectGetterInterface`

Additionally, it seems `StorageObjectInterface` is there just for testing so you're able to mock retrieving StorageObjects? If so, you can make it non-exported (lower case).

cli_tools/common/utils/storage/storage_object.go
```diff
@@ -0,0 +1,82 @@
...
+func (dst *StorageObject) Compose(srcs ...domain.StorageObjectInterface) (*storage.ObjectAttrs, error) {
+	var objs []*storage.ObjectHandle
+	for _, obj := range srcs {
+		objs = append(objs, obj.GetObjectHandle())
+	}
+	return dst.oh.ComposerFrom(objs...).Run(dst.ctx)
+}
+
+// Copy copies the src object into the dst object.
+func (dst *StorageObject) Copy(src domain.StorageObjectInterface) (*storage.ObjectAttrs, error) {
```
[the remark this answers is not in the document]
fiona-liu: sure, I will rename to CopyFrom

cli_tools/common/domain/interfaces.go
fiona-liu: I change the name change. Keeping the interface exported as its used by another package (as discussed)

cli_tools/common/utils/storage/storage_object.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// ObjectName returns the name of the object.
+func (so *StorageObject) ObjectName() string {
+	return so.oh.ObjectName()
+}
+
+// Compose takes in srcs as source objects and compose them into the destination object (dst).
+// Up to 32 objects can be composed into a one object.
+func (dst *StorageObject) Compose(srcs ...domain.StorageObjectInterface) (*storage.ObjectAttrs, error) {
```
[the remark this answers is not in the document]
fiona-liu: I added more doc. Let me know what you think. Thanks!

cli_tools/common/utils/storage/storage_object.go
```diff
@@ -0,0 +1,90 @@
...
+/*
+Compose takes in srcs as source objects and concatenates them into the destination storage object.
+Returns the object attributes of the destination object after composed.
+Up to 32 objects can be composed into a one object.
+This function is used during uploading to GCS. To increase upload speed of a file,
+the file is split into multiple parts and concurrently uploaded to storage objects.
+At the end of the upload, the function is called to compose the objects back into one
+storage object.
+*/
+
```
zoran15, line 78: remove extra blank line

cli_tools/common/utils/storage/storage_object.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/*
+Compose takes in srcs as source objects and concatenates them into the destination storage object.
+Returns the object attributes of the destination object after composed.
+Up to 32 objects can be composed into a one object.
+This function is used during uploading to GCS. To increase upload speed of a file,
+the file is split into multiple parts and concurrently uploaded to storage objects.
+At the end of the upload, the function is called to compose the objects back into one
+storage object.
```
zoran15, line 76:
```suggestion
Compose takes in srcs as source objects and concatenates them into the destination storage object.
Returns the object attributes of the destination object after composed.
Up to 32 objects can be composed into a one object.
This function can be used to make uploads to GCS faster. To increase upload speed of a file,
the file can be split into multiple parts and concurrently uploaded to storage objects.
At the end of the upload, this function should be called to compose the objects back into one
storage object.
```

We should not state how clients are using a functions as that can change. Stating it as an example is ok.

PR #1242, opened 2020-06-05
Added global timeout to OVF import

zoran15:


cli_tools/gce_ovf_import/ovf_importer/ovf_importer_test.go
```diff
@@ -741,6 +742,50 @@ func TestPopulateMissingParametersInvalidZone(t *testing.T) {
...
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	mockLogger := mocks.NewMockLoggerInterface(mockCtrl)
+	mockLogger.EXPECT().Log("Timeout 0s exceeded, stopping workflow \"wf\"")
+
+	params := getAllInstanceImportParams()
+	params.Timeout = "0s"
+
+	oi := OVFImporter{Logger: mockLogger, params: params}
```
dntczdx, line 755: The oi.Logger.Log doesn't impact the test - it has no return value, neither modifying anything. You may directly set a fake/empty logger. A mock logger is somehow overkill. But it's up to you.

zoran15: The logger output is used to validate the test using EXPECT()

dntczdx: OK I thought you were validating w.CancelReason. It's OK to validate log as plus, on the other hand, isn't w.CnacelReason more important to validate?

zoran15: That is tested in workflow_test.go in the previous (Daisy related) PR. 

What would make sense to validate here is that w.CancelWithReason() actually gets called. This is partially done in ovf_importer_test.go by making sure w.Cancel channel is closed.

cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -486,6 +490,17 @@ func (oi *OVFImporter) Import() (*daisy.Workflow, error) {
 	return w, nil
 }
 
+func (oi *OVFImporter) handleTimeout(w *daisy.Workflow) {
+	timeout, err := time.ParseDuration(oi.params.Timeout)
+	if err != nil {
+		oi.Logger.Log(fmt.Sprintf("Error parsing timeout `%v`", oi.params.Timeout))
+		return
+	}
```
EricEdens, line 498: It appears that the workflow may run indefinitely given a malformed timeout value. I can imagine a couple of other options; what was the motivation for picking this one?

Options:
1. Proceed without a timeout
2. Fail immediately and ask the user to fix the parameter
3. Fallback on a reasonable value

zoran15: Workflows will not run indefinitely as there are Daisy step workflows. The global timeout is in addition to Daisy step workflows. 

The alternative was to calculate step workflows in some way to achieve global timeout. Due to non-deterministic way of calculating these and the maintenance burden if workflows change, I've decided to introduce one global timeout.

EricEdens: makes sense

PR #1243, opened 2020-06-09
Daisy: Consolidate reading of serial port output.

### LINKED ISSUE #1160: A race between two serial logging pumps
Sorry if this has been logged, but I could not find a similar issue, and it has bothered be for a while.

This is a loop pumping serial port log to the log bucket. This is always on, and all fine (except it loses the tail of the log if the machine shuts town too quickly, but that's a GCE design defect, a whole another story--ideally, GCE should have kept the log open for another 15-30 second after the instance stops). I do not remember off the top of my head where exactly, but I seen in some of the workflows in this repo that a machine is even restarted with the only goal to extract its logs. I am talking about this loop:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/c776c023bd4e2908d5194634e7740f66119ec011/daisy/logger.go#L261-L268

The real race inside Daisy happens between this pump, and the one in `WaitForInstancesSignal` (WFIS). The latter may react so quickly and issue the .stop to instance that a solid trailing chunk of logs disappears, because the first pump nods it off. And you'd probably be much less than surprised that I'm going after these logs only when build fails, and the tail, the part that was probably the most crucial for diagnostics, would be missing. I'm talking about this loop:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/c776c023bd4e2908d5194634e7740f66119ec011/daisy/step_wait_for_instances_signal.go#L122-L175

I understand is not an easy problem to solve, and I have a solid chunk of boilerplate in every build script to work around these issues with a various degree of success: Two sleeps, essentially: one before logging the magic incantations for the WFIS step to match, and another after it but before a poweroff to give Daisy a break to suck out the remainder of the log--there is a surge of logging noise when the instance is stopping. Getting the timeout right to avoid false negatives is closer to magic than engineering, tho :)

 If it was only somehow possible to keep a single pump, so that what would have gone to the the bucket and/or stackdriver were exactly the same chunk of log that WFIS had seen and acted upon, that would have solved a half of the issue already. In other words, If WFIS were to somehow tap into the existing log stream pumping service inside Daisy instead of rolling the second one on its own. Do you think something like this is doable?

EricEdens:
Daisy reads serial port output (serial output) in two places: once in createInstances for log archival, and once in waitForInstancesSignal for control flow. These readers are independent: they both make GCP API calls, manage their own errors, and perform their own retries. 
Uncoordinated access causes two errors:
1. Failure to archive logs: Since each reader has its own event loop, there’s a race for createInstances to archive logs prior to waitForInstancesSignal killing the instance (1160). 
2. Subtle bugs: Over time, each set of reader code has been improved to handle transient failures and instance reboots. Since the code has evolved separately, we find subtle, difficult-to-debug errors.

This change introduces a pubsub mechanism to allow a single serial port reader to broadcast updates to multiple consumers.

Fixes #1160

### DIFF
--- a/daisy/serial_output_watcher.go
+++ b/daisy/serial_output_watcher.go
@@ -49,7 +49,7 @@ type SerialOutputWatcher interface {
 // compute API to make periodic polling requests to check for new serial output.
 func NewSerialOutputWatcher(client serialOutputClient, project, zone string) SerialOutputWatcher {
 	return &serialOutputWatcher{
-		subscriptions: map[instanceAndPort][]watch{},
+		subscriptions: map[serialPort][]watch{},
 		client:        client,
 		project:       project,
 		zone:          zone,
@@ -57,13 +57,13 @@ func NewSerialOutputWatcher(client serialOutputClient, project, zone string) Ser
 }
 
 type serialOutputWatcher struct {
-	subscriptions map[instanceAndPort][]watch
+	subscriptions map[serialPort][]watch
 	client        serialOutputClient
 	project, zone string
 }
 
 // serialOutputWatcher assumes that 404s are returned when an instance doesn't
-// exist. The value `deletionNum404s` determines how many subsequent HTTP 404 responses to
+// exist. The value `deletionRetry404s` determines how many subsequent HTTP 404 responses to
 // see prior to deciding the instance has been deleted (assuming the
 // instance was seen earlier).
 //
@@ -75,9 +75,9 @@ type serialOutputWatcher struct {
 //  - Once deletion finished: 404
 //  - Instance stopped, but not deleted: 400
 //  - Serial output successfully returned: 200
-const deletionNum404s = 3
+const deletionRetry404s = 3
 
-type instanceAndPort struct {
+type serialPort struct {
 	instance   string
 	portNumber int64
 }
@@ -89,7 +89,7 @@ type watch struct {
 
 func (s *serialOutputWatcher) Watch(
 	instanceName string, port int64, c chan<- string, interval time.Duration) {
-	sp := instanceAndPort{instanceName, port}
+	sp := serialPort{instanceName, port}
 	s.subscriptions[sp] = append(s.subscriptions[sp], watch{c, interval})
 }
 
@@ -114,23 +114,25 @@ func (s *serialOutputWatcher) start(instanceName string) {
 	}
 }
 
-// run starts a goroutine that polls for an instance's serial output. Output is written to the
+// watchOnePort starts a goroutine that polls for an instance's serial output. Output is written to the
 // subscribing channels. Channels are closed when the instance is deleted.
 func watchOnePort(client serialOutputClient, pollingFrequency time.Duration,
 	project, zone, instanceName string, port int64, subscribers []chan<- string) {
 	go func() {
 		seenInstance := false
 		consecutive404s := 0
-		offset := int64(0)
+		// The GCE serial port API implements paging using a cursor of byte offsets:
+		// https://cloud.google.com/compute/docs/reference/rest/v1/instances/getSerialPortOutput
+		readPosition := int64(0)
 		ticker := time.NewTicker(pollingFrequency)
 		for {
-			resp, err := client.GetSerialPortOutput(project, zone, instanceName, port, offset)
+			resp, err := client.GetSerialPortOutput(project, zone, instanceName, port, readPosition)
 			var httpCode int
 			if err != nil {
 				realErr := err.(*googleapi.Error)
 				httpCode = realErr.Code
 			} else {
-				offset = resp.Next
+				readPosition = resp.Next
 				httpCode = resp.HTTPStatusCode
 				seenInstance = true
 				if len(resp.Contents) > 0 {
@@ -144,7 +146,7 @@ func watchOnePort(client serialOutputClient, pollingFrequency time.Duration,
 			}
 			if seenInstance && httpCode == http.StatusNotFound {
 				consecutive404s++
-				if consecutive404s >= deletionNum404s {
+				if consecutive404s >= deletionRetry404s {
 					for _, c := range subscribers {
 						close(c)
 					}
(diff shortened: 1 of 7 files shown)


daisy/serial_output_watcher.go
```diff
@@ -0,0 +1,165 @@
...
+		callbacks := s.subscriptions[serialPort]
+		if len(callbacks) == 0 {
+			panic(fmt.Sprintf("Start called without subscribers for %v", serialPort))
+		}
+		var subscribers []chan<- string
+		pollingFrequency := callbacks[0].pollingInterval
+		for _, cb := range callbacks {
+			subscribers = append(subscribers, cb.channel)
+			if cb.pollingInterval < pollingFrequency {
+				cb.pollingInterval = pollingFrequency
```
zoran15, line 110: Related to the above comment : what if a watcher is registered after start? Should the polling interval be updated in `Watch()` as well?

daisy/serial_output_watcher.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+		}
+		watchOnePort(s.client, pollingFrequency, s.project, s.zone, serialPort.instance, serialPort.portNumber, subscribers)
+	}
+}
+
+// run starts a goroutine that polls for an instance's serial output. Output is written to the
```
zoran15, line 117:
```suggestion
// watchOnePort starts a goroutine that polls for an instance's serial output. Output is written to the
```

daisy/serial_output_watcher.go
```diff
@@ -0,0 +1,165 @@
...
+// subscribing channels. Channels are closed when the instance is deleted.
+func watchOnePort(client serialOutputClient, pollingFrequency time.Duration,
+	project, zone, instanceName string, port int64, subscribers []chan<- string) {
+	go func() {
+		seenInstance := false
+		consecutive404s := 0
+		offset := int64(0)
```
zoran15, line 124: What does this offset represents?

daisy/serial_output_watcher.go
```diff
@@ -0,0 +1,165 @@
...
+				if len(resp.Contents) > 0 {
+					for _, c := range subscribers {
+						select {
+						case c <- resp.Contents:
+						default:
+						}
+					}
+				}
+			}
+			if seenInstance && httpCode == http.StatusNotFound {
```
zoran15, line 145: Does this mean 404 is ignored until instance is created to avoid bailing on reading serial port output before instance is finished being created?

daisy/serial_output_watcher_test.go
```diff
@@ -0,0 +1,259 @@
...
+	watcher.Watch(instance, 2, c2, time.Nanosecond)
+	watcher.start(instance)
+
+	assert.Equal(t, "1_one", <-c1)
+	assert.Equal(t, "2_one", <-c2)
+	assert.Equal(t, "1_two", <-c1)
+	assert.Equal(t, "2_two", <-c2)
+}
+
+func TestSerialPortWatcher_StopsWhenInstanceDeleted(t *testing.T) {
```
zoran15, line 114: Does this test ignoring more than 3 404s before instance is created?

daisy/step_create_instances.go
```diff
@@ -51,73 +57,46 @@ func (ci *CreateInstances) UnmarshalJSON(b []byte) error {
...
-func logSerialOutput(ctx context.Context, s *Step, ii InstanceInterface, ib *InstanceBase, port int64, interval time.Duration) {
+func logSerialOutput(s *Step, name string, watcher *SerialOutputWatcher, wcProvider func() io.WriteCloser) {
 	w := s.w
 	w.stepWait.Add(1)
+	// The buffered channel ensures that we don't lose serial output if we have delays uploading
+	// the previous chunk to GCS. Assuming a poll interval of 3 seconds, the buffer of size 1024
+	// allows ~51 minutes of backlog before we start losing logs.
+	c := make(chan string, 1024)
+	(*watcher).Watch(name, serialPortToArchive, c, serialPortPollInterval)
+	(*watcher).start(name)
```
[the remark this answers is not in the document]
EricEdens: I'm glad you brought this up, as the current approach feels fragile to me as well. Any ideas on where that central place might be?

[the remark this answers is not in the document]
zoran15: Since serial port output can start as soon as an instance is created, it makes sense to have `start()` in step_create_instance.go. I'm not sure we can have `start()` somewhere else and ensure no serial output is lost.

Because of this, in the future, we might have subscribers call `Watch()` after `start()`. If we do allow this (currently it panics), we can handle it in a couple of ways:
* buffer output on the watcher side (up to a point to allow some later subcribers, but not have to store whole output for the duration of the execution), or
* allow subcribers to know if output has already started for given port/instance name before subcribing (and if any of it will be missed) and let them decide if they can live with this or not (by throwing panic on their own). If there was no output before late subcriber calls `Watch()` then nothing is lost and things should work as expected.

There are probably other possible solutions.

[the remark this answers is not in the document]
EricEdens: I've designed this system to specifically disallow late subscribers for two reasons:
1. We don't need it now.
2. We don't have an upcoming usecase for it.

Am I missing something here? Is there a use case that you're aware of?

Regarding how `Watch` and `start` are wired into the current steps I've addressed potential fragility in three ways:
1. I've documented the behavior: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1243/files#diff-364879f77dafcc960afbb3cc4ded0dabR35
2. If someone inadvertently performs a late subscription, they won't receive updates, so it will fail loudly.
3. If someone changes `step_wait_for_instances_signal.go` and moves `Watch` to an incorrect location, then tests will fail (unit and integration)

Regarding your two potential solutions (buffer or indicate that polling has already started), I think those are fantastic options *if* we need to support late subscribers. Do you have a use case in mind, though?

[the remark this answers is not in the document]
dntczdx: In my opinion "Watch" and "start" should be bounded together. No matter whether "start" has been done, it should be called again blindly without side effect.
It's also a reason why we should allow "late subscribe": it may have been started.

It's not a good idea to always rely on create_instance to start the watcher. The reason is that we definitely should be able to wait for a signal from a instance which is not created in the current daisy workflow.

PR #1244, opened 2020-06-09
Sanitize control characters from image filename.

EricEdens:
gsutil supports control characters, such as wildcards, that can't be escaped [1]. This change does the following:
1. Ignore the original filename (on the worker's disk, it will appear as `source_disk_file`)
2. Make OVA checking more robust:
  * Use magic numbers to check for a tar archive
  * Check contents for an ovf and vmdk file

Testing:
* Ran one of the OVA tests that was failing prior to the fixes described in (2).

1. https://cloud.google.com/storage/docs/gsutil/addlhelp/WildcardNames

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -83,12 +83,14 @@ function resizeDisk() {
   exit
 }
 
-# Determines whether a file is an OVA archive.
+# Determines whether a file is an OVA archive; the check isn't
+# comprehensive to *all* OVAs, but focuses on the OVAs supported
+# by this script (OVAs that use VMDKs).
 # Arguments:
 #   File to check.
 # Returns:
 #   0 if the file is a tar file, non-zero otherwise.
-function fileIsOva() {
+function fileIsOvaWithVmdk() {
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -84,6 +83,21 @@ function resizeDisk() {
...
 
+# Determines whether a file is an OVA archive.
+# Arguments:
+#   File to check.
+# Returns:
+#   0 if the file is a tar file, non-zero otherwise.
+function fileIsOva() {
+  if [[ $(file --mime-type "$1") =~ application/x-tar ]]; then
+    contents=$(tar --list -f "$1")
+    if [[ "$contents" =~ ovf && "$contents" =~ vmdk ]]; then
```
zoran15, line 94: OVAs can have disks in any format, not only vmdk

zoran15: That piece of legacy code was built on an incorrect assumption that OVAs contain disks only in VMDK format. We should fix this eventually. Can you please create a tracking bug for this fix?

PR #1252, opened 2020-06-15
Add script to export image and package metadata during linux builds

gaohannk:
- Add a "destination_prefix" as the destination gcs path to store "metadata.json"
- This field should be pass from top level workflow.
- The path would be looks like: gs://gce-image-archive/centos/centos-8-v20200316/metadata.json


daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+import time
+import uuid
+
+import utils
+
+
+def main():
+  # Get Parameters.
+  repo = utils.GetMetadataAttribute('google_cloud_repo', raise_on_not_found=True).strip()
```
hopkiw: why is .strip() here, but not on the others?

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,102 @@
...
+  else:
+    utils.Execute(['mount', '/dev/sdb1', '/mnt'])
+
+  logging.info('Installer root: %s' % os.listdir('/mnt'))
+
+  # Create and upload metadata of the image and packages
+  logging.info('Creating image metadata.')
+  release_time = time.time()
+  image = {
+      "id": uuid.uuid4(),
```
hopkiw, line 59: what's this uuid for?

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,102 @@
...
+      "family": el_release_name,
+      "name": el_release_name + build_date,
+      "version": build_date,
+      "location": "gs://gce-image-archive/centos-uefi/centos-7-v${build_date}",
+      "release_date": build_date,
+      "release_time": release_time,
+      "state": "Active",
```
hopkiw, line 66: is this meant to be active/deprecated? i think that you'll have to pull from GCE API, rather than here

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,102 @@
...
+      "environment": "prod",
```
hopkiw, line 67: similarly unsure what environment is for, esp if always hardcoded to prod

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+      "packages": []
+  }
+
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    # The URL is a github repo link which don't contain commit hash
+    cmd = "rpm -q --queryformat='%{NAME} %{VERSION} %{RELEASE} %{URL}\n'" \
```
hopkiw, line 77: should be `%{Vcs}` instead of `%{URL}`

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+          "name": package_name,
+          "version": package_version,
+          "commit_hash": "",
+          "release_date": package_release_time,
+          "stage": repo
+      }
+      image["packages"].append(metadata)
+
+      with open('/tmp/metadata.json', 'w') as f:
+        f.write(json.dumps(image))
```
hopkiw, line 96: uh, this will overwrite each time because mode `w` is overwriting. if you switch this to `a`, you'll append, but then you'll be appending full json blobs, which produces an invalid json blob.

i think you need to build up a python dict with all values and then write once at the end.

daisy_workflows/image_build/enterprise_linux/enterprise_linux_uefi.wf.json
```diff
@@ -145,6 +151,39 @@
...
+      "CreateInstances": [
+        {
+          "Name": "inst-exportmetadata",
+          "Disks": [{"Source": "disk-installerprep"}, {"Source": "disk-installer"}],
+          "MachineType": "n1-standard-4",
+          "Metadata": {
+            "files_gcs_dir": "${SOURCESPATH}/build_files",
+            "prefix": "ExportMetadata",
+            "script": "export_metadata.py",
+            "build_date": "${build_date}",
```
hopkiw, line 164: i don't think build_date is passed in to this workflow. i don't see it in the vars section. i think it's only available in the workflow which includes this one.

daisy_workflows/image_build/enterprise_linux/enterprise_linux_uefi.wf.json
```diff
@@ -33,6 +33,10 @@
     "rhel_uefi": {
       "Value": "true",
       "Description": "Create a RHEL UEFI image."
+    },
+    "destination_prefix": {
+      "Required": true,
+      "Description": "GCS path to export image metadata to"
```
hopkiw, line 39: i'm not sure why this is passed in. can't we deduce from the workflow that it should be placed alongside the image? or do we want to publish it somewhere else?

daisy_workflows/image_build/enterprise_linux/enterprise_linux_uefi.wf.json
```diff
@@ -44,6 +48,8 @@
     "build_files/kickstart": "./kickstart/",
     "build_files/ks_helpers.py": "./ks_helpers.py",
     "build_files/save_logs.py": "./save_logs.py",
+    "build_files/retrieve_metadata.py": "./retrieve_metadata.py",
```
hopkiw, line 51: i don't see this file in the PR, or see it used in the script

daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -175,6 +177,54 @@ def main():
...
+
+  for package in guest_packages:
+    cmd = "dpkg-query -W --showformat '${Package} ${Version} ${Git}\n'" \
+          + package
+    code, stdout = utils.Excute(cmd, capture_output=True)
+    if code == 0:
+      splits = stdout.decode('utf-8').split('\t\b')
+      package_name = splits[0]
+      package_version = splits[1]
+      package_commit_hash = splits[2][splits[2].rindex('/'):len(splits[2])]
```
hopkiw, line 206: this is quite a bit of work, mostly because of the format you passed to dpkg-query. if you changed the format of dpkg-query, you could produce output that's easier to parse.

daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -175,6 +177,54 @@ def main():
...
+          "stage": google_cloud_repo
+      }
+      image["packages"].append(metadata)
+      with open('/tmp/metadata.json', 'w') as f:
+        f.write(json.dumps(image))
+
+      logging.info('Uploading image metadata.')
+      metadata_dest = os.path.join(image_dest.strip("root.tar.gz"),
+                                   'metadata.json')
+      utils.UploadFile('/tmp/metadata.json', metadata_dest)
```
hopkiw, line 226: it's not entire clear in this diff interface, but is the indentation here suggesting we will upload this file once for every package?

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: I copy paste from build.py. But I think it's no need here.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: This is the image Id. for each build, we generate a unique id as primary key.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: The intent for this field is to record which environment we publish image to. However, I notice this should be determined during publish process (*.publish.json). So i think we should remove this field here.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: Gotcha.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: i think the mistake is "indent". the "package" metadata is a array field in the "image". 
json.dump(image) should produce 👍 
`{
  "id": "image-uuid",
  "family": "debian-8",
  "name": "gold",
  "release_date": "release_date",
  "state": "state",
  "environment" : "environment",
  "packages" : [
    {
      "id": "package-1-uuid",
      "name": "package_name",
      "version": "package_version",
      "commit_hash": "package_hash_1",
      "release_date": "package_release_date",
      "stage": "repo"
    },
    {
      "id": "package-2-uuid",
      "name": "package_name-2",
      "version": "package_version-2",
      "commit_hash": "package_hash_2",
      "release_date": "package_release_date-2",
      "stage": "repo"
    } 
   ]
}`

daisy_workflows/image_build/enterprise_linux/enterprise_linux_uefi.wf.json
gaohannk: Should it be explicitly included in the IncludWorkFlow.Vars?. I thought it's inherit from parent workflow.

gaohannk: Using datetime to directly generate it to avoid pass in.

daisy_workflows/image_build/enterprise_linux/enterprise_linux_uefi.wf.json
gaohannk: We want to place with the same gcs path with image. however the destination is defined in the "export-image" step.
But looks like it's not easy to construct the output path directly. So I thought it's may be easier to pass in the path.

`gs://gce-image-archive/centos-uefi/centos-7-v${build_date}/root.tar.gz
gs://gce-image-archive/centos-mbr/centos-8-mbr-v${build_date}/root.tar.gz`

daisy_workflows/image_build/debian/build_fai.py
gaohannk: My bad. We should upload one single json file per image.

daisy_workflows/image_build/debian/build_fai.py
gaohannk: This field is from "Git". The field is parse commit hash from git pull path.  For example: https://github.com/hopkiw/compute-image-packages/tree/33bed8dcff55a78a55bb589e071548fbc6b949ed

daisy_workflows/image_build/enterprise_linux/export_metadata.py
hopkiw: well builds are not performed by borgcron, but yes the uuid can be passed in which could then be used by both the build and the metadata. let's discuss offline

daisy_workflows/image_build/enterprise_linux/enterprise_linux_uefi.wf.json
hopkiw: that will require a lot of changes to the infrastructure surrounding these buils i think. let's discuss that offline too and revisit

daisy_workflows/image_build/debian/build_fai.py
hopkiw: I don't understand your reply, sorry. the link is to the whole repo at a point in time. and also we are not discussing a single field, but the lines 203 through 206 (at least) where we're doing string parsing. we control the output format, so we could produce a format that is explicitly designed to be easier to parse. for example we could use field separators instead of whitespace.

gaohannk: Will using separator, and remove parsing the github link.

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +124,50 @@ def main():
...
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "dpkg-query -W --showformat '${Package}:${Version}:${Git}'" \
+          + package
+    code, stdout = utils.Excute(cmd, capture_output=True)
+    if code == 0:
+      splits = stdout.decode('utf-8').split(':')
```
hopkiw, line 146: recommend:
```
package, version, git = stdout.decode('utf-8').split(':',2)
```

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +124,50 @@ def main():
...
+      "location": image_dest,
+      "release_date": datetime.now(timezone.utc),
+      "packages": []
+  }
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "dpkg-query -W --showformat '${Package}:${Version}:${Git}'" \
```
hopkiw, line 142: given the version has a colon in it, you might use a different separator

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +124,50 @@ def main():
...
+          + package
+    code, stdout = utils.Excute(cmd, capture_output=True)
+    if code == 0:
+      splits = stdout.decode('utf-8').split(':')
+      package_name = splits[0]
+      package_version = splits[1]
+      package_git_url = splits[2]
+      start = package_version.index(":")
+      end = package_version.rindex(".")
+      package_release_date = package_version[start: end]
```
hopkiw, line 152: uh given you did split on ':' already, you shouldn't have any more to index off of here. if you did, it looks like you're trying to pull a date out of a version, and i think you should just store the version as-is and not try to parse it. also the version contains the build date, not the release date.

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +124,50 @@ def main():
   synopsis_dest = os.path.join(outs_path, 'synopsis.json')
   utils.UploadFile('/tmp/synopsis.json', synopsis_dest)
 
+  # Create and upload metadata of the image and packages
+  logging.info('Creating image metadata.')
+  image = {
+      "family": "debian-9",
+      "name": re.match("debian-9-[\\w-]*", image_dest),
```
hopkiw, line 131: we can't know the image name at the time of source creation, as this one source could be associated with many images. think this will have to come from gce api

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +124,50 @@ def main():
...
+      "version": "stretch",
+      "location": image_dest,
+      "release_date": datetime.now(timezone.utc),
```
hopkiw, line 134: image build date, not release date

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +124,50 @@ def main():
...
+          "stage": repo
+      }
+      image["packages"].append(metadata)
+
+    # Write image metadata to a file
+    with open('/tmp/metadata.json', 'w') as f:
+      f.write(json.dumps(image))
+
+    logging.info('Uploading image metadata.')
+    metadata_dest = os.path.join(image_dest.strip("root.tar.gz"),
```
hopkiw, line 167: why don't we also pass in a `destination` variable for debian ? this approach makes debian somewhat less flexible than EL

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+      "packages": []
+  }
+
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    # The URL is a github repo link which don't contain commit hash
+    cmd = "rpm -q --queryformat='%{NAME}\n%{VERSION}\n%{RELEASE}\n%{URL}'" \
```
hopkiw, line 71: don't need `%{URL}`, should replace with `%{Vcs}`. also, why use colon in debian and newline in EL?

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+    if code == 0:
+      splits = stdout.decode('utf-8').split('\n')
+      package_name = splits[0]
+      package_version = splits[1] + "-" + splits[2]
+      package_url = splits[3]
+      package_release_date = package_version[0:package_version.index(".")]
+      package_metadata = {
+          "name": package_name,
+          "version": package_version,
+          # For el, we don't have commit hash
```
hopkiw, line 83: what does this comment mean? we should definitely have the commit hash from the `Vcs` field, but even if we didn't, why are we putting the package URL into the commit hash field?

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+          "commit_hash": package_url,
+          "release_date": package_release_date,
+          "stage": repo
```
hopkiw, line 86: i don't think you should include this, maybe just include once as 'build_repo' in the image metadata. it's not really information about the package

daisy_workflows/image_build/debian/build.py
hopkiw: maybe use newline as you did for EL?

daisy_workflows/image_build/debian/build.py
gaohannk: i thought this build.py is only invoked by debian9.wf.json workflow. so should I remove this field and populate later from gce api to image?

daisy_workflows/image_build/debian/build.py
gaohannk: Let's unified with "\n"

daisy_workflows/image_build/debian/build.py
gaohannk: I change the field "release_date" to "build_date" for both image and package. We store the version as-is. and I want to include a "build_date" field which the type in spanner is Date for query. But to me, the "version" is the only "source" we can get this. At least, we need it to parse/convert from it somewhere.

daisy_workflows/image_build/debian/build.py
gaohannk: the image_dest is already existed compared to EL. I can make it align with EL to avoid a strip here. But a little bit tricky here is, we can reuse "destination" variable in "build" and "export_image" steps for EL. however, for debian, we need to keep both "image_dest" and "destination" together even they has the same gcs folder prefix. 
imag_dest: "gcs://path/to/folder/root.tar.gz"
destination: "gcs://path/to/folder/metadata.json"

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: right. in realty, it's same for all packages in one image. we don't pull packages from different repos. and not exactly the attribute for a package. I will remove it.

daisy_workflows/image_build/debian/build.py
hopkiw: yes, the 'name' should be the name from the gce image resource. the build workflow only produces the binary source for an image, the `root.tar.gz` file. it has no name, only family and destination path.

daisy_workflows/image_build/debian/build.py
hopkiw: i should have made this two comments.

comment 1, code error: the `.index(":")` should not work since you have already called `.split(':')` on this string. 

comment 2, we should not parse "1:20200707.01.g1" into "20200707". Just pull the entire version field and don't bother about parsing it. In fact, you should append the release onto the version.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,95 @@
...
+import logging
+import os
+
+import utils
+
+
+def main():
+  # Get Parameters.
+  repo = utils.GetMetadataAttribute('google_cloud_repo',
+                                    raise_on_not_found=True)
```
EricEdens, line 37: Interesting -- so the new code for Debian and FAI lives in an existing `build` file, while for EL it gets its own new file. I'm left wondering why the same pattern wasn't used for all three (either combine into existing build files, or create entirely new files)

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,95 @@
...
+      "packages": []
+  }
+
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "rpm -q --queryformat='%{NAME}\n%{VERSION}\n%{RELEASE}\n%{Vcs}'" \
+          + package
```
EricEdens, line 71: Any thoughts on testing?

gaohannk: not aware of any unit test, but probably can add in integrated test?

daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -175,6 +178,40 @@ def main():
...
+      "packages": []
+  }
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "dpkg-query -W --showformat '${Package}\n${Version}\n${Git}'" \
+          + package
+    code, stdout = utils.Excute(cmd, capture_output=True)
```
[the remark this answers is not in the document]
gaohannk: Are you saying the build.py and build_fai.py? It's diverge somehow I don't know why

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +125,40 @@ def main():
...
+          + package
+    code, stdout = utils.Excute(cmd, capture_output=True)
+    if code == 0:
+      package, version, git = stdout.decode('utf-8').split(':', 2)
+      metadata = {
+          "name": package,
+          "version": version,
+          "commit_hash": git,
+      }
+      image["packages"].append(metadata)
```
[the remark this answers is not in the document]
gaohannk: Great catch, could log and error for other exit codes.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: While liam has explain this to me before. The debian and el workflow is not follow the same pattern. Make it simple, the el workflow separate the export image step to using export tool.

daisy_workflows/image_build/debian/build.py
gaohannk: I have fixed this. I  put the full version string.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,94 @@
...
+    utils.Execute(['mount', '/dev/sdb1', '/mnt'])
+
+  logging.info('Installer root: %s' % os.listdir('/mnt'))
+
+  # Create and upload metadata of the image and packages
+  logging.info('Creating image metadata.')
+  build_date = datetime.today().strftime('%Y%m%d')
+  image = {
+      "family": el_release_name,
+      "name": el_release_name + build_date,
```
hopkiw, line 57: i think we can just omit this field. it is only comprised of other existing fields, and it will be confused for the image name, which must be taken from gce

daisy_workflows/image_build/enterprise_linux/export_metadata.py
```diff
@@ -0,0 +1,94 @@
...
+      "build_repo": repo,
+      "packages": []
+  }
+
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "rpm -q --queryformat='%{NAME}\n%{VERSION}\n%{RELEASE}\n%{Vcs}'" \
```
hopkiw, line 70: think you need a space here too

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: I will remove this to keep consistent with debian.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: Fix

daisy_workflows/image_build/debian/debian_sid_fai.wf.json
```diff
@@ -3,6 +3,7 @@
   "Vars": {
     "build_date": {"Value": "${DATE}", "Description": "Build datestamp used to version the image."},
     "image_dest": {"Required": true, "Description": "The GCS path for the destination image tar.gz."},
+    "destination": {"Required": true, "Description": "The GCS path for the destination image metadata json"},
```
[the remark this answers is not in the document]
EricEdens: Looks like there are some more `destination` daisy variables, eg: daisy_workflows/image_build/enterprise_linux/rhel_6.wf.json

daisy_workflows/image_build/debian/build_fai.py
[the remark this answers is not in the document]
EricEdens: I see, yeah, it looks like there are already several parallel implementations.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
EricEdens: integ test sgtm :) (probably in another PR, I'm guessing?)

daisy_workflows/image_build/debian/debian_sid_fai.wf.json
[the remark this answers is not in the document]
gaohannk: oh. I thought you just want to change in debian. For EL, there is no "image_dest" var, instead it's named "destination" and it could be reused in different step. So It's not necessary to rename it as it's just used in export_metadata step. The meaning is clear in that step. Yea, that's another diverge from debian.

daisy_workflows/image_build/debian/build.py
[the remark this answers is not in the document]
gaohannk: I remove the exit code check in next revision as i just notice that util.Excute() method already handle the exit code and logging.

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +125,40 @@ def main():
...
+      "build_repo": repo,
+      "packages": []
+  }
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "dpkg-query -W --showformat '${Package}\n${Version}\n${Git}'" \
+          + package
```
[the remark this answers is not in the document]
gaohannk: Fix in the new revision.

daisy_workflows/image_build/enterprise_linux/export_metadata.py
gaohannk: Change to use format string. 
"rpm -q --queryformat='%%{NAME}\n%%{VERSION}\n%%{RELEASE}\n%%{Vcs}' %s" \
          % package

daisy_workflows/image_build/debian/build.py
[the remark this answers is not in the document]
gaohannk: Change to use format string.
 cmd = "dpkg-query -W --showformat '${Package}\n${Version}\n${Git}' %s" \
          % package

daisy_workflows/image_build/debian/build.py
```diff
@@ -123,6 +125,39 @@ def main():
...
+      "build_date,": datetime.now(timezone.utc),
+      "build_repo": repo,
+      "packages": []
+  }
+  # Read list of guest package
+  with open("guest_package") as f:
+    guest_packages = f.read().splitlines()
+
+  for package in guest_packages:
+    cmd = "dpkg-query -W --showformat '${Package}\n${Version}\n${Git}' " \
```
hopkiw, line 143: style guide suggests
```
cmd = ("",
"")
```

```
Do not use backslash line continuation except for with statements requiring three or more context managers.
Make use of Python’s implicit line joining inside parentheses, brackets and braces. If necessary, you can add an extra pair of parentheses around an expression.

```
https://google.github.io/styleguide/pyguide.html

PR #1254, opened 2020-06-16
Add external IP flag to daisy export workflows

jeffallen-google:
Citrix Machine Creation Services (MCS) uses Daisy Export workflows during VM image preparation. Customers have complained that the export instance launches with an external IP. This pull request adds new workflow vars allowing client to optionally disable external IP address for initial launch of export VM. New flag defaults to "false" to preserve previous behavior.

### DIFF
--- a/daisy/instance.go
+++ b/daisy/instance.go
@@ -103,6 +103,9 @@ type InstanceBase struct {
 	// RetryWhenExternalIPDenied indicates whether to retry CreateInstances when
 	// it fails due to external IP denied by organization IP.
 	RetryWhenExternalIPDenied bool `json:",omitempty"`
+	// NoExternalIP indicates whether to try CreateInstances with
+	// an external IP in the first place.
+	NoExternalIP string `json:",omitempty"`
 	// Should an existing instance of the same name be deleted, defaults to false
 	// which will fail validation.
 	OverWrite bool `json:",omitempty"`
@@ -116,6 +119,7 @@ type Instance struct {
 
 	// Additional metadata to set for the instance.
 	Metadata map[string]string `json:"metadata,omitempty"`
+
 }
 
 // InstanceBeta is used to create a GCE instance using Beta API.
@@ -500,7 +504,11 @@ func (i *Instance) populateNetworks() DError {
 	}
 	for _, n := range i.NetworkInterfaces {
 		if n.AccessConfigs == nil {
-			n.AccessConfigs = defaultAcs
+			if i.NoExternalIP == "true" {
+				n.AccessConfigs = []*compute.AccessConfig{}
+			} else {
+				n.AccessConfigs = defaultAcs
+			}
 		}
 
 		// Only set deafult if no subnetwork or network set.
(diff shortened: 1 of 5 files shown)


zoran15:
Ideally, we'd want customers to use cli_tools/gce_vm_image_export instead of using Daisy export workflow directly. This because gce_vm_image_export has additional logic to make export more reliable. This PR modifies only the native Daisy export path, while the users of gce_vm_image_export (gcloud, UI) will not have the ability to use this new feature.

zoran15:
Additionally, the `RetryWhenExternalIPDenied` flag (which is set to true by default) should already handle the scenario when external IPs are dissabled, by re-trying instance create step with no external IP. Is this not working for Citrix Machine Creation Services?

jeffallen-google:
> Additionally, the `RetryWhenExternalIPDenied` flag (which is set to true by default) should already handle the scenario when external IPs are dissabled, by re-trying instance create step with no external IP. Is this not working for Citrix Machine Creation Services?

Yes, `RetryWhenExternalIPDenied` should retry to handle the case when external IP's are denied by org policy. The intent here, by doing initial VM launch without external IP, is to avoid the necessity for retry and also handle the case that external IP's are not desired but also not explicitly denied by org policy.

jeffallen-google:
> Additionally, the `RetryWhenExternalIPDenied` flag (which is set to true by default) should already handle the scenario when external IPs are dissabled, by re-trying instance create step with no external IP. Is this not working for Citrix Machine Creation Services?



> Ideally, we'd want customers to use cli_tools/gce_vm_image_export instead of using Daisy export workflow directly. This because gce_vm_image_export has additional logic to make export more reliable. This PR modifies only the native Daisy export path, while the users of gce_vm_image_export (gcloud, UI) will not have the ability to use this new feature.

Thanks for the feedback.

Citrix is invoking these workflows directly using cloudbuild and the daisy container image, so this covers their requirements.

If you recommend also adding the new flag to gce_vm_image_export for other users, I'll be happy to look into including those additional changes to this PR.

zoran15:
> > Additionally, the `RetryWhenExternalIPDenied` flag (which is set to true by default) should already handle the scenario when external IPs are dissabled, by re-trying instance create step with no external IP. Is this not working for Citrix Machine Creation Services?
> 
> > Ideally, we'd want customers to use cli_tools/gce_vm_image_export instead of using Daisy export workflow directly. This because gce_vm_image_export has additional logic to make export more reliable. This PR modifies only the native Daisy export path, while the users of gce_vm_image_export (gcloud, UI) will not have the ability to use this new feature.
> 
> Thanks for the feedback.
> 
> Citrix is invoking these workflows directly using cloudbuild and the daisy container image, so this covers their requirements.
> 
> If you recommend also adding the new flag to gce_vm_image_export for other users, I'll be happy to look into including those additional changes to this PR.

Ideally, we don't want to add flags to Daisy for specific use cases that can be handled via custom workflow files. Fore example, no external IP can be achieved by using a modified export workflow file where AccessConfig on the instance is set to empty. Any custom logic should go in gce_vm_image_export to keep Daisy less complex. 

I see two approaches here: 
* Citrix switches to using gce_vm_image_export and we add a NoExternalIP flag there, or
* they start using a custom export workflow with AccessConfig set to empty

zoran15:
Closed by mistake.

jeffallen-google:
> I see two approaches here:
> 
> * Citrix switches to using gce_vm_image_export and we add a NoExternalIP flag there, or
> * they start using a custom export workflow with AccessConfig set to empty

I'll explore a custom export workflow first. If that doesn't pan out, we'll look into moving the flag out to gce_vm_image_export. Thanks for the guidance!

PR #1256, opened 2020-06-16
[windows upgrade] show cleanup guide again when success

dntczdx:
As per feedback from tech writer team, we show the cleanup guide again after the upgrade succeeded.

### DIFF
--- a/cli_tools/gce_windows_upgrade/upgrader/utils_test.go
+++ b/cli_tools/gce_windows_upgrade/upgrader/utils_test.go
@@ -84,12 +84,14 @@ func TestGetCleanupIntroduction(t *testing.T) {
 	}
 
 	for _, tc := range tcs {
-		u := upgrader{
-			derivedVars: &derivedVars{
-				originalWindowsStartupScriptURL: tc.scriptURLPtr,
-			},
-		}
-		_, err := getCleanupIntroduction(&u)
-		assert.NoError(t, err, "[test name: %v]", tc.name)
+		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cli_tools/gce_windows_upgrade/upgrader/utils_test.go
```diff
@@ -70,3 +70,26 @@ func TestGetIntroHelpText(t *testing.T) {
...
+	}
+
+	for _, tc := range tcs {
+		u := upgrader{
+			derivedVars: &derivedVars{
+				originalWindowsStartupScriptURL: tc.scriptURLPtr,
+			},
+		}
+		_, err := getCleanupIntroduction(&u)
+		assert.NoError(t, err, "[test name: %v]", tc.name)
```
EricEdens, line 93: Validate the content of the message?  (It can be tricky to find a balance here, since validating the *whole* message can make the test brittle. But *some* validation is a good sanity check)

EricEdens, line 93: With t.Run, you won't need to manually scope the error check: https://github.com/golang/go/wiki/TableDrivenTests

dntczdx: done

dntczdx: No need to do that. The only dynamic part in the msg is the 3 vars. I used "missingkey=error" so if vars are not fullfilled, there will be an error thrown. So check err == nil is enough.

PR #1257, opened 2020-06-16
Ensure destination image doesn't exist prior to running import.

EricEdens:
Prior to the #1227, import failed early when an image name was already used. Following #1227, the failure occurred after inflation, which introduces a significant delay.

Example invocation:

```
➜   ./main -image_name ubuntu  \
               -data_disk -client_id  edens-test \
               -source_image projects/compute-image-tools-test/global/images/debian-8-translate \
               -project=edens-test
[import-image] 2020/06/16 12:07:40 the image "ubuntu" already exists. Please remove it, or pick a name that isn't already used
```

The failure message appears instantly (at least to my eyes!).

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/validator.go
+++ b/cli_tools/gce_vm_image_import/importer/validator.go
@@ -45,8 +45,8 @@ func (v validateImageNameAvailable) validate() error {
 	// then the image name may be available.
 	image, _ := v.client.GetImage(v.project, v.name)
 	if image != nil {
-		return fmt.Errorf("the image %q already exists. Please remove it, "+
-			"or pick a name that isn't already used", v.name)
+		return fmt.Errorf("The resource '%s' already exists. "+
+			"Please pick an image name that isn't already used.", v.name)
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


cli_tools/gce_vm_image_import/importer/validator.go
```diff
@@ -0,0 +1,57 @@
...
+	client        getImageClient
+}
+
+func (v validateImageNameAvailable) validate() error {
+	// We ignore the error, with the assumption that if there's an error,
+	// then the image name may be available.
+	image, _ := v.client.GetImage(v.project, v.name)
+	if image != nil {
+		return fmt.Errorf("the image %q already exists. Please remove it, "+
+			"or pick a name that isn't already used", v.name)
```
zoran15, line 49: I don't think we should be recommending customers to remove existing images, as that sounds potentially dangerous and they could blame us if they lose any data. Stating image with the given name already exist should be enough. 

Also, it's a good idea to keep the existing error if any of the clients depend on a specific pattern. We can add more info, but we should keep existing error at least as an "intro" to the more verbose error.

PR #1262, opened 2020-06-17
Don't allocate an external IP during inflation.

EricEdens:
This change allows organizations that [disable external IPs](https://cloud.google.com/compute/docs/ip-addresses/reserve-static-external-ip-address#disableexternalip) to run image import.

### DIFF
--- a/daisy_workflows/image_import/inflate_file.wf.json
+++ b/daisy_workflows/image_import/inflate_file.wf.json
@@ -86,7 +86,8 @@
           "networkInterfaces": [
             {
               "network": "${import_network}",
-              "subnetwork": "${import_subnet}"
+              "subnetwork": "${import_subnet}",
+              "accessConfigs": []
             }
           ],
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/inflate_file.wf.json
```diff
@@ -86,7 +86,8 @@
           "networkInterfaces": [
             {
               "network": "${import_network}",
-              "subnetwork": "${import_subnet}"
+              "subnetwork": "${import_subnet}",
+              "accessConfigs": []
```
zoran15, line 90: I don't see why inflater would need an external IP so this is probably OK. We should just keep an eye on any unforseen issues. If there are non, we should do the same for the exporter.

EricEdens: sgtm. I did a few test runs and everything looked  good.

PR #1263, opened 2020-06-18
Image import global timeout handling

zoran15:


cli_tools/gce_vm_image_import/importer/data_disk_processor.go
```diff
@@ -53,14 +53,18 @@ func newDataDiskProcessor(pd persistentDisk, client createImageClient, project s
...
 	return []string{}
 }
 
-func (d dataDiskProcessor) process(ctx context.Context) (err error) {
+func (d *dataDiskProcessor) process(ctx context.Context) (err error) {
 	return d.client.CreateImage(d.project, &d.request)
 }
 
+func (d *dataDiskProcessor) cancel(reason string) {
+	//not possible to cancel direct calls to compute client and not likely to take long
```
dntczdx, line 65: If it's a long API call (like create instance/disk), there is a way to cancel techincally. The "operation" id can be used to cancel.

zoran15: Do you have any examples?

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -85,27 +95,45 @@ func (i *importer) runInflate(ctx context.Context) (err error) {
...
 		err = ctx.Err()
 	default:
-		err = processor.process(ctx)
-		i.traceLogs = append(i.traceLogs, processor.traceLogs()...)
+		finished := i.handleCancel(ctx, i.processor.cancel)
+		err = i.processor.process(ctx)
+		i.traceLogs = append(i.traceLogs, i.processor.traceLogs()...)
+		close(finished)
 	}
 	return err
```
EricEdens, line 121: My understanding of the context pattern is that it's the callee's responsibility to monitor the context, and to quit when the context is cancelled. The immediate result is that an implementation like the data disk processor will "just work" (assuming that the client that it's using also follows this pattern).

The other benefit is that we keep the `importer.go` file focused on business logic, both in its implementation, its future reading, and its  testers. This will become *more* important when we add upcoming features such as:
* UEFI
* native vmdk
* OS inspection
* capacity inspection

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -37,8 +37,8 @@ type bootableDiskProcessor struct {
 	OS              string
 }
 
-func (b bootableDiskProcessor) process(ctx context.Context) (err error) {
-	err = b.workflow.RunWithModifiers(ctx, b.preValidateFunc(), b.postValidateFunc())
+func (b *bootableDiskProcessor) process(ctx context.Context) (err error) {
```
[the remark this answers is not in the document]
zoran15: Because the receiver is copied on each call. What's the motivation not to use a pointer?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -112,16 +113,27 @@ func createDaisyInflater(args ImportArguments) (inflater, error) {
...
 
+func (inflater *daisyInflater) cancel(reason string) {
+	inflater.wf.IterateWorkflowSteps(func(step *daisy.Step) {
+		if step.CreateDisks != nil {
+			for _, disk := range *step.CreateDisks {
+				disk.NoCleanup = false
+			}
+		}
+	})
+	inflater.wf.CancelWithReason(reason)
```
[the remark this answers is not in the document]
zoran15: NoCleanup is set to true here because the inflated disk is needed for the processing step (translate or just image create). If the workflow is cancelled, this disk should be cleaned up. Otherwise, temporary disks would keep accumulating and customers would keep incurring charges. The only reason to keep temporary disks is debugging, which doesn't apply here.

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: Can you be more specific, in the context of this code?

EricEdens: Assuming you initialize the context's timeout prior to executing `importer.Run`, you need zeror changes to `importer.go` or its test.

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
EricEdens: I'm glad we're removing the disk; my concern is about perpetuating the pattern of modifying workflows after they've started.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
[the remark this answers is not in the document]
EricEdens: > Because the receiver is copied on each call.

Were you seeing pressure on the GC? The general guidance is to reserve pointers for mutability, which serves for self-documentation (danger! things change!), and safety, to ensure thing's *don't* change when when unexpected. 

go/gocomments#pass-values:  "Don't pass pointers as function arguments just to save a few bytes."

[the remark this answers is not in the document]
zoran15: That link is about func arguments. Receiver is a special type of an argument where function "belongs" to a struct and there's a high chance it will modify it. Using pointers consistently for receivers will reduce hard to trace errors where, a function that once only used the receiver, needs to modify something, but the modifications don't stick because they were made to a copy. I actually had this situation in this code where receivers are a mix of pointers and values. I find that mix very messy.

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
zoran15: Your second question to Andrew was how would he feel about deleting all disks, even if `NoCleanup` is set to `true`. I've answered that question in my previous comment.

Regarding the mechanism of deletion, I agree using workflow modification is not the prettiest and goes against what we agreed on regarding Daisy usage. I'll change it to avoid workflow modification.

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: There are a couple of issues with this approach:
- logging to GCS fails on context expiration leading to customer visible errors and a partial loss of logs
- I don't see an easy way to change Daisy step failure reason to indicate a timeout occurred, which is one of the main requirements for this change as it allows us to track timeouts vs. other failures.
- temporary disk cleanup won't happen because importer.diskCleanup() is initialized only after inflation finishes. Disk is created during inflation. If timeout happens during inflation, disk will be left hanging in the project. This might be solvable by initializing diskCleanup early and checking if the disk still exists or not.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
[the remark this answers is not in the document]
zoran15: https://tour.golang.org/methods/8

https://golang.org/doc/faq#methods_on_values_or_pointers

_If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used._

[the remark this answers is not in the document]
EricEdens: All of the receivers for bootableDiskProcessor currently *are* values, though -- what were you wanting to modify?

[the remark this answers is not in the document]
zoran15: I think for receivers we should always use pointers for consistency. I don't see a strong reason not to. More likely than not there will be a need to modify the receiver since functions belong to it. When that happens, all receivers need to be changed. It can also lead to hard to debug bugs where changes to a receiver don't stick because it's a value. 

The only benefit for using value receivers is to be explicit changes will not be made, which is dubious for receivers IMO. Any other benefits?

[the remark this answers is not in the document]
EricEdens: Mutable objects make even *harder*-to-debug situations, and there's a wealth of research and work to avoid them. What're you wanting to mutate here?

- https://github.com/google/guava/wiki/ImmutableCollectionsExplained
- https://github.com/google/auto/blob/master/value/userguide/index.md
- https://research.swtch.com/go2017#immutability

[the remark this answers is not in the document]
zoran15: I already explained it's for consistency across our code base, never said I want to mutate anything at this right moment.

cli_tools/gce_vm_image_import/importer/data_disk_processor.go
dntczdx: Here: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/compute/compute.go#L394
The c.raw.Disks.Insert actually returns a operation, which is a handle and can be used to cancel a long API.
Currently we don't have a mechanism to cancel it. If we want, we need to add some mechanism in daisy.Compute. You can decide whether it's worth doing it for this project.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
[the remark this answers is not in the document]
EricEdens: I'm very interested to learn about your view on exclusive-pointer receivers, and we should continue the discussion on its own. Given that this current PR doesn't need that functionality, though, it's probably best to forego those refactoring changes for now.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -58,15 +60,23 @@ type importer struct {
...
+	processor         processor
+	cancelReason      string
 }
 
-func (i importer) Run(ctx context.Context) (loggable service.Loggable, err error) {
-
+func (i *importer) Run(ctx context.Context) (loggable service.Loggable, err error) {
+	if i.timeout.Nanoseconds() > 0 {
+		var cancel func()
+		ctx, cancel = context.WithTimeout(ctx, i.timeout)
```
adjackura, line 71: For now don't cancel the context you pass to daisy, I have seen this cause behavior in cleanup that we would not want. Instead use the existing workflow cancel functions to tell a workflow to cancel and cleanup, you most likely will have to spin up a goroutine that just waits for a timer that then calls the cancel function

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -81,46 +92,83 @@ func (i importer) Run(ctx context.Context) (loggable service.Loggable, err error
...
+
+	disk, err := i.diskClient.GetDisk(i.project, i.zone, diskName)
+	if err != nil || disk == nil {
+		return
+	}
+	err = i.diskClient.DeleteDisk(i.project, i.zone, diskName)
+	if err != nil {
+		logger.Errorf("Failed to remove temporary disk %v: %e", i.pd, err)
 	}
 }
```
EricEdens, line 164: I took a look, and if the disk doesn't exist, Delete returns a 404. So, I'd recommend doing a single call, and skipping the error message if it's a 404 (and omitting the GetDisk call)

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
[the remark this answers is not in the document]
EricEdens: So, I'm still seeing receiver changes in this version. For this PR, let's focus on the current issue, and then save refactorings and other items for future PRs.

EricEdens, line 40: This package is written to follow the guidance on contexts, specifically that callees are responsible for watching for cancellations. This PR, however, deviates from that guidance, but leaves the package in a half-way state where contexts are passed around, but aren't used. Recall the confusion caused where Daisy accepts a context, but you're not allowed to cancel it. Since Daisy doesn't follow the context idiom, it's hard for future users to understand and use it. By following idiomatic guidance, I'm hoping to avoid that type of confusion.

I can see two approaches forward:
1. Move timeout logic fully into importer.go, and remove context from the signatures of downstream functions.
2. Keep the current architecture (see #1273)

I see two downsides with (1). First, it breaks with idiomatic patterns. So, if we do that, I think we should make a strong argument for why the idioms shouldn't apply here. Second, the change will have a large blast radius, so we'll need to justify that cost.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -81,46 +92,83 @@ func (i importer) Run(ctx context.Context) (loggable service.Loggable, err error
...
 		return err
 	}
+	return i.runStep(ctx, func(ctx context.Context) error { return i.processor.process(ctx) }, i.processor.cancel, i.processor.traceLogs)
+}
+
+func (i *importer) runStep(ctx context.Context, step func(context.Context) error, cancel func(string), getTraceLogs func() []string) (err error) {
+	e := make(chan error)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
```
EricEdens, line 114: I was hoping importer.go would focus on business logic, which will be important when we add more business logic steps. Think:
1. Native VMDK
2. UEFI tagging
3. OS detection
4. Pre-detecting disk size requirements
5. Switching to different translation logic

cli_tools/gce_vm_image_import/importer/processor.go
```diff
@@ -26,6 +26,7 @@ import (
 type processor interface {
 	process(ctx context.Context) error
 	traceLogs() []string
+	cancel(reason string)
```
EricEdens, line 29: have you envisioned other cancellation reasons?

cli_tools/gce_vm_image_import/importer/data_disk_processor.go
EricEdens: How about instantiating the compute client with a context that has the deadline applied? Or switching to a client that takes the context as a parameter?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
zoran15: What do you mean by large blast radius in case of removing context from inflater/processors? The change will be contained within the package.

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: In what way is this current solution blocking the addition of these new steps?

cli_tools/gce_vm_image_import/importer/processor.go
zoran15: Yes. For example, customer cancelled. We currently rely on Daisy for these cancellations. Certain steps already don't use Daisy and we might need to implement that feature in the importer itself.

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: Changed to pass a new context to Daisy and use importer context for importer timeout

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: Done

cli_tools/gce_vm_image_import/importer/processor.go
EricEdens: sgtm

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: Here, the `runStep` method assumes that steps will always have the signature `func(context.Context) error`. Although this is true today, these upcoming projects will challenge that assumption. Consider pre-detection of disk sizes. Using the straight-line design that I originally envisioned, an `inspector` would return its inspection results, and those results would be passed to `inflater.inflate` as a new parameter. 

    diskFile, err := i.inspector.inspect(ctx)

    // handle error

    if err = i.runInflate(ctx, diskFile.size); err != nil {
      return i.buildLoggable(), err
    }

You have any thoughts on what this would look like with your step framework?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
EricEdens: You're right, it's all in the same package. I just meant that the heavy-lifting interfaces (validation, inflation, and processing) would all need to be changed, along with their call sites and tests.

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: We already do something similar where inflater passes i.pd to processor. Step doesn't have to be `func(context) err` if we use closures (like currently for inflater).

EricEdens: very good point!

cli_tools/gce_vm_image_import/importer/data_disk_processor.go
zoran15: Based on my observations, when context expires, compute client will lose connectivity to the API endpoint. So having context expire will not do much for cancelling an operation. Instead, we would be unable to do any further clean-ups.

For this particular case, since image creation is fast, what we can do is wait for the operation to finish and, if it succeeded, delete the image that was created. 

I don't have any experience with cancelling compute operations, i.e. not sure how reliable it is. If we cancel an image create, are we guaranteed image will never be created? If not, we might still need to do a check ourselves and delete the image in our code.

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: 1. I'm not sure a single, global context would be appropriate for all RPCs. I imagine one would be used for the workflow, and would have the global timeout, and another would be used for individual operations such as logging. For example, you might want to limit logging to `n` amount of time. The [Polly framework](https://github.com/App-vNext/Polly/wiki/Transient-fault-handling-and-proactive-resilience-engineering) has a good discussion about various resiliency strategies, and does a good job of showing that we need multiple strategies since individual problems are slightly different.
2. For reporting the cancellation reason, doesn't the `context` package have typed errors for determining why the context was cancelled? See the [variables section](https://golang.org/pkg/context/#pkg-variables) in their docs. Also, we're free to implement the context.Context interface if those typed errors aren't sufficient.
3. Regarding cleanup of temporary disks, I like the solution that you've developed here; we can use that, right?
  * Install the disk cleanup deferal prior to starting inflation
  * If disk cleanup returns an HTTP 404, consider cleanup successful.

I have a proposal though! If we update Daisy so that cancellation [the context passed to Run](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/423907b63e7430a8e9dbe5edb296e4811f73797a/daisy/workflow.go#L266) is well-behaved, then your global timeout feature is nearly free.

I foresee the following changes:
* Daisy:
  1. Ensure the correct cancellation reason is surfaced.
  2. Ensure that cleanup occurs after cancellation.
* importer package:
  1. Tune disk cleanup to ensure disks are removed if timeout occurs during inflation.
  2. Ensure that logging occurs regardless of cancellation.

I'm happy to divide up work to make it go quicker.

Also, I want to say that I appreciate the amount of time you've spent thinking about this problem, and how you've developed several approaches and solutions. :)

cli_tools/gce_vm_image_import/importer/data_disk_processor.go
EricEdens: As I think about it, I like your current approach here. Given these two options, it seems like the first is the most user friendly (at leas to me!):
1. Exceed the timeout budget by a few seconds, but successfully finish the import and create an image.
2. Strictly follow the timeout budget, and delete the final image that would otherwise have been successful. 

A small irony in the second approach, too, is that deleting the image would require a nonzero amount of time :)

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: 1. What is the benefit of using separate context objects for each operation? Seems convoluted and not really plug&play as you describe it. We'd have a bunch of context objects, with different expiration times. It just seems like a future mess waiting to happen.
2. Contex has only cancelled/deadline variables. Also, we don't use context cancellation for user cancellation (we use w.Cancel channel). So that would need to be refactored as well.

I don't see how modifying Daisy to support context timeout is free: we are just shifting work from wrappers to Daisy with the amount of effort being currently unknown. We would be investing probably a week of extra work and possibly introducing bugs for benefits that are not really clear. 

What is your main concern with this PR? That timeout is not "free"?

cli_tools/gce_vm_image_import/importer/data_disk_processor.go
zoran15: I've implemented the option 1. from Eric's comment: if timeout happens during data disk processor step, wrapper will let it finish and not return timeout error. Since image create takes around a minute in my tests, with the extra buffer we have to perform cleanup, this should not lead to many terminated imports and will provide better customer experience with less failures that could have succeeded. I'll monitor the percentage of STARTED imports (which would indicate Cloud Build times out before the wrapper) and adjust the logic if needed.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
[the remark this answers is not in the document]
zoran15: Removed pointer receivers on processors.

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: > What is your main concern with this PR? That timeout is not "free"?

As a new teammate, I was intimidated by the [previous iteration](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/v20191206/cli_tools/gce_vm_image_import/importer/importer.go#L243), and my goal was to help future readers. The tactic was twofold. First streamline the program's entry, keeping it focused on business steps, while keeping mechanics (logging, timeout, retry) orthogonal.  Second, use idiomatic Go where possible.

That said, I've inferred that you are happy with the approach in this PR. To move us forward, I've given a few concrete review comments. Once those are done, I'd say it LGTM :)

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -81,42 +92,79 @@ func (i importer) Run(ctx context.Context) (loggable service.Loggable, err error
...
+	i.processor, err = i.processorProvider.provide(i.pd)
 	if err != nil {
 		return err
 	}
+	return i.runStep(ctx, func(ctx context.Context) error { return i.processor.process(ctx) }, i.processor.cancel, i.processor.traceLogs)
+}
+
+func (i *importer) runStep(ctx context.Context, step func(context.Context) error, cancel func(string) bool, getTraceLogs func() []string) (err error) {
+	e := make(chan error)
+	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
zoran15: ctx is needed in runStep as it's used for timeout but I've removed it from inflater/processors.

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -58,20 +63,26 @@ type importer struct {
 	processorProvider processorProvider
 	traceLogs         []string
 	diskClient        diskClient
+	timeout           time.Duration
+	processor         processor
```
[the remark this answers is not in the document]
zoran15: Hang onto what?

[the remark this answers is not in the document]
EricEdens: > Hang onto what?

Good question - the reference to `processor`

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -180,39 +328,74 @@ func (m *mockProcessorProvider) provide(pd persistentDisk) (processor, error) {
...
 type mockProcessor struct {
-	serialLogs   []string
-	err          error
-	interactions int
+	serialLogs     []string
+	err            error
+	interactions   int
+	processingTime time.Duration
+	processingChan chan bool
+	cancelError    error
```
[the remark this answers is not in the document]
zoran15: Removed cancelError

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -180,39 +328,74 @@ func (m *mockProcessorProvider) provide(pd persistentDisk) (processor, error) {
...
+	return true
+}
+
 type mockDiskClient struct {
-	interactions       int
-	project, zone, uri string
+	interactions                 int
+	project, zone, uri, diskName string
+	disk                         *compute.Disk
+	getDiskError                 error
```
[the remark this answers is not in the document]
zoran15: removed getDiskError

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -46,12 +47,18 @@ type daisyInflater struct {
 	wf              *daisy.Workflow
 	inflatedDiskURI string
 	serialLogs      []string
+	diskClient      diskClient
+	project         string
+	zone            string
```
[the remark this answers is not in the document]
zoran15: removed project and zone

cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -81,6 +85,65 @@ func TestRun_DeleteDisk(t *testing.T) {
 	assert.Equal(t, diskURI, mockDiskClient.uri)
 }
 
+func TestRun_DontDeleteDiskIfDoesntExist(t *testing.T) {
+	project := "project"
+	zone := "zone"
+	diskURI := "uri"
+	mockDiskClient := mockDiskClient{
+		disk: nil,
+	}
```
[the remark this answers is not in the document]
zoran15: I've updated all the disk deletion test to reflect the current logic where getDisk is not used.

cli_tools/gce_vm_image_import/importer/importer.go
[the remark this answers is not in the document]
zoran15: Removed `importer.processor`, thanks for spotting that.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -46,12 +47,18 @@ type daisyInflater struct {
...
-func (inflater daisyInflater) inflate(ctx context.Context) (persistentDisk, error) {
-	err := inflater.wf.Run(ctx)
+func (inflater *daisyInflater) inflate(ctx context.Context) (persistentDisk, error) {
+	err := inflater.wf.Run(context.Background())
+
 	if err != nil {
-		return persistentDisk{}, err
+		return persistentDisk{
+			uri: inflater.inflatedDiskURI,
+		}, err
```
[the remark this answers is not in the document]
zoran15: PD is returned here only to allow for cleanup on timeout if a disk actually got created. Do you see a use case where we care about disk sizes for timed out imports?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -84,7 +89,7 @@ func newBootableDiskProcessor(args ImportArguments, pd persistentDisk) (processo
 		return nil, err
 	}
 
-	return bootableDiskProcessor{
+	return &bootableDiskProcessor{
```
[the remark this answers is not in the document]
zoran15: That was my original goal. Do you still have concerns with pointer use even when used via interface references?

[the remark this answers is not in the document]
EricEdens: I don't think I have concerns, just curious!

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
EricEdens: If I were investigating timeouts during inflate, one thing I'd look at the filetype and file sizes to see if there's a correlation.  No?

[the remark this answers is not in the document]
zoran15: Changed to include all the additional info even if error is thrown during inflation

PR #1267, opened 2020-06-22
Daisy steps to respect context being canceled.

zoran15:


daisy/workflow.go
```diff
@@ -670,6 +670,13 @@ func (w *Workflow) runStep(ctx context.Context, s *Step) DError {
 	select {
 	case err := <-e:
 		return err
+	case <-ctx.Done():
```
dntczdx, line 673: it seems that my terminal is dead when I press ctrl+c during running daisy cli today, is this PR fixing that?

zoran15: No. This is to support using context expiration for timeout handling, related to #1263

Issue #1266, opened 2020-06-19
DetachDisks fails: cannot detach disk in project ""


tilmangiese:
It looks like DetachDisks has been completely broken for a few months.

Symptom:
```
[Daisy] Validating workflow "shrink-image"
[shrink-image]: 2020-06-19T21:44:07Z Validating workflow
[shrink-image]: 2020-06-19T21:44:07Z Validating step "setup"
[shrink-image]: 2020-06-19T21:44:07Z Validating step "run"
[shrink-image]: 2020-06-19T21:44:08Z Validating step "wait-finished"
[shrink-image]: 2020-06-19T21:44:08Z Validating step "wait-vm-shutdown"
[shrink-image]: 2020-06-19T21:44:08Z Validating step "detach-disk"
[shrink-image]: 2020-06-19T21:44:08Z Error validating workflow: step "detach-disk" validation error: Multiple errors:
* cannot detach disk in project "" from instance in project "<redacted>": "output"
* cannot detach disk in zone "" from instance in zone "us-central1-a": "output"
[Daisy] Error validating workflow "shrink-image": step "detach-disk" validation error: Multiple errors:
* cannot detach disk in project "" from instance in project "<redacted>": "output"
* cannot detach disk in zone "" from instance in zone "us-central1-a": "output"
```

Last working version: c776c023bd4e2908d5194634e7740f66119ec011 (March 18)
First failing commit: d4328ccb0e8120965ccf35f087db7d7132ccbc6b (March 24)

The detach step is very straight forward:

```json
    "detach-disk": {
      "DetachDisks": [
        {
          "DeviceName": "output",
          "Instance": "shrinker-vm"
        }
      ]
    },
```

Full workflow file attached: [shrink_image.wf.json.txt](https://github.com/GoogleCloudPlatform/compute-image-tools/files/4806854/shrink_image.wf.json.txt)

dntczdx:
Looking into it

dntczdx:
https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1269 to fix it

dntczdx:
It works fine from my side now, please verify, thanks!
@tilmangiese

tilmangiese:
Works as expected. Thank you!

PR #1269, opened 2020-06-23
[daisy] fixed 2 DetachDisks problem

dntczdx:
Issue 1: https://github.com/GoogleCloudPlatform/compute-image-tools/issues/1266
It's because a disk regex should be used to parse the disk link other than a device regex.
Issue 2: If a disk is attached with a customized device name (different from disk name), it can't be detached.
It's because from the very beginning, the daisy DetachDisks assumed device name == disk name, which is incorrect. In order to fix it, some logic are added to accept extra params in some functions. Also renamed some vars to avoid the confusion (for example, does dName mean deviceName or diskName?).

Added a e2e test to cover both issues.


daisy/disk.go
```diff
@@ -145,6 +145,7 @@ func (d *Disk) validate(ctx context.Context, s *Step) DError {
 }
 
 type diskAttachment struct {
+	diskName           string
```
dntczdx, line 148: this field is added so that we can find disk name by device name from attachments

daisy/disk.go
```diff
@@ -188,7 +189,7 @@ func (dr *diskRegistry) detachHelper(dName, iName string, isAttached bool, s *St
 		return nil
 	}
 
-	pre := fmt.Sprintf("step %q cannot detach disk %q from instance %q", s.name, dName, iName)
+	pre := fmt.Sprintf("step %q cannot detach disk with device name = %q from instance %q", s.name, dName, iName)
```
zoran15, line 192: minor: remove `=`. If you want to highlight it, consider putting it in quotes.

Also, dName can be diskName or deviceName. I thought you wanted to get rid of this var name?

daisy/disk.go
```diff
@@ -212,17 +213,17 @@ func (dr *diskRegistry) detachHelper(dName, iName string, isAttached bool, s *St
 }
 
 // registerAttachment is called by Instance.regCreate and AttachDisks.validate and marks a disk as attached to an instance by Step s.
-func (dr *diskRegistry) regAttach(dName, iName, mode string, s *Step) DError {
+func (dr *diskRegistry) regAttach(deviceName, diskName, iName, mode string, s *Step) DError {
 	dr.mx.Lock()
 	defer dr.mx.Unlock()
 
-	pre := fmt.Sprintf("step %q cannot attach disk %q to instance %q", s.name, dName, iName)
+	pre := fmt.Sprintf("step %q cannot attach disk %q to instance %q", s.name, deviceName, iName)
```
zoran15, line 220: The message is not in sync with the one above which uses "disk with device name %q". Error messages should be consistent.

daisy/instance.go
```diff
@@ -215,15 +215,19 @@ func (i *Instance) setSourceMachineImage(machineImage string) {}
...
 				continue
 			}
-			dName = d.InitializeParams.DiskName
+			diskName = d.InitializeParams.DiskName
 		}
-		errs = addErrs(errs, ir.w.disks.regAttach(dName, name, d.Mode, s))
+		deviceName := d.DeviceName
+		if deviceName == "" {
+			deviceName = diskName
+		}
```
zoran15, line 229: Should this logic maybe go to regAttach in case it's called from other places in the code base?

daisy/instance.go
```diff
@@ -322,15 +326,19 @@ func (i *InstanceBeta) setSourceMachineImage(machineImage string) {
...
 			if parts["disktype"] == "local-ssd" {
 				continue
 			}
-			dName = d.InitializeParams.DiskName
+			diskName = d.InitializeParams.DiskName
+		}
+		deviceName := d.DeviceName
+		if deviceName == "" {
+			deviceName = diskName
 		}
```
zoran15, line 340: related to the above comment, moving this to regAttach would avoid code duplication, unless there's a need for deviceName being empty in some scenarios

daisy/resource_registry.go
```diff
@@ -284,3 +284,15 @@ func (dr *diskRegistry) regUseDeviceName(deviceName, project, zone, instance str
 	res.users = append(res.users, s)
 	return res, isAttached, nil
 }
+
+func (dr *diskRegistry) findDiskResourceByDeviceName(deviceName, instance string) (*Resource, DError) {
+	attachmentMap, ok := dr.attachments[deviceName]
+	if !ok {
+		return nil, Errf("missing registered disk attachment for device name '%v'", deviceName)
```
zoran15, line 291: Any reason to switch away from the original error format (the one that uses dr.typeName)?

daisy/step_attach_disks.go
```diff
@@ -85,8 +85,12 @@ func (a *AttachDisks) validate(ctx context.Context, s *Step) (errs DError) {
 		ad.zone = disk["zone"]
 
 		// Register disk attachments.
-		errs = addErrs(errs, s.w.instances.w.disks.regAttach(ad.Source, ad.Instance, ad.Mode, s))
-
+		// If device name is not given explicitly, its device name will be the same as disk name.
+		deviceName := ad.Source
+		if ad.DeviceName != "" {
+			deviceName = ad.DeviceName
+		}
```
zoran15, line 92: ad.Source is disk name? Shouldn't the order of initialization be reversed here then? First set ad.DeviceName, if it's empty, set ad.Source? Or this is a different use case?

This is the third place where this device/disk default logic is implemented, so another reason to move it to regAttach, if it's possible to be generalized.

zoran15:
Approved by mistake.

daisy/disk.go
dntczdx: done

daisy/disk.go
dntczdx: Can't 100% sync with above actually, since detach is based on 'device name' while attach is based on 'disk name'. I should put 'diskName' here.

daisy/instance.go
dntczdx: That's a great suggestion, done

daisy/instance.go
dntczdx: done

daisy/step_attach_disks.go
dntczdx: done

daisy/resource_registry.go
dntczdx: After changed to the new lookup method, the original error msg is not a correct description any more. The new msg described what is missed.

dntczdx:
Let's bake it in test env for some days

PR #1272, opened 2020-06-25
[windows-upgrade e2e] add gcloud e2e tests

dntczdx:


cli_tools_e2e_test/common/utils/utils.go
```diff
@@ -131,9 +131,13 @@ func runCliTool(logger *log.Logger, testCase *junitxml.TestCase, cmdString strin
 }
 
 // RunTestCommand runs given test command
-func RunTestCommand(cmd string, args []string, logger *log.Logger, testCase *junitxml.TestCase) bool {
+func RunTestCommand(cmd string, args []string, logger *log.Logger, testCase *junitxml.TestCase, allowError bool) bool {
```
EricEdens, line 134: `allowError` is a big vague. Does this mean the test case is asserting that an error should occur? If it means "maybe an error can occur, and that's okay" -- what's an example of this?

dntczdx: In the past, we run a command and always expect success (return value=0).
Now, my new test needs to run a command and I know it will fail - which is what I am going to test, so it shouldn't mark the test case fail.
Maybe let me make a new function called "RunTestCommandIgnoringError".

cli_tools_e2e_test/common/utils/utils.go
```diff
@@ -139,6 +139,15 @@ func RunTestCommand(cmd string, args []string, logger *log.Logger, testCase *jun
 	return true
 }
 
+// RunTestCommandIgnoringError runs given test command. The test case won't be marked as fail even error happens.
+func RunTestCommandIgnoringError(cmd string, args []string, logger *log.Logger, testCase *junitxml.TestCase) bool {
```
EricEdens, line 143:
```suggestion
// RunTestCommandExpectingError runs given test command. Passes when the command fails.
func RunTestCommandExpectingError(cmd string, args []string, logger *log.Logger, testCase *junitxml.TestCase) bool {
```

cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
```diff
@@ -319,7 +394,15 @@ func runTest(ctx context.Context, image string, args []string, testType utils.CL
 			success = true
 		}
 	} else {
-		success = utils.RunTestForTestType(cmd, args, testType, logger, testCase)
+		isLatestGcloud := true
+		if testType == utils.GcloudProdWrapperLatest {
+			isLatestGcloud = false
+		}
```
EricEdens, line 400:
```suggestion
		isLatestGcloud := testType == utils.GcloudProdWrapperLatest 
```

zoran15: Should be negated

dntczdx: done

cli_tools_e2e_test/common/utils/utils.go
dntczdx: I don't always expect error, instead, it can be either success or fail.

PR #1273, opened 2020-06-25
Ensure image import respects the global timeout.

EricEdens:
An alternative approach to #1263.


EricEdens:
closing since this isn't intended for merging, just a demo

PR #1274, opened 2020-06-26
refactored flag value into common utils

fiona-liu:
moved implementations of flag.Value from image import into common utils

### DIFF
--- a/cli_tools/common/utils/flags/key_value_flag.go
+++ b/cli_tools/common/utils/flags/key_value_flag.go
@@ -1,0 +1,40 @@
+package flags
+
+import (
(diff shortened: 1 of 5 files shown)


cli_tools/common/utils/flags/key_value_flag.go
```diff
@@ -0,0 +1,40 @@
+package flags
```
EricEdens, line 1: It looks like a few of these files need the Apache license header

fiona-liu: added, thanks!

PR #1275, opened 2020-06-26
[windows upgrade e2e] delete instance after test

dntczdx:
CPU resource will be exhausted if we don't delete the test instance proactively. Even the cleanup tool will do the cleanup, the garbage is still too much and it has caused the quota problem yesterday.

### DIFF
--- a/cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
+++ b/cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
@@ -542,9 +542,10 @@ func verifyCleanup(instance *computeUtils.Instance, testCase *junitxml.TestCase,
 }
 
 func cleanupTestInstance(project, zone, instanceName string, logger *log.Logger, testCase *junitxml.TestCase) {
-	utils.RunTestCommandIgnoringError("gcloud", []string{
+	// Run gcloud to delete the instance, ignoring error.
+	utils.RunCliTool(logger, testCase, "gcloud", []string{
 		"compute", "instances", "delete", "--quiet",
 		fmt.Sprintf("--zone=%v", zone),
 		fmt.Sprintf("--project=%v", project), instanceName,
-	}, logger, testCase)
+	})
 }
--- a/cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
+++ b/cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
@@ -322,6 +322,17 @@ func runTest(ctx context.Context, image string, args []string, testType utils.CL
 		return
 	}
 
+	defer func() {
+		// delete the test instance when test is done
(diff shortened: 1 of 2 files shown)


cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
```diff
@@ -322,6 +322,17 @@ func runTest(ctx context.Context, image string, args []string, testType utils.CL
...
+		// delete the test instance when test is done
+		if !utils.RunTestCommandIgnoringError("gcloud", []string{
+			"compute", "instances", "delete",  "--quiet",
+			fmt.Sprintf("--zone=%v", testProjectConfig.TestZone),
+			fmt.Sprintf("--project=%v", testProjectConfig.TestProjectID), instanceName,
+		}, logger, testCase) {
+			return
+		}
+	} ()
+
```
zoran15, line 335: How does this actually delete test instances?

cli_tools_e2e_test/gce_windows_upgrade/test_suites/windows_upgrade/windows_upgrade_test_suite.go
```diff
@@ -322,6 +322,17 @@ func runTest(ctx context.Context, image string, args []string, testType utils.CL
...
 
+	defer func() {
+		// delete the test instance when test is done
+		if !utils.RunTestCommandIgnoringError("gcloud", []string{
+			"compute", "instances", "delete",  "--quiet",
+			fmt.Sprintf("--zone=%v", testProjectConfig.TestZone),
+			fmt.Sprintf("--project=%v", testProjectConfig.TestProjectID), instanceName,
+		}, logger, testCase) {
+			return
+		}
```
zoran15, line 333: Is there a more elegant way to delete instances, e.g. by calling compute client directly? I found this call very confusing as it appears it's running a test, but it's actually running `gcloud compute instances delete`. At minimum, this should be extracted in an appropriately named func, but it would be nicer not to yse RunTest syntax for cleanup tasks.

dntczdx: Thank you for the advice, I wrapped it with a "cleanupTestInstance" func.
My consideration here is that since it's created by gcloud successfully, it very probably can be deleted by gcloud successfully . Using a compute client requires us to init it by align credentials, which have a chance to make unnecessary mistake.

zoran15: OK. Just avoid calling RunTest if possible. RunCommand or similar is fine.

zoran15: Maybe export runCliTools from utils and use that?

dntczdx: Avoided in the main flow. The func name self explained the purpose.

dntczdx: done

PR #1277, opened 2020-06-27
An image import suite focused on translation logic.

EricEdens:
Historically, the majority of import tests are run as daisy integration tests, and therefore bypass the wrapper logic. This change adds a suite to the cli_tools_e2e_test framework that focuses on import logic. The existing e2e suite is recharacterized as "CLI tests", since it focuses on the API contracts of the CLI tools.

Key features, compared to existing solutions for image import tests:
1. Support for negative tests, which verify the output of error messages
1. Efficient test representation (compared to an entire JSON document)

To keep the PR simple, I added two tests to illustrate how the system works. I'll do a subsequent PR that migrates the existing integ tests in `daisy_integration_tests` to here.

Lastly, this adds some utilities to convert between daisy logs, standard library logs, and io.Writers.

### DIFF
--- a/cli_tools_e2e_test/common/logging/logging.go
+++ b/cli_tools_e2e_test/common/logging/logging.go
@@ -1,0 +1,69 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package logging
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"log"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// AsDaisyLogger returns a daisy.Logger that writes to a log.Logger.
+func AsDaisyLogger(logger *log.Logger) daisy.Logger {
+	return goLogToDaisyLog{logger: logger}
+}
+
+// AsWriter returns an io.Writer that writes to a log.Logger.
+func AsWriter(logger *log.Logger) io.Writer {
+	return goLogToIoWriter{logger: logger}
+}
+
+// goLogToDaisyLog is an implementation of daisy.Logger that writes to a log.Logger backend.
+type goLogToDaisyLog struct {
+	logger *log.Logger
+}
+
+func (l goLogToDaisyLog) WriteLogEntry(e *daisy.LogEntry) {
+	l.logger.Printf(e.Message)
+}
+
+func (l goLogToDaisyLog) WriteSerialPortLogs(w *daisy.Workflow, instance string, buf bytes.Buffer) {
+	// no-op
+}
+
+func (l goLogToDaisyLog) ReadSerialPortLogs() []string {
+	// no-op
+	return nil
+}
+
+func (l goLogToDaisyLog) Flush() {
+	// no-op
+}
+
+// goLogToDaisyLog is an implementation of io.Writer that writes to a log.Logger backend.
+type goLogToIoWriter struct {
+	logger *log.Logger
+}
+
+func (l goLogToIoWriter) Write(p []byte) (n int, err error) {
+	scanner := bufio.NewScanner(bytes.NewReader(p))
+	for scanner.Scan() {
+		l.logger.Println(scanner.Text())
+	}
+	return len(p), nil
+}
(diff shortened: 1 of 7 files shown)


cli_tools_e2e_test/gce_image_import_export/test_suites/import/post_translate_test.sh
```diff
@@ -0,0 +1 @@
+../../../../daisy_integration_tests/scripts/post_translate_test.sh
```
dntczdx, line 1: This is strange, you still need to copy both "cli_tools_e2e_test/.../post_translate_test.sh" and "daisy_integration_tests/.../post_translate_test.sh" to the docker, and need to maintain the relative path in the docker. Why not just copy ""daisy_integration_tests/.../post_translate_test.sh" in the docker build file?

EricEdens: Sounds good to me; removed the symlink in the latest commit.

The original purpose was for running the tests locally, but I've found it much easier (as you've suggested) to just use the docker container for running the tests.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -32,11 +32,11 @@ import (
 )
 
 const (
-	testSuiteName = "ImageImportTests"
+	testSuiteName = "CLI"
```
dntczdx, line 35: It's better to still have "import" keyword since export and import will be executed in one batch.

EricEdens: sgtm - updated to ImageImportCLI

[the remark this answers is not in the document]
EricEdens: sgtm - updated to ImageImportCLI

cli_tools_e2e_test/common/logging/logging.go
```diff
@@ -0,0 +1,69 @@
...
+	return goLogToDaisyLog{logger: logger}
+}
+
+// AsWriter returns an io.Writer that writes to a log.Logger.
+func AsWriter(logger *log.Logger) io.Writer {
+	return goLogToIoWriter{logger: logger}
+}
+
+// goLogToDaisyLog is an implementation of daisy.Logger that writes to a log.Logger backend.
+type goLogToDaisyLog struct {
```
zoran15, line 37: shouldn't the name be reversed, daisy logs (API) writing to Go log output?

cli_tools_e2e_test/common/logging/logging.go
```diff
@@ -0,0 +1,69 @@
...
+	"bufio"
+	"bytes"
+	"io"
+	"log"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// AsDaisyLogger returns a daisy.Logger that writes to a log.Logger.
+func AsDaisyLogger(logger *log.Logger) daisy.Logger {
```
zoran15, line 27: This is kind of minor right now as it's used in tests only, but, just looking at the method/package names, it's not clear that this returns a logger that writes to log.Logger, since it's in logging package (that's not related to log package). If it was log.AsDaisyLogger() that would be clear.

cli_tools_e2e_test/common/logging/logging.go
```diff
@@ -0,0 +1,69 @@
...
+func (l goLogToDaisyLog) ReadSerialPortLogs() []string {
+	// no-op
+	return nil
+}
+
+func (l goLogToDaisyLog) Flush() {
+	// no-op
+}
+
+// goLogToDaisyLog is an implementation of io.Writer that writes to a log.Logger backend.
```
zoran15, line 58: incorrect struct name in comment?

cli_tools_e2e_test/common/logging/logging.go
```diff
@@ -0,0 +1,69 @@
...
+type goLogToIoWriter struct {
```
zoran15, line 59: Similar to the comment above, isn't this name in reverse: IO writer that writes to Go Log?

cli_tools_e2e_test/common/logging/logging.go
EricEdens: The idea is that it *converts* a Go log to a Daisy log -- given a concrete `log.Logger`, you get an object that you can use wherever a daisy logger is required.

cli_tools_e2e_test/common/logging/logging.go
EricEdens: I'd agree -- this isn't intended to be a battle-tested API that's usable throughout the codebase, but rather a pragmatic set of shims that translate between a few different logging and IO interfaces. Regarding the package name `logging`, I hadn't intended for it to refer to any particular logger, but rather to logging in general.

cli_tools_e2e_test/common/logging/logging.go
EricEdens: thanks, updated

cli_tools_e2e_test/common/logging/logging.go
EricEdens: Similar to above, it's converting a go log to an IO writer.

cli_tools_e2e_test/common/logging/logging.go
zoran15: Still seems backwards to me, but the docs should clear any confusion

PR #1280, opened 2020-07-07
daisy/: added variable parsing to resize-disk.

hegao12:
Currently, the resize-disk commmand can only take integer input and
cannot parse variables in a workflow. In order to allow variables in
this command, a new field "SizeGbVar string" is added, and then its
value is converted into int64 and stored in the original "SizeGb int64"
field. In this way, everything else except the ResizeDisk struct fields and
populate function remains untouched.

### DIFF
--- a/daisy/step_resize_disks.go
+++ b/daisy/step_resize_disks.go
@@ -17,6 +17,7 @@ package daisy
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"sync"
 
 	"google.golang.org/api/compute/v1"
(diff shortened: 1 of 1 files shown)


daisy/step_resize_disks.go
```diff
@@ -30,13 +31,23 @@ type ResizeDisk struct {
 	compute.DisksResizeRequest
 	// Name of the disk to be resized
 	Name string
+	// SizeGb in string format to parse variables in workflow
+	SizeGbVar string "json:\"sizeGb,omitempty\""
```
zoran15, line 35: You can call it `SizeGb` (the same as the int field in DiskResizeRequest). In populate, you can refer to the int field as `DiskResizeRequest.SizeGb`.

hegao12: If so, all the others places that uses `SizeGb` will need to be changed to `DiskResizeRequest.SizeGb`. I'm not sure about all the places where it is used.

zoran15: In our code base it's only step_resize_disks and associated tests. They already reference it as DiskResizeRequest.SizeGb.

For other uses, since Daisy now uses Go modules, it will still be possible to reference the old version of the code by third parties.

hegao12: After I renamed it as `SizeGb`, the input was unmarshaled to the int64 `SizeGb` instead of the string one. The definition of the int64 one is inside of "google.golang.org/api/compute/v1", which I cannot change. So I think a different variable name is needed here.

hegao12: It's renamed now. Commit is amended.
BTW, if this is merged, when should I expect a new release of Daisy?

zoran15: This will be in the :latest build of Daisy within 10 minutes. Releases are manual so the earliest we can do it is tomorrow afternoon, once all the tests run.

PR #1282, opened 2020-07-08
[New VMDK API adoption] verify the new API on PROD

dntczdx:
This PR is a preparation for:
1. the new VMDK API's GA rollout.
2. verification of the adoption of new API.
It produces a shadow disk by new API in inflation stage, and compare the checksum of the disk content to ensure it matches the current inflation's output.

Added a "apiInflater" and use it as shadow. Added a "InflaterFacade" to coordinate the main and the shadow infalter.


cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -146,3 +294,146 @@ func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
...
+	pd.checksum, err = inflater.calculateChecksum(ctx, diskUri)
+	return pd, err
+}
+
+// run a workflow to calculate checksum, and cleanup the disk
+func (inflater apiInflater) calculateChecksum(ctx context.Context, diskUri string) (string, error) {
+	w := daisy.New()
+	w.Name = "shadow-disk-checksum"
+	checksumScript := checksumScriptConst
+	w.Steps = map[string]*daisy.Step{
```
EricEdens, line 352: Looks like the apiInflater has a lot of code; move to its own file? What do you think? That'd be nice since it'd also give you your own test file, which will be nice to avoid namespacing test names.

daisy_workflows/image_import/import_image.sh
```diff
@@ -141,6 +141,21 @@ function serialOutputKeyValuePair() {
   echo "<serial-output key:'$1' value:'$2'>"
 }
 
+function diskChecksum() {
+  set +x
+  CHECK_DEVICE=sdc
+  BLOCK_COUNT=$(cat /sys/class/block/$CHECK_DEVICE/size)
```
EricEdens, line 147: This duplicated code seems reasonable given that this is an experiment. Can you add a comment pointing to the duplicated version (so that we know to remove it, and so that we remember to keep them in sync if they change)

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -117,7 +117,9 @@ func (i *importer) cleanupDisk() {
 }
 
 func (i importer) buildLoggable() service.Loggable {
-	return service.SingleImageImportLoggable(i.pd.sourceType, i.pd.sourceGb, i.pd.sizeGb, i.traceLogs)
+	return service.SingleImageImportLoggable(i.pd.sourceType, i.pd.sourceGb, i.pd.sizeGb,
+		i.pd.matchResult, int64(i.pd.inflationTime.Seconds()),
+		int64(i.pd.shadowInflationTime.Seconds()), i.traceLogs)
```
EricEdens, line 122: Do you envision any of these metrics living beyond your test phase?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -40,6 +43,119 @@ type inflater interface {
...
+	var shadowPd persistentDisk
+	var shadowErr error
+	go func() {
+		shadowPd, shadowErr = facade.shadowInflater.inflate(ctx)
+		if shadowErr != nil {
+			inflaterChan <- sigShadowInflaterErr
+		} else {
+			inflaterChan <- sigShadowInflaterDone
+		}
+	}()
```
EricEdens, line 92: As I'm reading this, I'm curious about what happens if one or both inflaters keep going forever. My guess is that the children inflaters are responsible for managing their own lifecycle?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -40,6 +43,119 @@ type inflater interface {
...
+
+	// Return early if main inflater finished first.
+	result := <-inflaterChan
+	if result == sigMainInflaterDone || result == sigMainInflaterErr {
+		if result == sigMainInflaterDone {
```
EricEdens, line 97: It's *possible* that the shadow inflater could be blocking here, waiting to write its result to the channel, right? I'm not sure it's a big deal, since I think you're assuming that the API call will be much faster than the daisy worklfow.

I guess they'd both have to finish simultaneously to allow that to happen

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -40,6 +43,119 @@ type inflater interface {
...
+		} else {
+			facade.compareWithShadowInflater(ctx, &pd, &shadowPd)
+		}
+	} else if result == sigShadowInflaterErr && mainResult == sigMainInflaterDone {
+		if isCausedByUnsupportedFormat(shadowErr) {
+			pd.matchResult = "Shadow inflater doesn't support the format while main inflater supports"
+		} else {
+			pd.matchResult = "Shadow inflater failed while main inflater succeeded"
+		}
+	}
```
EricEdens, line 117: Are we keeping enough logs to help the PD team debug the native VMDK failures?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -60,19 +177,50 @@ func (inflater daisyInflater) inflate(ctx context.Context) (persistentDisk, erro
...
-	sourceGb   int64
-	sourceType string
+	uri                 string
+	sizeGb              int64
+	sourceGb            int64
+	sourceType          string
+	checksum            string
+	inflationTime       time.Duration
+	shadowInflationTime time.Duration
+	matchResult         string
```
EricEdens, line 199: These are temporary for the experiment, right? If so, can you add a comment mentioning that?

daisy/compute/compute.go
```diff
@@ -241,7 +245,17 @@ func shouldRetryWithWait(tripper http.RoundTripper, err error, multiplier int) b
...
-	o := []option.ClientOption{option.WithScopes(compute.ComputeScope, compute.DevstorageReadWriteScope)}
+	// Set these scopes to be align with compute.NewService
+	o := []option.ClientOption{
+		option.WithScopes(
+			compute.CloudPlatformScope,
+			compute.ComputeScope,
+			compute.ComputeReadonlyScope,
+			compute.DevstorageFullControlScope,
+			compute.DevstorageReadOnlyScope,
+			compute.DevstorageReadWriteScope,
```
EricEdens, line 256: I'm a bit of a noob in terms of scopes; if the account executing Daisy doesn't have a permission enabled (say: DevstorageFullControlScope), would there be a failure due to adding these new scopes?

In other words, is there a possibility that pushing this to Daisy will cause new breakages?

daisy/compute/compute.go
```diff
@@ -261,7 +275,15 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (Client, error)
 	if ep != "" {
 		rawBetaService.BasePath = ep
 	}
-	c := &client{hc: hc, raw: rawService, rawBeta: rawBetaService}
+	rawAlphaService, err := computeAlpha.New(hc)
+	if err != nil {
+		return nil, fmt.Errorf("alpha compute client: %v", err)
+	}
```
EricEdens, line 281: I recall @adjackura having some thoughts regarding Alpha APIs.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -146,3 +294,146 @@ func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
...
+		inflationTime: time.Since(startTime),
+	}
+
+	inflater.addTraceLog("Started checksum calculation.")
+	pd.checksum, err = inflater.calculateChecksum(ctx, diskUri)
+	return pd, err
+}
+
+// run a workflow to calculate checksum, and cleanup the disk
+func (inflater apiInflater) calculateChecksum(ctx context.Context, diskUri string) (string, error) {
```
EricEdens, line 348: The 'cleanup disk' comment is confusing since the caller has a `defer` to delete the disk.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -146,3 +294,146 @@ func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
...
+		},
+	}
+	w.Dependencies = map[string][]string{
+		"create-instance": {"create-disks"},
+		"wait-for-checksum": {"create-instance"},
+	}
+
+	daisycommon.SetWorkflowAttributes(w, inflater.args.Project, inflater.args.Zone, inflater.args.ScratchBucketGcsPath,
+		inflater.args.Oauth, "20m", inflater.args.ComputeEndpoint, inflater.args.GcsLogsDisabled,
+		inflater.args.CloudLogsDisabled, inflater.args.StdoutLogsDisabled)
```
EricEdens, line 414: I think '20m' means that you're allowing a maximum of 20 minutes to run the checksum workflow; is that right?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -146,3 +294,146 @@ func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
...
+	checksumScriptConst = `
+		function serialOutputKeyValuePair() {
+			echo "<serial-output key:'$1' value:'$2'>"
+		}
+
+		CHECK_DEVICE=sdb
+		BLOCK_COUNT=$(cat /sys/class/block/$CHECK_DEVICE/size)
+	
+		# Check size = 200000*512 = 100MB
+		CHECK_COUNT=200000
```
EricEdens, line 432: This means that you calculate the checksum of the first 100MB?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -60,19 +177,50 @@ func (inflater daisyInflater) inflate(ctx context.Context) (persistentDisk, erro
...
+	return inflaterFacade{
+		mainInflater:   di,
+		shadowInflater: ai,
+	}, nil
+}
+
+func createApiInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient *storage.Client) inflater {
+	if isImage(args.Source) {
+		return nil
+	}
```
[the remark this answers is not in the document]
dntczdx: sgtm, modified

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -146,3 +294,146 @@ func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
...
+	inflater.serialLogs = append(inflater.serialLogs, l)
+}
+
+func (inflater apiInflater) inflate(ctx context.Context) (persistentDisk, error) {
+	startTime := time.Now()
+	diskName := fmt.Sprintf("shadow-disk-%v", inflater.args.ExecutionID)
+	cd := computeAlpha.Disk{
+		Name:                diskName,
+		SourceStorageObject: inflater.args.SourceFile,
+	}
```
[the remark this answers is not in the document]
dntczdx: It will fail. Let me add a check to handle "alpha API not enabled" separately.
Currently, the plan is (confirmed with PD): 
Stage 1. land the PR so our e2e  (which uses the 'latest') can use the shadow verification; run for a couple of days so PD team collected enough data point so they can rollout to GA;
Stage 2. we change our code to call to GA API so the shadow verification can be done towards all live traffic;
Stage 3. Switch apiInflater as main. Stop shadow logic.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -146,3 +294,146 @@ func getDisk(workflow *daisy.Workflow, diskIndex int) *daisy.Disk {
...
+	err := inflater.computeClient.CreateDiskAlpha(inflater.args.Project, inflater.args.Zone, &cd)
+	if err != nil {
+		return persistentDisk{}, err
+	}
+	defer func() {
+		// Cleanup the shadow disk ignoring error
+		inflater.computeClient.DeleteDisk(inflater.args.Project, inflater.args.Zone, cd.Name)
+	}()
```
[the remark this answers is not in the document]
dntczdx: No difference. I reserved it for adding any extra logic if needed, which seems not happening. Let me remove it.

daisy/compute/compute.go
dntczdx: Alpha API won't work if they are not enabled. In our case, it's expected. It happens to only shadow logic and has no side effect.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: Deleted.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: Yes. Added comment to explain.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: As introduced in the design doc, we calculate 0~0.1GB, 9.9GB~10GB, 99.9GB~100GB, and max-0.1GB ~ max GB

cli_tools/gce_vm_image_import/importer/importer.go
dntczdx: No, I will delete them. Just filed  b/161177634 to track.

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
zoran15: At what point do we release to the general public? after stage 3?

cli_tools/gce_vm_image_import/importer/inflater.go
zoran15: Since shadow logic will not be used in customer projects, is it feasible to run calculate the full checksum on whole disks, since we're paying for it and any time penalty is not a big issue (unless it takes many hours to calculate)?

dntczdx: It will be used by customer in stage 2. And the penalty is high - as high as around 50% of inflating the disk in traditional way.

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
dntczdx: The shadow logic will be released to customer in stage 1.
The new API will be used in main path starting from stage 3.
BTW PD team had one more internal communication and they decided to release to beta so that they can get tested by all the customers. They will send a email to announce it.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: done

daisy_workflows/image_import/import_image.sh
dntczdx: done

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: This is a interesting topic.
For daisy inflater, it has a daisy timeout.
For API inflater, it relies on GCP operation (the long-running API handle) mechanism. We assume that waiting on 'operation' is safe and won't run forever.

We can also add an additional client-side timeout mechanism for the api inflater, however I chose not to do that, since the whole daisy workflow's timeout is based on the same assumption. For exmaple, step_create_instance. Even when daisy emit a timeout signal to the step_create_instance, it won't terminate at once. Instead, it wait for the API to finish. So, giving the API inflater a additional timeout won't obviously improve the whole import's safety. The whole import still mostly rely on the GCP API's mechanism.

In other words, one single API call is somehow 'atomic' for daisy.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: Both shadow and main inflater send a signal to inflaterChan. So nothing will be blocked:
1. If the main inflater finished first (corner case), the main thread will return without waiting for shadow.
2. If the shadow inflater finsihed first, the main thread will wait for main inflater.
Nobody will wait for shadow inflater.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: PD has their own log, everything we got here is sent from their side. We just need to let then know the rough time range.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: done

daisy/compute/compute.go
dntczdx: No. They are not related at all. This scope is just about which kind of operations can happen, it's a self-limitation from client side.

cli_tools/gce_vm_image_import/importer/api_inflater.go
```diff
@@ -0,0 +1,205 @@
...
+
+func isCausedByAlphaAPIAccess(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "Required 'Alpha Access' permission")
+}
+
+// apiInflater implements an inflater using google API. Its capability is
+// growing, depending on the API's capability.
```
EricEdens, line 45:
```suggestion
// apiInflater implements `importer.inflater` using the Compute Engine API.
```

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
EricEdens: Based on the staging description, it looks like it's not safe to deploy `compute-image-tools` while you're in stage 1? Or is there a gating mechanism that allows you to deploy in stage 1, while only running your new code in the e2e tests?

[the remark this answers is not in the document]
dntczdx: Hi Eric, no gate. There is no side effect calling alpha API for shadow - just got a failure. We will handle the failure properly.

cli_tools/gce_vm_image_import/importer/api_inflater.go
dntczdx: done

dntczdx:
Clarification for rollout stage 1:
There is no side effect for calling alpha API when it's not enabled for some cusotmers. We just received an error and will handle it properly (log as "alpha not enabled").The main path is not impacted at all.

cli_tools/gce_vm_image_import/importer/inflater.go
zoran15: Can we do the full disk checksum calculation for our internal tests and this faster logic for external customers? We should not worry about the extra time or cost in our internal projects and this will give us much higher confidence VMDK API works correctly.

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
zoran15: I thought shadow will use Beta API since PD team has said they will deploy to Beta yesterday? In that case, more customers will run this shadow logic as most (all?) have Beta API enabled. This is not a deal breaker, but it's something to take into account once releasing. 

I would echo Eric's comments about this change being in mainline potentially blocking our release. E.g. if we see any issues in our tests, you would need to revert these changes before we can make releases. We can avoid this by having the first stage being built from another branch, build docker images with a dedicated tag and use that tag in our e2e test (this will require a change to e2e tests).

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -263,6 +269,7 @@ func (l *Logger) sendLogToServerWithRetry(logExtension *ComputeImageToolsLogExte
 		}
 
 		logRequestJSON, err := l.constructLogRequest(logExtension)
+		fmt.Println(">>>>", string(logRequestJSON)) // TODO remove
```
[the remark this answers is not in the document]
dntczdx: it had been removed if you check out the latest cmmit

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -217,6 +217,12 @@ type OutputInfo struct {
 	// Serial output from worker instances; only populated
 	// if workflow failed.
 	SerialOutputs []string `json:"serial_outputs,omitempty"`
+	// Shadow disk match result for shadow disk inflater
+	ShadowDiskMatchResult string `json:"shadow_disk_match_result,omitempty"`
+	// Inflation time (seconds)
+	InflationTime []int64 `json:"inflation_time,omitempty"`
+	// Inflation time (seconds) of the shadow disk
+	ShadowInflationTime []int64 `json:"shadow_inflation_time,omitempty"`
```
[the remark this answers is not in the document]
dntczdx: 1. Currently only string and []int64 are supported by loggable.GetValues. It's not worth extracting the 1st element since it's harmless to use array here.
2. The array can be empty, but a int64 can't. We don't want a "0" value while it actually doesn't exist.

cli_tools/gce_vm_image_import/importer/inflater.go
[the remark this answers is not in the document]
dntczdx: Stage 1 will have 2 sub steps:
1.1 Alpha, which takes effect to mostly e2e tests.
1.2 beta, which takes effect to all customers. (customers don't need to enable a beta API)

So this PR will firstly run alpha.
I have tested those images which are used in e2e and there is no problem. I don't expect to see extra functionality problems in e2e, The purpose is just to make more calls for PD team.
Plus, it's on the shadow path.
So I don't feel it's risky enough to have an additional protection. The shadow-test itself is already a protection.

Before it's been to beta, we have already ran it towards e2e for some days.

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: I have manually tested full checksum for those internal tests used images.

zoran15: It would be still useful to run full checksum calculation, in case VMDK API is flaky (returns correct PD one time, not the next time) which can't be always caught by one-time manual test.

zoran15: Talked offline and agreed to run full checksum tests locally in a loop for a couple of days

cli_tools/common/utils/logging/service/literal_loggable.go
```diff
@@ -28,12 +28,17 @@ func (w literalLoggable) ReadSerialPortLogs() []string { return w.traceLogs }
...
+			importFileFormat:      fileFormat,
+			shadowDiskMatchResult: matchResult,
+		},
 		int64s: map[string][]int64{
-			sourceSizeGb: {sourceSize},
-			targetSizeGb: {resultSize},
+			sourceSizeGb:        {sourceSize},
+			targetSizeGb:        {resultSize},
+			inflationTime:       {inflationTimeInt64},
+			shadowInflationTime: {shadowInflationTimeInt64},
```
zoran15, line 41: Should add inflation type to correspond with the latest change in logs

cli_tools/gce_vm_image_import/importer/api_inflater.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return persistentDisk{}, daisy.Errf("Failed to get source file attributes: %v", err)
+	}
+	sourceFileSizeGb := (attrs.Size-1)/1073741824 + 1
+
+	diskURI := fmt.Sprintf("zones/%s/disks/%s", inflater.args.Zone, diskName)
+	pd := persistentDisk{
+		uri:           diskURI,
+		sizeGb:        cd.SizeGb,
+		sourceGb:      sourceFileSizeGb,
```
zoran15, line 99: Can this be switched to bytes instead of GB? Due to rounding, GB is not granular enough to detect small differences.

dntczdx: source file is the GCS file, it's not the inflation result. This number is just for log and won't be a part of comparison.
sizeGB is the disk size, which is always in GB granular.

cli_tools/common/utils/logging/service/literal_loggable.go
dntczdx: done

dntczdx:
Pending merge: need to wait for the log proto's rollout

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -41,6 +42,103 @@ const (
...
+	mainResult := <-inflaterChan
+	if result == sigShadowInflaterDone {
+		if mainResult == sigMainInflaterErr {
+			pd.matchResult = "Main inflater failed while shadow inflater succeeded"
+		} else {
+			facade.compareWithShadowInflater(&pd, &shadowPd)
+		}
+	} else if result == sigShadowInflaterErr && mainResult == sigMainInflaterDone {
+		if isCausedByUnsupportedFormat(shadowErr) {
+			pd.matchResult = "Shadow inflater doesn't support the format while main inflater supports"
```
zoran15, line 105:
```suggestion
			pd.matchResult = "Shadow inflater doesn't support the format while main inflater does"
```

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -41,6 +42,103 @@ const (
...
+		shadowPd, shadowErr = facade.shadowInflater.inflate()
+		if shadowErr != nil {
+			inflaterChan <- sigShadowInflaterErr
+		} else {
+			inflaterChan <- sigShadowInflaterDone
+		}
+	}()
+
+	// Return early if main inflater finished first.
+	result := <-inflaterChan
```
zoran15, line 87: should this be called shadowResult?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -81,6 +183,30 @@ type persistentDisk struct {
 	sizeGb     int64
 	sourceGb   int64
 	sourceType string
+
+	// Below fields are for experiment
```
zoran15, line 187:
```suggestion
	// Below fields are for shadow API inflation test
```

cli_tools/gce_vm_image_import/importer/inflater.go
dntczdx: this one can be either main result or shadow result. That is why I use a if-condition to distinguish them

PR #1289, opened 2020-07-16
Install OS config agent during import.

EricEdens:
This PR includes the [OS Config agent](https://cloud.google.com/compute/docs/manage-os) for imported images, instances, and machine images. To disable installation of the agent, users can specify the existing  `no-guest-environment` flag.

## What distros and versions are supported? 
All distro versions [supported by the OS Config agent](https://cloud.google.com/compute/docs/manage-os) will now include the OS config agent, if the agent is vended in GCE's official images. Specifically:
* Windows:
   * ISA: 64-bit only
   * Versions: All versions *after* Windows 2008r2
* Debian:
  * 9 (10 is not currently supported by image import)
* EL:
  * All importable versions

Ubuntu, SLES, and OpenSUSE will be added once the OS config agent is adopted by their upstream maintainers.

## Tests

The e2e tests are updated to verify the agent using the [suggested steps on their docs page](https://cloud.google.com/compute/docs/manage-os#check-install). For distro versions unsupported by the OS config agent (eg: Ubuntu 14.04), the e2e tests include an opt-out mechanism. Using an opt-out mechanism ensures that the test runs against newly-added distro versions in the future.

### How to disable installation of the OS config agent?

Specify the `no-guest-environment` flag during import. Depending on the import tool, the syntax of the flag may differ slightly.

### How do I use the OS config feature?

OS config requires that projects / instances are opted-in. If opt-in hasn't occurred, the agent runs idly, which is the same behavior as the [public images](https://cloud.google.com/compute/docs/images#os-compute-support) provided by Google. For more information, see https://cloud.google.com/compute/docs/manage-os .

### DIFF
--- a/daisy_integration_tests/scripts/post_translate_test.sh
+++ b/daisy_integration_tests/scripts/post_translate_test.sh
@@ -50,7 +50,8 @@ function wait_for_connectivity {
   exit 1
 }
 
-# Assert that a service is running; supports upstart and systemd.
+# Assert that a service is running; first trying `initctl`,
+# then trying `service`.
 function assert_running {
   local service="${1}"
   status "Checking $service"
@@ -209,10 +210,10 @@ function check_package_install {
 wait_for_connectivity
 
 # Run tests.
-check_package_install
 check_google_services
 check_google_cloud_sdk
 check_cloud_init
+check_package_install
 check_metadata_connectivity
 check_internet_connectivity
 
--- a/daisy_integration_tests/scripts/post_translate_test.sh
+++ b/daisy_integration_tests/scripts/post_translate_test.sh
@@ -15,6 +15,19 @@
 
 # Basic image tests to validate if Linux image translation was successful.
 
+
+# To disable checking for the OS config agent, add an instance metadata
+# value of osconfig_not_supported: true.
+url=http://metadata.google.internal/computeMetadata/v1/instance/attributes/osconfig_not_supported
+if command -v curl; then
+  OSCONFIG_NOT_SUPPORTED="$(curl --header "Metadata-Flavor: Google" $url)"
+elif command -v wget; then
+  OSCONFIG_NOT_SUPPORTED="$(wget --header "Metadata-Flavor: Google" -O - $url)"
+else
+  echo "TestFailed: Either curl or wget is required to run test suite."
+  exit 1
+fi
+
 FAIL=0
 FAILURES=""
 
@@ -35,6 +48,21 @@ function wait_for_connectivity {
 
   echo "FAILED: Unable to initialize network connection."
   exit 1
+}
+
+# Assert that a service is running; supports upstart and systemd.
+function assert_running {
+  local service="${1}"
+  status "Checking $service"
+  if command -v initctl; then
+    if ! initctl status "$service"; then
+      fail "$service not running."
+    fi
+  else
+    if ! service "$service" status; then
+      fail "$service not running."
+    fi
+  fi
 }
 
 function status {
@@ -71,58 +99,20 @@ function check_internet_connectivity {
 # Check Google services.
 function check_google_services {
   if [[ -f /usr/bin/google_guest_agent ]]; then
-    status "Checking google-guest-agent"
+    # Services expected when using the new guest environment (NGE).
+    assert_running google-guest-agent
+  else
+     # Services expected when using the legacy guest environment.
+    assert_running google-accounts-daemon
+    assert_running google-network-daemon
+    assert_running google-clock-skew-daemon
+  fi
 
-    # Upstart
-    if [[ -d /etc/init ]]; then
-      if initctl status google-guest-agent | grep -qv 'running'; then
-        fail "Google guest agent not running."
-      fi
-    else
-      service google-guest-agent status
-      if [[ $? -ne 0 ]]; then
-        fail "Google guest agent not running."
-      fi
-    fi
-
+  if [[ $OSCONFIG_NOT_SUPPORTED =~ "true" ]]; then
+    status "Skipping check for google-osconfig-agent, since it's not supported for this distro."
     return 0
   fi
-
-  # Upstart
-  if [[ -d /etc/init ]]; then
-    status "Checking google-accounts-daemon."
-    if initctl status google-accounts-daemon | grep -qv 'running'; then
-      fail "Google accounts daemon not running."
-    fi
-
-    status "Checking google-network-daemon."
-    if initctl status google-network-daemon | grep -qv 'running'; then
-      fail "Google Network daemon not running."
-    fi
-
-    status "Checking google-clock-skew-daemon."
-    if initctl status google-clock-skew-daemon | grep -qv 'running'; then
-      fail "Google Clock Skew daemon not running."
-    fi
-  else
-    status "Checking google-accounts-daemon."
-    service google-accounts-daemon status
-    if [[ $? -ne 0 ]]; then
-      fail "Google accounts daemon not running."
-    fi
-
-    status "Checking google-network-daemon."
-    service google-network-daemon status
-    if [[ $? -ne 0 ]]; then
-      fail "Google Network daemon not running."
-    fi
-
-    status "Checking google-clock-skew-daemon."
-    service google-clock-skew-daemon status
-    if [[ $? -ne 0 ]]; then
-      fail "Google Clock Skew daemon not running."
-    fi
-  fi
+  assert_running google-osconfig-agent
 }
 
 # Check Google Cloud SDK.
@@ -219,10 +209,10 @@ function check_package_install {
 wait_for_connectivity
 
 # Run tests.
+check_package_install
 check_google_services
 check_google_cloud_sdk
 check_cloud_init
-check_package_install
 check_metadata_connectivity
 check_internet_connectivity
 
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -152,7 +152,7 @@ def DistroSpecific(g):
       g.write_append(
           '/etc/yum.repos.d/google-cloud.repo', repo_sdk % el_release)
       yum_install(g, 'google-cloud-sdk')
-    yum_install(g, 'google-compute-engine')
+    yum_install(g, 'google-compute-engine', 'google-osconfig-agent')
 
   logging.info('Updating initramfs')
   for kver in g.ls('/lib/modules'):
@@ -193,7 +193,7 @@ def yum_install(g, *packages):
 
   Args:
     g (guestfs.GuestFS): A mounted GuestFS instance.
-    *packages (list of str): The YUM packages to be installed.
+    *packages: The YUM packages to be installed.
 
   Raises:
       RuntimeError: If there is a failure during installation.
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -193,7 +193,7 @@ def yum_install(g, *packages):
 
   Args:
     g (guestfs.GuestFS): A mounted GuestFS instance.
-    *packages: The YUM packages to be installed.
+    *packages (list of str): The YUM packages to be installed.
 
   Raises:
       RuntimeError: If there is a failure during installation.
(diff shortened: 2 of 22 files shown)


daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -45,11 +45,17 @@
...
 # Enabled for Google Cloud SDK
-deb http://archive.canonical.com/ubuntu {ubu_release} partner
+deb http://archive.canonical.com/ubuntu {release} partner
 '''
 
+# https://cloud.google.com/compute/docs/manage-os
+os_config_repo = '''
+# Enabled for GCE OS config agent
+deb http://packages.cloud.google.com/apt google-compute-engine-{release}-stable main
+'''  # noqa: E501
```
zoran15, line 57: what does this mean?

daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -45,11 +45,17 @@
 ConnectPort 563
 '''
 
-partner_list = '''
+partner_repo = '''
 # Enabled for Google Cloud SDK
-deb http://archive.canonical.com/ubuntu {ubu_release} partner
+deb http://archive.canonical.com/ubuntu {release} partner
```
zoran15, line 50: Is this ubuntu_release param from the docs at the top of the file?

daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -129,9 +135,7 @@ def DistroSpecific(g):
 
     remove_azure_agents(g)
 
-    g.write(
-        '/etc/apt/sources.list.d/partner.list',
-        partner_list.format(ubu_release=ubu_release))
+    add_gce_repositories(g, ubu_release)
```
zoran15, line 138: is this intended, to have both `release` and `ubu_release` params?

daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -163,6 +165,39 @@ def DistroSpecific(g):
...
+    g.write('/etc/apt/sources.list.d/google-compute-engine.list',
+            os_config_repo.format(release=release))
+
+
+def install_gce_packages(g, release):
+  logging.info('Installing GCE packages.')
+  utils.update_apt(g)
+  pkgs = ['gce-compute-image-packages', 'google-cloud-sdk']
+  if release in ['xenial', 'bionic']:
+    # OS Config agent is only supported for xenial and bionic.
```
zoran15, line 194: Maybe have this as info log, same as for Windows?

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: It's flake8's mechanism for bypassing a check; in this case, the line starting with `deb` exceeds the line length limit.

- https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html
- https://www.flake8rules.com/rules/E501.html

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: Good question. The comment at the top is referring to the interface exposed via instance metadata. Looking at that, along with your next comment, I'm going to consolidate all three into the same variable name for clarity. Good find!

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: Consolidated `release` and `ubu_release` to use `ubuntu_release`, which is the name of the metadata value passed in by daisy.

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: good idea! Updated here as well as in Debian (which doesn't support debian 8).

hopkiw:
"To disable installation of the agent, users can specify the existing `no-guest-environment` flag."
Please rename this flag - osconfig is not part of the guest environment per our public documentation on what that means. https://cloud.google.com/compute/docs/images/guest-environment

if we're going to change that and consider osconfig part of it, the flag should prevent installation of all guest environment packages, not only osconfig.

zoran15:
> "To disable installation of the agent, users can specify the existing `no-guest-environment` flag."
> Please rename this flag - osconfig is not part of the guest environment per our public documentation on what that means. https://cloud.google.com/compute/docs/images/guest-environment
> 
> if we're going to change that and consider osconfig part of it, the flag should prevent installation of all guest environment packages, not only osconfig.

This flag already controls the installation of all guest environment packages. I've discussed this briefly with Andrew and we agreed, for simplicity, it's better to keep a single flag that controls installation of all guest agents. If you feel this is not the right approach, let's set up a meeting and discuss it.

daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -37,6 +50,21 @@ function wait_for_connectivity {
   exit 1
 }
 
+# Assert that a service is running; supports upstart and systemd.
```
hopkiw, line 53: is this satisfactory? we don't need any sysvinit support for any imports? also is the `service` binary available everywhere? that's the legacy sysvinit tool, and on systems which have transitioned to systemd, it becomes a wrapper around systemd. if you're intending to support systemd specifically, you should use `systemctl` instead of `service`

daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -15,6 +15,19 @@
...
+
+# To disable checking for the OS config agent, add an instance metadata
+# value of osconfig_not_supported: true.
+url=http://metadata.google.internal/computeMetadata/v1/instance/attributes/osconfig_not_supported
+if command -v curl; then
+  OSCONFIG_NOT_SUPPORTED="$(curl --header "Metadata-Flavor: Google" $url)"
+elif command -v wget; then
+  OSCONFIG_NOT_SUPPORTED="$(wget --header "Metadata-Flavor: Google" -O - $url)"
+else
+  echo "TestFailed: Either curl or wget is required to run test suite."
```
hopkiw, line 27: i see a `fail` function is used elsewhere in this file. why not here?

daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -219,10 +209,10 @@ function check_package_install {
 wait_for_connectivity
 
 # Run tests.
+check_package_install
```
hopkiw, line 212: why reorder?

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -193,7 +193,7 @@ def yum_install(g, *packages):
 
   Args:
     g (guestfs.GuestFS): A mounted GuestFS instance.
-    *packages (list of str): The YUM packages to be installed.
+    *packages: The YUM packages to be installed.
```
hopkiw, line 196: what is the type of this parameter now?

daisy_workflows/image_import/suse/translate.py
```diff
@@ -171,6 +173,31 @@ def _disambiguate_suseconnect_product_error(g, product, error) -> Exception:
...
+  """Add Zypper repositories that are required for installing GCE packages."""
+  if not include_gce_packages:
+    return
+  # Installing OS Config agent via instructions:
+  #    https://cloud.google.com/compute/docs/manage-os
+  repo = """
+    # Enabled for GCE OS config agent
+    [google-compute-engine]
+    name=Google Compute Engine
+    baseurl=https://packages.cloud.google.com/yum/repos/google-compute-engine-sles{major}-stable
```
hopkiw, line 186: is this the first time we're adding this repo? i didn't even know it existed. this can bite us later if we ever accidentally upload packages to the repo other than osconfig, bc they may conflict. i think this might be the biggest change in here - have we discussed it with suse?

daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -45,11 +45,17 @@
...
+partner_repo = '''
 # Enabled for Google Cloud SDK
-deb http://archive.canonical.com/ubuntu {ubu_release} partner
+deb http://archive.canonical.com/ubuntu {ubuntu_release} partner
 '''
 
+# https://cloud.google.com/compute/docs/manage-os
+os_config_repo = '''
+# Enabled for GCE OS config agent
+deb http://packages.cloud.google.com/apt google-compute-engine-{ubuntu_release}-stable main
```
hopkiw, line 56: same question here about repos and if we have discussed this change with canonical

hopkiw:
> > "To disable installation of the agent, users can specify the existing `no-guest-environment` flag."
> > Please rename this flag - osconfig is not part of the guest environment per our public documentation on what that means. https://cloud.google.com/compute/docs/images/guest-environment
> > if we're going to change that and consider osconfig part of it, the flag should prevent installation of all guest environment packages, not only osconfig.
> 
> This flag already controls the installation of all guest environment packages. I've discussed this briefly with Andrew and we agreed, for simplicity, it's better to keep a single flag that controls installation of all guest agents. If you feel this is not the right approach, let's set up a meeting and discuss it.

oh there's no threading to these comments, how annoying.
i don't feel strongly about this, but we should make sure it won't cause any product confusion. this is the only place i know of where we treat osconfig as part of the guest environment, which is already a bit difficult of a concept to explain to customers.

daisy_integration_tests/scripts/post_translate_test.sh
EricEdens: It wasn't intended to describe what import supports, but rather my understanding of what the previous code was checking for. Perhaps a more general: "Assert that a service is running; first trying `initctl`, then trying `service`". Or maybe just "Assert that a service is running"

Regarding switching to using `systemctl`, or whether `service` is available everywhere, I'm fine with a pragmatic approach of adding more fallbacks if/when they become necessary (so far these two cover all of the test images we have, which roughly cover all of the supported distro/version combos)

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Reverted back.

More info: It's a vararg, where each member is expected to be a string. I originally removed this type signature since my static analyzer said it was invalid,  but I couldn't find any PEPs or examples showing the correct approach. Anyway, regardless of the static analyzer, I think it's clearer for the human.

daisy_workflows/image_import/suse/translate.py
EricEdens: This is the first time using the repo for import.

I like your overall observations; are they limited to this image import usecase? Or are they feedback on the overall OS Config distribution strategy on SUSE and Ubuntu?  (Here's the OS Config setup guide: https://cloud.google.com/compute/docs/manage-os)

Regarding import, I haven't talked with SUSE on this. Not sure about OS config overall.

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: Similar to above, regarding import, I haven't talked with Canonical on this. Not sure about OS config overall.

daisy_integration_tests/scripts/post_translate_test.sh
EricEdens: shoot, I re-ordered during development to get the results faster. Reverted to the old order.

daisy_integration_tests/scripts/post_translate_test.sh
EricEdens: The `fail` function accumulates test failures, but allows the suite to continue. In this case, if we fail to read from the metadata server, then we have insufficient information to proceed with testing, and exit immediately.

daisy_workflows/image_import/suse/translate.py
hopkiw: it's more of a risk with import, as we do it instead of the user. if we didn't leave the repo configured after installing packages, there wouldn't be any risk.

EricEdens: Talked offline with @hopkiw  and the OS Config team and decided to not install the OS Config agent on SLES, OpenSUSE, or Ubuntu until the agents are supported by upstream. Updated in latest version.

daisy_workflows/image_import/ubuntu/translate.py
EricEdens: Ditto to comment regarding SUSE

PR #1290, opened 2020-07-16
[new vmdk api adoption] daisy support new vmdk alpha API

dntczdx:
This is split from the https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1282/files?file-filters%5B%5D=.go
Since daisy is in a separate module, we need to split daisy-changes into a separate PR.

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
+	computeAlpha "google.golang.org/api/compute/v0.alpha"
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
@@ -36,6 +37,7 @@ type Client interface {
 	AttachDisk(project, zone, instance string, d *compute.AttachedDisk) error
 	DetachDisk(project, zone, instance, disk string) error
 	CreateDisk(project, zone string, d *compute.Disk) error
+	CreateDiskAlpha(project, zone string, d *computeAlpha.Disk) error
 	CreateForwardingRule(project, region string, fr *compute.ForwardingRule) error
 	CreateFirewallRule(project string, i *compute.Firewall) error
 	CreateImage(project string, i *compute.Image) error
@@ -64,6 +66,7 @@ type Client interface {
 	GetInstance(project, zone, name string) (*compute.Instance, error)
 	GetInstanceBeta(project, zone, name string) (*computeBeta.Instance, error)
 	GetDisk(project, zone, name string) (*compute.Disk, error)
+	GetDiskAlpha(project, zone, name string) (*computeAlpha.Disk, error)
 	GetForwardingRule(project, region, name string) (*compute.ForwardingRule, error)
 	GetFirewallRule(project, name string) (*compute.Firewall, error)
 	GetImage(project, name string) (*compute.Image, error)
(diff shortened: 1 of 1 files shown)


dntczdx:
> I thought PD team already rolled native VMDK to Beta. Do we have any benefit of doing Alpha first?

Not yet. As discussed offline, they need a couple of weeks to roll out it to beta. Right now we only have alpha, so we can get early test data.

daisy/compute/compute.go
```diff
@@ -261,7 +275,15 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (Client, error)
 	if ep != "" {
 		rawBetaService.BasePath = ep
 	}
-	c := &client{hc: hc, raw: rawService, rawBeta: rawBetaService}
+	rawAlphaService, err := computeAlpha.New(hc)
+	if err != nil {
+		return nil, fmt.Errorf("alpha compute client: %v", err)
+	}
```
zoran15, line 281: What happens if alpha API is not enabled? Will this fail? 

Maybe we can we ignore the error and throw error on attempted use of alpha API so client (wrappers) can deal with it.

dntczdx: Nothing bad will happen here. Client gets an error only when they actually call the alpha API.
If you look into the impl of computeAlpha.New, it always return a nil error, no real error will be returned, maybe because of the historic reason.

daisy/compute/compute.go
```diff
@@ -241,7 +245,17 @@ func shouldRetryWithWait(tripper http.RoundTripper, err error, multiplier int) b
 
 // NewClient creates a new Google Cloud Compute client.
 func NewClient(ctx context.Context, opts ...option.ClientOption) (Client, error) {
-	o := []option.ClientOption{option.WithScopes(compute.ComputeScope, compute.DevstorageReadWriteScope)}
+	// Set these scopes to be align with compute.NewService
+	o := []option.ClientOption{
```
adjackura, line 249: We can just drop this and use the deafaults.

daisy/compute/compute.go
adjackura: This is my biggest concern. Most users will not have alpha api access, so that should be the default assumption and not throw unneeded errors. Just creating the client may not do that at all though as im pretty sure it doesn't make any network call, so thi might just be at api call time and then an error is fine.

dntczdx: Yes Andrew as you said, error happens only when we call the API. It's as expected and we have the logic to properly handle the error.

dntczdx: This line just copied what beta and GA client creation logic, even though it is not actually catching anything

daisy/compute/compute.go
dntczdx: I followed what "NewService" do. The "NewService" always extends these scopes.
Imagine arg opts gave a narrow scope for any reason, without this extending, it will fail.


// NewService creates a new Service.
func NewService(ctx context.Context, opts ...option.ClientOption) (*Service, error) {
	scopesOption := option.WithScopes(
		"https://www.googleapis.com/auth/cloud-platform",
		"https://www.googleapis.com/auth/compute",
		"https://www.googleapis.com/auth/compute.readonly",
		"https://www.googleapis.com/auth/devstorage.full_control",
		"https://www.googleapis.com/auth/devstorage.read_only",
		"https://www.googleapis.com/auth/devstorage.read_write",
	)
	// NOTE: prepend, so we don't override user-specified scopes.
	opts = append([]option.ClientOption{scopesOption}, opts...)

PR #1292, opened 2020-07-18
Pass inflation disk sizes during OVF data disk import.

EricEdens:
#1283 caused a regression for OVF imports that contain multiple disks. When performing a multi-disk import, a user would see: "ImportFailed: Failed to resize disk. The Compute Engine default service account needs the role: roles/compute.storageAdmin"

This PR fixes the regression by:
1. Update import_image.sh to validate its required parameters.
2. Update OVF import to include the sizes of the disks that its passing to import_image.sh.

### DIFF
--- a/daisy_workflows/image_import/import_image.sh
+++ b/daisy_workflows/image_import/import_image.sh
@@ -41,6 +41,11 @@ SOURCE_SIZE_BYTES="$(gsutil du "${SOURCE_URL}" | grep -o '^[0-9]\+')"
 SOURCE_SIZE_GB=$(awk "BEGIN {print int(((${SOURCE_SIZE_BYTES}-1)/${BYTES_1GB}) + 1)}")
 IMAGE_PATH="/daisy-scratch/$(basename "${SOURCE_URL}")"
 
+# Validate input
+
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/import_image.sh
```diff
@@ -41,6 +41,11 @@ SOURCE_SIZE_BYTES="$(gsutil du "${SOURCE_URL}" | grep -o '^[0-9]\+')"
 SOURCE_SIZE_GB=$(awk "BEGIN {print int(((${SOURCE_SIZE_BYTES}-1)/${BYTES_1GB}) + 1)}")
 IMAGE_PATH="/daisy-scratch/$(basename "${SOURCE_URL}")"
 
+# Validate input
+
+[[ -z "$SCRATCH_DISK_SIZE_GB" ]] && echo "ImportFailed: metadata scratch_disk_size_gb is unset" && exit 1
+[[ -z "$INFLATED_DISK_SIZE_GB" ]] && echo "ImportFailed: metadata inflated_disk_size_gb is unset" && exit 1
```
zoran15, line 47: minor: unset -> not set. Unset sounds (at least to me) like we just cleared this metadata key (un-set it), while we want to say it's not set at all (missing).

EricEdens: done!

PR #1293, opened 2020-07-20
Increase timeout for el_import_dont_use_proxy.wf.json

EricEdens:
This change increases the timeout of el_import_dont_use_proxy.wf.json from 30 minutes to 60 minutes, since we've been seeing intermittent timeouts recently.

### DIFF
--- a/daisy_integration_tests/el_import_dont_use_proxy.wf.json
+++ b/daisy_integration_tests/el_import_dont_use_proxy.wf.json
@@ -54,7 +54,7 @@
       }
     },
     "translate-disk": {
-      "Timeout": "30m",
+      "Timeout": "60m",
       "IncludeWorkflow": {
         "Path": "../daisy_workflows/image_import/enterprise_linux/translate_centos_7.wf.json",
         "Vars": {


zoran15:
> The test in `daisy_integration_tests/el_import_dont_use_proxy.wf.json` has recently been exceeding its timeout of 30 minutes. This PR moves the test to the new image import suite, which achieves the following
> 
> 1. The test will inherit the default image import timeout (currently at 2 hours)

The timeout is set in test workflow so that's easy to change even without converting the test to a wrapper test.

daisy_integration_tests/daisy_e2e.test.gotmpl
```diff
@@ -53,9 +53,6 @@
     "debian9 translate": {
       "Path": "./debian_9_translate.wf.json"
     },
-    "Enterprise Linux - Don't use proxy": {
-      "Path": "./el_import_dont_use_proxy.wf.json"
-    },
```
zoran15, line 58: Removing Daisy translate tests means we don't support translate using native Daisy anymore. I think there might sitll be customers who are using translate workflows using Daisy so there's value in keeping these tests in addition to wrapper tests.

EricEdens: Oh my, you're right, we have docs on this: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/docs/image-import.md

Based on that, here's my intuition to fix the current timeout problem:
- Remove the new wrapper test
- Restore the old wf.json test
- Update the timeout in it to something larger

Thoughts?

zoran15: I'm ok with points 2 and 3. Regarding the removal of the new wrapper test, if it covers more than the original translate test and if this coverage will not be replicated in other wrapper tests you will add in the future, we can consider keeping it.

EricEdens: OK! If it's alright, my intuition is to remove the new test, and only adjust the timeout parameter in the `wf.json` file, since that's the root cause of the issue that we're trying to fix here.

PR #1297, opened 2020-07-21
EL build fixes and additions

zmarano:
- Add RHEL 8 for SAP build workflows.
- Also fix EL 8.2+ builds (need larger install disk size).
- Move the google-cloud-sdk install and remove older dependencies (which don't work on el 8.1+)
- Cleanup logging and output so scriptlet output is correctly logged to the serial console, remove unnecessary save_logs step.


daisy_workflows/image_build/enterprise_linux/ks_helpers.py
```diff
@@ -260,14 +260,21 @@ def BuildKsConfig(release, google_cloud_repo, byos, sap, uefi):
...
     logging.info('Building RHEL 8 image.')
     ks_packages = FetchConfigPart('el8-packages.cfg')
     ks_options = FetchConfigPart('el8-options.cfg')
     if uefi:
       logging.info('Building RHEL 8 for UEFI')
       ks_options = FetchConfigPart('el8-uefi-options.cfg')
     rhel_post = FetchConfigPart('rhel8-post.cfg')
+    if sap:
+      logging.info('Building RHEL 8 for SAP')
+      point = ''
```
hopkiw, line 273: maybe a nitpick but if you're just appending, you could just call this `rhel_post`

daisy_workflows/image_build/enterprise_linux/kickstart/rhel7-sap-post.cfg
```diff
@@ -31,5 +39,12 @@ unar
 "
 
 yum install -y ${SAP_PKGS}
+if [ $? -ne 0 ]; then
+  echo "BuildStatus: Failed to install SAP packages."
```
hopkiw, line 43: same here, aren't SAP packages required?

daisy_workflows/image_build/enterprise_linux/kickstart/el7-post.cfg
```diff
@@ -1,10 +1,22 @@
 # Google Compute Engine kickstart config for Enterprise Linux 7.
-%post --log=/dev/ttyS0
-set -x
+%onerror
+echo "BuildFailed!" > /dev/ttyS0
```
hopkiw, line 3: i still don't think we should be using daisy-isms in here. leaving this code independent is cleaner if possible.

hopkiw: also, el6-post doesn't have this %onerror section, is it not supported on el6?

daisy_workflows/image_build/enterprise_linux/kickstart/rhel7-sap-post.cfg
```diff
@@ -31,5 +29,6 @@ unar
 "
 
 yum install -y ${SAP_PKGS}
-yum install -y fence-agents-gce resource-agents-gcp
+yum install -y fence-agents-gce resource-agents-gcp || exit 0
```
hopkiw, line 32: the others use return, this uses exit

zmarano: Artifact of a failed experiment- changed.

daisy_workflows/image_build/enterprise_linux/kickstart/el7-post.cfg
zmarano: No unfortunately el6 doesn't support this wonderful feature :( So... basically it'll just crash but the logs will still be there and you can see why.

daisy_workflows/image_build/enterprise_linux/kickstart/el6-post.cfg
```diff
@@ -186,4 +186,6 @@ echo "(hd0) /dev/sda" > /boot/grub/device.map
 echo -e "root (hd0,0) \n setup (hd0)" | grub \
   --config-file /boot/grub/grub.conf --batch \
   --device-map /boot/grub/device.map
+
+) 2>&1 | /usr/bin/tee /dev/ttyS0
```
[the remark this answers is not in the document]
zmarano: So, its a combination of things. Moving to cmdline instead of text makes the output usable- redirecting tty3 (which is the live output in anaconda) to ttyS0 means its all logged to the serial console and therefore saved. Some of the commands that we catch are because there are conditions where they can fail and its ok. (I'll add comments)

daisy_workflows/image_build/enterprise_linux/kickstart/el6-post.cfg
```diff
@@ -1,10 +1,17 @@
 # Google Compute Engine kickstart config for Enterprise Linux 6.
-%post --log=/dev/ttyS0
-set -x
+%post --erroronfail
+exec < /dev/tty3 > /dev/tty3
+chvt 3
```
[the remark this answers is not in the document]
zmarano: Its the live execution output.

daisy_workflows/image_build/enterprise_linux/kickstart/el6-options.cfg
```diff
@@ -20,7 +20,7 @@ authconfig --useshadow --passalgo=sha512 --kickstart
 keyboard us
 network --bootproto=dhcp --noipv6 --hostname=localhost --device=eth0
 install
-text
+cmdline
```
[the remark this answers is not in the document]
zmarano: Makes the output usable- text is ncurses, cmdline is real text. https://wiki.centos.org/TipsAndTricks/KickStart

daisy_workflows/image_build/enterprise_linux/kickstart/cleanup.cfg
```diff
@@ -0,0 +1,5 @@
+# Cleanup.
+%post --nochroot --log=/dev/ttyS0
+set -x
+rm -Rf /mnt/sysimage/tmp/*
```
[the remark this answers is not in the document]
zmarano: Actually I need to verify this again because it may not actually be necessary. Will get back to you on this...

daisy_workflows/image_build/enterprise_linux/build_installer.py
```diff
@@ -58,6 +60,8 @@ def main():
   ks_content = ks_helpers.BuildKsConfig(release, repo, byos, sap, uefi)
   ks_cfg = 'ks.cfg'
   utils.WriteFile(ks_cfg, ks_content)
+  # Save the generated kickstart file to the build logs.
+  utils.UploadFile(ks_cfg, '%s/ks.cfg' % logs_path)
```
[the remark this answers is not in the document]
zmarano: Yeah- it should be daisy_logs_path (its not the outs path).

daisy_workflows/image_build/enterprise_linux/kickstart/el7-post.cfg
zmarano: Not a daisy-ism just a message for the fail conditions.

daisy_workflows/image_build/enterprise_linux/kickstart/el6-post.cfg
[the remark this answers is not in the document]
hopkiw: this doesn't redirect tty3, it redirects TO tty3. regardless of the tty3 lines at top, you are only getting serial port output from the `tee` command at the end. `exec > /dev/ttyS0` would have eliminated the need for the subshell and the tee command.

daisy_workflows/image_build/enterprise_linux/kickstart/el7-post.cfg
hopkiw: BuildFailed is a dasiy-ism, though. but i guess it's not too bad, so long as the behavior of the ks rules doesn't depend on being run by daisy.

daisy_workflows/image_build/enterprise_linux/kickstart/el6-post.cfg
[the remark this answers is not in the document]
zmarano: OH! Now I see why they are doing this. You're right, we only care about redirecting tty3 to ttyS0. These solutions all talk about getting live build logs to a screen and a log file. We don't actually care about that because we basically use the console as the log. So yeah- we can do this much easier.

daisy_workflows/image_build/enterprise_linux/kickstart/cleanup.cfg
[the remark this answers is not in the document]
zmarano: Yup- still necessary. A bunch of files end up in tmp that don't get cleaned up unless you do it.

daisy_workflows/image_build/enterprise_linux/kickstart/el7-post.cfg
```diff
@@ -31,20 +40,16 @@ EOL
...
+yum install -y google-cloud-sdk
 
 # Send /root/anaconda-ks.cfg to our logs.
 cp /run/install/ks.cfg /tmp/anaconda-ks.cfg
 
-# Remove files which shouldn't make it into the image.
-rm -f /etc/boto.cfg /etc/udev/rules.d/70-persistent-net.rules
+# Remove files which shouldn't make it into the image. These files may not
+# exist.
+rm -f /etc/boto.cfg /etc/udev/rules.d/70-persistent-net.rules || return 0
```
hopkiw, line 52: `rm -f` returns 0 on missing files. if that's the only fail scenario to handle, you can omit `|| return 0`
also do we want to *return* if one fails? i would think we'd want to continue?

daisy_workflows/image_build/enterprise_linux/kickstart/el8-post.cfg
```diff
@@ -1,10 +1,19 @@
 # Google Compute Engine kickstart config for Enterprise Linux 8.
-%post --log=/dev/ttyS0
-set -x
+%onerror
+echo "Build Failed!" > /dev/ttyS0
+shutdown -h now
+%end
 
+%post --erroronfail
+exec < /dev/tty3 > /dev/ttyS0
```
hopkiw, line 8: still don't understand taking input from tty3
this section doesn't require input, and no input will likely be available from tty3?

hopkiw: also you might want to redirect stderr to the console too

daisy_workflows/image_build/enterprise_linux/kickstart/el7-post.cfg
zmarano: Yeah- your'e right. Changed.

daisy_workflows/image_build/enterprise_linux/kickstart/el8-post.cfg
zmarano: Yeah- ok makes sense. I think the only use case is if you are using anaconda live or with interactive mode. We are not. Changed.

PR #1298, opened 2020-07-21
Don't install Cloud SDK when importing Debian 8.

EricEdens:
The Cloud SDK is currently broken for Debian 8 imports; the SDK dropped support for Python 3.4, which is Debian 8's version.

This change skips installing the Cloud SDK for Debian 8, and skips testing for it.

## Testing

Ran the following:
1. daisy -project edens-test -zone=us-central1-b daisy_integration_tests/debian_8_translate.wf.json
2. daisy -project edens-test -zone=us-central1-b daisy_integration_tests/debian_9_translate.wf.json

### DIFF
--- a/daisy_integration_tests/scripts/post_translate_test.sh
+++ b/daisy_integration_tests/scripts/post_translate_test.sh
@@ -135,7 +135,7 @@ function check_google_cloud_sdk {
     fi
   fi
 
-    # Skip for Debian 8: b/161567418
+    # Skip for Debian 8, (Cloud SDK requires Python 3.5+)
   if [ -f /etc/os-release ]; then
     grep -qi "jessie" /etc/os-release
     if [ $? -eq 0 ]; then
--- a/daisy_integration_tests/scripts/post_translate_test.sh
+++ b/daisy_integration_tests/scripts/post_translate_test.sh
@@ -135,6 +135,14 @@ function check_google_cloud_sdk {
     fi
   fi
 
+    # Skip for Debian 8: b/161567418
+  if [ -f /etc/os-release ]; then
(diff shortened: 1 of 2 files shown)


daisy_integration_tests/scripts/post_translate_test.sh
```diff
@@ -135,6 +135,14 @@ function check_google_cloud_sdk {
     fi
   fi
 
+    # Skip for Debian 8: b/161567418
```
zoran15, line 138: minor: change the bug link (which is not accessible outside Google) with a description (or have both)

daisy_workflows/image_import/debian/translate.py
```diff
@@ -67,13 +67,15 @@ def DistroSpecific(g):
       logging.warn('Could not uninstall Azure agent. Continuing anyway.')
 
     utils.update_apt(g)
-    pkgs = ['google-cloud-packages-archive-keyring', 'google-cloud-sdk',
-            'google-compute-engine']
-    # Debian 8 doesn't support the new guest agent, so we need to install
-    # the legacy Python version.
```
zoran15, line 73: Any consequences for the old Guest agent and running on python 2.7/3.4?

daisy_integration_tests/scripts/post_translate_test.sh
EricEdens: done!

daisy_workflows/image_import/debian/translate.py
EricEdens: AFAIK there aren't big changes planned for the old guest agent.

PR #1302, opened 2020-07-23
OVF import: use OS ID from OVF descriptor in addition to OS Type

zoran15:
OVF import: no OS Type specified in OVF descriptoris currently the highest rate error for instance import.

To solve this, use OS ID from OVF descriptor in addition to OS Type to support a wider range of OVFs that might not have OS Type set (or use OS Type from a different name space, like vbox).

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
+++ b/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
@@ -377,14 +377,14 @@ func GetOVFDescriptorAndDiskPaths(ovfDescriptorLoader domain.OvfDescriptorLoader
 func GetOSId(ovfDescriptor *ovf.Envelope) (string, error) {
 
 	if ovfDescriptor.VirtualSystem == nil {
-		return "", daisy.Errf("VirtualSystem must be defined to retrieve OS info")
+		return "", daisy.Errf("OVF descriptor error: VirtualSystem must be defined to retrieve OS info. Use --os flag to specify OS")
 	}
 	if ovfDescriptor.VirtualSystem.OperatingSystem == nil ||
 		len(ovfDescriptor.VirtualSystem.OperatingSystem) == 0 {
-		return "", daisy.Errf("OperatingSystemSection must be defined to retrieve OS info")
+		return "", daisy.Errf("OVF descriptor error: OperatingSystemSection must be defined to retrieve OS info. Use --os flag to specify OS")
 	}
 	if ovfDescriptor.VirtualSystem.OperatingSystem[0].OSType == nil && ovfDescriptor.VirtualSystem.OperatingSystem[0].ID == 0 {
-		return "", daisy.Errf("OperatingSystemSection.OSType or OperatingSystemSection.ID in OVF descriptor must be defined to retrieve OS info")
+		return "", daisy.Errf("OVF descriptor error: OperatingSystemSection.OSType or OperatingSystemSection.ID must be defined to retrieve OS info. Use --os flag to specify OS")
 	}
 
 	var osInfoFromOSType, osInfoFromOSID *osInfo
--- a/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
+++ b/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
@@ -35,42 +35,169 @@ const (
 	usbController          uint16 = 23
 )
 
-//Full list: https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html
-
-// Mapping OVF osType attribute to importer OS ID
-var ovfOSTypeToOSID = map[string]string{
-	"debian8Guest":          "debian-8",
-	"debian8_64Guest":       "debian-8",
-	"debian9Guest":          "debian-9",
-	"debian9_64Guest":       "debian-9",
-	"centos6Guest":          "centos-6",
-	"centos6_64Guest":       "centos-6",
-	"centos7Guest":          "centos-7",
-	"centos7_64Guest":       "centos-7",
-	"centos8Guest":          "centos-8",
-	"centos8_64Guest":       "centos-8",
-	"rhel6Guest":            "rhel-6",
-	"rhel6_64Guest":         "rhel-6",
-	"rhel7Guest":            "rhel-7",
-	"rhel7_64Guest":         "rhel-7",
-	"windows7Server64Guest": "windows-2008r2",
+type osInfo struct {
+	description      string
+	importerOSIDs    []string
+	nonDeterministic bool
 }
 
-// Mapping potentially supported OVF osType values to possible importer OS ID values
+func (oi *osInfo) isDeterministic() bool {
+	return len(oi.importerOSIDs) == 1 && !oi.nonDeterministic
+}
+
+func (oi *osInfo) hasImporterOSIDs() bool {
+	return len(oi.importerOSIDs) > 0
+}
+
+//Mapping OVF OS ID to OS info
+//Full list: http://schemas.dmtf.org/wbem/cim-html/2/CIM_OperatingSystem.html
+var ovfOSIDToImporterOSID = map[int16]osInfo{
+	1:   {description: "Other", importerOSIDs: []string{}},
+	2:   {description: "MACOS", importerOSIDs: []string{}},
+	3:   {description: "ATTUNIX", importerOSIDs: []string{}},
+	4:   {description: "DGUX", importerOSIDs: []string{}},
+	5:   {description: "DECNT", importerOSIDs: []string{}},
+	6:   {description: "Tru64 UNIX", importerOSIDs: []string{}},
+	7:   {description: "OpenVMS", importerOSIDs: []string{}},
+	8:   {description: "HPUX", importerOSIDs: []string{}},
+	9:   {description: "AIX", importerOSIDs: []string{}},
+	10:  {description: "MVS", importerOSIDs: []string{}},
+	11:  {description: "OS400", importerOSIDs: []string{}},
+	12:  {description: "OS/2", importerOSIDs: []string{}},
+	13:  {description: "JavaVM", importerOSIDs: []string{}},
+	14:  {description: "MSDOS", importerOSIDs: []string{}},
+	15:  {description: "WIN3x", importerOSIDs: []string{}},
+	16:  {description: "WIN95", importerOSIDs: []string{}},
+	17:  {description: "WIN98", importerOSIDs: []string{}},
+	18:  {description: "WINNT", importerOSIDs: []string{}},
+	19:  {description: "WINCE", importerOSIDs: []string{}},
+	20:  {description: "NCR3000", importerOSIDs: []string{}},
+	21:  {description: "NetWare", importerOSIDs: []string{}},
+	22:  {description: "OSF", importerOSIDs: []string{}},
+	23:  {description: "DC/OS", importerOSIDs: []string{}},
+	24:  {description: "Reliant UNIX", importerOSIDs: []string{}},
+	25:  {description: "SCO UnixWare", importerOSIDs: []string{}},
+	26:  {description: "SCO OpenServer", importerOSIDs: []string{}},
+	27:  {description: "Sequent", importerOSIDs: []string{}},
+	28:  {description: "IRIX", importerOSIDs: []string{}},
+	29:  {description: "Solaris", importerOSIDs: []string{}},
+	30:  {description: "SunOS", importerOSIDs: []string{}},
+	31:  {description: "U6000", importerOSIDs: []string{}},
+	32:  {description: "ASERIES", importerOSIDs: []string{}},
+	33:  {description: "HP NonStop OS", importerOSIDs: []string{}},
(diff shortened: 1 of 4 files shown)


cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -35,42 +35,169 @@ const (
...
-	"rhel7_64Guest":         "rhel-7",
-	"windows7Server64Guest": "windows-2008r2",
+func (oi *osInfo) hasImporterOSIDs() bool {
+	return len(oi.importerOSIDs) > 0
 }
 
-// Mapping potentially supported OVF osType values to possible importer OS ID values
+//Mapping OVF OS ID to OS info
+//Full list: http://schemas.dmtf.org/wbem/cim-html/2/CIM_OperatingSystem.html
+var ovfOSIDToImporterOSID = map[int16]osInfo{
```
dntczdx, line 54: would you help me understand:
1.  what will happen if the image has a unlisted OS ID?
2. is there a way to leverage the os-detection project here?

zoran15: 1. If user hasn't specified --os flag and there's no matching supported OS from OVF descriptor, error will be thrown here: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/766e13993361ec666ffd6a9e9d17c67dd560a709/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go#L433
2. Yes, OS detection will be utilized for OVF import once available. We'll probably throw warnings or errors if what is detected and what is in the OVF descriptor don't match.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -35,42 +35,169 @@ const (
...
+	113: {description: "Microsoft Windows Server 2012", importerOSIDs: []string{"windows-2012", "windows-2012-byol"}},
+	114: {description: "Microsoft Windows 8", importerOSIDs: []string{"windows-8-x86-byol"}, nonDeterministic: true},
+	115: {description: "Microsoft Windows 8 64-bit", importerOSIDs: []string{"windows-8-x64-byol"}, nonDeterministic: true},
+	116: {description: "Microsoft Windows Server 2012 R2", importerOSIDs: []string{"windows-2012r2", "windows-2012r2-byol"}},
+	117: {description: "Microsoft Windows Server 2016", importerOSIDs: []string{"windows-2016", "windows-2016-byol"}},
+	118: {description: "Microsoft Windows 8.1", importerOSIDs: []string{}},
+	119: {description: "Microsoft Windows 8.1 64-bit", importerOSIDs: []string{}},
+	120: {description: "Microsoft Windows 10", importerOSIDs: []string{"windows-10-x86-byol"}, nonDeterministic: true},
+	121: {description: "Microsoft Windows 10 64-bit", importerOSIDs: []string{"windows-10-x64-byol"}, nonDeterministic: true},
+	//TODO: windows-2019, windows-2019-byol
```
EricEdens, line 175: Is this TODO for a follow-up PR?

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -35,42 +35,169 @@ const (
 	usbController          uint16 = 23
 )
 
-//Full list: https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html
+type osInfo struct {
+	description      string
+	importerOSIDs    []string
+	nonDeterministic bool
```
winterfreshx, line 41: Please add a comment for what the flag signifies. Makes it easier to understand intended behavior of flags with comments

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -35,42 +35,169 @@ const (
...
+}
```
EricEdens, line 42: Can you add some comments for `description` and `nonDeterministic`? (For example, to help someone needing to add more OS mappings in a couple of years)

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -250,31 +377,66 @@ func GetOVFDescriptorAndDiskPaths(ovfDescriptorLoader domain.OvfDescriptorLoader
...
 	}
 
-	return osID, nil
+	if osInfoFromOSType == nil && osInfoFromOSID == nil {
+		return "", daisy.Errf("cannot determine OS from OVF descriptor. Use --os flag to specify OS")
+	}
+
+	osInfo := getMoreSpecificOSInfo(osInfoFromOSType, osInfoFromOSID)
+
+	if !osInfo.hasImporterOSIDs() {
```
winterfreshx, line 409: What about the "Other" OS return value? Shouldn't that mean that we haven't detected it?

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -35,42 +35,169 @@ const (
...
+	77:  {description: "Microsoft Windows Server 2008 64-Bit", importerOSIDs: []string{}},
+	78:  {description: "FreeBSD 64-Bit", importerOSIDs: []string{}},
+	79:  {description: "RedHat Enterprise Linux", importerOSIDs: []string{}},
+	80:  {description: "RedHat Enterprise Linux 64-Bit", importerOSIDs: []string{"rhel-6", "rhel-6-byol", "rhel-7", "rhel-7-byol", "rhel-8", "rhel-8-byol"}},
+	81:  {description: "Solaris 64-Bit", importerOSIDs: []string{}},
+	82:  {description: "SUSE", importerOSIDs: []string{}},
+	83:  {description: "SUSE 64-Bit", importerOSIDs: []string{"opensuse-15"}, nonDeterministic: true},
+	84:  {description: "SLES", importerOSIDs: []string{}},
+	85:  {description: "SLES 64-Bit", importerOSIDs: []string{"sles-12-byol", "sles-15-byol"}},
+	86:  {description: "Novell OES", importerOSIDs: []string{}},
```
EricEdens, line 140: As I read through this map, it appears that all entries are empty, marked as nonDetermistic, or have multiple possible matches. From my subsequent reading of the code, it looks like all such candidates wouldn't be used to allow the import to proceed. For example, if an OVF only specified id=83, then the import would stop and prompt the user to specify their `--os`.

Is my understanding correct? If so, will supporting osIDs allow imports to proceed that would have otherwise failed?

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -35,42 +35,169 @@ const (
...
-	"windows7_64Guest":      {"windows-7-x64-byol"},
-	"windows8Guest":         {"windows-8-x86-byol"},
-	"windows8_64Guest":      {"windows-8-x64-byol"},
-	"windows9Guest":         {"windows-10-x86-byol"},
-	"windows9_64Guest":      {"windows-10-x64-byol"},
-	"windows8Server64Guest": {"windows-2012", "windows-2012r2"},
-	"windows9Server64Guest": {"windows-2016", "windows-2019"},
+// correctness. All Windows Client imports are in this category due to the fact we can't assume BYOL licensing.
+//Full list: https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html
+var ovfOSTypeToOSID = map[string]osInfo{
```
winterfreshx, line 183: I noticed that we are missing some of the OS versions listed here that are on https://cloud.devsite.corp.google.com/compute/docs/images/os-details . Is that mostly because we have to do extra work to enable those new versions?

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: yes, once windows-2019 becomes available in OVF spec. I've created a tracker bug.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: Looking at the mapping, you're correct. The benefit to the customer is that, instead of throwing a "no mapping found" error, we will throw an error saying we know what is in the OVF descriptor but that it can't be mapped in a 1:1 manner to OS values GCE recognizes. 

For example, when ID=96 was provided, if the description didn't match VMWare specs (e.g. if OVF came from vBox), we would throw an error saying OS is not recognized. With this change, we'll instruct the user to provide either `debian-8` or `debian-9`.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: I think "Other" differs from "Unknown" (ID=0, not included). I'm not sure there's much value in keeping "Other" except that OVF descriptor is telling us there is an OS there but it's not listed. "Unknown" is basically `null` value and is thus not included.

I don't have strong preference on whether to keep "Other/Other 64-bit" or not.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: The list of operating systems in OVF descriptor spec is not something we can add to. E.g. COS would probably be listed as "Other". I'm not sure this answers your question?

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
dntczdx: Agree. I thought len(importerOSIDs)>1 means nonDeterministic, however after I read the whole PR I saw they are different. Would you explain why it's nonDeterministic when it's a 1:1 map?

zoran15: Added comments in the code, but the summary is that there are two main reasons:
1. Legal: we don't want to assume BYOL for Windows client because we don't know if a customer has BYOL agreement.
2. Possible multiple mappings but we only support one in import, so again we don't want to assume it's the version we support (e.g. "Suse 64-bit" to `opensuse-15` because there are other versions of Suse).

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: Added descriptions

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
winterfreshx: I meant Debian 10 and similar newer versions of OSes. It seems to be listed in the link that you have provided in the comment.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
EricEdens: I'd probably omit `Other`, since it would result in an error message of `cannot determine OS from OVF descriptor.`, which seems more helpful than `operating system "Other" detected but is not supported by...`

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
EricEdens: I see -- yeah, I agree, the new error messages that are introduced by this PR are a lot more helpful.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: Image import (and OVF import which relies on image import) currently don't support Debian 10 import, even though it's supported by GCE.

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
zoran15: Removed "Other" and "Other 64-bit"

zoran15:
> overall lgtm! In terms of consolidating data structures for future maintenance, another approach is to invert your data structures so that the maintainer only needs to specify a single instance of the `--os` parameter, and then one ore more OVF IDs and one or more OVF types, which (I think) would allow you to hide the manual specification of `deterministic`, since your code could calculate it at runtime. In pseudocode, an entry might look like:
> 
> ```
> {
>   osID: "debian-10",
>   ovfIDs: ["Debian 64-Bit"],
>   ovfTypes: ["debian10_64Guest"]
> },
> ```
> 
> On one hand, these data structures will be updated infrequently, so the extra refactoring might not be worth the effort. On the other hand, it also depends on the impact of someone inadvertently mis-specifying something (like the `nonDeterministic` flag), and whether you think the current tests would identify such an error.

Inverting the mapping would not help with avoiding the `nonDeterministic` field because of the reasons listed in one of the other comments above. Copy/paste:

1. Legal: we don't want to assume BYOL for Windows client because we don't know if a customer has BYOL agreement.

2.Possible multiple mappings but we only support one in import, so again we don't want to assume it's the version we support (e.g. "Suse 64-bit" to opensuse-15 because there are other versions of Suse).

Also, mapping from importer OS ID to OVF OS ID would prevent us from giving customers info on the OS detected when throwing an error that it's not supported by OVF import. Not a major feature but it does provide a bit better UX.

PR #1310, opened 2020-08-05
Image import: Include UEFI guest OS feature during inflation, if requested.

EricEdens:
Since Window's translation requires booting the instance, we need to add the UEFI_COMPATIBLE guest OS feature during inflation. Prior to this change, it was only added to the final GCP image.

## Testing
- Unit tests 
- Two new e2e tests (Windows and RHEL) where the source image requires UEFI for booting.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/api_inflater.go
+++ b/cli_tools/gce_vm_image_import/importer/api_inflater.go
@@ -43,19 +43,24 @@ func isCausedByAlphaAPIAccess(err error) bool {
 
 // apiInflater implements `importer.inflater` using the Compute Engine API
 type apiInflater struct {
-	serialLogs    []string
-	args          ImportArguments
-	computeClient daisyCompute.Client
-	storageClient storage.Client
+	serialLogs      []string
+	args            ImportArguments
+	computeClient   daisyCompute.Client
+	storageClient   storage.Client
+	guestOsFeatures []*computeAlpha.GuestOsFeature
 }
 
 func createAPIInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient storage.Client) inflater {
-	return &apiInflater{
+	inflater := apiInflater{
 		serialLogs:    []string{},
 		args:          args,
 		computeClient: computeClient,
 		storageClient: storageClient,
 	}
+	if args.UefiCompatible {
+		inflater.guestOsFeatures = []*computeAlpha.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}}
+	}
+	return &inflater
 }
 
 func (inflater *apiInflater) traceLogs() []string {
@@ -73,7 +78,9 @@ func (inflater *apiInflater) inflate() (persistentDisk, error) {
 	cd := computeAlpha.Disk{
 		Name:                diskName,
 		SourceStorageObject: inflater.args.SourceFile,
+		GuestOsFeatures:     inflater.guestOsFeatures,
 	}
+
 	err := inflater.computeClient.CreateDiskAlpha(inflater.args.Project, inflater.args.Zone, &cd)
 	if err != nil {
 		return persistentDisk{}, err
--- a/cli_tools/gce_vm_image_import/importer/inflater.go
+++ b/cli_tools/gce_vm_image_import/importer/inflater.go
@@ -234,11 +234,14 @@ func createDaisyInflater(args ImportArguments, fileInspector imagefile.Inspector
 		return nil, err
 	}
 
-	daisyUtils.UpdateAllInstanceNoExternalIP(wf, args.NoExternalIP)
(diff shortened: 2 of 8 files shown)


cli_tools/gce_vm_image_import/importer/api_inflater.go
```diff
@@ -43,19 +43,24 @@ func isCausedByAlphaAPIAccess(err error) bool {
...
 
 func createAPIInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient storage.Client) inflater {
-	return &apiInflater{
+	inflater := apiInflater{
 		serialLogs:    []string{},
 		args:          args,
 		computeClient: computeClient,
 		storageClient: storageClient,
 	}
+	if args.UefiCompatible {
```
dntczdx, line 60: Why not just do it inside apiInflater.inflate()? In my opinion, this is too detail and not worth show up in high-level logic.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -234,6 +234,9 @@ func createDaisyInflater(args ImportArguments, fileInspector imagefile.Inspector
 		return nil, err
 	}
 
+	if args.UefiCompatible {
+		daisyUtils.UpdateToUEFICompatible(wf)
```
zoran15, line 238: Maybe use addFeatureToDisk, similarly to how WINDOWS is added? This is fine as well but it does more than needed here. daisyUtils.UpdateToUEFICompatible(wf) was useful for monolithic Daisy workflows. With these fine grained workflows we can avoid it.

cli_tools/gce_vm_image_import/importer/api_inflater.go
```diff
@@ -43,19 +43,24 @@ func isCausedByAlphaAPIAccess(err error) bool {
...
+		inflater.guestOsFeatures = []*computeAlpha.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}}
```
zoran15, line 61: Not related to this bug, but are we setting "WINDOWS" guest OS feature in the API inflater? Maybe we should have a single func that adds these features (WINDOWS and UEFI_COMPATIBLE) in both API and regular inflater and just call it from both.

cli_tools/gce_vm_image_import/importer/api_inflater.go
EricEdens: I chose this location since it was the most unit-testable place that I could find.

Since you're actively working in this file, do you see a good place that's also testable?

cli_tools/gce_vm_image_import/importer/inflater_test.go
```diff
@@ -72,6 +72,30 @@ func TestCreateDaisyInflater_Image_NotWindows(t *testing.T) {
...
+
+	assert.Contains(t, getDisk(inflater.wf, 0).GuestOsFeatures, &compute.GuestOsFeature{
+		Type: "UEFI_COMPATIBLE",
+	})
+}
+
+func TestCreateDaisyInflater_Image_NotUEFI(t *testing.T) {
+	inflater := createDaisyInflaterForImageSafe(t, ImportArguments{
+		Source:         imageSource{uri: "image/uri"},
+		OS:             "ubunt-1804",
```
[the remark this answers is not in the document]
EricEdens: nice, fixed.

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: sgtm; updated.

cli_tools/gce_vm_image_import/importer/api_inflater.go
EricEdens: I like the idea of consolidating these modifications. Tactically, based on the current implementation, I *think* we'll need to wait until the api_inflater is using the GA GCE client.  Or did you see an opportunity now?

Regarding the WINDOWS guest os feature, it seems to be required as well. @dntczdx  is this your reading too? Separate PR?

zoran15: Both inflaters modify disk Guest OS features based on args. That logic can be in a shared func.

EricEdens: The structs have a different type though, right?
  - google.golang.org/api/compute/v1
  - google.golang.org/api/compute/v0.alpha

zoran15: Sure.

dntczdx: Sounds good, and by then we will have a better position to do the test. So far I have no problem to do the uefi logic in this layer temporarily.

PR #1318, opened 2020-08-10
Disable OS config test on windows_7_x64_byol.wf.json

EricEdens:
OS Config agent is only supported for Windows greater than 2008r2. Windows 7 and Windows 2008r2 are both NT 6.1.


EricEdens:
> Any idea why `Windows Translate: 7 32-bit BYOL` is not failing but 64-bit is?

It's already whitelisted -- I missed 64-bit it looks like.

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/5b81051e3e719b4db620a27ec8275b1d6f48fced/daisy_integration_tests/windows_7_x86_byol.wf.json#L44

PR #1322, opened 2020-08-12
[UEFI-compatible] auto detect & tag UEFI compatible

dntczdx:
Motivation: user may not have enough knowledge to determine whether their disk is UEFI-boot. Even they can determine it, gcloud & pantheon surface doesn't support importing a image with UEFI_COMPATIBLE support. (Only direct API supports that, which is not the main stream). We automatically detect it and mark it UEFI_COMPATIBLE based on the partition info.

Impl: after inflation stage & before translation stage, there is a inspection stage. We check the partition info in inspection stage to explore EFI partition. If any, we return a flag. The importer will consume it: if it's true, we will create a copy of the inflated disk with an additional guestOSFeature=UEFI_COMPATIBLE.

Changes including e2e tests, which covered:
MBR+BIOS (linux + windows)
MBR+UEFI (linux)
GPT+Hybrid (linux)
GPT+UEFI (linux + windows)

Not covered: GPT+BIOS, since GCVP doesn't support that.

### DIFF
--- a/cli_tools_cloudbuild.yaml
+++ b/cli_tools_cloudbuild.yaml
@@ -89,10 +89,10 @@ steps:
   args: 
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$_RELEASE
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
-  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$_RELEASE
-  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
-  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$_RELEASE
-  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/wrappers/gce_vm_image_import:$_RELEASE
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/wrappers/gce_vm_image_import:$COMMIT_SHA
+  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/wrappers/gce_vm_image_import:$_RELEASE
+  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/wrappers/gce_vm_image_import:$COMMIT_SHA
   - --context=/workspace
   - --dockerfile=gce_vm_image_import.Dockerfile
 


cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -128,9 +129,6 @@ func (b bootableDiskProcessor) postValidateFunc() daisy.WorkflowModifier {
 			}}
 		rl.LabelResources(w)
 		daisy_utils.UpdateAllInstanceNoExternalIP(w, b.noExternalIP)
-		if b.uefiCompatible {
```
dntczdx, line 131: This is deleted because UEFI_COMPATIBLE has been determine in inspection stage and will be inherited. No need to add it again in translation stage.
It's based on an assumption that "--inspect" is enabled automatically. If "--inspect" won't be enabled soon, I will have to add it back. @EricEdens please give some inputs here, thanks!

zoran15: I would keep it and add a TODO bug until it can safely be removed.

cli_tools/common/disk/inspect.go
```diff
@@ -55,8 +57,15 @@ type defaultInspector struct {
...
-	err := inspector.wf.Run(context.Background())
-	return InspectionResult{}, err
+	err = inspector.wf.Run(context.Background())
+	if err != nil {
+		return
+	}
+
+	if inspector.wf.GetSerialConsoleOutputValue("has_efi_partition") == "true" {
+		ir.HasEFIPartition = true
+	}
```
zoran15, line 69:
```suggestion
	ir.HasEFIPartition = inspector.wf.GetSerialConsoleOutputValue("has_efi_partition") == "true"
```

cli_tools/common/utils/logging/service/daisy_loggable.go
```diff
@@ -37,6 +37,14 @@ func (w workflowLoggable) GetValue(key string) string {
 	return w.wf.GetSerialConsoleOutputValue(key)
 }
 
+func (w workflowLoggable) GetValueAsBool(key string) bool {
```
zoran15, line 40: Can you use this above instead of comparing with "true"?

cli_tools/common/utils/logging/service/literal_loggable_test.go
```diff
@@ -57,6 +57,8 @@ func TestSingleImageImportLoggable(t *testing.T) {
 	inflationTimeValue := int64(10000)
 	shadowInflationTimeValue := int64(5000)
 	matchResultValue := "true"
+	isUEFICompatibleImageValue := true
+	isUEFIDetectedValue := true
```
zoran15, line 61: since there are only 2 possible values, it's a good idea to test different combos of these. E.g. if there's a bug where isUEFICompatibleImageValue is used instead of isUEFIDetectedValue, this test would not detect it as both are true here.

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -249,6 +249,10 @@ type OutputInfo struct {
 	ShadowInflationTime []int64 `json:"shadow_inflation_time_ms,omitempty"`
 	// Shadow disk match result for shadow disk inflater
 	ShadowDiskMatchResult string `json:"shadow_disk_match_result,omitempty"`
+	// Indicates whether the image is imported and marked as UEFI_COMPATIBLE
```
zoran15, line 252:
```suggestion
	// Indicates whether the imported image is marked as UEFI_COMPATIBLE
```

cli_tools/gce_vm_image_import/importer/processor.go
```diff
@@ -37,24 +42,62 @@ type processorProvider interface {
...
 
 func (d defaultProcessorProvider) provide(pd persistentDisk) (processor, error) {
 	if d.DataDisk {
-		return newDataDiskProcessor(pd, d.imageClient, d.Project,
+		return newDataDiskProcessor(pd, d.computeClient, d.Project,
 			d.Labels, d.StorageLocation, d.Description,
 			d.Family, d.ImageName), nil
 	}
+	var err error
+	pd, err = d.inspectDisk(pd)
```
zoran15, line 56: inspection doesn't naturally belog to provide() method. provide() is a processor factory method. Inspection should either be a separate step or a part of the processor.

cli_tools/gce_vm_image_import/importer/processor.go
```diff
@@ -37,24 +42,62 @@ type processorProvider interface {
...
+	// If UEFI_COMPATIBLE is enforced in user input args (by d.ImportArguments.UefiCompatible),
+	// then it has been honored in inflation stage, so no need to create a new disk here.
+	// Only create new disk with UEFI_COMPATIBLE when inspection result tells us to do it.
+	if !d.ImportArguments.UefiCompatible && inspectionResult.HasEFIPartition {
+		diskName := fmt.Sprintf("disk-%v-uefi", d.ImportArguments.ExecutionID)
+		err := d.computeClient.CreateDisk(d.ImportArguments.Project, d.ImportArguments.Zone, &compute.Disk{
+			Name:            diskName,
+			SourceDisk:      pd.uri,
+			GuestOsFeatures: []*compute.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}},
+		})
```
zoran15, line 86: Creating a new disk in order to set GuestOSFeatures doesn't belog to inspector. Inspector should just return info on the disk/image, it should not make updates. Updates belong to the processor or a separate step.

cli_tools/common/utils/logging/service/literal_loggable_test.go
dntczdx: It's covered by unit test in "processor_test.go". I may move it to another file after refactoring.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
dntczdx: Sounds good, let me add it back in this PR

dntczdx: I went through the scenarios again. Even without "--inspect", this logic can also be removed.
Today, when people enforced "isUefiCompatible" by arg (by API), it's honored in inflation stage, and will be inherited by the final image.

cli_tools/gce_vm_image_import/importer/processor.go
dntczdx: Sounds great, I moved it into bootableDiskProcessor.process()

cli_tools/gce_vm_image_import/importer/processor.go
dntczdx: Sounds great, I moved it into bootableDiskProcessor.process()

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -16,37 +16,103 @@ package importer
...
+
+	ir, err := b.inspectDisk()
+	if err != nil {
+		return b.pd, err
+	}
+
+	return b.recreateDisk(ir)
+}
+
+func (b bootableDiskProcessor) recreateDisk(ir disk.InspectionResult) (persistentDisk, error) {
```
zoran15, line 73: the name of this func should indicate it's handling UEFI, the current name is too generic. 

processUEFIDisk or similar (and document that, due to GuestOS features limitations, a new disk might be created)

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -16,37 +16,103 @@ package importer
...
 
 type bootableDiskProcessor struct {
-	workflow        *daisy.Workflow
-	userLabels      map[string]string
-	storageLocation string
-	uefiCompatible  bool
-	noExternalIP    bool
-	network         string
-	OS              string
+	ImportArguments
```
zoran15, line 35: Since this is basically inheritance in GoLang, it's a bit strange to have bootableDiskProcessor inherit from ImportArguments

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -16,37 +16,103 @@ package importer
...
+		b.pd.uri = fmt.Sprintf("zones/%v/disks/%v", b.Zone, diskName)
+	}
+
+	if b.UefiCompatible || ir.HasEFIPartition {
+		b.pd.isUEFICompatible = true
+	}
+
+	if ir.HasEFIPartition {
+		b.pd.isUEFIDetected = true
+	}
```
zoran15, line 101: Here's your updating copies of b.pd because both b and pd are passed as values, not references. The tests are failing for other reasons, but I assume it was not your intention to modify copies of objects since these changes will be lost once func ends.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -16,37 +16,103 @@ package importer
...
+
+	return b.pd, nil
+}
+
+func (b bootableDiskProcessor) inspectDisk() (disk.InspectionResult, error) {
+	log.Printf("Running experimental disk inspections on %v.", b.pd.uri)
```
zoran15, line 107: Why experimental? I would avoid using "experimental" in code that will be visible by customers. It will lead them to believe they are using an Alpha product.

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -249,6 +249,10 @@ type OutputInfo struct {
 	ShadowInflationTime []int64 `json:"shadow_inflation_time_ms,omitempty"`
 	// Shadow disk match result for shadow disk inflater
 	ShadowDiskMatchResult string `json:"shadow_disk_match_result,omitempty"`
+	// Indicates whether the imported image is marked as UEFI_COMPATIBLE
```
EricEdens, line 252:
```suggestion
	// Indicates whether the imported image is marked as UEFI_COMPATIBLE by the user.
```

cli_tools/gce_vm_image_import/importer/data_disk_processor.go
```diff
@@ -50,24 +52,20 @@ func newDataDiskProcessor(pd persistentDisk, client createImageClient, project s
...
+		pd: pd,
 	}
 }
 
 func (d dataDiskProcessor) traceLogs() []string {
 	return []string{}
 }
 
-func (d dataDiskProcessor) process() (err error) {
+func (d dataDiskProcessor) process() (persistentDisk, error) {
```
zoran15, line 63: Document that a new disk might be created and the old one would be deleted.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -16,37 +16,103 @@ package importer
...
+func (b bootableDiskProcessor) recreateDisk(ir disk.InspectionResult) (persistentDisk, error) {
+	// If UEFI_COMPATIBLE is enforced in user input args (by d.ImportArguments.UefiCompatible),
+	// then it has been honored in inflation stage, so no need to create a new disk here.
+	// Create new disk with UEFI_COMPATIBLE only when inspection result tells us to do.
+	if !b.UefiCompatible && ir.HasEFIPartition {
+		diskName := fmt.Sprintf("disk-%v-uefi", b.ExecutionID)
+		err := b.computeClient.CreateDisk(b.Project, b.Zone, &compute.Disk{
+			Name:            diskName,
+			SourceDisk:      b.pd.uri,
+			GuestOsFeatures: []*compute.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}},
```
EricEdens, line 82: Are we losing attributes here? for example:
 - Can there be existing guestOSFeatures? (If the import started as a GCE image, for example.)
 - licenses

cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -273,14 +293,113 @@ func runImageImportWithSubnetWithoutNetworkSpecified(ctx context.Context, testCa
...
+		},
+		utils.GcloudProdWrapperLatest: {"compute", "images", "import", imageName, "--quiet",
+			"--docker-image-tag=latest", fmt.Sprintf("--os=%v", os), fmt.Sprintf("--project=%v", testProjectConfig.TestProjectID),
+			getSourceArg(isSourceFile, source),
+			fmt.Sprintf("--zone=%v", testProjectConfig.TestZone), "--inspect",
+		},
+		utils.GcloudLatestWrapperLatest: {"compute", "images", "import", imageName, "--quiet",
+			"--docker-image-tag=latest", fmt.Sprintf("--os=%v", os), fmt.Sprintf("--project=%v", testProjectConfig.TestProjectID),
+			getSourceArg(isSourceFile, source),
+			fmt.Sprintf("--zone=%v", testProjectConfig.TestZone), "--inspect",
```
zoran15, line 372: Is there `--inspect` flag in gcloud?

cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -273,14 +293,113 @@ func runImageImportWithSubnetWithoutNetworkSpecified(ctx context.Context, testCa
...
+	if isUEFICompatible {
+		expectedGuestOsFeatures = guestOsFeatureForUEFICompatible
+	} else {
+		unexpectedGuestOsFeatures = guestOsFeatureForUEFICompatible
+	}
+
+	runImportTestWithExtraParams(ctx, argsMap[testType], testType, testProjectConfig, imageName, logger, testCase, expectedGuestOsFeatures, unexpectedGuestOsFeatures, "", "", nil)
+}
+
+func getSourceArg(isSourceFile bool, source string) string {
```
zoran15, line 387: getSourceFlag

daisy_workflows/image_import/inspection/inspect_disk.py
```diff
@@ -14,6 +14,22 @@
...
+
+import os
+
+try:
+  os.system('fdisk -l /dev/sdb -o type')
+  stream = os.popen('fdisk -l /dev/sdb -o type')
+  output = stream.read()
+
+  # 1. For GPT, the ESP output should be "EFI System", which matches partition
+  # GUID "C12A7328-F81F-11D2-BA4B-00A0C93EC93B".
```
zoran15, line 27: Why is this ID in the code?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
```diff
@@ -16,37 +16,103 @@ package importer
...
 type bootableDiskProcessor struct {
-	workflow        *daisy.Workflow
-	userLabels      map[string]string
-	storageLocation string
-	uefiCompatible  bool
-	noExternalIP    bool
-	network         string
-	OS              string
+	ImportArguments
+	computeClient daisyCompute.Client
```
EricEdens, line 36: Since this is such a huge interface, it makes understanding the intent of the calling code difficult. To solve that, specify the method that you're using, similar to below. This helps the reader understand which method you're using, and makes testing easier since you can create a mock without a framework.

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/514ebce5f1ed4baaa8769e5a590367006c8d20a5/cli_tools/gce_vm_image_import/importer/importer.go#L187-L186

cli_tools/gce_vm_image_import/importer/processor.go
```diff
@@ -37,24 +37,15 @@ type processorProvider interface {
...
-		log.Printf("Running experimental disk inspections on %v.", pd.uri)
-		inspectionResult, err := d.diskInspector.Inspect(pd.uri)
-		if err != nil {
-			log.Printf("Inspection error=%v", err)
-		} else {
-			log.Printf("Inspection result=%v", inspectionResult)
-		}
-	}
-	return newBootableDiskProcessor(d.ImportArguments, pd)
+	return newBootableDiskProcessor(d.computeClient, d.diskInspector, d.ImportArguments, pd)
```
EricEdens, line 50: My original vision was a modular approach where inspection results determine which processor to return. For example, if the image is inspected as non-UEFI and Linux, then the additional code from this PR isn't required.

Benefits:
-  Individual processors are smaller, and easier to grok / test
-  Future development won't continue to increase the size of bootable disk processor

Tactically, to make that happen, a couple of options:
1. This provider returns a list of providers. The consumer iterates over each, and passes the processed disk to the subsequent step.
2. Don't change the signature here; instead, use a facade that iterates over each sub-processor.

Psuedocode

```

if data_disk:
 return  [data_disk_processor]

run inspection

processors := []

if data disk and windows:
  processors += uefi_tagger

processors += bootable_disk_processor
return processors
```

cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -63,6 +63,20 @@ func CLITestSuite(
 			testSuiteName, fmt.Sprintf("[%v][CLI] %v", testType, "Import with different network param styles"))
 		imageImportWithSubnetWithoutNetworkSpecifiedTestCase := junitxml.NewTestCase(
 			testSuiteName, fmt.Sprintf("[%v][CLI] %v", testType, "Import with subnet but without network"))
+		imageImportLinuxUEFITestCase := junitxml.NewTestCase(
+			testSuiteName, fmt.Sprintf("[%v][CLI] %v", testType, "Import UEFI test for linux UEFI"))
+		imageImportLinuxUEFIFromImageTestCase := junitxml.NewTestCase(
+			testSuiteName, fmt.Sprintf("[%v][CLI] %v", testType, "Import UEFI from image test for linux UEFI"))
```
EricEdens, line 69: I *think* this is a good opportunity for this suite instead. Just copy the two UEFI tests, and replace the uefi_compatible arg with `inspect`, right?

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go#L70-L69

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
EricEdens: sgtm

daisy_workflows/image_import/inspection/inspect_disk.py
```diff
@@ -14,6 +14,22 @@
 # limitations under the License.
 
 # Example of communicating a key-value pair back to Daisy:
-#   print("Status: <serial-output key:'partition_scheme' value:'mbr'>")
+#
+
+import os
```
EricEdens, line 19: Here's the file you want to update. Since I'm still working on making it easier to run tests, I'm fine if you put this code directly in `inspect-disk.py`, and I'll backport it into the library later. Or you can put it in the library now. It's up to you!

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_import/inspection/inspect-disk.py

EricEdens: Also, the associated README has directions for running the inspection workflow directly.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
EricEdens: I wasn't planning on that being added to gcloud

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
zoran15: Personally, I found using limited interfaces like this annoying: whenever I needed to use a new operation, I had to change the interface. Understanding the intent of the code is not really easier: it's obvious from the calling code what is the intent even though there are other, not used functions in the big interface. 

Testing might be easier without a framework, but with a framework it's the same.

EricEdens: I start by reading the interface, so having well-defined dependencies helps me. But it's a good learning that we're not all like that :)

In this particular case, here's my guidance:
1. Implement a processor specifically for UEFI (there's a longer comment about that below)
2. Given that the new processor will only need the CreateDisk method, I'd strongly recommend using implicit interfaces to get the benefits that we've talked about. That said, if you and Zoran aren't fans, I'm fine to move forward with the consensus

daisy_workflows/image_import/inspection/inspect_disk.py
zoran15: So is this an example of a GUID or something that's always the same? If it's an example, make sure to document it as such.

cli_tools/gce_vm_image_import/importer/processor.go
dntczdx: See my new imp. After evaluation, a separate inspector makes the bst sense to me.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
dntczdx: Great! I moved all the new e2e tests to import_tests

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -83,26 +108,151 @@ var cases = []testCase{
...
+	guestOsFeatureForUEFICompatible := []string{"UEFI_COMPATIBLE"}
+	var expectedGuestOsFeatures, unexpectedGuestOsFeatures []string
+	if t.expectUEFICompatible {
+		expectedGuestOsFeatures = guestOsFeatureForUEFICompatible
+	} else {
+		unexpectedGuestOsFeatures = guestOsFeatureForUEFICompatible
+	}
+
+	t.verifyGuestOSFeatures(expectedGuestOsFeatures, image, junit, logger, unexpectedGuestOsFeatures)
+}
```
EricEdens, line 230: I'd recommend that we didn't introspect details such as guestOSFeatures, since guestOSFeatures are an *implementation detail*. What we care about (and what our customers care about) is that their image boots, and can perform some tasks. 

I counter argument to this would be: We don't want MBR images to be tagged UEFI since they boot slower. If we care about boot times, then, the correct test would boot images and measure times, rather than indirectly look at a tag.

That said, I could understand inspecting licenses, since those generally don't have an observable side effect, but are rather present as an opaque metrics mechanism.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -39,9 +41,18 @@ const (
 	suite = "ImageImport"
 )
 
+type testCaseInterface interface {
+	getTestCase() testCase
+	run(ctx context.Context, junit *junitxml.TestCase, logger *log.Logger,
+		testProjectConfig *testconfig.Project, testType utils.CLITestType)
+}
+
 type testCase struct {
```
EricEdens, line 50: I'd been expecting to see a few more cases, where you omit the user-specified UEFI args. That's a great test, since prior to your changes in this PR, the image wouldn't boot, and the test would time out. Once you've added the changes, the boot script runs, and the test passes.

cli_tools/gce_vm_image_import/importer/processor.go
EricEdens: Nice -- I think it's heading in the right direction. The current implementation that you have, though, mixes two *distinct* responsibilities: inspection and mutation. The tactical reason for separating these, is that we need the inspection results to make other decisions (which translator to run, for example), and therefore need inspection to finish prior to running your mutation steps.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
EricEdens: It sounds like Tony favors the large interface, so I'm fine with proceeding with that.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor.go
EricEdens: sgtm

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,114 @@
...
+	"google.golang.org/api/compute/v1"
+)
+
+type diskInspectionProcessor struct {
+	args              ImportArguments
+	computeDiskClient daisyCompute.Client
+	diskInspector     disk.Inspector
+}
+
+func (d *diskInspectionProcessor) process(pd persistentDisk) (persistentDisk, error) {
```
zoran15, line 33: I agree here with Eric that inspection and processing should be separate steps. Inspection should only do inspection while uefi disk processor could perform disk creation with UEFI_COMPATIBLE guest OS feature attached. Inspection would be a separate step after inflation (unless there are technical issues), while UEFI processor be in the list of processor that run on the disk.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
dntczdx: In most of the cases, that is perfectly correct. This case is special, since vanadium should not be considered as completely transparent in my opinion. We know they can boot any disk with UEFI_COMPATIBLE successfully no matter whether it's really a UEFI; we also know that behavior won't be there forever (backward-compatible for BIOS disk with UEFI_COMPATIBLE can be deprecated). So, in this case, it's insufficient to only check boot-success. We also want to make sure UEFI_COMPATIBLE is marked only when needed.

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
dntczdx: Moved the logic to the proper place as suggested, thanks!

EricEdens:
Regarding the pytest failure, the test is failing waiting for this: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1326

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
EricEdens: Let's meet in the middle! 
1. Undo the new interfaces, since I'm not sure there's truly a delineation between the two types (given your argument, don't we always want to very the guest os feature?)
1. Update `testCase` to have a field `expectedGuestOSFeatures` that is a list.
1. Move the assertion code outside this suite -- there's a common set of code for tests?.  I imagine a signature of `assertImageHasGuestOSFeatures(imageURI string, expectedFeatures []string)` that asserts the features are *exactly* what's passed in (nothing missing, nothing extra)
1. Call  `assertImageHasGuestOSFeatures`after `t.runPostTranslateTest`

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
EricEdens: Thanks Tony! Took a look, and the current version that's uploaded still has the UEFI work inside bootable_processor, rather than having a list of processors.

cli_tools/common/utils/logging/service/log_entry.go
EricEdens: That makes sense -- can you update the comment to explain that?  Here's the first idea that comes to mind: `Indicates whether UEFI_COMPATIBLE was added to the image's guestOSFeatures, either due to inspection or user request.`

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,86 @@
...
+	pd.isUEFIDetected = ir.HasEFIPartition
+	return pd, nil
+}
+
+func (d *diskInspectionProcessor) inspectDisk(uri string) (disk.InspectionResult, error) {
+	log.Printf("Running disk inspections on %v.", uri)
+	ir, err := d.diskInspector.Inspect(uri)
+	if err != nil {
+		log.Printf("Disk inspection error=%v", err)
+		return ir, daisy.Errf("Disk inspection error: %v", err)
```
EricEdens, line 49: does this result in the error being printed twice?

dntczdx: Yes, it's by design. It's similar to daisy workflow's behavior: print the error right away when it happens; then print again at the end of the cmd after cleanup is done.

EricEdens: hmmmm.  curious. Well, it's probably better to stay consistent with the rest of the system. sgtm.

cli_tools/common/utils/logging/service/log_entry.go
dntczdx: done

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
dntczdx: Split the logic to 3 processors now.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
dntczdx: Hi Eric
1. To simplify the code, I removed the interface and 2 concrete impl as you suggested. I didn't apply the validation for guestOSFeatures for the existing tests. The reason is because (1) it's not what those tests want to test (2) it can be flaky. All my added tests are designed to be stable regarding the UEFI behaviors - they are based on the vmdk files that I created. Meanwhile, other tests usually use a image path, or sometimes even image family, which can be different in the future (one example: windows image was non-UEFI but now they are UEFI, cutoff date on June 2020 ). The author of the other tests don't need to verify the UEFI tag, also they don't want to carefully set the expected UEFI result when they create/modify the tests since that is not what they care about.

2. So accordingly, I updated the inspect to add a "expectedGuestOSFeatures" list as suggested, meanwhile added a "unexpectedGuestOSFeatures". The reason is in # 3

3. I moved it to the common lib as suggested so people can reuse it. The functionality is slightly different from you said: What I am verifying is whether the image contains the given guestOSFeature or not. It's not a "exact" list. It's because I don't want the test to be impacted by other features, like "IS_WINDOWS". (still kept some image-specific logic in local, since the assertGuestOSFeatures can be reused for not only image but also disk and more)

4. "Call assertImageHasGuestOSFeatures after t.runPostTranslateTest" - done similarly. It has to be before runPostTranslateTest, otherwise the image was removed by runPostTranslateTest

zoran15: > "We don't want MBR images to be tagged UEFI since they boot slower." If that's the case, then we should establish a test framework that monitors boot times, since this focuses on the observable behavior that the customer cares about.

I think we need to test whether GuestOSFeatures are set correctly by our tests. This is not a black-box for us as we know what inputs lead to which GuestOSFeatures. This is for performance reasons, but, more importantly, for correctness: UEFI_COMPATIBLE tagged MBR formatted disks might not work in the future if legacy mode is removed from the firmware.

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/disk"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+type diskInspectionProcessor struct {
+	args          ImportArguments
+	diskInspector disk.Inspector
+}
+
+func (d *diskInspectionProcessor) process(pd persistentDisk) (persistentDisk, error) {
```
zoran15, line 29: minor: `d` sounds like it's short for disk or similar. Maybe `processor` or `dip`?

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Printf("Disk inspection error=%v", err)
+		return ir, daisy.Errf("Disk inspection error: %v", err)
+	}
+
+	log.Printf("Disk inspection result=%v", ir)
+	return ir, nil
+}
+
+func (d *diskInspectionProcessor) cancel(reason string) bool {
+	if d.diskInspector != nil {
```
zoran15, line 57: minor: if condition can be flipped and returned early to avoid one nesting level

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,86 @@
...
+			return true
+		}
+	}
+
+	//indicate cancel was not performed
+	return false
+}
+
+func (d *diskInspectionProcessor) traceLogs() []string {
+	if d.diskInspector != nil {
```
zoran15, line 70: minor: same as above, can be flipped and return early

cli_tools/gce_vm_image_import/importer/disk_mutation_processor.go
```diff
@@ -0,0 +1,79 @@
...
+func (d *diskMutationProcessor) process(pd persistentDisk) (persistentDisk, error) {
+	// If UEFI_COMPATIBLE is enforced in user input args (b.uefiCompatible),
+	// then it has been honored in inflation stage, so no need to recreate a new disk here.
+	if d.args.UefiCompatible {
+		return pd, nil
+	}
+
+	if !pd.isUEFIDetected {
+		return pd, nil
+	}
```
zoran15, line 40: minor: can be merged, but it's also ok to keep them separate for readability

cli_tools/gce_vm_image_import/importer/processor_test.go
```diff
@@ -0,0 +1,124 @@
...
+			pd, err = diskMutationProcessor.process(pd)
+			assert.NoError(t, err)
+			if tt.isUEFIDisk && !tt.isInputArgUEFICompatible {
+				assert.Truef(t, strings.HasSuffix(pd.uri, "uefi"), "UEFI Disk URI should have suffix 'uefi', actual: %v", pd.uri)
+			} else {
+				assert.Falsef(t, strings.HasSuffix(pd.uri, "uefi"), "Disk URI shouldn't have suffix 'uefi', actual: %v", pd.uri)
+			}
+
+			pd, err = bootableDiskProcessor.process(pd)
+			assert.NotEmpty(t, bootableDiskProcessor.workflow.Vars["source_disk"].Value)
```
zoran15, line 96: This is a bit of an E2E test as all 3 types of processors are partially tested here. This test func should only care if processor provider works correctly. The test sfor each of the 3 processors should be separated.

cli_tools_e2e_test/common/utils/utils.go
```diff
@@ -270,3 +270,32 @@ func ContainsSubString(strs []string, s string) bool {
 	}
 	return false
 }
+
+// ContainsAll checks whether all given strings exists
```
zoran15, line 274: it's not clear from the doc what is actually checked? if all from arr are in subarr or vice versa?

cli_tools_e2e_test/common/utils/utils.go
```diff
@@ -270,3 +270,32 @@ func ContainsSubString(strs []string, s string) bool {
...
+			}
+		}
+		if !exists {
+			return false
+		}
+	}
+	return true
+}
+
+// ContainsAny checks whether any given strings exists
```
zoran15, line 291: Same here

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
dntczdx: then there will be 2 "return false" and probably 2 comments "//indicate cancel was not performed", so I left it like this

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
dntczdx: similar to above, there will have to be 2 lines of "return []string{}", which is not as trivial as "return nil" in my opinion. SInce it's only 2 layers, I chose to keep it. It's more like a "only return a non-zero value when all these conditions are true"

cli_tools/gce_vm_image_import/importer/disk_mutation_processor.go
dntczdx: Yes I separate them so the comments can be read better

cli_tools_e2e_test/common/utils/utils.go
dntczdx: modified

cli_tools_e2e_test/common/utils/utils.go
dntczdx: modified

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
dntczdx: changed to "p"

cli_tools/gce_vm_image_import/importer/processor_test.go
dntczdx: split to 4: 1 for provider, 3 for 3 different processors

PR #1323, opened 2020-08-13
Test RHEL 8 import with e2e

EricEdens:
The existing `daisy_integration_tests/rhel_8_translate.wf.json` test wasn't wired to the E2E runner. This change migrates that test to the new e2e suite that also tests the wrapper.


EricEdens:
> Please don't delete test workflow JSONs for the time being. I'll be using them for the first version of M4CE V2V adaptation tests.

Ahh, makes sense. Re-added.

PR #1327, opened 2020-08-14
First batch of M4CE adaptation E2E tests

zoran15:
First batch of M4CE adaptation E2E tests which use existing Daisy translate tests. The tests added in this batch are succeeding adaptation but some are failing the testing phase. M4CE team is working actively on fixing these issues and running the tests will allow them to check the progress of their work.

### DIFF
--- a/daisy_integration_tests/scripts/v2v_adapt.sh
+++ b/daisy_integration_tests/scripts/v2v_adapt.sh
@@ -1,0 +1,1 @@
+docker run --privileged -v /dev/disk/by-id:/dev/disk/by-id:ro v2v -vv --log file=/dev/ttyS1,level=debug --log file=/dev/ttyS2,level=trace --report file=/dev/ttyS3 --report file=http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/v2v/report,headers=Metadata-Flavor=Google --drive file=/dev/disk/by-id/google-workload-disk1 --in-place
--- a/daisy_integration_tests/v2v_adapt_e2e.test.gotmpl
+++ b/daisy_integration_tests/v2v_adapt_e2e.test.gotmpl
@@ -1,6 +1,6 @@
 {
   "Name": "v2v-adapt-e2e-tests",
-  "TestParallelCount": 30,
+  "TestParallelCount": 10,
   "Tests": {
     "2012R2 vmware image translate": {
       "Path": "./2012r2_vmware_translate.wf.json",
--- a/daisy_integration_tests/v2v_adapt_e2e.test.gotmpl
+++ b/daisy_integration_tests/v2v_adapt_e2e.test.gotmpl
@@ -1,0 +1,26 @@
+{
+  "Name": "v2v-adapt-e2e-tests",
+  "TestParallelCount": 30,
+  "Tests": {
+    "2012R2 vmware image translate": {
+      "Path": "./2012r2_vmware_translate.wf.json",
+      "Vars": {"translate_workflow": "./v2v_adapt_windows.wf.json"}
+    },
+    "Windows Server 2019 translate": {
+      "Path": "./2019_translate.wf.json",
+      "Vars": {"translate_workflow": "./v2v_adapt_windows_byol.wf.json"}
+    },
+    "Windows Translate: 7 64-bit BYOL": {
+      "Path": "./windows_7_x64_byol.wf.json",
+      "Vars": {"translate_workflow": "./v2v_adapt_windows_byol.wf.json"}
+    },
+    "Windows Translate: 8.1 64-bit BYOL": {
+      "Path": "./windows_8_x64_byol.wf.json",
+      "Vars": {"translate_workflow": "./v2v_adapt_windows_byol.wf.json"}
+    },
+    "Windows Translate: 10 64-bit BYOL": {
+     "Path": "./windows_10_x64_byol.wf.json",
+     "Vars": {"translate_workflow": "./v2v_adapt_windows_byol.wf.json"}
+    }
+  }
+}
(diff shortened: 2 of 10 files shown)


daisy_integration_tests/v2v_adapt_windows.wf.json
```diff
@@ -0,0 +1,93 @@
...
+            {
+              "Source": "${source_disk}",
+              "DeviceName": "workload-disk1"
+            }
+          ],
+          "MachineType": "n1-standard-2",
+          "Metadata": {
+            "block-project-ssh-keys": "true",
+            "enable-guest-attributes": "true",
+            "startup-script": "docker run --privileged -v /dev/disk/by-id:/dev/disk/by-id:ro v2v -vv --log file=/dev/ttyS1,level=debug --log file=/dev/ttyS2,level=trace --report file=/dev/ttyS3 --report file=http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/v2v/report,headers=Metadata-Flavor=Google --drive file=/dev/disk/by-id/google-workload-disk1 --in-place"
```
EricEdens, line 48: To make this more readable, while retaining the inline format, you can put it in its own file, and use "source vars" as described here:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/docs/daisy-workflow-config-spec.md

daisy_integration_tests/v2v_adapt_windows_byol.wf.json
```diff
@@ -0,0 +1,92 @@
...
+  "Steps": {
+    "create-disk": {
+      "CreateDisks": [
+        {
+          "Name": "${worker_disk}",
+          "SourceImage": "projects/v2v-dev/global/images/family/v2v-candidate",
+	        "SizeGb": "200",
+          "Type": "pd-standard",
+          "Licenses": [
+            "https://compute.googleapis.com/compute/v1/projects/vm-options/global/licenses/enable-vmx"
```
EricEdens, line 24: To enable nested KVM, you also need to specify a CPU platform of haswell of newer. Either switching to n2 instances, or specifying [minCpuPlatform](https://cloud.google.com/compute/docs/reference/rest/v1/instances) in your instance definition will achieve that.

I usually use minCpuPlatform, since I've observed quota issues in my test account when using n2.

https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances

test-infra/prow/config.yaml
```diff
@@ -231,6 +231,33 @@ periodics:
        - name: compute-image-tools-test-service-account
          secret:
            secretName: compute-image-tools-test-service-account
+ - name: ci-v2v-adapt-e2e
+   interval: 24h
```
EricEdens, line 235: This specifies how often it runs? Since Sam is iterating, and the suite is tiny, maybe increase the frequency of runs?

daisy_integration_tests/v2v_adapt_e2e.test.gotmpl
```diff
@@ -0,0 +1,26 @@
+{
```
dntczdx, line 1: I expect to see a lot of v2v tests (>30, let me know if it's not the case), so we can consider 2 things:
1. put them in a subfolder
2. estimate and increase the quota of the test proj, or find a way to limit the parallel num

test-infra/prow/config.yaml
zoran15: Increase to 12hrs for the time being. The test count will match translate test count. We can tweak the frequency later on.

daisy_integration_tests/v2v_adapt_e2e.test.gotmpl
zoran15: The idea is to reuse existing translate tests by providing translate workflow as a variable. The only new files will be the actual adapt/translate workflows, which, so far, have only a couple of variants. If this grows, I'll consider extracting them outside the tests folder.

Regarding the parallel test execution. I've reduced the parallel count to 10. The frequency is currently 12hrs so this should add 12.5% resource use compared to the existing translate tests which run every 3hr and have parallel count of 30. Note that there are other tests so the real resource add is probably half of 12.5% (guestimate).

daisy_integration_tests/v2v_adapt_windows_byol.wf.json
zoran15: Specified minCpuPlatform as "Intel Haswell"

daisy_integration_tests/v2v_adapt_windows.wf.json
zoran15: Extracted to a file

daisy_integration_tests/scripts/v2v_adapt.sh
```diff
@@ -0,0 +1 @@
+docker run --privileged -v /dev/disk/by-id:/dev/disk/by-id:ro v2v -vv --log file=/dev/ttyS1,level=debug --log file=/dev/ttyS2,level=trace --report file=/dev/ttyS3 --report file=http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/v2v/report,headers=Metadata-Flavor=Google --drive file=/dev/disk/by-id/google-workload-disk1 --in-place
```
EricEdens, line 1: Usually you'd see a shebang line, a license, and the long line split up with slashes (\), and termination of the file with a newline. Did you need to omit those to make the `--byol` work below?

Example of the first two: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_integration_tests/scripts/post_translate_test.sh

zoran15: Yes. I could add a parameter to the script, but let's leave that for later as more tests are added as I feel this won't be the final version of this test suite.

EricEdens: sgtm.

PR #1332, opened 2020-08-25
Deploying image import wrappers to Artifact registry

zoran15:
Deploying image import wrappers to Artifact registry (two zones to start with for POC purposes, all zones and all wrappers to follow) to allow for regional wrapper deployment for reliability purposes.

### DIFF
--- a/cli_tools_cloudbuild.yaml
+++ b/cli_tools_cloudbuild.yaml
@@ -89,6 +89,10 @@ steps:
   args: 
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$_RELEASE
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$_RELEASE
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
+  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$_RELEASE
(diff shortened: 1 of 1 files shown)


cli_tools_cloudbuild.yaml
```diff
@@ -89,6 +89,10 @@ steps:
   args: 
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$_RELEASE
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$_RELEASE
```
EricEdens, line 92: Looking forward, how will this look on the consumer side? For example, in [1], the step uses `gcr.io/compute-image-tools/gce_vm_image_import:release`. Will the docs get updated to show each regional option? I'm curious about gcloud as well -- would gcloud pick the regional artifact based on the specified region/zone?

1. https://cloud.google.com/compute/docs/import/importing-virtual-disks#bootable

cli_tools_cloudbuild.yaml
```diff
@@ -89,6 +89,10 @@ steps:
...
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
+  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$_RELEASE
+  - --destination=us-west1-docker.pkg.dev/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
```
dntczdx, line 95: How's the final URQL look like? The current .pkg.dev looks like dev env...

PR #1335, opened 2020-08-28
fix bug, use date

gaohannk:


gaohannk:
Close PR as it's not needed.

PR #1338, opened 2020-09-01
[UEFI-compatible] auto detect & tag UEFI compatible

dntczdx:
Cont. of https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1322

Motivation: user may not have enough knowledge to determine whether their disk is UEFI-boot. Even they can determine it, gcloud & pantheon surface doesn't support importing a image with UEFI_COMPATIBLE support. (Only direct API supports that, which is not the main stream). We automatically detect it and mark it UEFI_COMPATIBLE based on the partition info.

Impl: after inflation stage & before translation stage, there is a inspection & mutation stage. We check the partition info in inspection stage to explore EFI partition. If any, we return a flag. The mutation stage will consume it: if it's true, we will create a copy of the inflated disk with an additional guestOSFeature=UEFI_COMPATIBLE.

Changes including e2e tests, which covered:
MBR+BIOS (linux + windows)
MBR+UEFI (linux)
GPT+Hybrid (linux)
GPT+UEFI (linux + windows)

Not covered: GPT+BIOS, since GCVP doesn't support that.


cli_tools/common/disk/inspect.go
```diff
@@ -31,10 +32,13 @@ type Inspector interface {
 	// Inspect finds partition and boot-related properties for a disk and
 	// returns an InspectionResult. The reference is implementation specific.
 	Inspect(reference string) (InspectionResult, error)
+	GetWorkflow() *daisy.Workflow
```
EricEdens, line 35: This is a considerable leak of implementation details :)

```suggestion
	Cancel() 
```

cli_tools/common/disk/inspect.go
```diff
@@ -55,8 +59,17 @@ type defaultInspector struct {
...
+func (inspector defaultInspector) Inspect(reference string) (ir InspectionResult, err error) {
 	inspector.wf.AddVar("pd_uri", reference)
-	err := inspector.wf.Run(context.Background())
-	return InspectionResult{}, err
+	err = inspector.wf.Run(context.Background())
+	if err != nil {
+		return
+	}
+
+	ir.HasEFIPartition, _ = strconv.ParseBool(inspector.wf.GetSerialConsoleOutputValue("has_efi_partition"))
```
EricEdens, line 69: It's *really* dangerous to throw away errors like this if you don't have metrics and alarms. My intuition is to propagate the error and fail the import. (If you silently throw it away, then most likely all UEFI imports are broken and will be unbootable)

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -162,24 +174,28 @@ func (i *importer) getCtxError(ctx context.Context) (err error) {
...
 		}
 	}
 }
 
 func (i *importer) buildLoggable() service.Loggable {
 	return service.SingleImageImportLoggable(i.pd.sourceType, i.pd.sourceGb, i.pd.sizeGb,
 		i.pd.matchResult, i.pd.inflationType, int64(i.pd.inflationTime.Milliseconds()),
-		int64(i.pd.shadowInflationTime.Milliseconds()), i.traceLogs)
+		int64(i.pd.shadowInflationTime.Milliseconds()), i.pd.isUEFICompatible, i.pd.isUEFIDetected,
+		i.traceLogs)
```
EricEdens, line 198: About `pd.isUEFI`:

I think you have the two different vars to support this metric. If so, then I think it's time to decouple the logging of key/value pairs. Right now, steps can expose traceLogs, and they're collected after running. You'd add another method to expose a map of key/value pairs, collecting them after each step runs.

Two big reasons:
1. SingleImageImportLoggable's signature is getting large, which leads to inadvertently swapped parameters.
2. Exposing implementation details on domain objects just for logging.

Pragmatically, if it were me, I'd do this to keep it simple:
1. Keep `SingleImageImportLoggable` with the high-level variables such as format, size, timing, etc. 
2. Add a method to insert additional key/value pairs
3. within `buildLoggable`, iterate over the key/value pairs that you collected from each step, and insert them.

daisy_workflows/image_import/inspection/src/boot_inspect/cli.py
```diff
@@ -45,6 +51,24 @@ def _output_human(results: model.InspectionResults):
   _output_json(results, indent=4)
 
 
+def _inspect_uefi(device):
+  try:
+    os.system('fdisk -l {} -o type'.format(device))
+    stream = os.popen('fdisk -l {} -o type'.format(device))
+    output = stream.read()
```
EricEdens, line 58: The modern way of starting a process is `subprocess.run` -- it handles stdout, stderr, and the process's return code. 

I *think* the debian 9 worker has Python 3.5: https://docs.python.org/3.5/library/subprocess.html

(`ps.open` here drops stderr and ignores the return code)

Also, you've got a duplicate call to fdisk :)

daisy_workflows/image_import/inspection/src/boot_inspect/cli.py
```diff
@@ -45,6 +51,24 @@ def _output_human(results: model.InspectionResults):
...
+
+    # 1. For GPT, the ESP output should be "EFI System", which matches
+    # partition GUID "C12A7328-F81F-11D2-BA4B-00A0C93EC93B".
+    # 2. For MBR, the ESP output should be "EFI (FAT-12/16/32)", which matches
+    # partition type "0xef"
+    if "EFI" in output:
+      return True
```
EricEdens, line 65: Log the output? Writing it to the console should end up in the trace logs.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -59,9 +64,18 @@ type testCase struct {
 
 	// Whether the image-under-test is expected to have the OS config agent installed.
 	osConfigNotSupported bool
+
+	// Expect to see all given strings in guestOsFeatures
+	expectedGuestOsFeatures []string
```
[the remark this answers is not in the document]
dntczdx: done

cli_tools/common/disk/inspect.go
```diff
@@ -59,8 +60,9 @@ type defaultInspector struct {
 // Inspect finds partition and boot-related properties for a GCP persistent disk, and
 // returns an InspectionResult. `reference` is a fully-qualified PD URI, such as
 // "projects/project-name/zones/us-central1-a/disks/disk-name".
-func (inspector defaultInspector) Inspect(reference string) (ir InspectionResult, err error) {
+func (inspector *defaultInspector) Inspect(reference string, inspectOS bool) (ir InspectionResult, err error) {
```
zoran15, line 63: Add docs for the new flag (inspectOS)

cli_tools/common/utils/logging/service/daisy_loggable.go
```diff
@@ -37,6 +37,14 @@ func (w workflowLoggable) GetValue(key string) string {
 	return w.wf.GetSerialConsoleOutputValue(key)
 }
 
+func (w workflowLoggable) GetValueAsBool(key string) bool {
+	v, err := strconv.ParseBool(w.wf.GetSerialConsoleOutputValue(key))
+	if err != nil {
```
zoran15, line 42: Can we propagate this error up the stack?

cli_tools/common/utils/logging/service/literal_loggable.go
```diff
@@ -17,18 +17,23 @@ package service
...
 func (w literalLoggable) GetValueAsInt64Slice(key string) []int64 { return w.int64s[key] }
 
 func (w literalLoggable) ReadSerialPortLogs() []string { return w.traceLogs }
 
 // SingleImageImportLoggable returns a Loggable that is pre-initialized with the metadata
 // fields that are relevant when importing a single image file.
-func SingleImageImportLoggable(fileFormat string, sourceSize, resultSize int64, matchResult string, inflationTypeStr string, inflationTimeInt64 int64, shadowInflationTimeInt64 int64, traceLogs []string) Loggable {
+func SingleImageImportLoggable(fileFormat string, sourceSize, resultSize int64, matchResult string,
+	inflationTypeStr string, inflationTimeInt64 int64, shadowInflationTimeInt64 int64,
+	isUEFICompatibleImageBool bool, isUEFIDetectedBool bool, traceLogs []string) Loggable {
```
zoran15, line 36: Consider refactoring to a builder as 10+ args is quite a lot for a func. Easy to make a mistake as well.

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,79 @@
...
+package importer
+
+import (
+	"log"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/disk"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+type diskInspectionProcessor struct {
```
zoran15, line 24: add docs on what this processor does

cli_tools/gce_vm_image_import/importer/disk_mutation_processor.go
```diff
@@ -0,0 +1,79 @@
...
+		return pd, nil
+	}
+
+	// Due to GuestOS features limitations, a new disk needs to be created to add the additional "UEFI_COMPATIBLE"
+	// and the old disk will be deleted.
+	diskName := fmt.Sprintf("disk-%v-uefi", p.args.ExecutionID)
+	err := p.computeDiskClient.CreateDisk(p.args.Project, p.args.Zone, &compute.Disk{
+		Name:            diskName,
+		SourceDisk:      pd.uri,
+		GuestOsFeatures: []*compute.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}},
```
zoran15, line 48: What of other features are already on the original PD? Shouldn't they be copied over and extended with UEFI_COMPATIBLE?

cli_tools/gce_vm_image_import/importer/disk_mutation_processor.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+	if err != nil {
+		return pd, daisy.Errf("Failed to create UEFI disk: %v", err)
+	}
+	log.Println("UEFI disk created: ", diskName)
+
+	// Cleanup the old disk after the new disk is created.
+	cleanupDisk(p.computeDiskClient, p.args.Project, p.args.Zone, pd)
```
zoran15, line 56: This is actually disk delete. I think that's closer to what's being done. Cleanup could also mean the disk continues to exist but is zeroed our or similar.

cli_tools_e2e_test/common/assert/assert.go
```diff
@@ -0,0 +1,47 @@
...
+	"log"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_e2e_test/common/utils"
+	"github.com/GoogleCloudPlatform/compute-image-tools/go/e2e_test_utils/junitxml"
+	"google.golang.org/api/compute/v1"
+)
+
+// GuestOSFeatures asserts expected & unexpected guestOSFeatures.
+func GuestOSFeatures(expectedGuestOsFeatures []string, unexpectedGuestOsFeatures []string, actualGuestOSFeatures []*compute.GuestOsFeature, junit *junitxml.TestCase, logger *log.Logger) {
```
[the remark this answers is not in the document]
zoran15: This is cli_tools_e2e_test/common, not cli_tools/common. The whole module is focused on tests, it already has hard dependencies on junit/logger

cli_tools_e2e_test/common/utils/utils.go
```diff
@@ -270,3 +270,32 @@ func ContainsSubString(strs []string, s string) bool {
 	}
 	return false
 }
+
+// ContainsAll checks whether all given strings in subarr exists in arr
+func ContainsAll(arr []string, subarr []string) bool {
```
zoran15, line 275: Can we have these two in common common module somewhere? It's not testing related and would be useful in prod code

cli_tools_e2e_test/common/assert/assert.go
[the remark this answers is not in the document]
EricEdens: Fair enough :)   I'd prefer to not have the hard dependencies, but I'm fine if you and Tony think the hard dependencies are the correct approach.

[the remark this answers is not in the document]
zoran15: We can still refactor it as you suggested. Just wanted to point out that it's not as a big of deal as if we were referencing junit package in e.g. import wrapper.

cli_tools_e2e_test/common/assert/assert.go
```diff
@@ -0,0 +1,47 @@
...
+	guestOsFeatures := make([]string, 0, len(actualGuestOSFeatures))
+	for _, f := range actualGuestOSFeatures {
+		guestOsFeatures = append(guestOsFeatures, f.Type)
+	}
+
+	if expectedGuestOsFeatures != nil {
+		if !utils.ContainsAll(guestOsFeatures, expectedGuestOsFeatures) {
+			junit.WriteFailure("GuestOsFeatures expect: %v, actual: %v", strings.Join(expectedGuestOsFeatures, ","), strings.Join(guestOsFeatures, ","))
+			logger.Printf("GuestOsFeatures expect: %v, actual: %v", strings.Join(expectedGuestOsFeatures, ","), strings.Join(guestOsFeatures, ","))
```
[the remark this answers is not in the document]
dntczdx: Hmm actually junit.WriteFailrue called junit.Logf, so I will still use logger.printf with junit.WriteFailrue

cli_tools/common/utils/logging/service/literal_loggable.go
```diff
@@ -17,30 +17,80 @@ package service
...
+// SetInflationAttributes sets inflation related attributes.
+func (b *SingleImageImportLoggableBuilder) SetInflationAttributes(matchResult string, inflationTypeStr string,
+	inflationTimeInt64 int64, shadowInflationTimeInt64 int64) *SingleImageImportLoggableBuilder {
+
+	b.matchResult = matchResult
+	b.inflationTypeStr = inflationTypeStr
+	b.inflationTimeInt64 = inflationTimeInt64
+	b.shadowInflationTimeInt64 = shadowInflationTimeInt64
+	return b
+}
```
zoran15, line 68: This is better than having one large func with 10+ args. 

Maybe a nit, but an even better improvement would be where builder functions would set each field directly into a loggable, instead of saving them in temporary fields within builder then copying them on Build() (Build() could just be basically a get() as it would return an already built loggable).

Something like:
b := service.NewSingleImageImportLoggableBuilder()

return b.WithMatchResult(matchResult).WithInflationType(inflationTypeStr)...WithFileFormat(fileFormat).WithUefiDetected(uefiDetected).Build()

where one With* func would be something like:

func (b *SingleImageImportLoggableBuilder) WithMatchResult(matchResult string) {
  b.loggable.strings[shadowDiskMatchResult] = matchResult
}

NewSingleImageImportLoggableBuilder() would have to initialise any maps etc. in b.loggable though.

You can even remove `With` prefix for brevity.

dntczdx: Sounds good. Modified mostly as you suggested. Still grouped them other than having a With func for each field.

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -71,88 +85,153 @@ var cases = []testCase{
...
+	}, {
+		caseName: "inspect-uefi-linux-uefi-rhel-7-from-image",
+		// image created from projects/gce-uefi-images/global/images/rhel-7-v20200403 and removed UEFI_COMPATIBLE
+		source:                  "projects/compute-image-tools-test/global/images/linux-uefi-no-guestosfeature-rhel7",
+		os:                      "rhel-7",
+		expectedGuestOsFeatures: []string{"UEFI_COMPATIBLE"},
+	}, {
+		caseName: "inspect-uefi-linux-nonuefi-debian-9",
+		// source created from projects/debian-cloud/global/images/debian-9-stretch-v20200714
+		source:                    "gs://compute-image-tools-test-resources/uefi/linux-nonuefi-debian-9.vmdk",
```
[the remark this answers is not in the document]
EricEdens: sgtm

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -59,9 +64,18 @@ type testCase struct {
...
 	osConfigNotSupported bool
+
+	// Expect to see all given strings in guestOsFeatures
+	expectedGuestOsFeatures []string
+
+	// Expect to see none of given strings in guestOsFeatures
+	unexpectedGuestOsFeatures []string
+
+	// Whether to add "inspect" arg
+	inspect bool
```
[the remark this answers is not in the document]
EricEdens: I guess that makes sense -- so the argument is that you want the tests to emulate what you'll be releasing. sgtm.

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -179,10 +179,12 @@ func (inflater *daisyInflater) inflate() (persistentDisk, error) {
...
-	uri        string
-	sizeGb     int64
-	sourceGb   int64
-	sourceType string
+	uri              string
+	sizeGb           int64
+	sourceGb         int64
+	sourceType       string
+	isUEFICompatible bool
+	isUEFIDetected   bool
```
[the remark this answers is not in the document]
EricEdens: It sounds like you want both values to be tracked in metrics, which sounds good to me. My overall feedback is to decouple your logging/metrics from these top-level objects.

Two actions:
1. Have a single bool `isUEFI` exposed on the PD struct
2. Emit your metrics independently of this domain object

zoran15 (approved):

cli_tools/common/disk/inspect.go
EricEdens: Here's a concrete proposal to ensure this doesn't fail silently: If the value returned by the inspector is not compliant with your spec (true|false), fail the import with a failure message that can be referenced back to this.

If the issue is caused by a programming bug, then all of your e2e tests will catch it immediately, and the failure message indicates where it needs to be fixed.

If the issue is caused by a regression in serial console logging (b/167607270), then we'll start seeing flaky e2e tests, and a spike in this error message.

cli_tools/gce_vm_image_import/importer/processor.go
```diff
@@ -25,35 +25,36 @@ import (
...
-	if d.Inspect && d.diskInspector != nil {
-		log.Printf("Running experimental disk inspections on %v.", pd.uri)
-		inspectionResult, err := d.diskInspector.Inspect(pd.uri)
-		if err != nil {
-			return nil, err
-		}
-		log.Printf("Inspection result=%v", inspectionResult)
+
+	processors = append(processors, newDiskInspectionProcessor(d.diskInspector, d.ImportArguments))
+	processors = append(processors, newDiskMutationProcessor(d.computeClient, d.ImportArguments))
```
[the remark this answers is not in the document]
EricEdens: Great! Latest refresh shows `newDiskMutationProcessor` still though

cli_tools/gce_vm_image_import/importer/processor_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+var uefiTests = []struct {
+	isUEFIDisk               bool
+	isInputArgUEFICompatible bool
+}{
+	{isUEFIDisk: true, isInputArgUEFICompatible: false},
+	{isUEFIDisk: false, isInputArgUEFICompatible: false},
+	{isUEFIDisk: true, isInputArgUEFICompatible: true},
+	{isUEFIDisk: false, isInputArgUEFICompatible: true},
+}
```
EricEdens, line 52: Unused in this file (it's referenced in other files, so you'll want to move/copy to the files where you're referencing it.)

cli_tools/gce_vm_image_import/importer/disk_inspection_processor_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDiskInspectionProcessor(t *testing.T) {
+	for i, tt := range uefiTests {
+		name := fmt.Sprintf("%v. inspect disk: disk is UEFI: %v, input arg UEFI compatible: %v", i+1, tt.isUEFIDisk, tt.isInputArgUEFICompatible)
```
EricEdens, line 27: Move/copy the test cases here -- shared data structures in tests have a sneaky way of causing confusing and hard-to-debug test failures.

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -0,0 +1,88 @@
...
+		return pd, nil
+	}
+
+	ir, err := p.inspectDisk(pd.uri)
+	if err != nil {
+		return pd, err
+	}
+
+	pd.isUEFICompatible = p.args.UefiCompatible || ir.HasEFIPartition
+	pd.isUEFIDetected = ir.HasEFIPartition
```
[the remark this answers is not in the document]
EricEdens: Resolving this to consolidate discussion to the other comment.

cli_tools/gce_vm_image_import/importer/uefi_processor_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUefiProcessor(t *testing.T) {
+	for i, tt := range uefiTests {
+		name := fmt.Sprintf("%v. inspect disk: disk is UEFI: %v, input arg UEFI compatible: %v", i+1, tt.isUEFIDisk, tt.isInputArgUEFICompatible)
+		t.Run(name, func(t *testing.T) {
```
EricEdens, line 28: Ditto about copy/move uefiTests

cli_tools/gce_vm_image_import/importer/uefi_processor_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUefiProcessor(t *testing.T) {
```
EricEdens, line 25:
```suggestion
func TestUefiProcessor_AddsUEFISuffixToPDName(t *testing.T) {
```

cli_tools/gce_vm_image_import/importer/uefi_processor.go
```diff
@@ -0,0 +1,82 @@
...
+	// GuestOSFeatures are immutable properties. Therefore:
+	// 1. Copy the existing disk, adding "UEFI_COMPATIBLE"
+	// 2. Update the reference
+	// 3. Delete the previous disk.
+	diskName := fmt.Sprintf("disk-%v-uefi", p.args.ExecutionID)
+	err := p.computeDiskClient.CreateDisk(p.args.Project, p.args.Zone, &compute.Disk{
+		Name:            diskName,
+		SourceDisk:      pd.uri,
+		GuestOsFeatures: []*compute.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}},
+	})
```
EricEdens, line 52: Unit test to verify the feature is added?

cli_tools/gce_vm_image_import/importer/uefi_processor.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return pd, daisy.Errf("Failed to create UEFI disk: %v", err)
+	}
+	log.Println("UEFI disk created: ", diskName)
+
+	// Cleanup the old disk after the new disk is created.
+	deleteDisk(p.computeDiskClient, p.args.Project, p.args.Zone, pd)
+
```
EricEdens, line 60: Unit test to verify it's deleted?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
```diff
@@ -16,25 +16,41 @@ package importer
...
 	"bytes"
+	"fmt"
 	"os"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestBootableDiskProcessor_Process(t *testing.T) {
```
EricEdens, line 27:
```suggestion
func TestBootableDiskProcessor_Process_WritesSourceDiskVar(t *testing.T) {
```

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
```diff
@@ -16,25 +16,41 @@ package importer
...
+	for i, tt := range uefiTests {
+		name := fmt.Sprintf("%v. inspect disk: disk is UEFI: %v, input arg UEFI compatible: %v", i+1, tt.isUEFIDisk, tt.isInputArgUEFICompatible)
```
EricEdens, line 29: Do we need all four of the test cases?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
```diff
@@ -45,9 +61,9 @@ func TestBootableDiskProcessor_PopulatesWorkflowVarsUsingArgs(t *testing.T) {
 	imageSpec.Region = "us-central"
 	imageSpec.SysprepWindows = true
 
-	actual := asMap(createAndRunPrePostFunctions(t, pd, imageSpec).workflow.Vars)
+	actual := asMap(createAndRunPrePostFunctions(t, imageSpec).workflow.Vars)
 	assert.Equal(t, map[string]string{
-		"source_disk":          "global/projects/test/pd/id",
+		"source_disk":          "", // not set yet
```
EricEdens, line 66:
```suggestion
		"source_disk":          "", // source_disk is written in process, since a previous processor may create a new disk
```

cli_tools/gce_vm_image_import/importer/processor_test.go
```diff
@@ -0,0 +1,176 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package importer
+
+import (
```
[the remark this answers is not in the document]
EricEdens: The tests that are left in this file won't hurt, so I'm fine if you leave them. It'd be good to observe the value they provide over time as processor.go changes.

cli_tools/gce_vm_image_import/importer/processor_test.go
dntczdx: done

cli_tools/gce_vm_image_import/importer/disk_inspection_processor_test.go
dntczdx: done

cli_tools/gce_vm_image_import/importer/uefi_processor_test.go
dntczdx: done

cli_tools/gce_vm_image_import/importer/uefi_processor_test.go
dntczdx: Renamed to TestUefiProcessor_RecreateDisk since suffix is the result of recreating

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
dntczdx: Renamed to ProcessUEFI since "WritesSourceDiskVar" is one thing to validate. There can be more var to validate

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
dntczdx: 2 is enough for this processor. let me remove the dup

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
dntczdx: done

cli_tools/gce_vm_image_import/importer/uefi_processor.go
dntczdx: That is added by the real API. The API in UT is a mock, so the validation is meaningless - none of our code got tested. All I can do is to verify whether the CreateDisk got called. Added that.

cli_tools/gce_vm_image_import/importer/uefi_processor.go
dntczdx: All I can do is to verify whether the DeleteDisk got called.Added.

cli_tools/gce_vm_image_import/importer/uefi_processor_test.go
```diff
@@ -0,0 +1,60 @@
...
+		{isUEFIDisk: false, isInputArgUEFICompatible: false},
+		{isUEFIDisk: true, isInputArgUEFICompatible: true},
+		{isUEFIDisk: false, isInputArgUEFICompatible: true},
+	}
+
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+	mockComputeClient := mocks.NewMockClient(mockCtrl)
+	mockComputeClient.EXPECT().CreateDisk(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+	mockComputeClient.EXPECT().DeleteDisk(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
EricEdens, line 42: any reason to not verify the arguments? (Given you confidence that the guestOSFeature was included)

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
```diff
@@ -16,25 +16,48 @@ package importer
...
 	"bytes"
+	"fmt"
 	"os"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestBootableDiskProcessor_ProcessUEFI(t *testing.T) {
```
EricEdens, line 27: Earlier I suggested `TestBootableDiskProcessor_Process_WritesSourceDiskVar` since that helped clarify what you're testing. You mentioned there'd be other variables to verify? Are you still planning on adding those?

Also, if you'd not put the expectation in the test name, can you write a comment describing the test's intent.

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
```diff
@@ -16,25 +16,48 @@ package importer
...
+		t.Run(name, func(t *testing.T) {
+			args := ImportArguments{
+				WorkflowDir:    "../../../daisy_workflows",
+				OS:             "ubuntu-1804",
+				UefiCompatible: tt.isInputArgUEFICompatible,
+			}
+			p, err := newBootableDiskProcessor(args)
+			assert.NoError(t, err)
+			_, err = p.process(persistentDisk{uri: "uri"})
+			assert.NotEmpty(t, p.(*bootableDiskProcessor).workflow.Vars["source_disk"].Value)
```
EricEdens, line 46: Any reason to not assert the actual value?

cli_tools/gce_vm_image_import/importer/bootable_disk_processor_test.go
EricEdens: In bootable_disk_processor, I'm not logic that differentiates between whether the disk has an EFI partition. Can you point me to that?

cli_tools/gce_vm_image_import/importer/processor.go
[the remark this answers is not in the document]
EricEdens: Still seeing `newDiskMutationProcessor`

dntczdx:
Refactored how to call the loggable to emit metrics

EricEdens (approved):
Nice!

PR #1341, opened 2020-09-01
Add tests for EL 7.8 and EL 8.2

EricEdens:
This includes images for EL 7.8 and EL 8.2 to ensure that the centos-7, centos-8, rhel-7, and rhel-8 `--os` flags work with the new minor versions.

Ran all of the EL tests, and they pass.

### DIFF
--- a/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
+++ b/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
@@ -96,14 +96,36 @@ var cases = []testCase{
 		source:        "projects/compute-image-tools-test/global/images/debian-9-translate",
 		os:            "opensuse-15",
 		expectedError: "\"debian-9\" was detected on your disk, but \"opensuse-15\" was specified",
+	},
+	// EL
+	{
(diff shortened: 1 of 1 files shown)


cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -96,14 +96,36 @@ var cases = []testCase{
...
+		caseName: "el-centos-8-0",
+		source:   "projects/compute-image-tools-test/global/images/centos-8-import",
+		os:       "centos-8",
+	}, {
+		caseName: "el-centos-8-2",
+		source:   "projects/compute-image-tools-test/global/images/centos-8-2",
+		os:       "centos-8",
 	}, {
-		caseName:  "rhel-7-uefi",
+		caseName:  "el-rhel-7-uefi",
```
zoran15, line 114: Since`el` is already included in `rhel`, why the duplication in case names? Is it just for sorting purposes?

EricEdens:
Yeah, sorting. Also, it helped with running all of the enterprise linux
tests by specifying a pattern of `el-`

PR #1347, opened 2020-09-09
Update 3-disk Ubuntu OVF import test verification script

zoran15:
Bring 3-disk Ubuntu OVF import test verification script in line with post_translate_test.sh to fix flaky failures like unable to ping google.com

### DIFF
--- a/cli_tools_e2e_test/gce_ovf_import/scripts/ovf_import_test_ubuntu_3_disks.sh
+++ b/cli_tools_e2e_test/gce_ovf_import/scripts/ovf_import_test_ubuntu_3_disks.sh
@@ -15,8 +15,56 @@
 
 # Tests to validate if Ubuntu OVF with 3 disks was imported successfully.
 
+
+# To disable checking for the OS config agent, add an instance metadata
+# value of osconfig_not_supported: true.
+url=http://metadata.google.internal/computeMetadata/v1/instance/attributes/osconfig_not_supported
+if command -v curl; then
+  OSCONFIG_NOT_SUPPORTED="$(curl --header "Metadata-Flavor: Google" $url)"
+elif command -v wget; then
+  OSCONFIG_NOT_SUPPORTED="$(wget --header "Metadata-Flavor: Google" -O - $url)"
+else
+  echo "TestFailed: Either curl or wget is required to run test suite."
+  exit 1
+fi
+
 FAIL=0
 FAILURES=""
+
+function wait_for_connectivity {
+  if grep -q 'CentOS release 6' /etc/centos-release && command -v nmcli; then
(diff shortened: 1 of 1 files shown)


dntczdx:
Would you explain what is the root cause and what is fixed in this PR?

zoran15:
> Would you explain what is the root cause and what is fixed in this PR?

test verification script for 3-disk Linux OVF imports is based on post_translate_test.sh. post_translate_test.sh was updated since the 3-disk version was created, but the 3-disk version was never updated with improvements and new features. This brings the 3-disk version on the same level as post_translate_test.sh

The reason we are doing it now is that OVF import tests are flaky with not being able to ping google.com sometimes. This change has a retry implemented when pinging google.com so this should eliminate flakiness.

cli_tools_e2e_test/gce_ovf_import/scripts/ovf_import_test_ubuntu_3_disks.sh
```diff
@@ -29,63 +77,43 @@ function fail {
...
+function check_metadata_connectivity {
   status "Checking metadata connection."
-  ping -q -c 2 metadata.google.internal
-  if [[ $? -ne 0 ]]; then
-    fail "Failed to connect to metadata.google.internal."
-  fi
+  for i in $(seq 1 20) ; do
+    ping -q -w 10 metadata.google.internal && return 0
+    status "Waiting for connectivity to metadata.google.internal."
+    sleep $((i**2))
```
dntczdx, line 85: If I understood it correctly, the total wait can be up to 1+4+9+16+....+400, which is too long

zoran15: It's the exact same logic as post_translate_test.sh. The only difference between the two scripts is the check for the 2 data disks.

Why is it too long?

dntczdx: The sum is 2870, which is 40+min. Is it worth waiting for so long for internal service recovery?

zoran15: This will only run in our E2E tests, so I don't see an issue for waiting that long. Whether it's too long, maybe.

FWIW, this exact code was already reviewed and merged 6 months ago: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1153

dntczdx: Sounds good. Let's keep them sync, and we can discuss whether it's proper separately.

dntczdx (approved):

PR #1350, opened 2020-09-10
Update multi-regional mapping in import/export wrappers

zoran15:
Update multi-regional mapping in import/export wrappers when determining the best zone to run import in when multi-regional storage is used: add new regions and sort based on VM cost.


EricEdens (approved):

EricEdens:
I'm curious, what's the impact if this goes stale? (Either in terms of missing regions, or having stale cost data?)

zoran15:
> I'm curious, what's the impact if this goes stale? (Either in terms of missing regions, or having stale cost data?)

* Missing regions: if all other regions are down and a missing region is up, import will fail as it didn't attempt to run in it. Highly unlikely all other regions will be down so not a concern, but it's good to have an up-to-date list of regions just in case.
* Stale cost data: import costs to the user will be more than the lowest price possible (if the cheapest region was selecged).

PR #1352, opened 2020-09-12
SLES import: Support conversion to on-demand

EricEdens:
This PR adds:
 - Support for importing on-demand SLES and SLES for SAP guests.
   The imported guest is not required to have an active
   SLES subscription.
 - 15.2 support for SLES, SLES for SAP, and OpenSUSE
 - Parity for SLES and SLES for SAP: 12.4, 12.5, 15.1, and 15.2
 
## What's supported

All importable versions of SLES are supported:
1. SLES 12.4, 12.5
2. SLES for SAP 12.4, 12.5
3. SLES 15.1, 15.2
4. SLES for SAP 15.1, 15.2

## OS targets

Eight targets are available for the `--os` flag, covering all combinations of:
- Major version: 12, 15
- Flavor: SLES, SLES for SAP
- Subscription: BYOL, convert to on-demand

See `cli_tools/common/utils/daisy/daisy_utils.go` for the new targets.

## Noteworthy changes

* The on-demand conversion runs in a chroot. This is required since SLES's registration code
  fingerprints the instance. Running in a guestfs appliance fails registration.
* Similarly, to support fingerprinting, the Debian 9 import worker includes
  the license that will be applied to the final image.

## Testing:
1. Added e2e tests:
 * Each minor version has an unregistered image that is imported and verified (8 total)
 * Each major version has one registered image that is imported as BYOL (4 total)
2. Ran SLES and OpenSUSE tests from [ci-daisy-e2e](https://k8s-testgrid.appspot.com/google-gce-compute-image-tools#ci-daisy-e2e)
3. Added Python unit test support in the `utils` package in linux_common.

PR #1356, opened 2020-09-16
SUSE import: Support 15.2; Parity for SLES for SAP

EricEdens:
After this PR, the following versions of SLES will be importable:

  | SLES | SLES for SAP | OpenSUSE
-- | -- | -- | --
12.4 | Existing | Existing | *
12.5 | Existing | New | *
15.1 | Existing | New | Existing
15.2 | New | New | New

`*: Not importable; Version is EOL.`

This PR is split one of two from https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1352, which I'm splitting to simplify the review process.

Testing:

* Added e2e tests:
   * Each major version has one registered image that is imported as BYOL (4 total), this PR.
   * Each minor version has an unregistered image that is imported and verified (8 total), in the on-demand PR split from https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1352
* Ran existing SLES and OpenSUSE tests from ci-daisy-e2e
* Added Python unit test support in the utils package in linux_common. Currently they run manually; once this is pushed I'll add a prow config similar to [the boot inspect pytest](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml#L16)


cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -41,7 +41,8 @@ func (p *diskInspectionProcessor) process(pd persistentDisk,
 		return pd, err
 	}
 
-	pd.isUEFICompatible = p.args.UefiCompatible || ir.HasEFIPartition
+	// Hardcoding to false due to b/168671324
+	pd.isUEFICompatible = false
```
EricEdens, line 45: This could also land in a PR of its own.

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -82,6 +84,18 @@ var (
 	privacyTagRegex = regexp.MustCompile(`(\[Privacy\->)|(<\-Privacy\])`)
 )
 
+// OSChoices returns the supported OS identifiers, sorted.
+func OSChoices() []string {
```
zoran15, line 88: GetSortedOSIDs or similar. Choice is not specific enough. Also, might get confused with a field without Get

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -82,6 +84,18 @@ var (
...
+	choices := make([]string, len(osChoices))
+	i := 0
```
zoran15, line 90: Minor: you can avoid having a counter: https://yourbasic.org/golang/slice-of-keys-values-from-map/

daisy_workflows/image_import/suse/translate.py
```diff
@@ -25,98 +24,126 @@
...
 class _Package:
   """A Zypper package to be installed.
 
   Attributes:
-      name (str): Zypper's name for the package.
-      gce (bool): Is the package specific to GCP? When set to true, this
-                  package will only be installed if when `install_gce_packages`
-                  is also true.
+      name: Zypper's name for the package.
+      gce: Is the package specific to GCP? When set to true, this
```
zoran15, line 36: Any reason for removing types from the doc string?

EricEdens: Good question -- I was reading go/python-tips/035, and it encouraged using PEP-484 annotations instead of docstring annotations.

daisy_workflows/linux_common/utils/configs.py
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import shlex
+
+
+def update_grub_conf(original: str, **configs: str) -> str:
```
dntczdx, line 18: To which OSs this func can apply? Can it fix the hybrid issue?

EricEdens: All of the linux translators update the grub config; this new function is only used for SLES at the moment.

> Can it fix the hybrid issue?

Talked offline -- this PR doesn't address any of the findings in b/168671324 with respect to SLES 12.5's hybrid booting.

dntczdx (approved):
Looks good from my side.

zoran15 (approved):

EricEdens:
Adds support for 15.2 (SLES, SLES SAP, and OpenSUSE), and adds increases supported versions of SLES SAP so that SLES and SLES SAP have the same supported versions.

PR #1361, opened 2020-09-17
Deploy image import wrapper Docker image to Artifact registry repos in Asia/Australia

zoran15:


dntczdx (commented):
Since it's not released after you added Europe's setting, may I know what's the benifits splitting them into multiple PRs?

zoran15:
> Since it's not released after you added Europe's setting, may I know what's the benifits splitting them into multiple PRs?

Monitor build times due to having to build a Docker image for each region (because of a Kaniko bug)

dntczdx:
> > Since it's not released after you added Europe's setting, may I know what's the benifits splitting them into multiple PRs?
> 
> Monitor build times due to having to build a Docker image for each region (because of a Kaniko bug)

Sounds good. I don't know much details about this config; if it's executed in sequence (I guess so since you are adding timeout every time), is it possible to set timeout for each sections?

zoran15:
> > > Since it's not released after you added Europe's setting, may I know what's the benifits splitting them into multiple PRs?
> > 
> > 
> > Monitor build times due to having to build a Docker image for each region (because of a Kaniko bug)
> 
> Sounds good. I don't know much details about this config; if it's executed in sequence (I guess so since you are adding timeout every time), is it possible to set timeout for each sections?

There's no need to time out each step. I just didn't want to dump 100 new build steps in one go (4 wrappers x 23 regions), but do a more gradual approach. Hopefully Kaniko team will fix their bug so we can specify multiple repos in one build steps. Our build times won't increase nearly as much at that point.

dntczdx (approved):
LGTM

PR #1362, opened 2020-09-17
Add e2e test to validate rollout of ubuntu-2004 to gcloud.

EricEdens:
We're preparing to add `ubuntu-2004` as an option for `--os` in gcloud. This PR creates an e2e test that imports an `ubuntu-2004`, which we will use to track the deployment of the new flag value.

## Testing
- Ran the  "ubuntu-2004 should be an option for --os" suite, and confirmed that the `gce_vm_image_import` case passes, while the gcloud cases fail. Prior to failing, they emit: "argument --os: Invalid choice: 'ubuntu-2004'. Did you mean 'ubuntu-1804'?"

### DIFF
--- a/cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
+++ b/cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	testSuiteName = "CLI"
+	testSuiteName = "ImageImportCLI"
 )
 
 // CLITestSuite ensures that gcloud and the wrapper have consistent behavior for image imports.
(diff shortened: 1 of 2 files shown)


cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -132,6 +137,39 @@ func runImageImportOSTest(ctx context.Context, testCase *junitxml.TestCase, logg
...
+				"-source_image", sourceImageURI,
+				"-zone", testProjectConfig.TestZone,
+			},
+			utils.GcloudProdWrapperLatest: {"beta", "compute", "images", "import", imageName, "--quiet",
+				"--docker-image-tag=latest", "--os", osID, "--project", testProjectConfig.TestProjectID,
+				"--source-image", sourceImageURI,
+				"--zone", testProjectConfig.TestZone,
+			},
+			utils.GcloudLatestWrapperLatest: {"beta", "compute", "images", "import", imageName, "--quiet",
+				"--docker-image-tag=latest", "--os", osID, "--project", testProjectConfig.TestProjectID,
```
dntczdx, line 162: Just noticed all tests here are "beta". Is that suppose to be? If not we may need to make it an arg just as Zoran did in another PR today

zoran15: --docker-image-tag is in beta only. If a test needs to use the latest wrapper, we have to use beta gcloud. 

Can we have --docker-image-tag in GA, still hidden?

EricEdens: Good discussion -- I realized I actually _don't_ want to use the latest wrapper, since this is intended to verify that gcloud will work once pushed.

Updated to remove `--docker-image-tag=latest`

zoran15 (approved):

PR #1364, opened 2020-09-18
EL import: Don't fail when unexpected directory in /lib/modules

EricEdens:
Prior to this, if a user had extra directories in /lib/modules, the import would fail. In EL6:  'modules must be specified using absolute paths.'  In EL8: 'depmod: ERROR: Bad version passed'

Root cause: Although each directory in /lib/modules typically corresponds to a kernel version, that may not always be true. kernel-abi-whitelists, for example, creates extra directories in /lib/modules. Prior to this change, each directory was blindly passed to depmod. 

## Testing
- Added a new e2e test that repro'd  the issue, and passes after the fix.

### DIFF
--- a/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
+++ b/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
@@ -160,7 +160,7 @@ var basicCases = []*testCase{
 	// EL - Error cases
 	{
 		caseName: "el-allow-extra-dirs-in-lib-modules", // b/168774581
-		source:   "projects/compute-image-tools-test/global/images/el-depmod-lib-modules",
+		source:   "projects/compute-image-tools-test/global/images/el-depmod-extra-lib-modules",
 		os:       "centos-8",
 		inspect:  true,
 	},
--- a/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
+++ b/cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
@@ -117,6 +117,7 @@ var basicCases = []*testCase{
 		expectedError: "\"debian-9\" was detected on your disk, but \"opensuse-15\" was specified",
 		inspect:       true,
 	},
+
 	// EL
 	{
 		caseName: "el-centos-7-8",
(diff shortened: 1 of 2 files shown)


cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -154,7 +155,18 @@ var basicCases = []*testCase{
 		source:   "projects/compute-image-tools-test/global/images/rhel-8-2",
 		os:       "rhel-8",
 		inspect:  true,
-	}, {
+	},
+
+	// EL - Error cases
+	{
+		caseName: "el-allow-extra-dirs-in-lib-modules", // b/168774581
+		source:   "projects/compute-image-tools-test/global/images/el-depmod-lib-modules",
```
zoran15, line 163: minor: this image could have been named e.g. el-depmod-extra-lib-modules to indicate there are more lib modules than just the ones representing the kernel

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -161,8 +162,29 @@ def DistroSpecific(g: guestfs.GuestFS):
...
+    # /lib/modules.
+    #
+    # Skip building initramfs if the directory doesn't look like a
+    # kernel version. Emulates the version matching from depmod [2].
+    #
+    # 1. https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/lib.html
+    # 2. https://kernel.googlesource.com/pub/scm/linux/kernel/git/mmarek/kmod
+    # /+/tip/tools/depmod.c#2537
+    if not re.match(r'^\d+.\d+', kver):
+      logging.debug('Skipping {}'.format(kver))
```
zoran15, line 177: minor: Add what is being skipped to the message

zoran15 (approved):

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
EricEdens: done

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: done

PR #1365, opened 2020-09-21
[VMDK API Adoption] cleanup shadow disk when error happens

dntczdx:
Fix b/169073057

Original design:
1. when main inflater finished later than shadow (which is 99% cases), no problem.
2. when main inflater finished earlier than shadow with an error, the main thread got terminated and shadow inflater missed the chance to cleanup shadow PD.

New design:
1. when main inflater finished later than shadow (which is 99% cases), still no problem.
2. when main inflater finished earlier than shadow, the main thread needs to call "shadowInfalter.cancel()" in order to:
   * cancel the shadowInfalter.Infalte(). The inflate() will try to return early, but needs to guarantee the cleanup is done
   * wait for shadowInfalter.Infalte() to finish, in order to guarantee cleanup won't be interrupted before main thread exits.

PR #1374, opened 2020-09-24
[uefi] exclude hybrid disk from the scope of UEFI_COMPATIBLE

dntczdx:
1. exclude "hybrid mbr" disk by detecting with gdisk; exclude "protective mbr" with BIOS boot partition
2. adjust e2e to expect "UEFI_COMPATIBLE" not to show up for imported hybrid image
3. root_fs is not populated yet since it's a preparation of the next stage - make hybrid work for btrfs (and potentially more), and it's nontrivial.


cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -254,6 +254,21 @@ type OutputInfo struct {
 	IsUEFICompatibleImage bool `json:"is_uefi_compatible_image,omitempty"`
 	// Indicates whether the image is auto-detected to be UEFI compatible
 	IsUEFIDetected bool `json:"is_uefi_detected,omitempty"`
+	// InspectionResults contains metadata determined using automated inspection
+	InspectionResults InspectionResults `json:"inspection_results,omitempty"`
+}
+
+// InspectionResults contains metadata determined using automated inspection
+type InspectionResults struct {
+	// UEFIBootable indicates whether the disk has a UEFI boot loader.
```
EricEdens, line 263: This field should indicate whether the identified OS is bootable with UEFI. The next field is similar, but should indicate whether it's bootable with BIOS

cli_tools_e2e_test/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -230,55 +230,52 @@ var basicCases = []*testCase{
...
+	{
+		caseName: "inspect-uefi-linux-uefi-rhel-7",
+		// source created from projects/gce-uefi-images/global/images/rhel-7-v20200403
+		source:                  "gs://compute-image-tools-test-resources/uefi/linux-uefi-rhel-7.vmdk",
+		os:                      "rhel-7",
+		requiredGuestOsFeatures: []string{"UEFI_COMPATIBLE"},
+	}, {
+		caseName: "inspect-uefi-linux-uefi-rhel-7-from-image",
+		// image created from projects/gce-uefi-images/global/images/rhel-7-v20200403 and removed UEFI_COMPATIBLE
+		source:                  "projects/compute-image-tools-test/global/images/linux-uefi-no-guestosfeature-rhel7",
```
EricEdens, line 243: In addition, can you add test cases to [1]? That allows you to assert the inspection results themselve:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/ef8b0df743b71aa0c1edee7973277b1bbaea08e0/cli_tools_e2e_test/diskinspect/disk_inspect_test.go#L57

daisy_workflows/image_import/inspection/src/boot_inspect/cli.py
```diff
@@ -51,22 +55,31 @@ def _output_human(results: model.InspectionResults):
...
   try:
     stream = os.popen('fdisk -l {} -o type'.format(device))
     output = stream.read()
     print(output)
 
+    bios_bootable = False
+    uefi_bootable = False
+    root_fs = ""
+
+    if "BIOS boot" in output:
```
EricEdens, line 68: String matching the human-readable output of a command is typically error prone and fragile. What alternatives have you investigated? Does libguestfs expose the information that you need? Is there a shared library for fdisk that you can call through a python binding?

daisy_workflows/image_import/inspection/src/boot_inspect/cli.py
```diff
@@ -51,22 +55,31 @@ def _output_human(results: model.InspectionResults):
   _output_json(results, indent=4)
 
 
-def _inspect_uefi(device):
+def _inspect_boot_loader(device):
```
EricEdens, line 58: Since the test framework is ready, move this into one of the subdirectories and include tests for it.

cli_tools/common/disk/inspect.go
```diff
@@ -70,13 +78,14 @@ func (inspector *defaultInspector) Inspect(reference string, inspectOS bool) (ir
 		return
 	}
 
-	// TODO: Mute the UEFI detection results for now. Restore it till we got the reason of b/168671324
-	// ir.HasEFIPartition, _ = strconv.ParseBool(inspector.wf.GetSerialConsoleOutputValue("has_efi_partition"))
```
zoran15, line 74: We first need to roll out https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1365 before rolling this one out, correct?

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -41,8 +41,13 @@ func (p *diskInspectionProcessor) process(pd persistentDisk,
 		return pd, err
 	}
 
-	pd.isUEFICompatible = p.args.UefiCompatible || ir.HasEFIPartition
-	loggableBuilder.SetUEFIMetrics(pd.isUEFICompatible, ir.HasEFIPartition)
+	isHybrid := ir.UEFIBootable && ir.BIOSBootable
+	if !p.args.UefiCompatible && isHybrid {
+		log.Printf("This disk can boot with both BIOS and UEFI. It will be imported as BIOS booting. " +
+			"If you want to use UEFI booting, please specify arg '--uefi_compatible'.")
```
zoran15, line 47: This flag is currently not available in gcloud and probably it will be called -uefi-compatible (since we don't use underscores in gcloud flags). Might be confusing before we make this flag available in gcloud unless users are using API. 

It's probably a better idea to link to https://cloud.google.com/compute/docs/import/importing-virtual-disks#importing_a_virtual_disk_with_uefi_bootloader

It already documents how to achieve this with API. Once gcloud flag is available, docs will be updated.

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -41,8 +41,13 @@ func (p *diskInspectionProcessor) process(pd persistentDisk,
 		return pd, err
 	}
 
-	pd.isUEFICompatible = p.args.UefiCompatible || ir.HasEFIPartition
-	loggableBuilder.SetUEFIMetrics(pd.isUEFICompatible, ir.HasEFIPartition)
+	isHybrid := ir.UEFIBootable && ir.BIOSBootable
+	if !p.args.UefiCompatible && isHybrid {
+		log.Printf("This disk can boot with both BIOS and UEFI. It will be imported as BIOS booting. " +
```
zoran15, line 46: "It will be imported as BIOS booting" doesn't sound right. Maybe get Shawnique to review this message?

daisy_workflows/image_import/inspection/src/boot_inspect/cli.py
dntczdx: I moved _inspect_boot_loader into inspection.py. Would you route me to the test folder? is it inspection/tests?

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
```diff
@@ -41,8 +41,13 @@ func (p *diskInspectionProcessor) process(pd persistentDisk,
 		return pd, err
 	}
 
-	pd.isUEFICompatible = p.args.UefiCompatible || ir.HasEFIPartition
-	loggableBuilder.SetUEFIMetrics(pd.isUEFICompatible, ir.HasEFIPartition)
+	isHybrid := ir.UEFIBootable && ir.BIOSBootable
+	if !p.args.UefiCompatible && isHybrid {
+		log.Printf("This disk can boot with either BIOS or a UEFI bootloader. The default setting for booting is BIOS. " +
+			"If you want to boot using UEFI, please reference to https://cloud.google.com/compute/docs/import/importing-virtual-disks#importing_a_virtual_disk_with_uefi_bootloader'.")
```
zoran15, line 47:
```suggestion
		log.Printf("This disk can boot with either BIOS or a UEFI bootloader. The default setting for booting is BIOS. " +
			"If you want to boot using UEFI, please see https://cloud.google.com/compute/docs/import/importing-virtual-disks#importing_a_virtual_disk_with_uefi_bootloader'.")
```

cli_tools_tests/module/diskinspect/disk_inspect_test.go
```diff
@@ -75,6 +79,38 @@ func TestBootInspect(t *testing.T) {
 				Distro:       "windows",
 				Major:        "6",
 				Minor:        "1",
+				UEFIBootable: false,
+				BIOSBootable: false,
```
EricEdens, line 83: This and Debian 9 have both UEFI and BIOS as false. That looks suspicious.

daisy_workflows/image_import/inspection/src/boot_inspect/inspection.py
```diff
@@ -117,6 +117,42 @@ def inspect_device(g, device: str) -> model.InspectionResults:
...
+    part_list = g.part_list('/dev/sda')
+    for part in part_list:
+      guid = g.part_get_gpt_type('/dev/sda', part['part_num'])
+
+      # It covers both GPT "EFI Sstem" and BIOS "EFI (FAT-12/16/32)"
+      if guid == 'C12A7328-F81F-11D2-BA4B-00A0C93EC93B':
+        uefi_bootable = True
+        # TODO: detect root_fs (b/169245755)
+      if guid == '21686148-6449-6E6F-744E-656564454649':
+        bios_bootable = True
```
EricEdens, line 148: Recall that there are  four combos of firmware and partition scheme: [UEFI, BIOS]  with [MBR, GPT]. Here you cover only GPT.

My key concern is that anything that was previously bootable without UEFI_COMPATIBLE should not be marked as UEFI_COMPATIBLE. I don't know whether BIOS on GPT is currently bootable; if so, that'd be good to inspect for.

What you really want to identify is whether the inspected OS is bootable by BIOS on MBR.

Good reading:
 - https://wiki.archlinux.org/index.php/partitioning#Choosing_between_GPT_and_MBR
 - https://www.rodsbooks.com/gdisk/hybrid.html
 - https://systemd.io/DISCOVERABLE_PARTITIONS/

daisy_workflows/image_import/inspection/src/boot_inspect/inspection.py
```diff
@@ -117,6 +117,42 @@ def inspect_device(g, device: str) -> model.InspectionResults:
   )
 
 
+def inspect_boot_loader(g) -> (bool, bool, str):
+  """Finds boot-loader properties for the device using offline inspection.
```
EricEdens, line 121: A tuple is difficult to understand and test, especially when there are more than two values. In terms of unit testing, I suspect having one method per inspection will be easier. You could also consider returning an object with fields.

cli_tools/common/disk/inspect.go
```diff
@@ -38,8 +38,18 @@ type Inspector interface {
 
 // InspectionResult contains the partition and boot-related properties of a disk.
 type InspectionResult struct {
-	// HasEFIPartition indicates whether the disk has a EFI partition.
-	HasEFIPartition                    bool
+	// UEFIBootable indicates whether the disk is bootable with UEFI.
+	UEFIBootable bool
+
+	// BIOSBootable indicates whether the disk is bootable with BIOS.
+	// It's only useful for detecting hybrid (UEFI+BIOS) mode. A BIOS-only disk
```
EricEdens, line 45: My understanding of hybrid is that the disk supports BIOS booting with MBR partitioning, and plus some other GPT booting (usually I've seen GPT+UEFI).

https://www.rodsbooks.com/gdisk/hybrid.html

Do you have a different understanding?

cli_tools/common/disk/inspect.go
```diff
@@ -38,8 +38,18 @@ type Inspector interface {
 
 // InspectionResult contains the partition and boot-related properties of a disk.
 type InspectionResult struct {
-	// HasEFIPartition indicates whether the disk has a EFI partition.
-	HasEFIPartition                    bool
+	// UEFIBootable indicates whether the disk is bootable with UEFI.
```
EricEdens, line 41: This field should indicate whether the identified OS is bootable with UEFI. The next field is similar, but should indicate whether it's bootable with BIOS.

The current docs talk about the *disk*.

daisy_workflows/image_import/inspection/src/boot_inspect/cli.py
EricEdens: daisy_workflows/image_import/inspection/tests

daisy_workflows/image_import/inspection/src/boot_inspect/inspection.py
dntczdx: done

cli_tools/gce_vm_image_import/importer/disk_inspection_processor.go
dntczdx: "reference to" is not from Shawnique, let me change to "see"

cli_tools/common/disk/inspect.go
EricEdens: If that's the case, then shouldn't the variable be: "HasEFIPartition" or something like that?

zoran15 (approved):

cli_tools/common/disk/inspect.go
dntczdx: done. for "hybrid mbr" we exclude it; for "protectvie mbr" we exclude if "BIOS boot" partition exists.

cli_tools_tests/module/diskinspect/disk_inspect_test.go
```diff
@@ -75,6 +79,74 @@ func TestBootInspect(t *testing.T) {
...
+		}, {
+			// UEFI inspection test for MBR-only
+			"projects/debian-cloud/global/images/debian-9-stretch-v20200714",
+			disk.InspectionResult{
+				Architecture: "x64",
+				Distro:       "debian",
+				Major:        "9",
+				Minor:        "12",
+				UEFIBootable: false,
+				BIOSBootable: false,
```
EricEdens, line 105: When I read this I think: "The disk isn't bootable, since it can't boot with UEFI and it can't boot with BIOS."

I propose two options:

1. Update your detection to determine if a disk is bootable with MBR and BIOS 
2. Split `BIOSBootable` into two variables, documenting that a disk *may* be bootable, even when all three are false
   * GPT BIOS bootable
   * Hybrid MBR bootable

dntczdx: renamed to BIOSBootableWithHybridMBROrProtectiveMBR for better understandable. We don't need to split since we won't treat them differently.

EricEdens (approved):

PR #1373, opened 2020-09-24
Remove resources when bucket owned by different project

EricEdens:
If `scratchBucketCreator.IsBucketInProject` returns false, delete `sourceFile` if it's in the scratch bucket.

Use 'hide whitespace' to see the relevant changes to the test.

PR #1375, opened 2020-09-25
e2e test for: Remove resources when bucket owned by different project

EricEdens:
e2e tests for #1373.

This tests two scenarios (both assume that the scratch bucket is owned by a different project).

1. Disk file in scratch bucket: Expect the file is deleted.
2. Disk file **not** in scratch bucket: File should not be deleted.

To facilitate testing, I needed to move the logic of `gce_vm_image_import.main` to an importable package.

Running locally:

```
cd cli_tools_e2e_test/import
go test
```


cli_tools_e2e_test/go.mod
```diff
@@ -5,13 +5,11 @@ go 1.13
...
 	github.com/aws/aws-sdk-go v1.34.4
-	golang.org/x/net v0.0.0-20200813134508-3edf25e44fcc // indirect
-	golang.org/x/tools v0.0.0-20200813203630-136574234359 // indirect
 	google.golang.org/api v0.30.0
-	google.golang.org/genproto v0.0.0-20200813001606-1ccf2a5ae4fd // indirect
-	google.golang.org/grpc v1.31.0 // indirect
-	honnef.co/go/tools v0.0.1-2020.1.5 // indirect
 )
+
+replace github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_e2e_test/common => ./common
```
zoran15, line 15: Is this replacement needed in the final code or it's just for your local testing?

cli_tools_e2e_test/import/scratch_ownership_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/api/googleapi"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_vm_image_import/cli"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_e2e_test/common/config"
+)
+
+const (
+	// For this test to work, it needs to be run within
+	// a project *other* than compute-image-test-resources.
```
zoran15, line 34: Why is compute-image-test-resources project a factor here? Gcloud copies files from source bucket to scratch bucket, which, in the case of e2e tests, will be in compute-image-test-pool-001.

cli_tools/gce_vm_image_import/cli/cli.go
```diff
@@ -0,0 +1,145 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
```
zoran15, line 1: Why is this refactoring needed for this test? We already test image import wrapper directly in cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go where we call the binary itself.

zoran15, line 1: Why is this refactoring needed for this test? We already test image import wrapper directly in cli_tools_e2e_test/gce_image_import_export/test_suites/import/cli_tests.go where we call the binary itself.

cli_tools_e2e_test/go.mod
EricEdens: It is required for the final code; it instructs go's module resolver to use a local directory: https://golang.org/ref/mod#go-mod-file-replace

Specifically, I want the new tests to have compile-time dependency on `cli_tools` in the current code tree. The current modules in `cli_tools_e2e_test` depends on a snapshot of that directory.

cli_tools_e2e_test/import/scratch_ownership_test.go
EricEdens: I used `compute-image-test-resources` as the owner of the bucket so that the test could be run in any other project (a personal project for test, one of the pool projects, `compute image tools test`, etc.

cli_tools/gce_vm_image_import/cli/cli.go
EricEdens: resolving (duplicate of previous comment)

EricEdens: By using a compile-time dependency, rather than a snapshot dependency, this test asserts the correctness of the current codebase at a particular commit. That helps when developing locally (I'm testing what I'm working on), and provides flexibility for running this at different stages in a CI system.

cli_tools_e2e_test/go.mod
zoran15: This is go.mod for cli_tools_e2e_test package and the replacement is for cli_tools_e2e_test/common to common, which is in the same package, so it looks like a no-op to me. What am I missing?

cli_tools_e2e_test/import/scratch_ownership_test.go
zoran15: Makes sense. FYI, we don't (or shouldn't) run anything in compute-image-test-resources, it's only for storing resources.

cli_tools/gce_vm_image_import/cli/cli.go
zoran15: Our test runners are built at the same snapshot as the code they are testing, so I don't see a benefit for our prod tests. 

I can see a benefit in this refactoring when running locally, because it skips the need for `go build` on wrappers themselves and copy them to the same folder as test runner. On the other hand, these two steps can be easily automated with a simple script that also runs a test.

cli_tools_e2e_test/go.mod
EricEdens: I made `common` into a module so that I could depend on it directly, rather than depending on all of the dependencies of the other test suites. See 'cli_tools_e2e_test/common/go.mod'

zoran15: >... rather than depending on all of the dependencies of the other test suites...

Can you please elaborate why is this needed?

cli_tools/gce_vm_image_import/cli/cli.go
EricEdens: Here are some benefits I see:

- When running locally, you can execute a test in a terminal using `go test`
- In an IDE you can step through the test
- Running locally is the same as prow (they both run `go test`), so you don't have inconsistencies due to local workarounds being different than how the test is executed on the CI system
- When tests pass on testgrid, it's clear what code was being tested (the git commit of the test code is the same as the git commit of the code-under-test)
  - https://github.com/GoogleCloudPlatform/testgrid/blob/master/config.md#column-headers

cli_tools_e2e_test/go.mod
EricEdens: Tactically, this appears in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1378, where the feature is landed with the e2e test; this only works since the module in that PR depends on the current source tree of `cli_tools` rather than a pre-pushed snapshot.

More info: The `gce_*` test suites share the [same set of dependencies](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools_e2e_test/go.mod). If the dependencies don't make sense for a particular suite, the developer has two subpar options: (1) modify all test suites to work with the different dependency, (2) modify their test suite to work with the other test suites.

Currently, I'm

EricEdens:
Moving to draft while we re-work directory structures. See https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1379 .

EricEdens:
Closing to:
1. Rebase against  https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1379 .
2. Add feature to avoid recursive deletion.

Issue #1279, opened 2020-07-06
Error in replacing variables in command resize-disk


hegao12:
Hi,
I'm trying to create and resize a disk with Daisy, but I cannot use a variable in command "resize-disk".
The following is a sample workflow, my execution command and the error message.

$ cat problem-resizing-disk.wf.json 
{
    "Name":"my-try",
    "Project":  "cos-oss-interns-playground",
    "Vars":{
        "size":{"Required":true}
    },
    "Steps":{
        "create-disk":{
            "CreateDisks":[{
                "Name":"mydisk",
                "SizeGb": "10",
                "SourceImage":"projects/cos-cloud/global/images/family/cos-stable"
            }]
        },
        "resize-disk":{
                "ResizeDisks": [
                        {
                          "Name": "mydisk",
                          "SizeGb": "${size}"
                         }
                ]
        }
    },
    "Dependencies":{
        "resize-disk":["create-disk"]
    }
}

$ daisy -var:size 12 -zone=us-east1-b problem-resizing-disk.wf.json 
2020/07/06 19:10:47 error parsing workflow "problem-resizing-disk.wf.json": json: invalid use of ,string struct tag, trying to unmarshal "${size}" into int64

Thanks!

zoran15:
ResizeDisks currently doesn't support setting size from variables (unlike CreateDisks). This is something that can be added in the future. 

Is there any specific reason why you're not creating the disk with the correct size in CreateDisks step?

hegao12:
Because of the special characteristics of COS, I need extra disk space to modify the disk partitions after the instance boots.

zoran15:
I can see two solutions here:

1. Modify step_resize_disks.go in a similar way to step_create_disks.go and the way it uses disks.go which allows string values for SizeGb field.
2. Use another way to modify the size field. Depending on how you run Daisy, you could modify the field once workflow is loaded and parsed. Or, you can modify the source workflow json.

PR #1376, opened 2020-09-25
make ResizeDisk inherit Disk like CreateDisk

gaohannk:
Fix https://github.com/GoogleCloudPlatform/compute-image-tools/issues/1279


gaohannk:
The original issue has been resolved. No need to change.

PR #1379, opened 2020-09-30
CLI testing: Support module and e2e tests

EricEdens:
This modifies the directory structure to support two types of tests that share common utility code.

- e2e tests are executed on pre-compiled binaries. They validate that modules are combined correctly, and that common scenarios work as expected.
- module tests exercise a single module, typically with live dependencies. They are typically faster than e2e tests, allowing for increased coverage at a lower cost.

Prior to pushing this, I'll submit a PR to update [1], which will update the associated presubmits.

1. https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml#L116

## Testing

- Built `prowjobs_cloudbuild.yaml`: https://pantheon.corp.google.com/cloud-build/builds/49e6f867-ba7a-4a0b-9758-c4332d1c11ff?project=edens-test
- Executed tests using the three test binaries (image import/export, OVF, and windows upgrade)


EricEdens:
Prow updates: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/524

cli_tools_tests/go.mod
```diff
@@ -1,15 +1,21 @@
...
+	github.com/stretchr/testify v1.5.1
 	google.golang.org/api v0.31.0
+	honnef.co/go/tools v0.0.1-2020.1.5 // indirect
 )
 
-replace github.com/GoogleCloudPlatform/compute-image-tools/cli_tools => ../../cli_tools
+replace github.com/GoogleCloudPlatform/compute-image-tools/cli_tools => ../cli_tools
 
-replace github.com/GoogleCloudPlatform/compute-image-tools/daisy => ../../daisy
+replace github.com/GoogleCloudPlatform/compute-image-tools/daisy => ../daisy
```
zoran15, line 19: With Daisy leaving compute-image-tools repo, how do you see this working in the future?

cli_tools_tests/module/README.md
```diff
@@ -1,4 +1,4 @@
-This is an e2e test for `cli_tools/common/disk/inspect.go`.
+Module tests: Larger than unit tests, but smaller than e2e tests.
```
zoran15, line 1: E2E vs. module tests is not about size, but what is being tested. A module test can test the same behavior as unit test. In our case, the main difference is whether a whole module is tested using real data (module) or if anything is mocked (unit)

cli_tools_tests/go.mod
```diff
@@ -1,15 +1,21 @@
-module github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_integ_test/boot_inspect
+module github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_tests
```
zoran15, line 1: I thought e2e and module tests will be separate packages? We don't need to have a code dependency on cli_tools in E2E tests, only module tests need it. This dependency can potentially lead to e2e tests picking up this dependency as well (possibly by mistake).

cli_tools_tests/common/config/config.go
```diff
@@ -0,0 +1,36 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
```
zoran15, line 1: With the new folder structure, i.e.:
cli_tools_tests/module
cli_tools_tests/e2e

Where each subfolder holds certain category of tests, common seems out of place, as it could be mistaken for 'common' tests, not common code shared between different test categories. 

Is there overlap between common code in e2e and module tests? If not, we can split the code and move it into e2e/common and module/common. If there is an overlap, renaming common to indicate it doesn't actually hold tests is an option.

EricEdens:
Hmm. GitHub is keeping the old cli-tools-e2e-tests results. Closing this and opening a new PR.

cli_tools_tests/module/README.md
EricEdens: Updated; see latst in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1381

cli_tools_tests/go.mod
EricEdens: You'd remove the 'replace' and pin to a specific version of daisy.

cli_tools_tests/common/config/config.go
EricEdens: Copying to https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1381

cli_tools_tests/go.mod
EricEdens: Copying to https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1381

PR #1381, opened 2020-09-30
CLI testing: Support module and e2e tests

EricEdens:
This modifies the directory structure to support two types of tests that share common utility code. There are no logic changes.

- e2e tests are executed on pre-compiled binaries. They validate that modules are combined correctly, and that common scenarios work as expected.
- module tests exercise a single module, typically with live dependencies. They are typically faster than e2e tests, allowing for increased coverage at a lower cost.

### Previous

```
cli_tools_e2e_test
├── common
│   ├── assert
│   ├── compute
│   ├── logging
│   ├── storage
│   └── utils
├── diskinspect
│   ├── disk_inspect_test.go
│   ├── go.mod
│   ├── go.sum
│   └── README.md
├── gce_image_import_export
│   ├── main.go
│   └── test_suites
├── gce_ovf_import
│   ├── main.go
│   ├── scripts
│   └── test_suites
├── gce_windows_upgrade
│   ├── main.go
│   └── test_suites
├── go.mod
└── go.sum
```

### This PR:

```
cli_tools_tests
├── e2e
│   ├── assert.go
│   ├── gce_image_import_export
│   ├── gce_ovf_import
│   ├── gce_windows_upgrade
│   └── utils.go
├── go.mod
├── go.sum
└── module
    ├── diskinspect
    └── README.md
common
├── gcp
│   ├── image.go
│   ├── instance.go
│   └── storage.go
├── go.mod
├── go.sum
├── logging
│   └── logging.go
└── runtime
    └── config.go
```

## Testing
- Presubmits updated to reflect new directory structure: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/524
- Built `prowjobs_cloudbuild.yaml`: https://pantheon.corp.google.com/cloud-build/builds/49e6f867-ba7a-4a0b-9758-c4332d1c11ff?project=edens-test
- Executed tests using the three test binaries (image import/export, OVF, and windows upgrade)


cli_tools_tests/go.mod
```diff
@@ -1,15 +1,20 @@
-module github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_integ_test/boot_inspect
+module github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_tests
```
EricEdens, line 1: @zoran15 from #1379 

I thought e2e and module tests will be separate packages? We don't need to have a code dependency on cli_tools in E2E tests, only module tests need it. This dependency can potentially lead to e2e tests picking up this dependency as well (possibly by mistake).

cli_tools_tests/common/config/config.go
```diff
@@ -0,0 +1,36 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
```
EricEdens, line 1: @zoran15  from #1379 

With the new folder structure, i.e.:
cli_tools_tests/module
cli_tools_tests/e2e

Where each subfolder holds certain category of tests, common seems out of place, as it could be mistaken for 'common' tests, not common code shared between different test categories.

Is there overlap between common code in e2e and module tests? If not, we can split the code and move it into e2e/common and module/common. If there is an overlap, renaming common to indicate it doesn't actually hold tests is an option.

cli_tools_tests/go.mod
EricEdens: Looking back, `cli_tools` has been a dependency of the e2e tests since the original version of the package:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/3c09ffd2f054e8f65153b22a8df712507ff1e7bc/cli_tools_e2e_test/go.mod

Given that, I'm curious about your concerns regarding the e2e tests having a dependency on `cli_tools`. What are some issues that you're thinking of?

cli_tools_tests/common/config/config.go
EricEdens: Updated naming in current version:

```
cli_tools_tests
├── common
├── e2e_tests
└── module_tests
```

Regarding common code, some is specific to Tony's custom junitxml work, and that wouldn't be used in `module_tests`. But there are some utilities for creating GCP resources that are agnostic to the testing suite.

cli_tools_tests/go.mod
zoran15: The original dependency to cli_tools was when OVF import e2e tests tested the importer instead of the actual binary. Now, this dependency is used for some utils that are used for testing, but not for actual tests. As these are CLI tools e2e tests, ideally, we would be testing CLIs (binaries/gcloud), and not depend on the code in cli_tools.

cli_tools_tests/common/config/config.go
zoran15: Since we are moving folders around, maybe we can take a step back and think where this folder should naturally go. It doesn't contain actual tests, but testing framework. Maybe it should be a package of its own? Maybe put it in test-infra?

cli_tools_tests/go.mod
EricEdens: Talked offline -- there are existing dependencies on cli_tools from e2e, and we decided to leave that unchanged for this PR

cli_tools_tests/common/config/config.go
EricEdens: Talked offline -- created a new root directory for common code

cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go
```diff
@@ -29,14 +29,12 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
-	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_e2e_test/common/assert"
-	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_e2e_test/common/compute"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools_tests/e2e"
+	"github.com/GoogleCloudPlatform/compute-image-tools/common/gcp"
+	logging2 "github.com/GoogleCloudPlatform/compute-image-tools/common/logging"
```
zoran15, line 34: Just logging? Or commonlogging?

common/runtime/config.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// GetConfig fetches a value, first looking up $envKey. If that's not found, it tries
+// `gcloud config get-value $gcloudConfig`.
+func GetConfig(envKey, gcloudConfig string) string {
```
zoran15, line 26: It's not clear, without looking at the implementation, that this is fetching config using `gcloud config`. Docs and func name should be more descriptive.

cli_tools_tests/e2e/assert.go
```diff
@@ -12,17 +12,16 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-// Package assert contains e2e tests assertion functions
-package assert
+package e2e
```
zoran15, line 15: I don't see anything CLI tools e2e specific in here. Do you think this file can go to e.g. common/testing?

cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go
EricEdens: thanks; changed to 'logging' (poor automatic refactoring on my part)

common/runtime/config.go
EricEdens: If you were using this and your IDE said `GetConfig(envKey, gcloudConfig string)`, and you wrote `GetConfig("ZONE", "compute/zone")`, where would you have assumed that "compute/zone" be used?

cli_tools_tests/e2e/assert.go
EricEdens: The assertions here are highly coupled to the e2e tests since they use the custom `junitxml` testcase that is only used in the e2e tests.

common/runtime/config.go
zoran15: I probably wouldn't know how to write `GetConfig("ZONE", "compute/zone")` in the first place :)

dntczdx:
Would you summarize the impl of this change?
1. package renaming?
2. new directory & file moved?
3. other actual code change?

It's hard to find where are #3 located (which needs to be reviewed) since there are too many #1 and #2 files.
The correctness of #1 and #2 should be guaranteed by compiler.

EricEdens:
Hey @dntczdx 

> 1. package renaming?
> 2. new directory & file moved?

`cli_tools_e2e_test` is split into two packages `cli_tools_test` and `common`. 

* `cli_tools_test` has a directory for e2e tests, and a directory for module tests. In addition to tests, there are two files that contain utility code that's only used by the `e2e` tests.
* `common` includes code from `cli_tools_e2e_test` that *could* be used elsewhere. Eventually `cli_tools/common` should get consolidated here as well.

> 3. other actual code change?

There are no logic changes; all changes are related to moving, renaming, or package changes.

cli_tools_tests/e2e/assert.go
zoran15: These functions are pretty generic, I don't see anything E2E/module test specific in them. What do you think about moving them to `common/testing`?

EricEdens: `go/e2e_test_utils/junitxml` is a legacy way of formatting golang test results for prow, and therefore I'd like to keep it isolated to where it's used. Otherwise, an unsuspecting dev in the future will try to use it from `cli_tools` or `cli_tools_tests/modules` and find themselves in a pickle.

dntczdx:
> * `common` includes code from `cli_tools_e2e_test` that _could_ be used elsewhere. Eventually `cli_tools/common` should get consolidated here as well.

I believe these files (image.go, instance.go, storage/file.go, etc) will only be used by test code, since they are just a very simple wrapper of api lib, in order to call google API without injecting a existing compute/storage client. Also, it used "replace ...daisy" so it shouldn't be a dev module

dntczdx (approved):

cli_tools_tests/e2e/assert.go
zoran15: > legacy way of formatting golang test results for prow

What do you mean by legacy here? What is the current way of doing this?

EricEdens: By legacy, I mean that this code's usecase became popular, so its functionality has been absorbed into the Prow ecosystem:
* https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md
* https://github.com/jstemmer/go-junit-report

For new projects, the current approach is to use the [gotest](https://github.com/GoogleCloudPlatform/guest-test-infra/tree/master/container_images/gotest) runner. It checks out a repo, executes `go test` against a directory, and formats and uploads the results. The project owner focuses on writing tests using Go's test framework, and the platform ensures it works with testgrid and spyglass.

In addition to gotest, [gointegtest](https://github.com/GoogleCloudPlatform/guest-test-infra/pull/37) is new, and it executes tests across a parallel fleet of workers.

zoran15: It seems we can consider refactoring our tests if there's a reason to do so, or at least consider the new framework for new tests.

Regarding assert.go I still think that the new `common` package is a more appropriate spot, but I won't block this PR just because of that.

zoran15 (approved):

PR #1383, opened 2020-09-30
[scratch bucket not in project] collect metrics

dntczdx:
This is a part of the scratch alerting system: collect "scratch bucket not in project" data point.

This is a draft in order for design preview. To land it, we need to:
1. add metrcis field to proto and rollout
2. remove the "replace". split the daisy files and cli_tools files into 2 PRs.


dntczdx:
Unnecessary since we can simply filter by failure message content

PR #1387, opened 2020-10-01
Building OVF import wrapper and windows upgrader into all repos. 

zoran15:
Building OVF import wrapper and windows upgrader into all repos. Switched to kaniko executor:v1.1.0 which is the official latest version

### DIFF
--- a/cli_tools_cloudbuild.yaml
+++ b/cli_tools_cloudbuild.yaml
@@ -85,7 +85,7 @@ steps:
   dir: 'cli_tools/gce_vm_image_import'
   args: ['go', 'build', '-o=/workspace/linux/gce_vm_image_import']
   env: ['CGO_ENABLED=0']
-- name: 'gcr.io/kaniko-project/executor:kavala'
+- name: 'gcr.io/kaniko-project/executor:v1.1.0'
   args:
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$_RELEASE
   - --destination=gcr.io/$PROJECT_ID/gce_vm_image_import:$COMMIT_SHA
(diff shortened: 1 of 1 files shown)


cli_tools_cloudbuild.yaml
```diff
@@ -201,10 +201,56 @@ steps:
...
   args: ['go', 'build', '-o=/workspace/linux/gce_ovf_import']
   env: ['CGO_ENABLED=0']
-- name: 'gcr.io/kaniko-project/executor:v1.0.0'
+- name: 'gcr.io/kaniko-project/executor:v1.1.0'
   args:
   - --destination=gcr.io/$PROJECT_ID/gce_ovf_import:$_RELEASE
   - --destination=gcr.io/$PROJECT_ID/gce_ovf_import:$COMMIT_SHA
+  - --destination=us-docker.pkg.dev/$PROJECT_ID/wrappers/gce_ovf_import:$_RELEASE
+  - --destination=us-docker.pkg.dev/$PROJECT_ID/wrappers/gce_ovf_import:$COMMIT_SHA
+  - --destination=us-central1-docker.pkg.dev/$PROJECT_ID/wrappers/gce_ovf_import:$_RELEASE
```
EricEdens, line 210: What happens if a single destination is unavailable? Do the remaining destinations get deployed? Do they all fail?

zoran15: I believe they will all fail as permission check would fail so push would not get called:

https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/push.go#L104

EricEdens: Talked offline -- it'd be good to monitor how often our deployments are blocked due to region unavailability, and if it becomes an issue to innovate around it.

EricEdens (approved):

PR #1388, opened 2020-10-03
Image import: When scratch is shared, ensure file deletion isn't recursive

EricEdens:
In #1373 we added deletion of the source file when the source file is contained in a scratch bucket that is contained in a different project.

This PR adds the following:
 - Validate that the source file is a single file. If there are other files that have the same prefix, don't perform deletion.
 - Module tests

### DIFF
--- a/cli_tools/common/utils/param/param_utils.go
+++ b/cli_tools/common/utils/param/param_utils.go
@@ -85,8 +85,8 @@ func populateScratchBucketGcsPath(scratchBucketGcsPath *string, zone string, mgc
 					errorParts = append(errorParts, fmt.Sprintf("Deleted %q", file))
 				} else {
 					errorParts = append(errorParts, fmt.Sprintf(
-						"Failed to delete %q. Check with the owner of gs://%q for more information: %v",
-						file, scratchBucketName, err))
+						"Failed to delete %q: %v. Check with the owner of gs://%q for more information.",
+						file, err, scratchBucketName))
 				}
 			}
 
--- a/cli_tools/common/utils/param/param_utils.go
+++ b/cli_tools/common/utils/param/param_utils.go
@@ -80,13 +80,13 @@ func populateScratchBucketGcsPath(scratchBucketGcsPath *string, zone string, mgc
 			}
 
 			if strings.HasPrefix(file, fmt.Sprintf("gs://%s/", scratchBucketName)) {
-				err := storageClient.DeleteGcsPath(file)
+				err := storageClient.DeleteObject(file)
 				if err == nil {
 					errorParts = append(errorParts, fmt.Sprintf("Deleted %q", file))
 				} else {
 					errorParts = append(errorParts, fmt.Sprintf(
-						"Failed to delete %q. Check with the owner of gs://%q for more information",
-						file, scratchBucketName))
+						"Failed to delete %q. Check with the owner of gs://%q for more information: %v",
+						file, scratchBucketName, err))
 				}
 			}
 
--- a/cli_tools/common/utils/param/param_utils.go
+++ b/cli_tools/common/utils/param/param_utils.go
@@ -45,9 +45,14 @@ func GetProjectID(mgce domain.MetadataGCEInterface, projectFlag string) (string,
 	return projectFlag, nil
 }
 
+// populateScratchBucketGcsPath validates the scratch bucket, creating a new one if not
+// provided, and returns the region of the scratch bucket. If the scratch bucket is
+// already populated, and the owning project doesn't match `project`, then an error is returned.
+// In that case, if `file` resides in the non-owned scratch bucket and `removeFileWhenScratchNotOwned`
+// is specified, then `file` is deleted from GCS.
 func populateScratchBucketGcsPath(scratchBucketGcsPath *string, zone string, mgce domain.MetadataGCEInterface,
 	scratchBucketCreator domain.ScratchBucketCreatorInterface, file string, project *string,
-	storageClient domain.StorageClientInterface, cleanupSharedScratch bool) (string, error) {
+	storageClient domain.StorageClientInterface, removeFileWhenScratchNotOwned bool) (string, error) {
 
 	scratchBucketRegion := ""
 	if *scratchBucketGcsPath == "" {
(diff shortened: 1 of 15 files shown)


cli_tools/common/utils/storage/storage_client.go
```diff
@@ -129,6 +130,43 @@ func (sc *Client) DeleteGcsPath(gcsPath string) error {
...
+	if err == nil {
+		return daisy.Errf("Error deleting `%v`: Multiple objects with prefix", gcsPath)
+	}
+	if err != iterator.Done {
+		return daisy.Errf("Error deleting `%v`: Failed querying GCS: `%v`", gcsPath, err)
+	}
+	if firstObject.Name != objectPath {
+		return daisy.Errf("Error deleting `%v`: Object not found", gcsPath)
+	}
+	if err := sc.GetObject(bucketName, firstObject.Name).Delete(); err != nil {
```
zoran15, line 164: GetObject().Delete() should delete only one object. I don't believe it works on prefixes, so all the iteration and ensuring there's only one object is unnecessary IMO. 

One of the reason why DeleteObjects() had to iterate is because of no prefix deletion.

EricEdens: If `-source_file` is an object reference, but also a prefix for other objects, then `GetObject().Delete()` would remove the directly-referenced object, while leaving the partially-matched objects. There are some examples of this in `scrach_ownership_test`; see:
  - Test_DontDeleteSourceFile_WhenAnotherFileHasSamePrefix
  - Test_DontDeleteSourceFile_WhenDirectory when `makeDirObject` is true

I suppose there's an argument that it's alright to delete in both of these cases (it's only one file). I decided against that since neither of those cases occur from  `gcloud compute images import`.

Regarding the new library method `DeleteObject`, I'd argue that this implementation is the safest choice. Consider when a dev that doesn't read the docs, and passes a directory object ('path/' referes to an empty object), assuming that the entire directory will get removed. Without the safety of the current method, the unsuspective dev would think the directory had been cleaned up, but in reality all of the children objects are still present.

zoran15: DeleteObject() should delete one object, irrespective if another object has the same file prefix. It should not throw an error, as that makes it impossible to delete a single file in all cases with this API. We already have DeleteGcsPath() for the prefix deletion. Func name and docs should be enough to clarify that DeleteObject() will delete a single file only.

dntczdx: I think Zoran's point is that, if you found a exact match, then you don't need to care about any prefix match.

EricEdens: Talked with Zoran offline; the current version has the following modifications:
1. storage_client.DeleteObject does an exact lookup on the object path
2. Scratch cleanup only occurs when the client is gcloud.

cli_tools/common/utils/param/populator.go
```diff
@@ -45,15 +45,15 @@ type populator struct {
...
+func (p populator) PopulateMissingParameters(project *string, clientID string, zone *string,
+	region *string, scratchBucketGcsPath *string, file string, storageLocation *string) error {
 
 	if err := PopulateProjectIfMissing(p.metadataClient, project); err != nil {
 		return err
 	}
 
 	scratchBucketRegion, err := populateScratchBucketGcsPath(scratchBucketGcsPath, *zone, p.metadataClient,
-		p.scratchBucketClient, file, project, p.storageClient)
+		p.scratchBucketClient, file, project, p.storageClient, clientID == "gcloud")
```
zoran15, line 56: We should consider cleaning up for empty client IDs as well. Those are either old gcloud versions or API calls that are not following the pattern in our examples AND are importing files that they've copied to a scratch bucket. I think both of these criteria being true at the same is unlikely, though not impossible. I'm not 100% sure which approach is better here TBH.

cli_tools/common/utils/param/param_utils.go
```diff
@@ -47,7 +47,7 @@ func GetProjectID(mgce domain.MetadataGCEInterface, projectFlag string) (string,
 
 func populateScratchBucketGcsPath(scratchBucketGcsPath *string, zone string, mgce domain.MetadataGCEInterface,
 	scratchBucketCreator domain.ScratchBucketCreatorInterface, file string, project *string,
-	storageClient domain.StorageClientInterface) (string, error) {
+	storageClient domain.StorageClientInterface, cleanupSharedScratch bool) (string, error) {
```
zoran15, line 50: cleanupSharedScratch sounds like the whole scratch bucket is deleted. It also sounds like no other criteria is used to perform this operation, while in reality current project has to be a mismatch for scratch bucket project. 
Maybe: cleanupFileOnMismatchedProject?

cli_tools/common/utils/param/populator.go
EricEdens: Good news: we don't allow empty client IDs, either currently or historically:

* Current: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/gce_vm_image_import/importer/args_test.go#L75
* Previous architecture: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/v20190523/cli_tools/gce_vm_image_import/main.go#L104

cli_tools/common/utils/param/param_utils.go
EricEdens: good point. Renamed to 'removeFileFromSharedScratch'

zoran15: Also consider adding docs for it so it's even clearer what it does without looking at the code.

zoran15 (approved):

cli_tools/common/utils/param/param_utils.go
EricEdens: added docs to populateScratchBucketGcsPath

dntczdx (approved):

EricEdens:
In #1373 we added deletion of the source file when the source file is contained in a scratch bucket that is contained in a different project.

This PR adds the following:

- Validate that the source file is a single file. If there are other files that have the same prefix, don't perform deletion.
- Module tests

PR #1391, opened 2020-10-13
Windows inspection: Use marketing versions

EricEdens:
Prior to this, boot-inspect returned NT versions for Windows' major and minor versions. There are a couple of problems with this:
 1. Desktop and server use the same NT versions (server 2012r2 and desktop 8.1 have the same NT version)
 2. NT versions are sometimes reused. Server 2016 and Server 2019 use the same NT version

NT versions 6.0 and newer are included. Version mappings from https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions

### DIFF
--- a/daisy_workflows/image_import/inspection/src/boot_inspect/model.py
+++ b/daisy_workflows/image_import/inspection/src/boot_inspect/model.py
@@ -36,8 +36,6 @@ class Distro(enum.Enum):
   SLES = 'sles'
   UBUNTU = 'ubuntu'
   WINDOWS = 'windows'
-  WINDOWS_DESKTOP = 'windows-desktop'
-  WINDOWS_SERVER = 'windows-server'
 
 
 def distro_for(name: str):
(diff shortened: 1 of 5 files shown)


daisy_workflows/image_import/inspection/src/boot_inspect/model.py
```diff
@@ -36,6 +36,8 @@ class Distro(enum.Enum):
   SLES = 'sles'
   UBUNTU = 'ubuntu'
   WINDOWS = 'windows'
+  WINDOWS_DESKTOP = 'windows-desktop'
+  WINDOWS_SERVER = 'windows-server'
```
zoran15, line 40: Doesn't WINDOWS include the other two?

EricEdens: You're right, good catch -- that was a previous iteration that I intended to remove.

Removed WINDOWS_DESKTOP and WINDOWS_SERVER

zoran15 (approved):

dntczdx (approved):

PR #1066, opened 2020-01-03
Improve error message for failed EL translation.

EricEdens:
This improves error messaging for two failure modes, where the user currently sees an error such as 

1. Incorrect `--os` such that EL's translate.py gets executed. Current error message is `TranslateFailed: error: command:`.
2. Correct `--os` using BYOL, but importing a VM that doesn't have a subscription. Current error message is `TranslateFailed: error: command: Importing GPG key 0xA7317B0F`.

Testing
 - Ran:
    - daisy_integration_tests/centos_6_qcow2_translate.wf.json
    - daisy_integration_tests/centos_7_qcow2_translate.wf.json
 - Verified error message by importing a RHEL 6 image that didn't have a subscription as BYOL.
 - Successfully imported same image without using BYOL.

PR #1394, opened 2020-10-14
EL import: Surface yum_install errors

EricEdens:
Audited ten recent failures, and none were related to the incorrect OS  being specified. This makes sense, since the newer change [1] would have overwritten this error message with correct OS if it were indeed mispecified. After this change, the actual error will be more prominent to the user (rather than logged as the second-to-last line).

Failures from audit: 
1. Repositories unreachable (local file, uncreachable URL)
2. Corrupt rpm database
3. Package conflicts

__

1. https://github.com/GoogleCloudPlatform/compute-image-tools/blob/f5a91c65376c26d14c93854a42d408fc2da41a81/cli_tools/gce_vm_image_import/importer/errors.go#L28-L27

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -226,6 +226,7 @@ def yum_install(g, *packages):
   Raises:
       RuntimeError: If there is a failure during installation.
   """
+  last_exception = None
   for i in range(6):
     try:
       # There's no sleep on the first iteration since `i` is zero.
@@ -239,11 +240,10 @@ def yum_install(g, *packages):
           '"{0}"'.format(p) for p in packages))
       return
     except Exception as e:
+      last_exception = e
       logging.debug('Failed to install {}. Details: {}.'.format(packages, e))
-  raise RuntimeError(
-      'Verify that you have specified the correct operating system '
-      'in the `--os` flag.  If you are bringing your own license (BYOL), '
-      'also verify that your subscription is eligible to be run on GCP.')
+  raise RuntimeError('Failed to install {}. Details: {}.'.format(
+      packages, last_exception))
 
 
 def main():


daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -239,11 +240,10 @@ def yum_install(g, *packages):
...
       return
     except Exception as e:
+      last_exception = e
       logging.debug('Failed to install {}. Details: {}.'.format(packages, e))
-  raise RuntimeError(
-      'Verify that you have specified the correct operating system '
-      'in the `--os` flag.  If you are bringing your own license (BYOL), '
-      'also verify that your subscription is eligible to be run on GCP.')
+  raise RuntimeError('Failed to install {}. Details: {}.'.format(
+      packages, last_exception))
```
dntczdx, line 247: So is the previous error msg gone completely? It looks like the old one doesn't make sense at all

zoran15 (approved):

daisy_workflows/image_import/enterprise_linux/translate.py
dntczdx: "The previous msg" means "Verify that you have specified the correct operating system in the `--os` flag..."

EricEdens: > So is the previous error msg gone completely?

The error message that you're referring to was added in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1066 . At that time, our analysis showed that the error message was appropriate. Since then, we added [1] which is a massive improvement, since it shows what's actually on the disk.

> It looks like the old one doesn't make sense at all

That's why we're removing it :)

1. https://github.com/GoogleCloudPlatform/compute-image-tools/blob/f5a91c65376c26d14c93854a42d408fc2da41a81/cli_tools/gce_vm_image_import/importer/errors.go#L28-L27

dntczdx (approved):

PR #1395, opened 2020-10-14
SLES import: Improve error messaging when SUSE not found on disk

EricEdens:
When no operating system is detected during SLES import, the user currently sees: `TranslateFailed: error: exits with return code 1`

The cause was:

```
ValueError: Import of unknown-0.0 is not supported. The following versions are supported: [opensuse-15.1|2, sles-15.1|2, sles-12.4|5]
```

Which appears in the logs, but is not surfaced to the user.

Two items are fixed:

1. Recognize that a distro of 'unknown' means that no operating system was found.
2. Fix surfacing of the actual error message.

### DIFF
--- a/cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go
+++ b/cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go
@@ -86,6 +86,12 @@ var basicCases = []*testCase{
 		expectedError: "\"debian-9\" was detected on your disk, but \"opensuse-15\" was specified",
 		inspect:       true,
 	},
+	{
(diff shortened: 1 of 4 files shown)


zoran15 (commented):
Can this be applied for other operating systems, or is this SUSE specific?

EricEdens:
> Can this be applied for other operating systems, or is this SUSE specific?

It's specific to SUSE, in that it's the only translator that uses inspection results to guide its logic. The remaining translators rely on the versions that are passed using metadata.

Once OS detection is finished, I'm planning on removing this inspection, so that we only inspect in a single place. Tactically, that means SUSE will rely on what's passed using metadata, instead of this inspection.

zoran15 (approved):

PR #1397, opened 2020-10-15
Inflation: Verify disk size prior to resizing

EricEdens:
Includes the following changes:
* import_image.sh: Prior to resizing a disk, check its size to determine whether it needs resizing.
* module tests: Added a test for `gce_vm_image_import/importer/inflater.go`
   * This required reverting the `Inflater.inflate(loggableBuilder *service.SingleImageImportLoggableBuilder)` prototype back to `Inflater.inflate()`; the loggable builder is only used in the facade implementation.
* Create a file `cli_tools_tests/module/import/pkg_test.go` to contain `zone` and `project` variables that are shared within the package.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/inflater.go
+++ b/cli_tools/gce_vm_image_import/importer/inflater.go
@@ -202,13 +202,13 @@ type shadowTestFields struct {
 	inflationType string
 }
 
-func createInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient storage.Client,
+func newInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient storage.Client,
 	inspector imagefile.Inspector, loggableBuilder *service.SingleImageImportLoggableBuilder) (Inflater, error) {
-	return CreateDaisyInflater(args, inspector)
+	return NewDaisyInflater(args, inspector)
 }
 
-// CreateDaisyInflater returns an Inflater implementation that uses a Daisy workflow.
-func CreateDaisyInflater(args ImportArguments, fileInspector imagefile.Inspector) (Inflater, error) {
+// NewDaisyInflater returns an Inflater that uses a Daisy workflow.
+func NewDaisyInflater(args ImportArguments, fileInspector imagefile.Inspector) (Inflater, error) {
 	diskName := "disk-" + args.ExecutionID
 	var wfPath string
 	var vars map[string]string
--- a/cli_tools/gce_vm_image_import/importer/inflater.go
+++ b/cli_tools/gce_vm_image_import/importer/inflater.go
@@ -45,8 +45,9 @@ const (
 
 // inflaterFacade implements an inflater using other concrete implementations.
 type inflaterFacade struct {
-	mainInflater   inflater
-	shadowInflater inflater
+	mainInflater    Inflater
+	shadowInflater  Inflater
+	loggableBuilder *service.SingleImageImportLoggableBuilder
 }
 
 // signals to control the verification towards shadow inflater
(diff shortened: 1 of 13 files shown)


cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -194,23 +195,20 @@ type persistentDisk struct {
...
 }
 
-func createInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient storage.Client, inspector imagefile.Inspector) (inflater, error) {
-	di, err := createDaisyInflater(args, inspector)
-	if err != nil {
-		return nil, err
-	}
-
-	return di, nil
+func createInflater(args ImportArguments, computeClient daisyCompute.Client, storageClient storage.Client,
```
zoran15, line 205: Quite a few unused args. Is this to support API inflater creation?

cli_tools/gce_vm_image_import/importer/inflater.go
```diff
@@ -194,23 +195,20 @@ type persistentDisk struct {
...
+	inspector imagefile.Inspector, loggableBuilder *service.SingleImageImportLoggableBuilder) (Inflater, error) {
+	return CreateDaisyInflater(args, inspector)
 }
 
-func createDaisyInflater(args ImportArguments, fileInspector imagefile.Inspector) (inflater, error) {
+// CreateDaisyInflater returns an Inflater implementation that uses a Daisy workflow.
+func CreateDaisyInflater(args ImportArguments, fileInspector imagefile.Inspector) (Inflater, error) {
```
zoran15, line 211: For "constructor" type funcs, we've been mostly using New...() naming. I guess Create...() was picked up earlier, but it would be nice to have consistent naming as it could lead to confusion otherwise.

cli_tools_tests/module/import/file_inflation_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/storage"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_vm_image_import/importer"
+	daisycompute "github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
+)
+
+const (
+	workflowDir = "../../../daisy_workflows"
+)
+
+func TestDaisyFileInflation(t *testing.T) {
```
zoran15, line 35: For module tests, due to the possibility of modules being defined loosely, it's not always clear what they tests without reading through test code. Please add docs with big picture scope of the test (and for future tests)

zoran15 (commented):
What triggered this change?

EricEdens:
@zoran15 :
> What triggered this change?

An import failed with message  "Invalid value for field 'sizeGb': '12'. New disk size '12' GiB must be larger than existing size '12' GiB." I *think* this occurred due to the inflation worker being rebooted while `inflate_image.sh` was being executed. 

See b/170868974 for details.

cli_tools_tests/module/import/file_inflation_test.go
EricEdens: Good feedback!
 - Renamed to `inflater_test.go`
 - Added a file-level comment
 - Renamed the `Test` function to indicate what's being tested
 - Added a comment talking about the worker's disk resizing

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: Updated to `NewDaisyInflater`

cli_tools/gce_vm_image_import/importer/inflater.go
EricEdens: Yeah; @dntczdx's VMDK project uses these.

zoran15 (approved):

PR #1398, opened 2020-10-20
Image import: Include logs in a single directory

EricEdens:
Since image import uses multiple daisy workflows, the logs of each workflow are placed in an independent directory under the GCS scratch path. This PR adds an intermediate directory, which serves to namespace the current import.

## Testing

Ran an import. Here's the path that daisy is using for inflation and translation:

```
[import-image]: 2020-10-20T08:12:14-07:00 Daisy scratch path: https://console.cloud.google.com/storage/browser/edens-test-daisy-bkt-us/gce-image-import-2020-10-20T08:12:09-07:00-r946p/daisy-import-image-20201020-15:12:12-30bjr
[import-image]: 2020-10-20T08:14:02-07:00 Daisy scratch path: https://console.cloud.google.com/storage/browser/edens-test-daisy-bkt-us/gce-image-import-2020-10-20T08:12:09-07:00-r946p/daisy-import-image-20201020-15:14:00-9r2v0
```

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/args.go
+++ b/cli_tools/gce_vm_image_import/importer/args.go
@@ -111,9 +111,7 @@ func (args *ImportArguments) ValidateAndPopulate(populator param.Populator,
 		return err
 	}
 
-	// Appending a namespace to the scratch bucket ensures that all logs and artifacts
-	// are contained in the same directory.
-	args.addNamespaceToScratchBucket()
+	args.populateNamespacedScratchDirectory()
 	if err := args.populateNetwork(); err != nil {
 		return err
 	}
@@ -121,7 +119,10 @@ func (args *ImportArguments) ValidateAndPopulate(populator param.Populator,
 	return args.validate()
 }
 
-func (args *ImportArguments) addNamespaceToScratchBucket() {
+// populateNamespacedScratchDirectory updates ScratchBucketGcsPath to include a directory
+// that is specific to this import, formulated using the start timestamp and the execution ID.
+// This ensures all logs and artifacts are contained in a single directory.
+func (args *ImportArguments) populateNamespacedScratchDirectory() {
 	if !strings.HasSuffix(args.ScratchBucketGcsPath, "/") {
 		args.ScratchBucketGcsPath += "/"
 	}
--- a/cli_tools/gce_vm_image_import/importer/args.go
+++ b/cli_tools/gce_vm_image_import/importer/args.go
@@ -17,12 +17,13 @@ package importer
 import (
 	"flag"
 	"fmt"
-	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
 
 	daisy_utils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/daisy"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/flags"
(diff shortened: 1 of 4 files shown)


cli_tools/gce_vm_image_import/importer/args.go
```diff
@@ -108,13 +111,25 @@ func (args *ImportArguments) ValidateAndPopulate(populator param.Populator,
...
+	// are contained in the same directory.
+	args.addNamespaceToScratchBucket()
 	if err := args.populateNetwork(); err != nil {
 		return err
 	}
 
 	return args.validate()
 }
 
+func (args *ImportArguments) addNamespaceToScratchBucket() {
```
zoran15, line 124: Using "Namespace" for  the import top level folder might be confusing, since Google storage docs talk about a single namespace for all buckets which this PR doesn't (and can't) address: https://cloud.google.com/storage/docs/naming-buckets

Maybe call it "ParentFolder", "TopLevelFolder" or something along those lines?

cli_tools/gce_vm_image_import/importer/args_test.go
```diff
@@ -136,9 +137,26 @@ func TestPopulateRegion(t *testing.T) {
 	assert.Equal(t, "us-west2", actual.Region)
 }
 
-func TestTrimScratchBucket(t *testing.T) {
-	assert.Equal(t, "gcs://bucket",
-		expectSuccessfulParse(t, "-scratch_bucket_gcs_path", "  gcs://bucket  ").ScratchBucketGcsPath)
+func TestTrimAndIncludeExecutionIDInScratchPath(t *testing.T) {
```
zoran15, line 140: minor: Ideally, this should test for the full folder name, not just a part of it. The test func name should be updated accordingly in that case. This is sufficient if injecting start time into workflow parsing is difficult/ugly to implement.

cli_tools/gce_vm_image_import/importer/args.go
EricEdens: Makes sense; I could see how someone skimming the linked doc might get confused and think these are related concepts.

Updated the helper method to refer to "path" instead of "bucket" and added some docs.

cli_tools/gce_vm_image_import/importer/args_test.go
EricEdens: sgtm. Updated to use a hard-coded date.

zoran15 (approved):

cli_tools/gce_vm_image_import/importer/args.go
```diff
@@ -134,6 +150,9 @@ func (args ImportArguments) validate() error {
 		return fmt.Errorf("-%s and -%s can't be both specified",
 			osFlag, customWorkflowFlag)
 	}
+	if !strings.HasSuffix(args.ScratchBucketGcsPath, args.ExecutionID) {
+		panic("Scratch bucket should have been namespaced with execution ID.")
```
dntczdx, line 154: If no suffix, the following logic won't be broken. Why don't we allow it? The worst case is it behaves as before. It's not worth failing customers for this reason.

EricEdens: It's a programming bug if this panic is reached; so, if someone submits a regression, all of our e2e tests will fail with a stack trace pointing here.

dntczdx (approved):

cli_tools/gce_vm_image_import/importer/args.go
```diff
@@ -17,13 +17,14 @@ package importer
...
 	"fmt"
-	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/daisycommon"
+
```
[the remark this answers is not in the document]
EricEdens: oh interesting, `goimports` was detecting them as different packages. Fixed.

PR #1405, opened 2020-10-26
Update translate.ps1 to better handle being unable to obtain metadata setting values

bkatyl:
- Add additional logging around obtaining metadata settings
- Move obtain the setting until after Configure-Network and to right before where they are needed.
- Add check for if any of the metadata values are null and reboot if so to attempt to correct the issue connecting to the metadata server.
- Removed $script:outs_dir as the variable was not being used
- Added 64-bit OS detection and removed $script:is_x86

### DIFF
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -45,7 +45,7 @@ function Get-MetadataValue {
     $client = New-Object Net.WebClient
     $client.Headers.Add('Metadata-Flavor', 'Google')
     $value = ($client.DownloadString($url)).Trim()
-    Write-Host "Retrieved metadata key $key with value $value."
+    Write-Host "Retrieved metadata for key $key with value $value."
     return $value
   }
   catch [System.Net.WebException] {
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -45,16 +45,16 @@ function Get-MetadataValue {
     $client = New-Object Net.WebClient
     $client.Headers.Add('Metadata-Flavor', 'Google')
     $value = ($client.DownloadString($url)).Trim()
-    Write-Output "Retrieved metadata key $key with value $value."
+    Write-Host "Retrieved metadata key $key with value $value."
     return $value
   }
   catch [System.Net.WebException] {
     if ($default) {
-      Write-Output "Failed to retrieve metadata for $key, returning default $default."
+      Write-Host "Failed to retrieve metadata for $key, returning default $default."
       return $default
     }
     else {
-      Write-Output "Translate: Failed to retrieve value from metadata for $key, returning null."
+      Write-Host "Failed to retrieve value from metadata for $key, returning null."
       return $null
     }
   }
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -44,14 +44,17 @@ function Get-MetadataValue {
   try {
     $client = New-Object Net.WebClient
     $client.Headers.Add('Metadata-Flavor', 'Google')
-    return ($client.DownloadString($url)).Trim()
+    $value = ($client.DownloadString($url)).Trim()
+    Write-Output "Retrieved metadata key $key with value $value."
+    return $value
   }
   catch [System.Net.WebException] {
     if ($default) {
+      Write-Output "Failed to retrieve metadata for $key, returning default $default."
       return $default
     }
     else {
-      Write-Output "Failed to retrieve value for $key."
+      Write-Output "Translate: Failed to retrieve value from metadata for $key, returning null."
       return $null
     }
   }
@@ -265,11 +268,6 @@ function Install-32bitPackages {
 
 try {
   Write-Output 'Translate: Beginning translate PowerShell script.'
-  $script:outs_dir = Get-MetadataValue -key 'daisy-outs-path'
-  $script:install_packages = Get-MetadataValue -key 'install-gce-packages'
-  $script:sysprep = Get-MetadataValue -key 'sysprep'
-  $script:is_byol = Get-MetadataValue -key 'is_byol'
-  $script:is_x86 = Get-MetadataValue -key 'is_x86'
 
   Remove-VMWareTools
   Change-InstanceProperties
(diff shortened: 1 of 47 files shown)


daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -265,25 +268,28 @@ function Install-32bitPackages {
 
 try {
   Write-Output 'Translate: Beginning translate PowerShell script.'
-  $script:outs_dir = Get-MetadataValue -key 'daisy-outs-path'
-  $script:install_packages = Get-MetadataValue -key 'install-gce-packages'
-  $script:sysprep = Get-MetadataValue -key 'sysprep'
-  $script:is_byol = Get-MetadataValue -key 'is_byol'
-  $script:is_x86 = Get-MetadataValue -key 'is_x86'
```
zoran15, line 272: Should this be removed from translate Daisy workflows as well?

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -44,14 +44,17 @@ function Get-MetadataValue {
   try {
     $client = New-Object Net.WebClient
     $client.Headers.Add('Metadata-Flavor', 'Google')
-    return ($client.DownloadString($url)).Trim()
+    $value = ($client.DownloadString($url)).Trim()
+    Write-Host "Retrieved metadata key $key with value $value."
```
zoran15, line 48:
```suggestion
    Write-Host "Retrieved metadata for key $key with value $value."
```

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -265,25 +268,28 @@ function Install-32bitPackages {
...
   Configure-Network
   Setup-NTP
   Configure-RDPSecurity
 
-  if ($script:is_x86.ToLower() -ne 'true') {
+  $script:install_packages = Get-MetadataValue -key 'install-gce-packages'
+  $script:sysprep = Get-MetadataValue -key 'sysprep'
+  $script:is_byol = Get-MetadataValue -key 'is_byol'
+  if ($script:install_packages -eq $null -or $script:sysprep -eq $null -or $script:is_byol -eq $null) {
+    Write-Output "Translate: failed to obtain at least one of the required value from metadata, rebooting in an attempt to resolve issue. install_packages=$script:install_packages, sysprep=$script:sysprep, is_byol=$script:is_byol"
```
zoran15, line 282:
```suggestion
    Write-Output "Translate: failed to obtain at least one of the required values from metadata, rebooting in an attempt to resolve issue. install_packages=$script:install_packages, sysprep=$script:sysprep, is_byol=$script:is_byol"
```

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -265,25 +268,28 @@ function Install-32bitPackages {
...
   Configure-RDPSecurity
 
-  if ($script:is_x86.ToLower() -ne 'true') {
+  $script:install_packages = Get-MetadataValue -key 'install-gce-packages'
+  $script:sysprep = Get-MetadataValue -key 'sysprep'
+  $script:is_byol = Get-MetadataValue -key 'is_byol'
+  if ($script:install_packages -eq $null -or $script:sysprep -eq $null -or $script:is_byol -eq $null) {
+    Write-Output "Translate: failed to obtain at least one of the required values from metadata, rebooting in an attempt to resolve issue. install_packages=$script:install_packages, sysprep=$script:sysprep, is_byol=$script:is_byol"
+    Restart-Computer -Force
+  }
```
EricEdens, line 284: Any idea why a reboot helps? Thoughts on limiting the number of reboots allowed before quitting?

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -44,14 +44,17 @@ function Get-MetadataValue {
   try {
     $client = New-Object Net.WebClient
```
EricEdens, line 45: Thoughts on retrying failures?

bkatyl: That's a great idea. I added 3 retries and a 10 second sleep between each attempt.

daisy_workflows/image_import/windows/translate.ps1
bkatyl: Yep, good catch. I've added that change below.

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -265,25 +271,28 @@ function Install-32bitPackages {
...
   Change-InstanceProperties
   Configure-Network
   Setup-NTP
   Configure-RDPSecurity
 
-  if ($script:is_x86.ToLower() -ne 'true') {
+  $script:install_packages = Get-MetadataValue -key 'install-gce-packages'
+  $script:sysprep = Get-MetadataValue -key 'sysprep'
+  $script:is_byol = Get-MetadataValue -key 'is_byol'
+  if ($script:install_packages -eq $null -or $script:sysprep -eq $null -or $script:is_byol -eq $null) {
```
dntczdx, line 284: it would be better to know why reboot is the solution. It sounds more like a bug if reboot is occasionally required. Maybe you can dig into and found a big hole.
Previously we saw calling to metadata service failed, but I don't remember in any case it became null. Maybe the powershell lib doesn't work as expected (for example, handle metadata service error incorrectly?)

daisy_workflows/image_import/windows/translate.ps1
bkatyl: Limiting to one reboot is pretty reasonable. Since the startup scripts use the temp directory, I'll add a file exist check and creation of the file in the temp directory to apply this limit.

This is a non-scientific answer, but in Windows especially when dealing with driver changes a reboot goes a long way to fixing many things. It's also better than this being a fatal error where the entire import is aborted.

bkatyl: I dug into this further and realized that at least one of the failure had been after removing vmware tools which also reboots the system. Which got me thinking that in these situations the script was starting right at boot instead of after setupcomplete.cmd, so also added the 60 second delay the other scripts have to help deal with timing issues.

EricEdens: sgtm (I see that we're doing the `ping 127.0.0.1 -n 60` pattern in several other Windows startup scripts in this repo). It sounds like there's an opportunity to investigate the root cause of why `ping 127.0.0.1 -n 60` helps, but that's probably beyond the scope of this PR.

EricEdens (approved):

daisy_workflows/image_import/windows/translate.ps1
bkatyl: Thanks for the feedback. I dug in a bit further and realized that specifically the vmware tools removal and reboot could have been the underlying issue. Which meant that the variables were read right at boot potentially before the network was fully ready which causes one of the variables to return a non-200 

The reboot approach was inspired by another customer issue where the network was unusable until the system was rebooted. It's meant to be a last ditch effort to help the import success by giving it one more fresh attempt instead of just aborting.

PR #1407, opened 2020-10-26
[log] move the loc of inspection_results due to the inconsistency of proto v.s. design

dntczdx:
Tested from local, verified that db received the log & parsed correctly


zoran15 (approved):

EricEdens (commented):
lgtm

PR #1410, opened 2020-10-29
Inspection refactor: Use protobufs for transport

EricEdens:
To support inspection metrics, this PR refactors the Python worker and the Go client to use protobufs internally and for transport. Transport still uses the serial console. However, instead of passing literal values for each field, the entire message is serialized and passed as a base64-encoded string. (Protobufs were in the original design, but I punted on it to unblock @dntczdx auto-UEFI feature.)

In this review, the public interface from `cli_tools/common/disk/inspect` is unchanged. The next PR will send the metrics to the logging server by:
1. Have `cli_tools/common/disk/inspect` return the proto 
2. Replace this struct with the proto: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/a7b746366a9a6672d75e763607956338feb74bc3/cli_tools/common/utils/logging/service/log_entry.go#L261

One review follows to send the metrics to the logging server. Once completed, five hand-maintained data models will be eliminated:

1. Python class in boot_inspect/model
2. Python serial output keys
3. Go serial output keys
4. Go struct for inspection results
5. Go struct for the logging server

## Testing
The module tests and e2e tests pass:
 - cli_tools_tests/e2e/gce_image_import_export/test_suites/import`
 - `cli_tools_tests/module/diskinspect`

Added unit test coverage for `cli_tools/common/disk/inspect`.

## Example

Serial log output from inspection worker, while inspecting a Windows 2019 GPT disk that only allows booting with UEFI:

```
 Inspecting OS
 GPT fdisk (gdisk) version 1.0.1

 Partition table scan:
   MBR: protective
   BSD: not present
   APM: not present
   GPT: present

 Found valid GPT with protective MBR; using GPT.
 Disk /dev/sdb: 67108864 sectors, 32.0 GiB
 Logical sector size: 512 bytes
 Disk identifier (GUID): 20DFB4C5-2E01-4E36-9C60-2B058892F1DC
 Partition table holds up to 128 entries
 First usable sector is 34, last usable sector is 67108830
 Partitions will be aligned on 2-sector boundaries
 Total free space is 2015 sectors (1007.5 KiB)

 Number  Start (sector)    End (sector)  Size       Code  Name
    1              34           32767   16.0 MiB    0C01  Microsoft reserved ...
    2           32768          237567   100.0 MiB   EF00  EFI system partition
    3          237568        67106815   31.9 GiB    0700  Basic data partition

 Results:
 os_release {
   major_version: "2019"
   architecture: X64
   distro_id: WINDOWS
 }
 uefi_bootable: true
 os_count: 1

 Status: <serial-output key:'inspect_pb' value:'CgsaBDIwMTkoAjDoBxgBOAE='>
 Success: Done!
 + echo Success:
 Success:
```


zoran15:
Before reviewing the code, it would be good to understand why protobufs are needed for the communication between the worker and Go in the context of inspection metrics? What piece of inspection metrics require protobufs?

EricEdens:
> Before reviewing the code, it would be good to understand why protobufs are needed for the communication between the worker and Go in the context of inspection metrics? What piece of inspection metrics require protobufs?

That's a good question.  It's not a strict requirement, but rather an investment to ease future maintenance.

In terms of metrics, I'm especially interested *when* there's a failure. The `ErrorWhen` enum captures this. Since errors can occur on the worker (in Python) or in the CLI (in Go), both codebases need up-to-date definitions of that enum.

My first iteration hardcoded those fields in Python and Go. I found myself updating the five data structures described in the overview ("Once completed, five hand-maintained data models will be eliminated:"), and I reflected on:

1. It's costly to keep these in sync
2. It's difficult to test that they're all setup properly (the best I know: run an import and look at the metrics server)
3. If they're not setup properly, we lose entire sets of logs (like we saw recently)

I'm still brainstorming on how to test that the logs are accepted by the log server. That said, I have higher confidence that the schema is correct in all of the right places following this change.

cli_tools/common/disk/inspect.go
```diff
@@ -63,50 +71,197 @@ func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, s
...
-	wf *daisy.Workflow
+// bootInspect implements disk.Inspector using the Python boot-inspect package,
+// executed on a worker VM using Daisy.
+type bootInspector struct {
+	traceLogs []string
+	worker    daisyWorker
+}
+
+// daisyWorker is a facade over daisy.Workflow to facilitate mocking.
+type daisyWorker interface {
```
zoran15, line 85: Consider renaming. Since this abstracts daisy.Workflow and is not exported, using daisyWorkflow should be ok.

cli_tools/common/disk/inspect.go
```diff
@@ -63,50 +71,197 @@ func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, s
...
 	wf.Vars["network"] = daisy.Var{Value: network}
 	wf.Vars["subnet"] = daisy.Var{Value: subnet}
-	return &defaultInspector{wf}, nil
+	return &bootInspector{[]string{}, &defaultDaisyWorker{wf}}, nil
 }
 
-// defaultInspector implements disk.Inspector using a Daisy workflow.
-type defaultInspector struct {
-	wf *daisy.Workflow
+// bootInspect implements disk.Inspector using the Python boot-inspect package,
```
zoran15, line 77: bootInsector to match the name of the struct?

Also, would bootDiskInspector be more appropriate?

cli_tools/common/disk/inspect.go
```diff
@@ -63,50 +71,197 @@ func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, s
...
 
 // Inspect finds partition and boot-related properties for a GCP persistent disk, and
 // returns an InspectionResult. `reference` is a fully-qualified PD URI, such as
 // "projects/project-name/zones/us-central1-a/disks/disk-name". `inspectOS` is a flag
 // to determine whether to inspect OS on the disk.
-func (inspector *defaultInspector) Inspect(reference string, inspectOS bool) (ir InspectionResult, err error) {
-	inspector.wf.AddVar("pd_uri", reference)
-	inspector.wf.AddVar("is_inspect_os", strconv.FormatBool(inspectOS))
-	err = inspector.wf.Run(context.Background())
+func (i *bootInspector) Inspect(reference string, inspectOS bool) (InspectionResult, error) {
```
zoran15, line 107: Related to the comment about renaming bootInspector to maybe bootDiskInspector, looking at this func signature, it's not clear what is being inspected, since neither the func, nor reference arg mention disk. Both could have "disk" in their names to improve readability.

cli_tools/common/disk/inspect.go
```diff
@@ -63,50 +71,197 @@ func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, s
...
+			results.Architecture = strings.ToLower(pbResults.OsRelease.Architecture.String())
+		}
+	}
+	results.UEFIBootable = pbResults.GetUefiBootable()
+	results.BIOSBootableWithHybridMBROrProtectiveMBR = pbResults.GetBiosBootable()
+	results.RootFS = pbResults.RootFs
+	return results
+}
+
+// validate checks the fields from a pb.InspectionResults object for logical consistency, returning
```
zoran15, line 178:
```suggestion
// validate checks the fields from a pb.InspectionResults object for consistency, returning
```

daisy_workflows/image_import/inspection/README.md
```diff
@@ -57,13 +57,14 @@ boot-inspect /images/ubuntu-16.04-server-cloudimg-amd64-disk1.vmdk
 The safe and quick option is to run the prow job's container locally:
 
 ```shell script
+cd $COMPUTE_IMAGE_TOOLS
```
zoran15, line 60: Is $COMPUTE_IMAGE_TOOLS defined?

proto/py/compute_image_tools_proto/inspect_pb2.pyi
```diff
@@ -0,0 +1,151 @@
+# @generated by generate_proto_mypy_stubs.py.  Do not edit!
```
zoran15, line 1: Why is this file (and all the imports) called inspect_pb2? Why the "2"?

zoran15, line 1: Copyright?

proto/py/compute_image_tools_proto/inspect_pb2.py
```diff
@@ -0,0 +1,382 @@
...
+from google.protobuf import message as _message
+from google.protobuf import reflection as _reflection
+from google.protobuf import symbol_database as _symbol_database
+# @@protoc_insertion_point(imports)
+
+_sym_db = _symbol_database.Default()
+
+
+
+
```
zoran15, line 16: All these blank spaces were autogenerated?

proto/inspect.proto
```diff
@@ -0,0 +1,120 @@
...
+  X64 = 2;
+}
+
+// OsRelease records the name and version of an operating system.
+message OsRelease {
+  // cli_formatted is a concatenation of distro, major_version, and
+  // minor_version using the format expected by the `--os` flag.
+  // For examples, see:
+  //  https://cloud.google.com/sdk/gcloud/reference/compute/images/import#--os
+  string cli_formatted = 1;
```
zoran15, line 50: Maybe it makes sense to refer to CLI as gcloud. Looking at this code base, one can assume CLI is actually the wrapper (which in a way it is, but we use it as backend in our services).

EricEdens: That's a good point. The fields map to //google3/logs/proto/compute_image_tools/compute_image_tools_log.proto, and `cli_formatted` is used there, though.

proto/py/compute_image_tools_proto/inspect_pb2.py
EricEdens: yeah -- `updatepb.sh` adds a footer to disable flake8, but otherwise leaves the generated code as-is.

proto/py/compute_image_tools_proto/inspect_pb2.pyi
EricEdens: I was following this guidance: "License headers are generally not required for generated code..." from go/releasing/preparing

EricEdens: The `_pb2` suffix comes from the protoc compiler: https://developers.google.com/protocol-buffers/docs/reference/python-generated

daisy_workflows/image_import/inspection/README.md
EricEdens: The example assumed that variable pointed to the root of the compute-image-tools repo. Clearly that was a bad assumption :)  I removed the variable, and wrote a comment to run the example from the root of the repo.

cli_tools/common/disk/inspect_test.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+			inspector := bootInspector{
+				worker: mockWorker,
+			}
+			assert.Equal(t, tt.cancelled, inspector.Cancel(tt.reason))
+		})
+	}
+}
+
+type mockDaisyWorker struct {
```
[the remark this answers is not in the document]
EricEdens: You're right -- this custom mock was growing hairy.  Three changes:
1. Moved DaisyWorker to its own package
2. Used gomock to generate the mocks for it
3. Replaced custom mock with [2]

cli_tools/common/disk/inspect_test.go
```diff
@@ -0,0 +1,330 @@
...
+func encode(results *pb.InspectionResults) string {
+	if results == nil {
+		return ""
+	}
+	bytes, err := proto.Marshal(results)
+	if err != nil {
+		panic(err)
+	}
+	return base64.StdEncoding.EncodeToString(bytes)
+}
```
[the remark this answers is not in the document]
EricEdens: I hear you on common code. At the moment we're only using the base64 encoding in the serial console output of the worker, and I was hoping to hide the base64 encoding as an implementation detail of this package.

This is what the encoding looks like:

` Status: <serial-output key:'inspect_pb' value:'CgsaBDIwMTkoAjDoBxgBOAE='>`

proto/inspect.proto
```diff
@@ -0,0 +1,120 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
EricEdens: good catch - done

cli_tools/common/disk/inspect_test.go
```diff
@@ -0,0 +1,330 @@
...
+			if err == nil {
+				t.Fatal("err must be non-nil")
+			}
+			assert.Contains(t, err.Error(), tt.expectErrorToContain)
+			assert.Equal(t, tt.expectResults, actual)
+		})
+	}
+}
+
+func TestBootInspector_Inspect_InvalidWorkerResponses(t *testing.T) {
```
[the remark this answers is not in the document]
EricEdens: Updated them all (you're right, they were all different :/  )

cli_tools/common/disk/inspect_test.go
```diff
@@ -0,0 +1,330 @@
...
+	assert.Equal(t, realWorker.wf.Project, "project-id")
+	assert.Equal(t, realWorker.wf.Zone, "zone-id")
+	assert.Equal(t, realWorker.wf.Vars["network"].Value, "network-id")
+	assert.Equal(t, realWorker.wf.Vars["subnet"].Value, "subnet-id")
+}
+
+func Test_bootInspector_PassesVarsWhenInvokingWorkflow(t *testing.T) {
+	for caseNumber, tt := range []struct {
+		inspectOS bool
+		pdURI     string
```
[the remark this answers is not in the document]
EricEdens: Changed all references of pdURI to reference

cli_tools/common/disk/inspect_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+	assert.NoError(t, err)
+	realWorker := inspector.(*bootInspector).worker.(*defaultDaisyWorker)
+	assert.Equal(t, realWorker.wf.Project, "project-id")
+	assert.Equal(t, realWorker.wf.Zone, "zone-id")
+	assert.Equal(t, realWorker.wf.Vars["network"].Value, "network-id")
+	assert.Equal(t, realWorker.wf.Vars["subnet"].Value, "subnet-id")
+}
+
+func Test_bootInspector_PassesVarsWhenInvokingWorkflow(t *testing.T) {
```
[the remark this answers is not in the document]
EricEdens: Updated all tests in the file to have the same prefix (TestBootInspector)

cli_tools/common/disk/inspect_test.go
```diff
@@ -0,0 +1,330 @@
+package disk
```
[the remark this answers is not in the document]
EricEdens: good catch - done

cli_tools/common/disk/inspect.go
```diff
@@ -63,50 +71,197 @@ func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, s
...
+	return createLegacyResults(results), nil
+}
+
+// tracef formats according to a format specifier and appends the results to the trace logs.
+func (i *bootInspector) tracef(format string, a ...interface{}) {
+	i.traceLogs = append(i.traceLogs, fmt.Sprintf(format, a...))
+}
+
+// assembleErrors sets the errorWhen field, and generates an error object.
+func (i *bootInspector) assembleErrors(pdURI string, results *pb.InspectionResults,
```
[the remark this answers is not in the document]
EricEdens: done

cli_tools/common/disk/inspect.go
```diff
@@ -63,50 +71,197 @@ func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, s
...
-	ir.RootFS = inspector.wf.GetSerialConsoleOutputValue("root_fs")
-	return
+	return nil
+}
+
+// populate fills the fields in the pb.InspectionResults that are not returned by the worker.
+// This is required since the worker is unaware of import-specific idioms, such as the formatting
+// used by gcloud's --os argument.
+func (i *bootInspector) populate(results *pb.InspectionResults) error {
+	if results.ErrorWhen == pb.InspectionResults_NO_ERROR && results.OsCount == 1 {
```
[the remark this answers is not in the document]
EricEdens: It's valid to have NO_ERROR with nonzero OsCount. (If the disk is empty, it's not an inspection error).

See this for docs: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/a320efcf431ba3d1a1571179d8b25d39349772b6/proto/inspect.proto#L107

cli_tools/common/disk/inspect.go
EricEdens: done

cli_tools/common/disk/inspect.go
EricEdens: Similar to above, I don't have a huge preference. That said, I took this approach since the symbol would be `disk.Inspector` when consumed outside of this package.

cli_tools/common/disk/inspect.go
EricEdens: Fixed the comment!

Regarding `bootDiskInspector`, I don't have a huge preference. Since the package name is `disk` already, I was thinking it'd be implicit. More in the next comment on this....

zoran15 (approved):

dntczdx (approved):
For a long PR like this, I mostly rely on tests to check the correctness. The test set looks good to me.

It's a good idea to publish a weekly release before landing this one, since this change is relatively big, and potentially have things to do with log content.

EricEdens:
> have things to do with log content.

The logging behavior change comes in the next PR -- I agree that'll it'll be good to have a release prior to that.

PR #1411, opened 2020-10-31
Fixed an edge case in image importer run step that can cause a deadlock on timeout

zoran15:
Fixed an edge case in image importer run step that can cause deadlock to occur on timeout due to WaitGroup being added to before checking for context expiration/timeout. This can cause wg.Wait() in the cancellation handler below to wait forever.

### DIFF
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -145,9 +145,16 @@ func (i *importer) runStep(ctx context.Context, step func() error, cancel func(s
 			e <- i.getCtxError(ctx)
 		default:
 			wg.Add(1)
-			stepErr := step()
-			wg.Done()
-			e <- stepErr
+			var stepErr error
+			defer func() {
+				// error should only be returned after wg is marked as done. Otherwise,
+				// a deadlock can occur when handling a timeout in the select below
+				// because cancel() causes step() to finish, then waits for wg, while
+				// writing to error chan waits on error chan reader which never happens
+				wg.Done()
+				e <- stepErr
+			}()
+			stepErr = step()
 		}
 	}()
 
--- a/cli_tools/gce_vm_image_import/importer/importer.go
+++ b/cli_tools/gce_vm_image_import/importer/importer.go
@@ -137,7 +137,6 @@ func (i *importer) runProcess(ctx context.Context) error {
 func (i *importer) runStep(ctx context.Context, step func() error, cancel func(string) bool, getTraceLogs func() []string) (err error) {
 	e := make(chan error)
 	var wg sync.WaitGroup
-	wg.Add(1)
 	go func() {
 		//this select checks if context expired prior to runStep being called
 		//if not, step is run
@@ -145,6 +144,7 @@ func (i *importer) runStep(ctx context.Context, step func() error, cancel func(s
 		case <-ctx.Done():
 			e <- i.getCtxError(ctx)
 		default:
+			wg.Add(1)
 			stepErr := step()
 			wg.Done()
 			e <- stepErr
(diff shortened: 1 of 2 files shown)


cli_tools/gce_vm_image_import/importer/importer_test.go
```diff
@@ -352,6 +352,72 @@ func TestRun_ProcessCantTimeoutImportSucceeds(t *testing.T) {
...
+		importer.runStep(ctx,
+			func() error {
+				// step
+				select {
+				case <-cancelChan:
+					break
+				case <-time.After(time.Second * 5):
+					break
+				}
+				return nil
```
EricEdens, line 386: Is it worth asserting that the step isn't run? (And that the cancellation is run?) Or does that make the test too fragile?

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -137,14 +137,14 @@ func (i *importer) runProcess(ctx context.Context) error {
 func (i *importer) runStep(ctx context.Context, step func() error, cancel func(string) bool, getTraceLogs func() []string) (err error) {
 	e := make(chan error)
 	var wg sync.WaitGroup
-	wg.Add(1)
 	go func() {
 		//this select checks if context expired prior to runStep being called
 		//if not, step is run
 		select {
 		case <-ctx.Done():
 			e <- i.getCtxError(ctx)
```
EricEdens, line 145: Is my understanding right? Here's my interpretation: 

1. Is the context cancelled? If so, return the context's cancellation error.
2. Otherwise, the step will run
3. Wait until either the step finishes, or the context is cancelled.

(This isn't feedback or a change request. Just seeing if my reading is right :)  )

zoran15: Yes, that's correct. The problem is WaitGroup got updated before it was determined if the context is already cancelled. This issue only occurs if import actually timed out. It doesn't affect imports that have enough time to finish.

cli_tools/gce_vm_image_import/importer/importer_test.go
zoran15: The main purpose of the test is to make sure there are no deadlocks on timeout. However, in this case (0ms timeout) it makes sense asserting step didn't actually run, so I added the check.

EricEdens (approved):

cli_tools/gce_vm_image_import/importer/importer.go
EricEdens: cool!

cli_tools/gce_vm_image_import/importer/importer.go
```diff
@@ -137,14 +137,14 @@ func (i *importer) runProcess(ctx context.Context) error {
...
 		default:
+			wg.Add(1)
 			stepErr := step()
 			wg.Done()
```
dntczdx, line 149: in order to fix the potential deadlock better, I suggest to use "defer wg.done()" in case step() triggered a panic. There are do some panic code added.

zoran15: if `wg.Done()` was called from defer, it would be invoked even if the first case happened (ctx.Done()). That would result in a panic of it's own due to negative wg counter.

Also, deferred is called even if panic occurs. E.g. this code:
'''
func main() {
	defer fmt.Println("deferred")
	panic("panicking")
}
'''

Will print:
'''
deferred
panic: panicking
'''

dntczdx: correct. Let me give a LGTM then you decide modify or not

dntczdx (approved):

cli_tools/gce_vm_image_import/importer/importer.go
zoran15: Done

dntczdx (approved):

EricEdens (approved):

PR #1414, opened 2020-11-04
distro.Release: Support Windows

EricEdens:
This updates cli_tools/common/distro to support Windows.

### DIFF
--- a/cli_tools/common/distro/distro.go
+++ b/cli_tools/common/distro/distro.go
@@ -36,6 +36,8 @@ const (
 
 	archX86 = "x86"
 	archX64 = "x64"
+
+	r2 = "r2"
 )
 
 var osFlagExpression = regexp.MustCompile(
@@ -45,7 +47,7 @@ var osFlagExpression = regexp.MustCompile(
 	"^(?P<distro>[a-z]+(?:-[a-z]+)?)" +
 		// Version is required, and is at least one word character.
 		// Examples: 2004, 2008, 2008r2
-		"-(?P<version>\\w+)" +
+		"-(?P<version>[a-z0-9]+)" +
 		// Architecture is optional. The only options are `x86` and `x64`.
 		"(?:-(?P<arch>x86|x64))?" +
 		// License is optional. The only value is `byol`.
@@ -54,6 +56,14 @@ var osFlagExpression = regexp.MustCompile(
 var architectures = map[string]bool{
 	archX64: true,
 	archX86: true,
+}
+
+// Flags that don't follow `osFlagExpression` and have been
+// replaced with newer versions. Mapping is from legacy to modern.
+var legacyFlags = map[string]string{
+	// windows-8-1-x64-byol includes an extra hyphen between its
+	// major and minor version. The non-legacy flag is windows-8-x64-byol.
+	"windows-8-1-x64-byol": "windows-8-x64-byol",
 }
 
 // Release encapsulates product and version information
@@ -90,6 +100,10 @@ func FromComponents(distro string, major string, minor string, architecture stri
 		return newWindowsRelease(major, minor, architecture)
 	}
 
+	return newLinuxRelease(distro, major, minor)
+}
+
+func newLinuxRelease(distro string, major string, minor string) (Release, error) {
 	majorInt, e := strconv.Atoi(major)
 	if e != nil || majorInt < 1 {
 		return nil, fmt.Errorf(
@@ -132,11 +146,8 @@ func FromComponents(distro string, major string, minor string, architecture stri
 //
 // https://cloud.google.com/sdk/gcloud/reference/compute/images/import#--os
 func FromGcloudOSArgument(osFlagValue string) (r Release, e error) {
-	if osFlagValue == "windows-8-1-x64-byol" {
-		// windows-8-1-x64-byol is a legacy flag value, and it's the only value that
-		// includes an extra hyphen between its major and minor version. The non-legacy
-		// flag is windows-8-x64-byol.
-		osFlagValue = "windows-8-x64-byol"
+	if legacyFlags[osFlagValue] != "" {
+		osFlagValue = legacyFlags[osFlagValue]
 	}
 	match := osFlagExpression.FindStringSubmatch(strings.ToLower(osFlagValue))
 	if match == nil {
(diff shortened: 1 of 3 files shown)


cli_tools/common/distro/distro.go
```diff
@@ -52,8 +75,21 @@ type Release interface {
...
+	architecture = strings.ToLower(architecture)
+
+	if architecture != "" && !architectures[architecture] {
+		return nil, fmt.Errorf("Architecture `%s` is not supported for import", architecture)
+	}
+
+	if distro == windows {
+		return newWindowsRelease(major, minor, architecture)
+	}
+
```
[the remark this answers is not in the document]
EricEdens: Right now each of the `new` methods return a struct, so you can think of `FromComponents` as determining which struct to return.

cli_tools/common/distro/distro.go
```diff
@@ -29,10 +30,32 @@ const (
...
 )
 
+var osFlagExpression = regexp.MustCompile(
+	// Distro is required and is comprised of at least one letter.
+	// There may be two segments, separated by a hyphen.
+	// Examples: ubuntu, sles, sles-sap, windows
+	"^(?P<distro>[a-z]+(?:-[a-z]+)?)" +
+		// Version is required, and is at least one word character.
+		// Examples: 2004, 2008, 2008r2
+		"-(?P<version>\\w+)" +
```
zoran15, line 48: \w will match upper and lower case characters, while distro matches only lower case characters. Is this inconsistency needed for any OS version?

cli_tools/common/distro/distro.go
```diff
@@ -52,8 +75,21 @@ type Release interface {
...
 	distro = strings.ToLower(distro)
+	architecture = strings.ToLower(architecture)
+
+	if architecture != "" && !architectures[architecture] {
+		return nil, fmt.Errorf("Architecture `%s` is not supported for import", architecture)
+	}
+
+	if distro == windows {
+		return newWindowsRelease(major, minor, architecture)
+	}
```
zoran15, line 91: It would be more readable if windows was handled in the main switch below. Is the break in logic because the rest of the code is Linux specific? 

If it's not possible to have windows in the switch, consider extracting the rest of this func in e.g. newLinuxRelease()

cli_tools/common/distro/distro.go
```diff
@@ -106,25 +132,38 @@ func FromComponents(distro string, major string, minor string) (r Release, e err
...
 func FromGcloudOSArgument(osFlagValue string) (r Release, e error) {
-	os := strings.ToLower(osFlagValue)
-	if strings.HasSuffix(os, "-byol") {
-		os = strings.TrimSuffix(os, "-byol")
+	if osFlagValue == "windows-8-1-x64-byol" {
+		// windows-8-1-x64-byol is a legacy flag value, and it's the only value that
+		// includes an extra hyphen between its major and minor version. The non-legacy
+		// flag is windows-8-x64-byol.
+		osFlagValue = "windows-8-x64-byol"
+	}
```
zoran15, line 140: Constants. 

Maybe introduce a generic mapping if we have this issue in the future, e.g. a map from input OS flag value to "true" flag value.

cli_tools/common/distro/distro.go
```diff
@@ -167,6 +206,44 @@ func newCommonLinuxRelease(distro string, major, minor int) (Release, error) {
...
+// 2012 and 2012r2 and *not* import compatible, and have different
+// minor versions. In the first case, the minor version is empty.
+// In the second it is "r2".
+type windowsRelease struct {
+	major, minor, architecture string
+}
+
+func (w windowsRelease) ImportCompatible(other Release) bool {
+	actualOther, ok := other.(windowsRelease)
+	compat := ok &&
```
zoran15, line 220: compat->compatible. Not much longer and others won't read it as `combat` :)

cli_tools/common/distro/distro.go
```diff
@@ -106,25 +132,38 @@ func FromComponents(distro string, major string, minor string) (r Release, e err
...
+	var major, minor string
 	if distro == ubuntu {
 		// In gcloud, major and minor are combined as MMmm, such as ubuntu-1804
 		if len(version) != 4 {
 			return r, fmt.Errorf("expected version with length four. Actual: `%s`", version)
 		}
 
-		return FromComponents("ubuntu", version[:2], version[2:])
+		major, minor = version[:2], version[2:]
+	} else if distro == windows && strings.HasSuffix(version, "r2") {
```
zoran15, line 161: "r2" is used in multiple places in multiple functions, extract as a constant

cli_tools/common/distro/distro.go
```diff
@@ -167,6 +206,44 @@ func newCommonLinuxRelease(distro string, major, minor int) (Release, error) {
...
+		w.architecture == actualOther.architecture
+	if w.minor == "r2" || actualOther.minor == "r2" {
+		compat = compat && (w.minor == actualOther.minor)
+	}
+	return compat
+}
+
+func (w windowsRelease) AsGcloudArg() string {
+	arg := fmt.Sprintf("windows-%s", w.major)
+	if w.minor == "r2" {
```
zoran15, line 231: Should it throw fatal if it's not "r2"?

cli_tools/common/distro/distro.go
EricEdens: Extracted out a method to handle Linux.

(The Linux code shares version validation logic)

cli_tools/common/distro/distro.go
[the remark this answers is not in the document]
EricEdens: Extracted out a method to handle Linux.

cli_tools/common/distro/distro.go
EricEdens: sgtm. Switched to use `[a-z0-9]`

cli_tools/common/distro/distro.go
EricEdens: Switched to a map

cli_tools/common/distro/distro.go
EricEdens: done!

cli_tools/common/distro/distro.go
EricEdens: done!

cli_tools/common/distro/distro.go
EricEdens: There could be other minor versions (eg: 8 vs 8.1), but "r2" is the only minor version that we treat separately. Added a comment to describe this.

cli_tools/common/distro/distro.go
```diff
@@ -29,10 +30,42 @@ const (
...
+		// License is optional. The only value is `byol`.
+		"(?:-(?P<license>byol))?$")
+
+var architectures = map[string]bool{
+	archX64: true,
+	archX86: true,
+}
+
+// Flags that don't follow `osFlagExpression` and have been
+// replaced with newer versions.
```
zoran15, line 62: Minor: state explicitly that the mapping is from legacy to current flag versions

dntczdx (approved):

zoran15 (approved):

PR #1415, opened 2020-11-04
Wait if W32Time is in a pending state

bkatyl:
If the time service is in a pending state (PendingState or PendingStop) the import will currently fail due to the Stop-Service or Start-Service cmdlets returning an error.

This checks the W32Time service status and if it includes pending it will wait and recheck every second. For the instance having this issue, it took 13 seconds until it was not in a pending state. This delay allowed the Stop-Service cmdlet to succeed.

### DIFF
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -63,6 +63,27 @@ function Get-MetadataValue {
   return $null
 }
 
+function Wait-Service {
+  <#
+  .SYNOPSIS
+    Wait until the service is not in a pending state.
+  .DESCRIPTION
+    This function waits until the specified service is not in a pending state
+    as you cannot start or stop a service in a pending state.
+  .PARAMETER ServiceName
+    The name of the service to check the state of.
+  #>
+    param (
+    [parameter(Mandatory=$true)]
+      [string]$ServiceName
+  )
+  #
+  while((Get-Service -Name $ServiceName).Status -match 'pending') {
+    Write-Host "$ServiceName service is in a pending state, waiting 1 second."
+    Start-Sleep -s 1
+  }
+}
+
 function Remove-VMWareTools {
   Get-ChildItem HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall | Foreach-Object {
     if ((Get-ItemProperty $_.PSPath).DisplayName -eq 'VMWare Tools') {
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -93,6 +93,11 @@ function Setup-NTP {
   Set-ItemProperty -Path $tzi_path -Name RealTimeIsUniversal -Value 1
 
   # Set up time sync...
+  # Ensure the time service isn't in a pending state; wait until it is not in a pending state.
+  while((Get-Service -Name "W32time").Status -match 'pending') {
+    Write-Host "W32time service is in pending state, waiting 1 second."
+    Start-Sleep -s 1
+  }
```
zoran15, line 100: Looks like these two pieces of code are the same. If so, extract them in a separate function and re-use.

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -93,6 +93,11 @@ function Setup-NTP {
   Set-ItemProperty -Path $tzi_path -Name RealTimeIsUniversal -Value 1
 
   # Set up time sync...
+  # Ensure the time service isn't in a pending state; wait until it is not in a pending state.
+  while((Get-Service -Name "W32time").Status -match 'pending') {
```
EricEdens, line 97: It looks like there are several possible statuses.  Without knowing _anything_ about Windows services, I'd presume you'd want to assert the service is running (rather than stopped, for example). 

Here's the path I took to find the docs:

* https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-service
* https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontroller
* https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontrollerstatus

Lastly, I see this utility method; is it available from PowerShell?

https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontroller.waitforstatus

bkatyl: WaitForStatus is available, but it is only able to wait for a single state. This isn't bad however it means you need to be more specific as there are [4 pended states](https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontrollerstatus?view=dotnet-plat-ext-3.1#fields), so the state you're waiting for would need to match the corresponding pending state.

WaitForStatus is likely more useful when your application/script does a state changes in such a way were you don't inherit the WaitForStatus behavior and that you're only looking for one specific state.

The goal of this is to ensure the Start|Stop-Service cmdlets are successful and for that just need to make sure the service isn't in one of the pending states. So I'd prefer to keep this as it's simplier and accomplishes the same goal.

daisy_workflows/image_import/windows/translate.ps1
bkatyl: Absolutely, I'll make this into a utility function.

The service controller has global timeouts, so we shouldn't need to. By default it gives a service 20000ms to complete a state change, though that can be customized in the OS's settings.

daisy_workflows/image_import/windows/translate.ps1
EricEdens: makes sense!

EricEdens (approved):

zoran15 (approved):

dntczdx (approved):

PR #1417, opened 2020-11-06
EL import: Update ca-certificates when translation starts

EricEdens:
Import fails when `ca-certificates`  is stale such that a yum repo cannot by verified. A concrete example occurs when `epel` is added to an older version of EL 6, as described in https://stackoverflow.com/questions/26734777 .

After this PR, all EL imports will have `ca-certificates` upgraded, _if_ `ca-certificates` is installed.  

## Testing

* Create a CentOS 6 image that reproduces the issue.
* Ran all EL tests in `import_tests.go`

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -87,6 +87,40 @@ terminal --timeout=0 serial console
 '''
 
 
+def upgrade_ca_certificates(g: guestfs.GuestFS):
+  """Upgrade ca-certificates, if it is installed.
+
+  Stale or missing CA certificates can cause yum operations to fail.
+  """
+
+  try:
+    g.sh('rpm -V ca-certificates')
+  except RuntimeError as e:
+    logging.debug('ca-certificates not found. Skipping upgrade.', e)
+    return
+
+  upgraded = False
+  try:
+    g.sh('yum upgrade -y ca-certificates')
+    upgraded = True
+  except RuntimeError as e:
+    logging.debug('Failed to upgrade ca-certificates', e)
+    try:
+      # A common failure is older versions of EL where the user has added the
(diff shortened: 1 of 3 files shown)


dntczdx (approved):
lgtm. The logic looks correct to me, once "ca-certificates" is not a required thing for the whole translation.

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -87,6 +87,40 @@
...
+        g.sh('yum upgrade -y ca-certificates --disablerepo=epel')
+        upgraded = True
+    except RuntimeError as e:
+      logging.debug('Failed second attempt to upgrade ca-certificates', e)
+
+  if upgraded:
+    logging.info('Upgraded ca-certificates package.')
+  else:
+    logging.info('Failed to upgrade ca-certificates package. If import '
+                 'fails, update the package manually and try again.')
```
zoran15, line 121: Minor: if ca-certificates upgrade fails, it will print 3 different "Failed to upgrade messages", which can be confusing. I think the first one can be expanded with "...retrying with disabled epel..." type of message. The second ("Failed second attempt...") seems redundant, as the main failed message will anyway be printed.

zoran15 (approved):

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Originally I had them all going to `logging.info`, and indeed that was noisy. By using `logging.debug`, the failures (and exceptions) only go in the serial console logs. A typical gcloud user should only see the message at line 118, or the message at line 120.

Here's the log configuration: 

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/676b7dcad24ebaf15363c8bfaaf1ae815fa96efe/daisy_workflows/linux_common/utils/common.py#L397

PR #1418, opened 2020-11-06
Ensure shell output from guest is retained

EricEdens:
While debugging recent yum-related failures, half had not-helpful failure messages.

Examples:

* `sh:"""`
* `sh: Error: Nothing to do."""`
* `sh: Traceback (most recent call last):"""`

The problem is guestfs.GuestFS's shell methods. They discard either stdout or stderr, depending on whether the command fails or is successful (respectfully). 

This library is motivated by subprocess.run, which returns an object containing stdout, stderr, and the return code of the process.

### DIFF
--- a/daisy_workflows/linux_common/utils/guestfsprocess.py
+++ b/daisy_workflows/linux_common/utils/guestfsprocess.py
@@ -45,7 +45,7 @@ def run(g: 'GuestFSInterface', command) -> 'CompletedProcess':
     >>> run(g, 'printf hi; exit 1')
     {'stdout': 'hi', 'stderr': '', 'code': 1, 'cmd': 'printf hi; exit 1'}
   """
-  tmp_dir = g.mkdtemp('gprocXXXXXX')
+  tmp_dir = g.mkdtemp('/tmp/gprocXXXXXX')
   program_path = os.path.join(tmp_dir, 'program.sh')
   stdout_path = os.path.join(tmp_dir, 'stdout.txt')
   stderr_path = os.path.join(tmp_dir, 'stderr.txt')
--- a/daisy_workflows/linux_common/utils/guestfsprocess.py
+++ b/daisy_workflows/linux_common/utils/guestfsprocess.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+# Copyright 2020 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


daisy_workflows/linux_common/utils/guestfsprocess.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class CompletedProcess:
+  def __init__(self, stdout: str, stderr: str, code: int, cmd: str):
+    self.stdout = stdout
+    self.stderr = stderr
+    self.code = code
+    self.cmd = cmd
+
+  def __eq__(self, o: object) -> bool:
```
hopkiw, line 119: what are the use cases for comparing CompletedProcess instances?

daisy_workflows/linux_common/utils/guestfsprocess.py
```diff
@@ -0,0 +1,127 @@
...
+    Thu 05 Nov 2020 06:53:55 PM PST
+
+    >>> run(g, 'printf hi; exit 1')
+    {'stdout': 'hi', 'stderr': '', 'code': 1, 'cmd': 'printf hi; exit 1'}
+  """
+  tmp_dir = g.mkdtemp('gprocXXXXXX')
+  program_path = os.path.join(tmp_dir, 'program.sh')
+  stdout_path = os.path.join(tmp_dir, 'stdout.txt')
+  stderr_path = os.path.join(tmp_dir, 'stderr.txt')
+  return_code_path = os.path.join(tmp_dir, 'return_code.txt')
```
hopkiw, line 52: i may be missing some unique requirements here but programs on linux can pipe their stdout, stderr etc. file descriptors to each other. it should not be necessary to create temporary files to capture these.

daisy_workflows/linux_common/utils/guestfsprocess.py
```diff
@@ -0,0 +1,127 @@
...
+
+  Returns:
+    String containing the shell script.
+  """
+  return textwrap.dedent("""
+  touch {stdout_path}
+  touch {stderr_path}
+  touch {return_code_path}
+
+  bash -c {cmd} 1> {stdout_path} 2> {stderr_path}
```
hopkiw, line 88: no quoting around cmd? have you validated this will work for all sorts of commands you want to call? what are the commands being run now?

EricEdens: It's quoted by `shlex.quote` prior to being written to the template: https://docs.python.org/3/library/shlex.html#shlex.quote

See the  `test_escape_array_members`, for example.

daisy_workflows/linux_common/utils/guestfsprocess.py
EricEdens: Good question -- it's just for testing.

daisy_workflows/linux_common/utils/guestfsprocess.py
EricEdens: This Python code runs on a worker instance, but `command` is executed in the guest. 

Typically we've been using `guestfs.sh` or `guestfs.command` for this functionality. The problem with these APIs, though, is that you get *either* stdout, or stderr, depending on the return code of the program. Specifcally:

1. When the return code is zero, return a string containing stdout.
2. When the return code is nonzero, throw an exception that captures stderr.

The goal is to have stdout, stderr, and the return code for every execution.

hopkiw (approved):

PR #1420, opened 2020-11-09
Enable-RemoteDesktop move up further

eissi:
When --syspre-windows flag is used, Enable-RemoteDesktop is not called so that an instance from the imported image can't be accessed unless RDP was enabled before importing.


zoran15 (approved):

bkatyl (approved):
LGTM. Thanks for catching this and sorry if it caused any inconvenience.

I added a commit to remove the blank spaces between the if statements.

PR #1422, opened 2020-11-10
EL import: Improve yum failure messages

EricEdens:
This PR improves failure messaging when `yum` commands fails. Current failure messages include:

- sh: Error: Package: 1:google-compute-engine-oslogin-20200819.00-g1.el7.x86_64 (google-compute-engine)"""
- sh: Importing GPG key 0xAxyz:"""
- sh: Traceback (most recent call last):"""
- sh: file:///path/repomd.xml: [Errno 14] curl#37 - \""Couldn't open file /path/repomd.xml\"""""

To improve error messaging, this PR adds validation for common failure scenarios. Although validation occurs eagerly, we allow the import to continue. If the import fails, then we create a customized error message using the validation results. Scenarios are:

1. A custom repo is not accessible
2. BYOL is specified, but an active license isn't found
3. The `yum` command is not found

## Testing

1. Created disk images and tests to repro each of the three scenarios.
2. Ran all of the tests in cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go


daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -87,16 +88,75 @@
...
   """
 
-  try:
-    g.sh('rpm -V ca-certificates')
-  except RuntimeError as e:
-    logging.debug('ca-certificates not found. Skipping upgrade.', e)
+  p = run(g, 'rpm -qa ca-certificates | grep ca-certificates')
+  if p.code != 0:
+    logging.debug(
+        'ca-certificates not found. Skipping upgrade. {}'.format(p))
```
zoran15, line 159: What's the difference between -V and -qa? Can exception still occur with this command?

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -285,10 +351,31 @@ def yum_install(g, *packages):
       packages, last_exception))
 
 
+def customize_error_message(e: BaseException,
+                            errs: NonFatalErrors) -> BaseException:
+  """Create an error message based on PotentialErrors."""
+  logging.debug('Translation failed due to: {}'.format(e))
+  logging.debug('Potential errors: {}'.format(errs))
```
zoran15, line 358: Potential causes?

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -285,10 +351,31 @@ def yum_install(g, *packages):
       packages, last_exception))
 
 
+def customize_error_message(e: BaseException,
+                            errs: NonFatalErrors) -> BaseException:
+  """Create an error message based on PotentialErrors."""
```
zoran15, line 356: PotentialErrors is formatted like a class name. Should it be "potential errors" or NonFatalErrors?

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -285,10 +351,31 @@ def yum_install(g, *packages):
...
+  logging.debug('Potential errors: {}'.format(errs))
+  if errs.yum_command_not_found:
+    return RuntimeError('Verify the disk\'s OS: `yum` not found. {}'.format(e))
+  if errs.subscription_manager_status_error:
+    return RuntimeError(
+        'subscription-manager did not find an active subscription. Omit '
+        '`-byol` to register with on-demand licensing. {}'.format(e))
+  if errs.yum_unreachable_repo:
+    return RuntimeError('Ensure that all configured repos are reachable. '
+                        '{}'.format(e))
```
zoran15, line 367: Can multiple potential causes happen at the same time? If so, this logic will only return the first one.

daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -87,16 +88,75 @@
 '''
 
 
+class NonFatalErrors:
```
zoran15, line 91: I find this class, with boolean fields instead of a list of errors a bit strange. I would have expected a list of errors instead. That way, each error type can have it's own formatting and the processing of potential causes can be done in a loop without the knowledge of details of each error (the IFs starting in line 359). With 3 different causes this is not that bad right now, but, if many more potential causes are added in the future, this piece of code and the return error IFs will become less pretty.

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: `-V` checks that the package is installed, and then performs validations (think: checksums, file permissions). This caused false negatives when the package was installed, but failed one of the validations.

That said, your comment prompted me to dig some more, and I created separate PR that improves the performance of the ca-certificates update: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1423

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Removed this log in the current version (I liked your idea of not using the `NonFatalErrors` class)

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Removed this log in the current version (I liked your idea of not using the NonFatalErrors class)

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Theoretically multiple errors could occur. I'd like to try this simple version first, though, since it's a big step forward from our current error messages. 

Good news: Given this PR, it's a straightforward change to identify multiple errors in a future modification if we find that helpful.

daisy_workflows/image_import/enterprise_linux/translate.py
EricEdens: Good point. In the current version, I switched to a list of matcherse.

zoran15 (approved):

PR #1423, opened 2020-11-11
EL import: Tune upgrading of ca-certificates

EricEdens:
In #1417 we updated EL import to update ca-certificates. This PR makes two improvements, both found by reading testgrid translation logs:

* Avoid unnecessary latency by updating ca-certificates only when translating EL6 guests that use the epel repo. In the EL6 test [1], the previous PR added about 80 seconds. In the RHEL7 [2] test, it added about 15 minutes. In the current version, the time penalty only occurs when:
   * The guest is EL6
   * It uses the epel repo
   * It uses an old version of ca-certificates 
* Improve debugability of the feature by using the library added in #1418.

1. https://github.com/GoogleCloudPlatform/compute-image-tools/blob/5a2d47d1138373d1ee2a2152eb3a96dffee7ccfd/cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go#L288
2. https://github.com/GoogleCloudPlatform/compute-image-tools/blob/5a2d47d1138373d1ee2a2152eb3a96dffee7ccfd/cli_tools_tests/e2e/gce_image_import_export/test_suites/import/import_tests.go#L317

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -30,6 +30,7 @@ import time
 import guestfs
 import utils
 import utils.diskutils as diskutils
+from utils.guestfsprocess import run
 
 repo_compute = '''
 [google-compute-engine]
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -148,12 +115,21 @@ def DistroSpecific(g: guestfs.GuestFS):
...
   if install_gce == 'true':
     logging.info('Installing GCE packages.')
     g.write('/etc/yum.repos.d/google-cloud.repo', repo_compute % el_release)
     if el_release == '6':
+      # yum operations fail when the epel repo is used with stale
+      # ca-certificates, causing translation to fail. To avoid that,
+      # update ca-certificates when the epel repo is found.
+      #
+      # The `--disablerepo` flag does the following:
+      #  1. Skip the epel repo for *this* operation only.
```
zoran15, line 127: The old logic had a check (on error) that epel is used. The PR description mentions update step is done when epel is used. Where is this step performed in this PR?

EricEdens: It's at line 129. The update will only occur if epel is installed; if it's not installed, the yum command will fail, and the update won't occur.

zoran15 (approved):

PR #1427, opened 2020-11-13
Use anonymized errors in populateScratchBucketGcsPath

EricEdens:
Ensure that anonymized error is used as the first argument to `daisy.Errf`.

### DIFF
--- a/cli_tools/common/utils/param/populator_test.go
+++ b/cli_tools/common/utils/param/populator_test.go
@@ -17,6 +17,7 @@ package param
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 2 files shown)


cli_tools/common/utils/param/populator_test.go
```diff
@@ -267,6 +268,7 @@ func TestPopulator_DeleteResources_WhenScratchBucketInAnotherProject(t *testing.
 		deleteResult         error
 		deleteExpected       bool
 		expectedError        string
+		expectedAnonError    string
```
zoran15, line 271: What is "anon"?

EricEdens:
Anon as in anonymous or anonymized. Updated to expectedAnonymizedError

>

cli_tools/common/utils/param/populator_test.go
EricEdens: Anon as in anonymous or anonymized. Updated to expectedAnonymizedError

zoran15: I still see `anon` in tests and in param_utils.go

EricEdens: ahh, good find. Updated all.

zoran15 (approved):

PR #1428, opened 2020-11-15
Revert "Inspection refactor: Use protobufs for transport (#1410)"

EricEdens:
#1410 caused a panic in `el-yum-not-found` at [1]. Rolling back to ensure master is green while I add more test coverage.

1. https://github.com/GoogleCloudPlatform/compute-image-tools/blob/93b6de86409bc966e4cabdff93c2e852c469ab4a/cli_tools/common/disk/inspect.go#L181-L180


EricEdens:
Rolled forward: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1429

PR #1432, opened 2020-11-16
Migrate cli_tools to use Debian 10 worker

EricEdens:
@zmarano  is deprecating the Debian 9 worker.

### DIFF
--- a/cli_tools/common/utils/daisy/daisy_utils.go
+++ b/cli_tools/common/utils/daisy/daisy_utils.go
@@ -86,6 +86,8 @@ var (
 	}
 	privacyRegex    = regexp.MustCompile(`\[Privacy\->.*?<\-Privacy\]`)
 	privacyTagRegex = regexp.MustCompile(`(\[Privacy\->)|(<\-Privacy\])`)
+
+	debianWorkerRegex = regexp.MustCompile("projects/compute-image-tools/global/images/family/debian-\\d+-worker")
 )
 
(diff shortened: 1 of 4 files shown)


cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -146,14 +148,14 @@ func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool)
...
+// compatible by adding "UEFI_COMPATIBLE" to GuestOSFeatures. Debian workers
 // are excluded until UEFI becomes the default boot method.
 func UpdateToUEFICompatible(workflow *daisy.Workflow) {
 	workflow.IterateWorkflowSteps(func(step *daisy.Step) {
 		if step.CreateDisks != nil {
 			for _, disk := range *step.CreateDisks {
-				// for the time being, don't run Debian 9 worker in UEFI mode
-				if strings.Contains(disk.SourceImage, "projects/compute-image-tools/global/images/family/debian-9-worker") {
+				// for the time being, don't run Debian worker in UEFI mode
+				if debianWorkerRegex.MatchString(disk.SourceImage) {
```
zoran15, line 158: This restriction was added more than a year ago. Might not be necessary anymore since UEFI is well tested. We can leave it as is, but should probably follow up

zoran15 (approved):

zmarano (approved):

PR #1444, opened 2020-12-01
CLI: Add new common logger

EricEdens:
`ToolLogger` replaces existing logging interfaces. It supports explicit log destinations, and type-safe collection of metrics, and prepares us for a future refactoring that shares code between image import and OVF import.

Two existing loggers are replaced in this PR. Those are shown below, along with candidates for future  replacement.

 * `logWriter` in `gce_vm_image_import`  -- replaced in this PR
 * `logging.LoggerInterface`  -- replaced in this PR
 * `logger.Loggable` - replacement candidate
 * `service.SingleImageImportLoggableBuilder` -  replacement candidate

## Log line format

Log lines are formatted to match Daisy's standard logger. This improves our current logs, which emit at least three different formats, depending on where the log is emitted:

1. Image import: Colon after prefix, no UTC offset: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/gce_vm_image_import/cli/cli.go#L97
2. Daisy: Colon after prefix, UTF offset: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy/logger.go#L293
3. OVF import (and other consumers of LoggerInterface): No colon after import,  no UTC offset: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/common/utils/logging/logger.go

Here's what it looks like:

**OVF import - Before PR**
 * Inconsistent colon
 * No vertical alignment of message
 * Different days and hours are printed

```
CLI code: [import-ovf] 2020-12-01T04:36:22Z Will create instance of `g1-small` machine type.
Daisy:    [import-ovf]: 2020-11-30T20:36:22-08:00 Validating workflow
```

**OVF import - After PR**

```
CLI code: [import-ovf]: 2020-11-30T20:36:22-08:00 Will create instance of `g1-small` machine type.
Daisy:    [import-ovf]: 2020-11-30T20:36:22-08:00 Validating workflow
```

**Image import - Before PR**
 * No vertical alignment of message
 * Different days and hours are printed

```
CLI code: [import-image]: 2020-12-01T04:36:22Z Running disk inspections on zones/us-central1-c/disks/disk-86l3d.
Daisy:    [inspect-disk]: 2020-11-30T20:36:22-08:00 Validating workflow
```

**Image import - After PR**
```
CLI code: [import-image]: 2020-11-30T20:36:22-08:00 Running disk inspections on zones/us-central1-c/disks/disk-dltdg.
Daisy:    [inspect-disk]: 2020-11-30T20:36:22-08:00 Validating workflow
```


cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,179 @@
...
-// NewStdoutLogger creates a new logger which uses prefix for all messages logged.
-// All messages are sent to stdout.
-func NewStdoutLogger(prefix string) LoggerInterface {
-	return stdoutLogger{prefix: prefix}
+// ToolLogger is a logger for interactive tools. It supports
+// string messages and structured metrics.
+//
+// Structured metrics are accumulated over the lifespan of the logger.
+type ToolLogger interface {
+	LogWriter
```
dntczdx, line 73: I don't see an obvious benefits to split the interface to writer+reader. Would you explain the design? Just to hide the reader interface for CLI program?

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,179 @@
...
+	l.trace = []string{}
+	proto.Merge(ret, l.outputInfo)
+	l.outputInfo = &pb.OutputInfo{}
+	return ret
+}
+
+// write a message to the underlying logger, and buffer it for inclusion in OutputInfo.SerialLogs.
+func (l *defaultToolLogger) write(prefix, message string) {
+	var logLineParts []string
+	if prefix != "" {
```
dntczdx, line 164: if prefix has no ":" as suffix, attach a ":". Then blindly append it to logLineParts. That would be more natural in my opinion.

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,179 @@
...
+			logLineParts = append(logLineParts, prefix)
+		} else {
+			logLineParts = append(logLineParts, prefix+":")
+		}
+	}
+	logLineParts = append(logLineParts, l.timeProvider().Format(l.timestampFormat), message)
+	logLine := strings.Join(logLineParts, " ")
+	l.output.Print(logLine)
+	l.userAndDebugBuffer.WriteString(logLine)
+	if !strings.HasSuffix(logLine, "\n") {
```
dntczdx, line 175: what about l.output? DO you think a guaranteed "\n" is reasonable for output?

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,179 @@
...
+		if strings.HasSuffix(prefix, ":") {
+			logLineParts = append(logLineParts, prefix)
+		} else {
+			logLineParts = append(logLineParts, prefix+":")
+		}
+	}
+	logLineParts = append(logLineParts, l.timeProvider().Format(l.timestampFormat), message)
+	logLine := strings.Join(logLineParts, " ")
+	l.output.Print(logLine)
+	l.userAndDebugBuffer.WriteString(logLine)
```
dntczdx, line 174: I don't feel it's guaranteed to operate userAndDebugBuffer in a single thread. We may add thread-safe mechanism.

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,179 @@
...
-	Message        string    `json:"message"`
+// LogWriter is a logger for interactive tools. It supports
+// string messages and structured metrics.
+//
+// Structured metrics are accumulated over the lifespan of the logger.
+//
+// To rebuild the mock, run `go generate ./...`
+//go:generate go run github.com/golang/mock/mockgen -package mocks -source $GOFILE -destination ../../../mocks/mock_logger.go
+type LogWriter interface {
+	// WriteUser messages are shown in Pantheon or Gcloud to the user.
```
dntczdx, line 35: and also in OutputInfo.serial_outputs?
I feel there are overlaps among these 4 methods right? They are not all exclusive.

cli_tools/common/utils/logging/logger.go
EricEdens: sgtm - done

cli_tools/common/utils/logging/logger.go
EricEdens: The standard library guarantees a newline with `log.Logger.Print`; that said, I think it's clearer to put the newline prior to the write so that future readers aren't confused. Updated.

cli_tools/common/utils/logging/logger.go
EricEdens: Good call, I hadn't thought about concurrent usage :/  Updated to serialize all reads and writes with a mutex.

cli_tools/common/utils/logging/logger.go
EricEdens: serialized in the latest version.

dntczdx: I meant the comment line - "WriteUser messages are shown in Pantheon or Gcloud to the user", per my understanding it's also passed into OutputInfo.serial_outputs, so I feel the description is not that accurate. Let me know if I misunderstood it.

dntczdx (approved):
overall lgtm

cli_tools/common/utils/logging/logger.go
EricEdens: I see -- good point :)  Clarified in the latest version where each message will be shown.

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+
+	// mutationLock should be taken when reading or writing trace, userAndDebugBuffer, or outputInfo.
+	mutationLock sync.Mutex
 }
 
-func (e *LogEntry) String() string {
-	return fmt.Sprintf("%s %s", e.LocalTimestamp.Format("2006-01-02T15:04:05Z"), e.Message)
+// WriteUser writes message to the underlying log.Logger, and then buffers the message
+// for inclusion in ReadOutputInfo().
+func (l *defaultToolLogger) WriteUser(message string) {
```
zoran15, line 129: Have you considered using "Info" instead of "User" for this category of messages? This would be more in line with existing logging framework's categorizations (e.g. INFO, WARNING, DEBUG, TRACE). We can explicitly state INFO is shown to users. This way, we can also use these messages for other purposes except for showing them to users.

Renaming it to "Info" would also avoid confusion on why `WriteUser` is not writing a user(name) but is actually logging.

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+// Structured metrics are accumulated over the lifespan of the logger.
+//
+// To rebuild the mock, run `go generate ./...`
+//go:generate go run github.com/golang/mock/mockgen -package mocks -source $GOFILE -destination ../../../mocks/mock_logger.go
+type LogWriter interface {
+	// WriteUser messages appear in the following places:
+	//  1. Web UI and gcloud.
+	//  2. Standard output of the CLI tool.
+	//  3. Backend trace logs (all debug and user logs are combined to a single trace log).
+	WriteUser(message string)
```
zoran15, line 40: Consider using Log* prefix or no prefix at all. Eg. Info(), Debug(), Trace()

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+}
+
+func (l redirectShim) Write(p []byte) (n int, err error) {
+	l.writer.WriteUser(string(p))
+	return len(p), nil
+}
+
+// LogReader exposes pb.OutputInfo to a consumer.
+type LogReader interface {
+	ReadOutputInfo() *pb.OutputInfo
```
zoran15, line 71: I don't see this called anywhere. What is the plan for this func?

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
-	LocalTimestamp time.Time `json:"localTimestamp"`
-	Message        string    `json:"message"`
+// LogWriter is a logger for interactive tools. It supports
+// string messages and structured metrics.
+//
+// Structured metrics are accumulated over the lifespan of the logger.
+//
+// To rebuild the mock, run `go generate ./...`
+//go:generate go run github.com/golang/mock/mockgen -package mocks -source $GOFILE -destination ../../../mocks/mock_logger.go
+type LogWriter interface {
```
zoran15, line 35: Logger might be a more natural name, even in the presence of LogWritter interface. It doesn't only write logs, it also collects them. It's also mroe commonly used name for a class of these responsibilities.

cli_tools/common/utils/logging/logger.go
zoran15: +1 on this

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+	// Prepend stdout logs (user, debug) to the trace logs.
+	var combinedTrace []string
+	if l.userAndDebugBuffer.Len() > 0 {
+		combinedTrace = []string{l.userAndDebugBuffer.String()}
+		l.userAndDebugBuffer.Reset()
+	}
+	ret := &pb.OutputInfo{SerialOutputs: append(combinedTrace, l.trace...)}
+	l.trace = []string{}
+	proto.Merge(ret, l.outputInfo)
+	l.outputInfo = &pb.OutputInfo{}
```
zoran15, line 176: minor: this would be more readable if output building and buffer reset were in two separate functions that are called from here. 
e.g.:
ret := buildOutput()
resetBuffers()
return ret

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+	return ret
 }
 
-type defaultLogger struct{}
+// write a message to the underlying logger, and buffer it for inclusion in OutputInfo.SerialLogs.
+func (l *defaultToolLogger) write(prefix, message string) {
```
zoran15, line 181: Since it's adding a new line character, `writeLine()`?

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+	var logLineBuilder strings.Builder
+	// If there's a prefix, ensure it ends with a colon.
+	if prefix != "" {
+		logLineBuilder.WriteString(prefix)
+		if !strings.HasSuffix(prefix, ":") {
+			logLineBuilder.WriteByte(':')
+		}
```
zoran15, line 188: Consider updating the prefix on construction if `:` is always appended anyway, to avoid having to do it for each line

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+		logLineBuilder.WriteByte(' ')
```
zoran15, line 189: here as well, if prefix is not used for other purposes where `: ` would be extra.

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,205 @@
...
+// LogReader exposes pb.OutputInfo to a consumer.
+type LogReader interface {
+	ReadOutputInfo() *pb.OutputInfo
+}
+
+// ToolLogger is a logger for interactive tools. It supports
+// string messages and structured metrics.
+//
+// Structured metrics are accumulated over the lifespan of the logger.
+type ToolLogger interface {
```
zoran15, line 78: If this was named just `Logger` and had all functions from the two interfaces inlined, it would be a simpler structure.Related to the question below on the split between these two interfaces.

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,200 @@
...
+	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pb"
+	"github.com/golang/protobuf/proto"
 )
 
-// LogEntry encapsulates a single log entry.
-type LogEntry struct {
-	LocalTimestamp time.Time `json:"localTimestamp"`
-	Message        string    `json:"message"`
+// LogWriter is a logger for interactive tools. It supports
+// string messages and structured metrics.
```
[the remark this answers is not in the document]
EricEdens: This PR makes a design choice: Method names indicate the destination/consumer of a message. To show a message to a user, call `WriteUser`. Other systems use generic names such as `INFO`; this PR consciously avoids that, since `INFO` is vague and requires extra work for future developers to understand where a message will be written.

Given that, `WriteMetric` signals that the destination is the metrics server.

Regarding structured messages -- the OVF consolidation project doesn't require that, so I didn't design that into the logger. From the top of my head: if we wanted to support other types in the future, we could change the parameter type to `interface {}` in the future.

Does that sound okay?

cli_tools/common/utils/logging/logger.go
EricEdens: Resolving in favor of the "From the context of logging" thread

cli_tools/common/utils/logging/logger.go
EricEdens: Good question: It's a preventative measure to hide the information in `ReadOutputInfo`. Since this accumulates information, and will be available throughout the life of the request, there's a temptation for code to use these fields to perform control flow and logic. 

This creates hard-to-understand code, since "business information" can come from two places: function parameters, and this magical shared context.

Concretely, in other projects I've seen this manifest in as:
1. Putting  information that's required for logic into the global context, but shouldn't be logged. 
2. Dependencies are implicit, so re-ording tasks can break code that assumes a value will be set.
3. A developer thinks it's safe to modify metrics or logging, without knowing that some downstream code will be affected. Since the components are far away, this typically won't be found in unit tests. If it's found in integration tests, it typically results in hard-to-debug failures.

cli_tools/common/utils/logging/logger.go
EricEdens: The only log consumer is the logging backend -- the `LogReader` interface will replace current usage of `Loggable`:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/e16e8e02ae7fa1c3e2940987368500b7d821ea61/cli_tools/common/utils/logging/service/logger.go#L184

cli_tools/common/utils/logging/logger.go
EricEdens: Resolving, since I think the thread "If this was named just " also addresses this

cli_tools/common/utils/logging/logger.go
EricEdens: Resolving since the thread "I don't see an obvious benefits to split the interface to writer+reader." is open for this s

cli_tools/common/utils/logging/logger.go
zoran15: The reading part is more like building the final result than just reading it, since it's a bit involved and it clears the caches. What about having the main `Logger` interface (that would do what `LogWriter` does) and `LogOutputInfoBuilder` that does what LogReader does.

cli_tools/common/utils/logging/logger.go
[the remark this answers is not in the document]
zoran15: The consumer of messages is up to clients calling our backend. gcloud and Pantheon do show user messages to users, but CLI tools could potentially be run via API or integrated into other clients where there might be no users to see these messages. Ideally, CLI tools should not be aware of gcloud and pantheon, if possible.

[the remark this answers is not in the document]
EricEdens: > CLI tools could potentially be run via API or integrated into other clients where there might be no users to see these messages. Ideally, CLI tools should not be aware of gcloud and pantheon, if possible.

I like that you're thinking of future extensions. Would you be alright trying this interface for our current requirements? In the future, if this code is deployed such that there isn't an end user, then we can re-evaluate whether the interface makes sense?

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,200 @@
...
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pb"
+	"github.com/golang/protobuf/proto"
 )
 
-// LogEntry encapsulates a single log entry.
-type LogEntry struct {
-	LocalTimestamp time.Time `json:"localTimestamp"`
-	Message        string    `json:"message"`
+// LogWriter is a logger for interactive tools. It supports
```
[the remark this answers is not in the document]
EricEdens: Changed to "Logger is a logger for CLI tools. It supports"

cli_tools/common/utils/logging/logger.go
EricEdens: resolving in favor of the above thread

cli_tools/common/utils/logging/logger.go
EricEdens: I'd definitely agree if the struct members were immutable, but given that they could change, it seems most safe (to me) to ensure the correct output when the prefix is written.

That said, I'm not wedded to it. If you prefer doing the formatting once in `NewToolLogger`, I can do that!

cli_tools/common/utils/logging/logger.go
```diff
@@ -15,50 +15,200 @@
...
+	// WriteDebug messages are shown in the console output of the CLI tool.
+	WriteDebug(message string)
+	// WriteTrace messages are saved to the logging backend  (OutputInfo.serial_outputs)
+	WriteTrace(message string)
+	// WriteMetric merges all non-default fields into a single OutputInfo instance.
+	WriteMetric(metric *pb.OutputInfo)
+}
+
+// RedirectGlobalLogsToUser redirects the standard library's static logger
+// to logWriter.User.
```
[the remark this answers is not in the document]
EricEdens: Thanks -- changed to "All messages are written to logger's User level."

cli_tools/common/utils/logging/logger.go
EricEdens: Changed to no prefix

cli_tools/common/utils/logging/logger.go
EricEdens: done

cli_tools/common/utils/logging/logger.go
EricEdens: done

zoran15 (approved):

PR #1447, opened 2020-12-02
Image import: Renaming to generalize UEFI processor

EricEdens:
No behavioral changes: just renaming, and increasing test coverage.

This prepares for upcoming work to modularize import. Since OVF import doesn't create disk images, the license needs to be added to the disk. That happens during `metadata_processor`, since the OS is possibly unknown until inspection runs.


cli_tools/gce_vm_image_import/importer/metadata_processor.go
```diff
@@ -25,13 +25,18 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
-// uefiProcessor mutate the disk in order to make the image compatible with UEFI boot.
-type uefiProcessor struct {
+// metadataProcessor ensures metadata is present on a GCP disk. If
+// metadata is missing, the disk will be recreated.
+type metadataProcessor struct {
```
dntczdx, line 30: I don't know whether diskMetadataProcessor is a better name (since there are also many operations on instance metadata in our team). Leave that to you. Other than that LGTM.

dntczdx (approved):

zoran15:
> No behavioral changes: just renaming, and increasing test coverage.
> 
> This prepares for upcoming work to modularize import. Since OVF import doesn't create disk images, the license needs to be added to the disk. That happens during `metadata_processor`, since the OS is possibly unknown until inspection runs.



> No behavioral changes: just renaming, and increasing test coverage.
> 
> This prepares for upcoming work to modularize import. Since OVF import doesn't create disk images, the license needs to be added to the disk. That happens during `metadata_processor`, since the OS is possibly unknown until inspection runs.

Is the plan to have metadata processor do UEFI detection and license setting? I wonder if having a separate processor for license setting that would be used only in OVF import make things cleaner, instead of mixing responsibilities within single processor.

EricEdens:
There's a pipeline of processors: first inspection, then metadata modifications, and finally translation:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/1e15f592117f0eba3886ec86a819b1ef436da5c1/cli_tools/gce_vm_image_import/importer/processor.go#L54

This same pipeline will get used for image import and OVF import. The only difference is what happens *after* the pipeline runs. For image import, an image is created from the PD and the PD is deleted. For OVF import, the disk is retained and no image is created.

I propose we attache licenses to the disk in the middle of the pipeline for two reasons:
1. Since specifying the OS will be optional, this is the first moment where we'll definitely know the OS on the disk.
2. OVF import needs licenses on the disk, not the image.

zoran15:
> There's a pipeline of processors: first inspection, then metadata modifications, and finally translation:
> 
> https://github.com/GoogleCloudPlatform/compute-image-tools/blob/1e15f592117f0eba3886ec86a819b1ef436da5c1/cli_tools/gce_vm_image_import/importer/processor.go#L54
> 
> This same pipeline will get used for image import and OVF import. The only difference is what happens _after_ the pipeline runs. For image import, an image is created from the PD and the PD is deleted. For OVF import, the disk is retained and no image is created.
> 
> I propose we attache licenses to the disk in the middle of the pipeline for two reasons:
> 
> 1. Since specifying the OS will be optional, this is the first moment where we'll definitely know the OS on the disk.
> 2. OVF import needs licenses on the disk, not the image.

Wouldn't a different (or customized) processor provider solve this by adding another processor to the pipeline that performs license attachment?

I'm woried this new metadata processor will become catch all for different and potentially unrelated operations on imported disks. With the way we currently have processor pipeline set up, it seems like adding another processor is an easy way to keep different operations contained in different classes.

EricEdens:
> I'm woried this new metadata processor will become catch all for different and potentially unrelated operations on imported disks.

That's fair, and I like the foresight :) 

To help clarify, this processor has a (hopefully) well-defined scope. It's limited to the following:

1. Changes are limited to fields on the compute.Disk resource.
2. All proposed changes must be known *prior* to running the processor.
3. Control flow is the same, regardless if there are one or many proposed changes:
  a. Fetch the current disk
  b. Determine if the requested changes are already present
  c. If no changes are required, return
  d. Otherwise, make the changes by cloning the disk, and then delete the old disk

Tactically, if we created individual processors for each metadata change, then could be one clone per metadata change. Consider the following pipeline:

1. Set guest OS feature to WINDOWS
2. Set guest OS feature to UEFI_COMPATIBLE
3. Add license for the particular OS
4. Add license for tracking the import method (OVF vs image)

This would require four clones (or two if there was one processor for each metadata type).

zoran15:
Consider the following pipeline:
> 
> 1. Set guest OS feature to WINDOWS
> 2. Set guest OS feature to UEFI_COMPATIBLE
> 3. Add license for the particular OS
> 4. Add license for tracking the import method (OVF vs image)
> 
> This would require four clones (or two if there was one processor for each metadata type).

We should definitely avoid multiple clones of a disk. At the same time, we should try to separate the code that does different metadata changes. We have at least one new change in the future (setting BARE_METAL Guest OS feature). Something to keep in mind when the time to implement all these changes comes.

EricEdens:
The BARE_METAL example is helpful. I've used this, and your other feedback when working on #1451!

PR #1448, opened 2020-12-02
Add SerialPortToLog parameter to InstanceBase

same-id:
When using the WaitForInstancesSignal with a port other than 1:

  "WaitForInstancesSignal": [{
    "Name": "builder",
    "SerialOutput": {
      "Port": 3,
      "SuccessMatch": "BuilderSuccess:",
      "FailureMatch": "BuilderFailure:",
      "StatusMatch": "BuilderStatus:"
    }
  }]

It is not possible to see the entire serial port logs in GCS due to the
fact that the logging is actually an attribute of the CreateInstances
and the port number there is hardcoded to 1.

Introduce SerialPortToLog parameter for instance base to configure which
serial port is being logged for the given instance.

### DIFF
--- a/daisy/instance.go
+++ b/daisy/instance.go
@@ -106,10 +106,12 @@ type InstanceBase struct {
 	// Should an existing instance of the same name be deleted, defaults to false
 	// which will fail validation.
 	OverWrite bool `json:",omitempty"`
+	// Serial port to log to GCS bucket, defaults to 1
+	SerialPortsToLog []int64 `json:",omitempty"`
 }
 
 // Instance is used to create a GCE instance using GA API.
-// Output of serial port 1 will be streamed to the daisy logs directory.
+// By default, output of serial port 1 will be streamed to the daisy logs directory.
 type Instance struct {
 	InstanceBase
 	compute.Instance
@@ -119,7 +121,7 @@ type Instance struct {
 }
 
 // InstanceBeta is used to create a GCE instance using Beta API.
-// Output of serial port 1 will be streamed to the daisy logs directory.
+// By default, output of serial port 1 will be streamed to the daisy logs directory.
 type InstanceBeta struct {
 	InstanceBase
 	computeBeta.Instance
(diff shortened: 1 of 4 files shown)


zoran15 (commented):
Add test coverage to instance_test.go and step_create_instance_test.go

same-id:
Added test cases

daisy/instance.go
```diff
@@ -632,6 +643,13 @@ func (i *InstanceBeta) getComputeDisks() []*computeDisk {
 	return computeDisks
 }
 
+func (ib *InstanceBase) validateSerialPortToLog() (errs DError) {
+	if ib.SerialPortToLog < 0 || ib.SerialPortToLog > 4 {
+		errs = addErrs(errs, Errf("cannot create instance: SerialPortToLog must be between 1-4"))
```
EricEdens, line 648:
```suggestion
		errs = addErrs(errs, Errf("cannot create instance: SerialPortToLog must be between 1-4, inclusive"))
```

zoran15:
> Added test cases

I don't see the change. Maybe you didn't push?

EricEdens (commented):
thanks for the PR! lgtm (ditto Zoran's feedback about tests)

same-id:
> > Added test cases
> 
> I don't see the change. Maybe you didn't push?

Yep, my bad

daisy/instance.go
same-id: Added

EricEdens (approved):

zoran15 (approved):

daisy/instance.go
```diff
@@ -106,10 +106,12 @@ type InstanceBase struct {
 	// Should an existing instance of the same name be deleted, defaults to false
 	// which will fail validation.
 	OverWrite bool `json:",omitempty"`
+	// Serial port to log to GCS bucket, defaults to 1
+	SerialPortToLog int64 `json:",omitempty"`
```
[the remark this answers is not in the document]
same-id: Sure, done

adjackura (approved):

PR #1455, opened 2020-12-11
Configuring WinRM on Windows clients

bkatyl:
Configuring WinRM on Windows clients. This is to provide the same WinRM management capability to clients that is present in Windows Server images.

### DIFF
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -303,6 +303,13 @@ function Install-32bitPackages {
   }
 }
 
+function Enable-WinRM {
+  if ($pn -like '*Enterprise') {
+    Write-Host 'Translate: Windows Client detected, enabling WinRM (including on Public networks).'
+    & winrm quickconfig -quiet -force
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -303,6 +303,13 @@ function Install-32bitPackages {
   }
 }
 
+function Enable-WinRM {
+  if ($pn -like '*Enterprise') {
```
zoran15, line 307: Is this matching correct: "*Enterprise" means Windows Client? I would have expected it to be the opposite.

bkatyl: Yes, Windows 7/8.1/10 Enterprise are the client version that is available via volume licence agreements. They started usinng the Enterprise edition SKU in vista.

A full list of the client and server product names is at https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions

zoran15 (approved):

EricEdens (approved):

PR #1462, opened 2020-12-15
Image import: Move importer package to common/disk

EricEdens:
Expose image import code so that OVF can call it.


zoran15 (commented):
Some of the classes in this move don't belong to a common package. E.g. args.go are specific to gce_vm_image_import. source.go is also specific to image import and won't apply to OVF import (e.g. there's no image to VM import, while there's image to image import).

zoran15 (commented):
Another one is importer.go obviously.

cli_tools/common/disk/importer/bootable_disk_processor_test.go
```diff
@@ -33,7 +33,7 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-	opensuse15workflow = path.Join("../../../daisy_workflows/image_import", settings.WorkflowPath)
+	opensuse15workflow = path.Join("../../../../daisy_workflows/image_import", settings.WorkflowPath)
 }
 
 func TestBootableDiskProcessor_Process_WritesSourceDiskVar(t *testing.T) {
```
EricEdens, line 39: > Some of the classes in this move don't belong to a common package. E.g. args.go are specific to gce_vm_image_import. source.go is also specific to image import and won't apply to OVF import (e.g. there's no image to VM import, while there's image to image import).

Thanks for the feedback! I was intending all of these files to be moved over; the subsequent change will be to allow the result to be either an image or a disk. OVF will specify disk, while image import specifies image. (I wanted this review to be a clean directory move.)


You mentioned some specific files; notes below:

- `args.go`: I *do* want ImportArguments to be exported, and the caller will be responsible for providing the fields within that struct. The CLI parsing code, though: I'm ambivalent on exporting that. Thoughts?
- `source.go`: The key type is `Source`, which is agnostic of whether it's an image or file. In the case of OVF, it'll only use the file type. 
- `importer.go`: OVF will use the `Importer.Run` method for each of the disks in the OVF descriptor.

zoran15: This change is making image importer a library. I think the responsibilities should be clear between the CLI tool (gce_vm_image_import) and the importer library (this change). As you mentioned, CLI arg parsing doesn't belong to the importer library. But even other pieces of the imporer might not make sense to belong to a library. This can be addressed in a refactor that you'll do later. 

Importer was designed as a CLI tool. Copy/pasting it to a library will probably lead to a poorly designed library.

For example, what kind of import is done is controlled by args. Some combos of args are invalid (e.g. os provided and dataDisk set to true). If importer was a library designed from ground up, it would probably have methods such as:
* importBootableDisk
* importDataDisk
* ImportBootableImage
* ImportDataImage

Since this PR is just a folder copy, we can discuss how to make importer a proper library before the next PR. A short doc with proposed changes and the new architecture makes sense here.

EricEdens: The proposed interface is in the doc titled "OVF Consolidation". I went ahead and closed this PR, since it sounds like attacking the problem from the opposite direction makes more sense:

1. PR that creates the interface from "OVF Consolidation"
2. PR that moves the code from gce_vm_image_import to disk/import and implements that interface

PR #1465, opened 2020-12-18
[import] support compute_service_account for image import + onstep import

dntczdx:
e2e is also added accordingly. Followed the pattern of image export.


cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -94,48 +94,51 @@ type InputParams struct {
...
 type ImageExportParams struct {
 	*CommonParams
 
-	DestinationURI string `json:"destination_uri,omitempty"`
-	SourceImage    string `json:"source_image,omitempty"`
-	Format         string `json:"format,omitempty"`
+	DestinationURI        string `json:"destination_uri,omitempty"`
+	SourceImage           string `json:"source_image,omitempty"`
+	Format                string `json:"format,omitempty"`
+	ComputeServiceAccount string `json:"compute_service_account,omitempty"`
```
zoran15, line 122: This was missed in image export PR?

cli_tools/gce_ovf_import/main.go
```diff
@@ -69,6 +69,7 @@ var (
 	hostname                    = flag.String(ovfimportparams.HostnameFlagKey, "", "Specify the hostname of the instance to be created. The specified hostname must be RFC1035 compliant.")
 	machineImageStorageLocation = flag.String(ovfimportparams.MachineImageStorageLocationFlagKey, "", "GCS bucket storage location of the machine image being imported (regional or multi-regional)")
 	buildID                     = flag.String("build-id", "", "Cloud Build ID override. This flag should be used if auto-generated or build ID provided by Cloud Build is not appropriate. For example, if running multiple imports in parallel in a single Cloud Build run, sharing build ID could cause premature temporary resource clean-up resulting in import failures.")
+	computeServiceAccount       = flag.String("compute_service_account", "", "Compute service account to be used by importer. When empty, the Compute Engine default service account is used.")
```
zoran15, line 72: Is this PR implementing compute service account for OVF import as well? I thought it's only image/one-step import?

Btw, OVF importer uses "-" instead of "_" in flags (should be `compute-service-account`).

cli_tools/gce_vm_image_import/importer/args_test.go
```diff
@@ -180,6 +180,13 @@ func TestTrimComputeEndpoint(t *testing.T) {
 		parseAndValidate(t, "-compute_endpoint_override", "  http://endpoint ").ComputeEndpoint)
 }
 
+func TestTrimComputeServiceAccount(t *testing.T) {
+	assert.Equal(t, "",
+		parseAndValidate(t, "-compute_service_account", " 	").ComputeServiceAccount)
+	assert.Equal(t, "default",
+		parseAndValidate(t, "-compute_service_account", " default	").ComputeServiceAccount)
```
zoran15, line 187: minor: use a non-default CE service account for testing (since default is handled in a specific way in the code)

cli_tools_tests/e2e/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -364,14 +379,96 @@ func runImageImportShadowDiskCleanedUpWhenMainInflaterFails(ctx context.Context,
 	}
 }
 
+func runImageImportOSDefaultServiceAccountWithMissingPermissionsTest(ctx context.Context, testCase *junitxml.TestCase, logger *log.Logger,
```
zoran15, line 382: It's not clear from either test case names or this func names if this should suceed or fail. E.g. missing permissions should probably fail but I don't see any assertion for test failure.

zoran15 (commented):
Passing compute_service_account variable to each translate workflow seems like something that could have been done in Daisy's postValidateModifier. Even though we are planning to deprecate this helper func, Andrew is working on an alternate solution where we would be able to run this code in preValidateModifier (e.g. after loading). 

I'm not strongly opposed to the current implementation though, as it works.

cli_tools_tests/e2e/gce_image_import_export/test_suites/import/cli_tests.go
```diff
@@ -37,6 +37,10 @@ const (
 	testSuiteName = "ImageImportCLI"
 )
 
+var (
```
EricEdens, line 40: Can you add a comment describing what this is?

cli_tools_tests/e2e/gce_image_import_export/test_suites/import/cli_tests.go
EricEdens: Related: I'm thinking for each tool, you'd want coverage for all of these combinations (at least the first two, but for full coverage the third as well):

- Service account specified:  default, custom
- Does the specified account exist?
- Does the specified account have sufficient permissions?

Some will generate error messages, and that's good, since it will allow you to assert the error message that is returned.

daisy_workflows/image_import/debian/translate_debian.wf.json
```diff
@@ -20,6 +20,10 @@
     "import_subnet": {
       "Value": "",
       "Description": "SubNetwork to use for the import instance"
+    },
+    "compute_service_account": {
+      "Value": "default",
+      "Description": "Compute service account to run as"
```
EricEdens, line 26:
```suggestion
      "Description": "Service account that will be used by the translation worker instance."
```

(I tried to use phrasing from: https://cloud.google.com/compute/docs/access/service-accounts#associating_a_service_account_to_an_instance )

daisy_workflows/image_import/inflate_file.wf.json
```diff
@@ -38,6 +38,10 @@
     "import_license": {
       "Value": "projects/compute-image-tools/global/licenses/virtual-disk-import",
       "Description": "Import License used for tracking migration workflow use."
+    },
+    "compute_service_account": {
+      "Value": "default",
+      "Description": "Compute service account to run as"
```
EricEdens, line 44: Similar to the comment for translation workflows, let's specify which instance the service account will be use diwth.

daisy_workflows/image_import/inspection/boot-inspect.wf.json
```diff
@@ -16,6 +16,10 @@
     },
     "is_inspect_os": {
       "Description": "Whether to inspect OS."
+    },
+    "compute_service_account": {
+      "Value": "default",
+      "Description": "Compute service account to run as"
```
EricEdens, line 22: Similarly, specify which instance the account will be used with.

daisy_workflows/image_import/debian/translate_debian.wf.json
```diff
@@ -20,6 +20,10 @@
     "import_subnet": {
       "Value": "",
       "Description": "SubNetwork to use for the import instance"
+    },
```
EricEdens, line 23: Given the large number of files changed, I'd recommend adding a test similar to the one below that parses each workflow and validates the changes that you've made:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/4f42f758697b956153b95d4a216b5154f6636246/cli_tools/common/utils/daisy/daisy_utils_test.go#L95

daisy_workflows/image_import/debian/translate_debian.wf.json
zoran15: We should use https://source.corp.google.com/piper///depot/google3/third_party/py/googlecloudsdk/surface/compute/images/export.py;l=168?q=export.py%20images&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental as a base for help text as it has been reviewed by doc writers

zoran15: We should use the help text for --compute-service account in `gcloud beta compute images export` as a base since it was reviewed by doc writers. Can't link the source here as it's in piper.

cli_tools/common/disk/inspect.go
```diff
@@ -47,14 +47,19 @@ type Inspector interface {
 // NewInspector creates an Inspector that can inspect GCP disks.
 // A GCE instance runs the inspection; network and subnet are used
 // for its network interface.
-func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, subnet string) (Inspector, error) {
+func NewInspector(wfAttributes daisycommon.WorkflowAttributes, network string, subnet string,
+	computeServiceAccount string) (Inspector, error) {
```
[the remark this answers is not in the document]
zoran15: SetWorkflowAttributes sets actual attributes on a workflow (e.g. w.GCSPath, w.Project...), while compute service account is just a workflow variable. Without renaming SetWorkflowAttributes, this doesn't seem to be the right place to move this logic. 

I see two options:
* make workflow attributes more generic to include variables as well, or
* have a set of common workflow variables in e.g. WorkflowVariables/SetWorkflowVariables()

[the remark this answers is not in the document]
EricEdens: Good point about attributes vs variables.

I like the first option (struct that contains attributes and variables that we always want to support), to reinforce that we always want to support non-default service accounts.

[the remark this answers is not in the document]
dntczdx: Can we refactor that later? Right now it's pretty case-by-case, we don't always set those vars. Let's do it later to make it more clear due to the size of this PR.

cli_tools/gce_vm_image_import/importer/api_inflater.go
```diff
@@ -171,6 +171,10 @@ func (inflater *apiInflater) calculateChecksum(ctx context.Context, diskURI stri
 	w := daisy.New()
 	w.Name = "shadow-disk-checksum"
 	checksumScript := checksumScriptConst
+	computeServiceAccount := inflater.args.ComputeServiceAccount
+	if computeServiceAccount == "" {
+		computeServiceAccount = "default"
+	}
```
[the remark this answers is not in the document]
dntczdx: added

dntczdx:
For some reason, some comments disappeared. But I addressed all of them.

zoran15 (approved):

cli_tools/common/disk/inspect.go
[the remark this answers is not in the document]
EricEdens: yeah, refactor later is fine with me

EricEdens (approved):

cli_tools_tests/e2e/gce_image_import_export/test_suites/import/cli_tests.go
EricEdens: I'm glad we have the core scenarios. For the error message scenarios, I'm fine classifying those as "nice to have"

PR #1467, opened 2020-12-21
Image import: Take logging.Logger as a dependency

EricEdens:
Prepares for modularizing image import. OVF import allows multiple disks to be imported, so a single logger needs to be shared across all importers.

Key changes:
* Take `logging.Logger` as a parameter; use the logger throughout with all dependencies.
* Added module test that verifies important logs (serial console, inspection results)
* Steps no longer expose their trace logs with the method `TraceLogs`; instead, steps write their trace logs to the injected logger.

Testing:
* Added module tests to verify logs and metrics
* Verified log creation in backend

### DIFF
--- a/cli_tools/common/assert/assert.go
+++ b/cli_tools/common/assert/assert.go
@@ -14,7 +14,11 @@
 
 package assert
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/files"
+)
 
 // GreaterThanOrEqualTo asserts that value is greater than or equal to limit.
 func GreaterThanOrEqualTo(value int, limit int) {
@@ -31,4 +35,11 @@ func Contains(element string, arr []string) {
 		}
 	}
 	panic(fmt.Sprintf("%s is not a member of %v", element, arr))
+}
+
+// DirectoryExists asserts that a directory is on the current filesystem.
+func DirectoryExists(dir string) {
+	if !files.DirectoryExists(dir) {
+		panic(fmt.Sprintf("%s: Directory not found", dir))
+	}
 }
--- a/cli_tools/common/utils/logging/service/outputinfo_loggable_test.go
+++ b/cli_tools/common/utils/logging/service/outputinfo_loggable_test.go
@@ -1,0 +1,114 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package service
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
+	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pb"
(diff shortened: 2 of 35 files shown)


cli_tools/common/assert/assert.go
```diff
@@ -32,3 +35,13 @@ func Contains(element string, arr []string) {
...
+
+// DirectoryExists asserts that a directory is on the current filesystem.
+func DirectoryExists(dir string) {
+	stat, err := os.Stat(dir)
+	if os.IsNotExist(err) {
+		panic(fmt.Sprintf("%s: Directory not found", dir))
+	} else if !os.IsNotExist(err) && !stat.IsDir() {
+		panic(fmt.Sprintf("%s: Not a directory", dir))
+	}
+}
```
zoran15, line 47: Since this doesn't use only input parameters to determine the output, the location doesn't seem appropriate. FileSystemUtils or similar would be a better location, and outside assert package. Otherwise, any function that returns a boolean could end up in this package.

cli_tools/common/utils/logging/service/outputinfo_loggable_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
+	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pb"
+	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pbtesting"
+)
+
+func Test_outputInfoLoggable_GetInspectionResults(t *testing.T) {
```
zoran15, line 28: Capitalize output? Even though it's not exported, I think words after "_" get capitalized.

cli_tools/gce_vm_image_import/cli/cli.go
```diff
@@ -89,6 +92,15 @@ func Main(args []string) error {
...
 
+func validateDirectoryAndMakeAbsolute(workflowRoot string) string {
+	abs, err := filepath.Abs(workflowRoot)
+	if err != nil {
+		panic(err)
+	}
+	assert.DirectoryExists(abs)
+	return abs
+}
+
```
zoran15, line 103: This seems like a generic function. Some comments:
* workflowRoot is too specific from the context of this func, it can be e.g. `directory`
* This would fit in a FileSystemUtils mentioned in a comment regarding assert.DirectoryExists
* This func has two responsibilities, consider splitting validation and conversion to absolute path. Otherwise, consider giving it a more descriptive name, e.g. `validateDirectoryExistsAndConvertPathToAbsolute` (even a bigger reason to split in two funcs)

cli_tools/common/assert/assert.go
EricEdens: A couple of items:
 * In common, all of the functions are precondition assertions that "should" always pass, and if they don't pass, it's a programming or packaging bug.
 * From a reader's point of view, the symbol is: `assert.DirectoryExists`
 * There isn't a return value for this or the other functions

Given this, would you be alright:
1. Keep the package `assert`
2. Move to a different file, eg: `filesystem_assertions.go`

cli_tools/common/utils/logging/service/outputinfo_loggable_test.go
EricEdens: sgtm, done!

cli_tools/gce_vm_image_import/cli/cli.go
EricEdens: Makes sense -- moved to `cli_tools/common/utils/files/files.go`

cli_tools/common/assert/assert.go
zoran15: `DirectoryExists` seems like a generic function that is currently used as an assertion but could be used to control program flow in the future. I think it would be more discoverable in a package that is not focused on assertions.

Maybe have both, one in file utils package that returns a boolean and one in assert that throws panic based on the output from the first?

zoran15 (approved):

dntczdx (approved):
LGTM overall

EricEdens:
Prepares for modularizing image import. OVF import allows multiple disks to be imported, so a single logger needs to be shared across all importers.

PR #1476, opened 2020-12-29
Show a more user friendly error when external IP addresses are prohibited

zoran15:
Show a more user friendly error when external IP addresses are prohibitedited by org policy with a link to image import docs.

Currently, the error shown to users is:

`googleapi: Error 412: Constraint constraints/compute.vmExternalIpAccess violated for project 111111111111. Add instance projects/a-project/zones/europe-west3-c/instances/inst-importer-import-image-87wd3 to the constraint to use external IP with it., conditionNotMet`

This is not useful as the recommendation to add importer VM to an exclusion list won't work because importer VM names change for each import.

After this change, error will be:

`constraint constraints/compute.vmExternalIpAccess violated for project a-project. For more information about importing disks using networks that don't allow external IP addresses, see https://cloud.google.com/compute/docs/import/importing-virtual-disks#no-external-ip`

### DIFF
--- a/cli_tools/gce_vm_image_import/cli/cli.go
+++ b/cli_tools/gce_vm_image_import/cli/cli.go
@@ -94,6 +94,9 @@ func Main(args []string, toolLogger logging.ToolLogger, workflowDir string) erro
 }
 
 func userFriendlyError(err error, importArgs importer.ImportArguments) error {
+	if err == nil {
+		return err
+	}
 	if strings.Contains(err.Error(), "constraints/compute.vmExternalIpAccess") {
 		return fmt.Errorf("constraint constraints/compute.vmExternalIpAccess "+
 			"violated for project %v. For more information about importing disks using "+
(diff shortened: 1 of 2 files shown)


cli_tools/gce_vm_image_import/cli/cli.go
```diff
@@ -92,6 +93,17 @@ func Main(args []string, toolLogger logging.ToolLogger, workflowDir string) erro
 	return nil
 }
 
+func userFriendlyError(err error, importArgs importer.ImportArguments) error {
+	if strings.Contains(err.Error(), "constraints/compute.vmExternalIpAccess") {
```
EricEdens, line 97: probably want a nil check since you'll be dereferencing `err`.

EricEdens (approved):

cli_tools/gce_vm_image_import/cli/cli.go
zoran15: Yep, missed that. Done

PR #1478, opened 2021-01-05
Disk inspection: Honor no-external-ip

EricEdens:
Prior to this, when VMs are disallowed to use external IPs, inspection fails with an error `Constraint constraints/compute.vmExternalIpAccess violated`. If `--os` is supplied, the error is logged to the console. If `--os` is omitted, import will fail.

While fixing this, I fixed the action item from #1465 to consolidate features that we always want to support (no external IP, custom network, custom service account, etc)

## Testing
1. Reproduced issue with Zoran's test account, then confirmed fix
2. Created a module test that reproduced the issue, and confirms the  fix.


cli_tools/daisycommon/daisy_utils.go
```diff
@@ -50,37 +50,53 @@ Loop:
...
+	DisableGCSLogs, DisableCloudLogs, DisableStdoutLogs     bool
+
+	// Worker instance customizations
+	Network, Subnet       string
+	ComputeServiceAccount string
+	NoExternalIP          bool
+}
+
+// SetWorkerCustomizations sets variables that are used when creating worker instances.
+func SetWorkerCustomizations(wf *daisy.Workflow, env EnvironmentSettings) {
```
zoran15, line 69: Can this be a method on EnvironmentSettings?

cli_tools/daisycommon/daisy_utils.go
```diff
@@ -50,37 +50,53 @@ Loop:
...
+	}
 }
 
 // SetWorkflowAttributes sets workflow running attributes.
-func SetWorkflowAttributes(w *daisy.Workflow, attrs WorkflowAttributes) {
-	w.Project = attrs.Project
-	w.Zone = attrs.Zone
-	if attrs.GCSPath != "" {
-		w.GCSPath = attrs.GCSPath
+func SetWorkflowAttributes(w *daisy.Workflow, env EnvironmentSettings) {
```
zoran15, line 78: Same here

cli_tools_tests/infra/README
```diff
@@ -0,0 +1,11 @@
+This directory contains terraform definitions that are used with module and e2e testing.
```
zoran15, line 1: Add some background on Terraform for those who are not aware of what Terraform is.

Also, definitions for what? GCP projects?

cli_tools_tests/infra/README
```diff
@@ -0,0 +1,11 @@
...
+
+One project definition per subdirectory.
```
zoran15, line 3: GCP project?

cli_tools_tests/infra/no-external-ip/README
```diff
@@ -0,0 +1,9 @@
+Creates a GCP project that disables external IPs for VMs. Internet traffic is routed through
+a Cloud NAT router.
+
+Specifically the project has:
+ 1. No default network
+ 2. One network with default firewall
+ 3. One subnet (single region)
```
zoran15, line 7: Private Google access enabled or disabled?

cli_tools_tests/infra/no-external-ip/README
```diff
@@ -0,0 +1,9 @@
+Creates a GCP project that disables external IPs for VMs. Internet traffic is routed through
```
zoran15, line 1: it's not only no-external-ip but also no default network

cli_tools_tests/infra/no-external-ip/main.tf
```diff
@@ -0,0 +1,65 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+variable "wrapper_executors" {
+  description = "Accounts that will execute binaries directly. The cloud build service account is included automatically."
+  default = [
+    "user:ericedens@google.com",
+    "user:zoranl@google.com",
+    "user:tzz@google.com",
```
zoran15, line 20: Having our emails in a public repo is not ideal. It's also likely we will need to store additional sensitive info (like projects names) in the future in these configs.

Can we have Terraform configurations in google3?

EricEdens: Very good point. To unblock this PR, I'll remove the Terraform configurations, and we can investigate long-term solutions.

cli_tools_tests/infra/no-external-ip/README
EricEdens: Updated to clarify that the default network is removed.

cli_tools_tests/infra/no-external-ip/README
EricEdens: Private Google access is unmodified (disabled then?). AFAIK it isn't required since Cloud NAT is allowing access to the public endpoints.

Added text to clarify that PGA is not enabled.

cli_tools_tests/infra/README
EricEdens: done!

cli_tools_tests/infra/README
EricEdens: punting on this since I removed the Terraform configs in this PR

cli_tools/daisycommon/daisy_utils.go
EricEdens: I was worried it'd be ambiguous which struct is being updated, since setters (often) update the receiving object. What do you think?

cli_tools/daisycommon/daisy_utils.go
```diff
@@ -50,37 +50,53 @@ Loop:
 	return w, nil
 }
 
-// WorkflowAttributes holds common attributes that are required to instantiate a daisy workflow.
-type WorkflowAttributes struct {
-	Project, Zone, GCSPath, OAuth, Timeout, ComputeEndpoint, WorkflowDirectory string
-	DisableGCSLogs, DisableCloudLogs, DisableStdoutLogs, NoExternalIP          bool
+// EnvironmentSettings controls the resources that are used during import.
```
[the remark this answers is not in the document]
EricEdens: good catch. Changed to "tool execution"

cli_tools/daisycommon/daisy_utils.go
zoran15: Maybe rename the method at the same time? Like `UpdateWorkflow` or similar?

cli_tools_tests/infra/no-external-ip/main.tf
zoran15: PR description should be updated as well

cli_tools/daisycommon/daisy_utils.go
EricEdens: sgtm. Changed to "Apply" instead of "Set".

cli_tools_tests/infra/no-external-ip/main.tf
EricEdens: updated!

zoran15 (approved):

PR #1484, opened 2021-01-11
Reflect deprecation of SECURE_BOOT

EricEdens:
The SECURE_BOOT flag (when creating a new VM instance) has been deprecated and has no effect.

The only usages were in tests for merging slices.

### DIFF
--- a/cli_tools/common/utils/string/string_utils_test.go
+++ b/cli_tools/common/utils/string/string_utils_test.go
@@ -48,14 +48,14 @@ func TestCombineStringSlices(t *testing.T) {
 			want: []string{"WINDOWS"},
 		},
 		{
-			s1:   []string{"SECURE_BOOT"},
+			s1:   []string{"MULTI_IP_SUBNET"},
 			s2:   []string{"WINDOWS"},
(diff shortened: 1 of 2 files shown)


cli_tools/common/utils/string/string_utils_test.go
```diff
@@ -48,14 +48,14 @@ func TestCombineStringSlices(t *testing.T) {
 			want: []string{"WINDOWS"},
 		},
 		{
-			s1:   []string{"SECURE_BOOT"},
+			s1:   []string{"MULTI_IP_SUBNET"},
```
zoran15, line 51: Is `MULTI_IP_SUBNET` here just to have something in the list so it's not empty?

zoran15 (approved):

cli_tools/common/utils/string/string_utils_test.go
EricEdens: All of the instances of `SECURE_BOOT` were opaque references used in these utility tests. So I swapped them for a feature that wasn't deprecated.

PR #1486, opened 2021-01-11
Updated READMEs for image importer/exporter and OVF importer.

zoran15:
Updated READMEs for image importer/exporter and OVF importer. Updated help text for a couple of args in image importer.


EricEdens (approved):
Thanks for updating these!

PR #1496, opened 2021-01-20
Add disk space warning for Windows imports

bkatyl:
- If imported system volume has less than 1GB disk space free, output warning to TranslateBootstrap.
- Output system volume utilization and amount of free space to detailed log.

### DIFF
--- a/daisy_workflows/image_import/windows/translate_bootstrap.ps1
+++ b/daisy_workflows/image_import/windows/translate_bootstrap.ps1
@@ -148,6 +148,13 @@ try {
   Write-Output "Detected BCD folder drive letter: ${bcd_drive}"
   Write-Output "Detected Windows folder drive letter: ${script:os_drive}"
 
+  # Output system volume utilization and warn if less than 1GB free disk space.
+  $diskInfo = get-WmiObject win32_logicaldisk -Filter "DeviceID='${script:os_drive}'"
+  Write-Output "${script:os_drive} has $([math]::Round($diskInfo.FreeSpace / 1GB,2))GB free and is $([math]::Round(($diskInfo.FreeSpace / $diskInfo.Size)*100,2))% used."
+  if (($diskInfo.FreeSpace / 1GB) -lt 1) {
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/windows/translate_bootstrap.ps1
```diff
@@ -148,6 +148,13 @@ try {
   Write-Output "Detected BCD folder drive letter: ${bcd_drive}"
   Write-Output "Detected Windows folder drive letter: ${script:os_drive}"
 
+  # Output system volume utilization and warn if less than 1GB free disk space.
+  $diskInfo = get-WmiObject win32_logicaldisk -Filter "DeviceID='${script:os_drive}'"
+  Write-Output "${script:os_drive} has $([math]::Round($diskInfo.FreeSpace / 1GB,2))GB free and is $([math]::Round(($diskInfo.FreeSpace / $diskInfo.Size)*100,2))% used."
+  if (($diskInfo.FreeSpace / 1GB) -lt 1) {
+    Write-Output "TranslateBootstrap: Warning imported system volume has less than 1GB free. $([math]::Round($diskInfo.FreeSpace / 1MB,2))MB disk space free. This may cause the import to fail."
```
zoran15, line 155: Since warnings are often buried in logs, if translate fails and if this warning was printed, it would be helpful if transalte failure message included this additional info as a potential reason. 

What is currently printed as translate bootstrap failure reason? If it's too generic, consider saving the fact free space is low in a var and use it at the end of this file when printing failure reason.

This will also help us figure out what percentage of failed Windows imports are due to low free space by looking at our debug logs

zoran15 (approved):

daisy_workflows/image_import/windows/translate_bootstrap.ps1
bkatyl: Thanks for the feedback, I like that approach more. Updating the PR with warning message function and if the script fails emiting the warning messages in the failure. Also adding this to translate.ps1 so both scripts have this capability.

PR #1502, opened 2021-01-27
Fixing powershell syntax issues

bkatyl:
Fixing powershell syntax issues


zoran15 (approved):
Are these syntax issues causing customer facing problems?

bkatyl:
This will resolve the test build failures. Customer using latest would have been impacted.

PR #1504, opened 2021-01-28
OVF export E2E tests

zoran15:
This is the first version of OVF export E2E tests. It includes only one test that does the following:

* Creates a VM from a GMI stored in compute-image-test-pool-001
* Runs OVF export wrapper to export this VM to a GCS bucket
* Verifies the OVF by:
  * Checking that all the expected files are in the GCS path provided
  * Ensuring the instance that was used as a source for OVF export is intact (running and has all the disks attached)
  * Checking that exported OVF can be imported back into GCE and that the imported instance passes OVF import tests.
* Cleans up temporary resources (temporary VM used for export, imported VM, GCS resources)

In the future we want to add verification using VMWare virtualization environments. This is not included in this PR.


cli_tools_tests/e2e/gce_ovf_export/scripts/ovf_import_test_ubuntu_3_disks.sh
```diff
@@ -0,0 +1,235 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests to validate if Ubuntu OVF with 3 disks was imported successfully.
+
+
+# To disable checking for the OS config agent, add an instance metadata
```
EricEdens, line 19: Thoughts on referencing the other script? (Rather than duplicating it)

cli_tools_tests/e2e/gce_ovf_import/scripts/ovf_import_test_ubuntu_3_disks.sh

cli_tools_tests/e2e/gce_ovf_export/test_suites/instance_ovf_export/instance_ovf_export_test_suite.go
```diff
@@ -0,0 +1,386 @@
...
+	logger.Printf("Exported instance verified to be in the original sate...")
+	return true
+}
+
+func verifyExportedOVFContent(ctx context.Context, testCase *junitxml.TestCase,
+	logger *log.Logger, props *instanceOvfExportTestProperties) (bool, []*gcp.File) {
+
+	splits := strings.Split(props.sourceGMI, "/")
+	sourceGMIName := splits[len(splits)-1]
+	sourceGMIName = "machine-image-import-s2mtj"
```
EricEdens, line 221: I'm guessing the second `sourceGMIName` was from a test run and can be removed?

cli_tools_tests/e2e/gce_ovf_export/test_suites/instance_ovf_export/instance_ovf_export_test_suite.go
```diff
@@ -0,0 +1,386 @@
...
+	filePaths := []string{
+		fmt.Sprintf("%v/%v-%v.vmdk", props.exportPath, props.instanceName, sourceGMIName),
```
EricEdens, line 223: AFAICT, the name of the VMDK is derived from the name of the disk? Is that's the case, you could verify the names within the loop `for _, disk := range verificationInstance.Disks {` . That would help the reader understand how the VMDK's name is generated.

cli_tools_tests/e2e/gce_ovf_export/test_suites/instance_ovf_export/instance_ovf_export_test_suite.go
zoran15: This was hardcoded based on the GMI used (which changed due to it being deleted by cleanerupper). 

I've refactored this to use instance disk device names, since that is used by the exported to generate file names.

cli_tools_tests/e2e/gce_ovf_export/test_suites/instance_ovf_export/instance_ovf_export_test_suite.go
zoran15: Already mentioned above: I've refactored this to use instance disk device names, since that is used by the exported to generate file names.

cli_tools_tests/e2e/gce_ovf_export/scripts/ovf_import_test_ubuntu_3_disks.sh
zoran15: changed

EricEdens (approved):

PR #1505, opened 2021-01-28
VPC image format E2E test

Quintonamore:
Added a vpc test to both daisy and image import e2e tests. I decided to keep the azure vhd test since it is still testing an import of an azure exported vhd file.

@pjh


zoran15 (approved):
VPC can refer to https://cloud.google.com/vpc so it's a good idea to have a more descriptive test name as well as PR title to avoid any confusion. This is because we do have logic to handle shared VPCs in import/export.

EricEdens (approved):

PR #1506, opened 2021-01-29
CLI logging: Trap panics in application code

EricEdens:
`runWithServerLogging` executes a function, then logs the results. Prior to this, a panic would cause the application to crash. Following this PR, `runWithServerLogging` traps the panic and sends a "failure" message using the panic's details and a stack trace.

Follow-up from https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1497#discussion_r562353775

### DIFF
--- a/cli_tools/common/utils/logging/service/logger.go
+++ b/cli_tools/common/utils/logging/service/logger.go
@@ -232,10 +232,7 @@ func (l *Logger) runWithServerLogging(function func() (Loggable, error),
 		l.logStart()
 	}()
 
-	// Execute the closure and ensure that panic won't crash the program.
-	r := &panicSafeRunner{inner: function}
-	r.runInner()
-	loggable, err := r.loggable, r.err
+	loggable, err := runWithRecovery(function)
 	l.updateParams(projectPointer)
 	if err != nil {
 		wg.Add(1)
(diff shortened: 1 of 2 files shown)


cli_tools/common/utils/logging/service/logger.go
```diff
@@ -381,6 +386,34 @@ func (l *Logger) constructLogRequest(logExtension *ComputeImageToolsLogExtension
 	return reqStr, err
 }
 
+// safeRunner supports running a function that may panic. If there's a panic, execution is
+// recovered and the panic's details are captured.
+type safeRunner struct {
```
zoran15, line 391: panicSafeRunner? this would make it clear what the runner is making the inner safe from.

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -381,6 +386,34 @@ func (l *Logger) constructLogRequest(logExtension *ComputeImageToolsLogExtension
...
+		if err := recover(); err != nil {
+			log.Printf("Fatal error: %v", err)
+			p.loggable = literalLoggable{
+				traceLogs: []string{
+					fmt.Sprintf("Captured panic: %v", err),
+					"stacktrace from panic: \n" + string(debug.Stack()),
+				},
+			}
+			p.err = errors.New("A fatal error has occurred. " +
+				"Please submit an issue at https://github.com/GoogleCloudPlatform/compute-image-tools/issues")
```
zoran15, line 411: Does this overwrite what p.inner() returned as an error? Does that one get lost? Or is that one nil in case of panic?

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -229,7 +232,9 @@ func (l *Logger) runWithServerLogging(function func() (Loggable, error),
 		l.logStart()
 	}()
 
-	loggable, err := function()
+	capture := &safeRunner{inner: function}
```
zoran15, line 235: why is safeRunner variable called capture? Seems unintuitive.

EricEdens: Changed to `r` ("capture" was intended to signify the captured loggable and error)

cli_tools/common/utils/logging/service/logger.go
EricEdens: sgtm, done

cli_tools/common/utils/logging/service/logger.go
EricEdens: From my reading of [1]:
1. If inner panics, it doesn't return, so `p.err` won't be written at 414.
 1. If inner didn't panic, then `recover()` returns nil, so this block won't get executed.

zoran15 (approved):

cli_tools/common/utils/logging/service/logger.go
```diff
@@ -229,7 +232,10 @@ func (l *Logger) runWithServerLogging(function func() (Loggable, error),
 		l.logStart()
 	}()
 
-	loggable, err := function()
+	// Execute the closure and ensure that panic won't crash the program.
+	r := &panicSafeRunner{inner: function}
+	r.runInner()
+	loggable, err := r.loggable, r.err
```
dntczdx, line 238: why not just return loggable & err from runInner()?

EricEdens: I didn't think it would be possible, but I'm glad you suggested it! Updated to remove the intermediate struct.

dntczdx (approved):

PR #1510, opened 2021-02-01
VHD/VPC file formats shadow test gate

Quintonamore:
Added a gate to only shadow test files that match the VPC file format. Passing the inspect image metadata in case we start shadow testing more than one file format.

@pjh


cli_tools/common/image/importer/inflater.go
```diff
@@ -60,6 +61,8 @@ const (
 	sigShadowInflaterErr  = "shadow err"
 )
 
+var allowedAPIFormatsEx = regexp.MustCompile("^(vpc)$")
```
EricEdens, line 64: Interesting, I would have expected to see `vmdk` include here as well.

cli_tools/common/image/importer/inflater.go
```diff
@@ -230,7 +233,18 @@ func newInflater(request ImageImportRequest, computeClient daisyCompute.Client,
 		return di, nil
 	}
 
-	ai := createAPIInflater(request, computeClient, storageClient, logger)
+	deadline, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(inspectionTimeout))
```
EricEdens, line 236: This will cause inspection to run twice -- once here, and once in the daisy inflater. Can you run it once, and share the results?

cli_tools/common/image/importer/inflater.go
EricEdens, line 64: Interesting, I would have expected to see vmdk include here as well.

Quintonamore: I hadn't thought of that and may have had my blinders on here, @dntczdx should we still be testing vmdk formats as well?

dntczdx: We still need to test VMDK - it's not even in GA yet.

EricEdens (approved):

EricEdens:
Overall looks good to me! Let's sync with Tony this afternoon regarding rollout.

cli_tools/common/image/importer/api_inflater_test.go
```diff
@@ -151,7 +152,7 @@ func TestAPIInflater_getCalculateChecksumWorkflow(t *testing.T) {
 		ExecutionID:  "1234",
 		NoExternalIP: false,
 		WorkflowDir:  daisyWorkflows,
-	}, nil, &storage.Client{}, logging.NewToolLogger(t.Name()), true)
+	}, nil, &storage.Client{}, logging.NewToolLogger(t.Name()), imagefile.Metadata{}, true)
```
dntczdx, line 155: in my opinion there should be one testcase to validate whether the "sourceType" is populated correctly

cli_tools/common/image/importer/inflater.go
```diff
@@ -23,6 +25,8 @@ import (
 	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pb"
 )
 
+var allowedAPIFormatsEx = regexp.MustCompile("^(vpc)$")
```
dntczdx, line 28: both vpc and vhd? I am not sure whether vhd is also a possible output, maybe not.

Quintonamore: I tested the valid outputs using qemu, and it classifies vhd/vpc strictly to `vpc`.

cli_tools/common/image/importer/inflater.go
dntczdx: It's a bit tricky - once we switched off shadow test, we want to go back to the current logic - only run inspection for daisy inflater. Plus, I guess we will need the shadow test every time when PD team release a new format. So let's not simply remove the inspection logic from daisy inflater; instead, use a condition to determine whether to run it.

dntczdx (approved):

cli_tools/common/image/importer/api_inflater_test.go
Quintonamore: I wasn't able to mock the storage client appropriately to test the PD fields, but instead I added a check in `inflater_test.go` to make sure the metadata is being passed correctly.

PR #1512, opened 2021-02-02
Machine image import: Use import module for alpha and beta

EricEdens:
This is a follow-up to #1497 where we refactored instance import to use import modules.

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_importer/ovf_importer_test.go
+++ b/cli_tools/gce_ovf_import/ovf_importer/ovf_importer_test.go
@@ -215,18 +215,23 @@ func TestSetUpMachineImageWorkflowHappyPathFromOVANoExtraFlags(t *testing.T) {
 	assert.Equal(t, "us-west2", machineImage.StorageLocations[0])
 }
 
-type ModuleImportTestCase struct {
-	descriptorFilenames []string
-	fileURIs            []string
-	imageURIs           []string
-	diskURIs            []string
-}
-
 func Test_InstanceImport_SetupWorkflow_HappyCase_PreGA(t *testing.T) {
 	wfPath := "../../../daisy_workflows/" + createInstanceWorkflow
-	track := domain.Beta
-	project := defaultProject
-	noExternalIP := false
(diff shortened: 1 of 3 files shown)


cli_tools/gce_ovf_import/ovf_importer/ovf_importer_test.go
```diff
@@ -288,11 +260,22 @@ func Test_InstanceImport_SetupWorkflow_HappyCase_PreGA(t *testing.T) {
 	assert.Equal(t, params.Oauth, w.OAuthPath)
 	assert.Equal(t, params.Ce, w.ComputeEndpoint)
 	assert.Equal(t, params.ScratchBucketGcsPath, w.GCSPath)
-	assert.Len(t, w.Steps, 2, "Expect two steps: create-instance and cleanup")
+	if createMachineImage {
+		// Creating the machine image adds two steps:
+		//  1. Stop the instance
+		//  2. Create the GMI
```
zoran15, line 266: It also deletes the instance.

EricEdens: Good point! Updated the comment, and added an assertion that the cleanup step includes the instance when a GMI is being created.

zoran15 (approved):

daisy_workflows/ovf_import/create_gmi.wf.json
```diff
@@ -0,0 +1,98 @@
+{
+  "Name": "create-instance",
+  "DefaultTimeout": "240m",
```
dntczdx, line 3: Hmm I don't remember what's the original number, but I feel I don't see 240m anywhere before. Is it align with the original design?

EricEdens: Good question -- it's the value that's used in the current OVF workflows, eg: 

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/22dbfbdebb2353295698e93c2089ded45c56707b/daisy_workflows/ovf_import/import_ovf_to_instance.wf.json#L3

PR #1513, opened 2021-02-04
[vmdk api adoption] siwtch to GA shadow test

dntczdx:
This test will be ran in order to get at least 50 calls.


cli_tools/import_precheck/check_sha2.go
```diff
@@ -36,7 +37,8 @@ func (s *sha2DriverSigningCheck) run() (*report, error) {
 		return r, nil
 	}
 
-	pkgs, err := packages.GetInstalledPackages()
```
dntczdx, line 39: The new OSConfig packages changed the interface so I have to add this line, otherwise it can't pass the build

zoran15 (approved):

PR #1514, opened 2021-02-05
[vmdk api adoption] Switch from shadow-test mode to fallback mode

dntczdx:
This PR will start using API inflater as the main inflater.
Shadow test mode: daisy inflater as main inflater; API inflater as shadow test.
Fallback mode: api inflater as main inflater; daisy inflater as fallback inflater (expect to be used by non-vmdk).

Changes:
1. api_inflater: use a flag "isShadowTest" to determine whether to run shadow-test-only logic (calculate checksum for example).
2. Split "inflater.go" to "daisy_inflater.go" and "inflater.go"; "inflater_test.go" is split accordingly.
3. Renamed original "inflaterFacade" to "shadowTestInflaterFacade", which handles shadow-test mode; added a new "inflaterFacade" to handle fallback mode. Added tests accordingly.

New logic is only in api_inflater.go and inflater.go. Others are only code movement or trivial changes.

### DIFF
--- a/cli_tools/common/image/importer/api_inflater.go
+++ b/cli_tools/common/image/importer/api_inflater.go
@@ -171,6 +171,8 @@ func (inflater *apiInflater) getShadowDiskName() string {
 
 func (inflater *apiInflater) Cancel(reason string) bool {
 	if !inflater.isShadowInflater {
+		// We don't have to do any actual cancellation for the single CreateDisk API call.
+		// Only the daisy workflow is worth cancelling.
 		return false
 	}
 
--- a/cli_tools/common/image/importer/api_inflater.go
+++ b/cli_tools/common/image/importer/api_inflater.go
@@ -47,22 +47,24 @@ func isCausedByAlphaAPIAccess(err error) bool {
 
 // apiInflater implements `importer.inflater` using the Compute Engine API
 type apiInflater struct {
-	request         ImageImportRequest
-	computeClient   daisyCompute.Client
-	storageClient   domain.StorageClientInterface
-	guestOsFeatures []*compute.GuestOsFeature
-	wg              sync.WaitGroup
-	cancelChan      chan string
-	logger          logging.Logger
+	request          ImageImportRequest
+	computeClient    daisyCompute.Client
+	storageClient    domain.StorageClientInterface
+	guestOsFeatures  []*compute.GuestOsFeature
+	wg               sync.WaitGroup
+	cancelChan       chan string
+	logger           logging.Logger
+	isShadowInflater bool
 }
 
-func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Client, storageClient domain.StorageClientInterface, logger logging.Logger) Inflater {
+func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Client, storageClient domain.StorageClientInterface, logger logging.Logger, isShadowInflater bool) *apiInflater {
 	inflater := apiInflater{
-		request:       request,
-		computeClient: computeClient,
-		storageClient: storageClient,
-		cancelChan:    make(chan string, 1),
-		logger:        logger,
+		request:          request,
+		computeClient:    computeClient,
+		storageClient:    storageClient,
+		cancelChan:       make(chan string, 1),
+		logger:           logger,
+		isShadowInflater: isShadowInflater,
 	}
 	if request.UefiCompatible {
 		inflater.guestOsFeatures = []*compute.GuestOsFeature{{Type: "UEFI_COMPATIBLE"}}
@@ -71,6 +73,23 @@ func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Cl
 }
 
 func (inflater *apiInflater) Inflate() (persistentDisk, shadowTestFields, error) {
+	if inflater.isShadowInflater {
+		return inflater.inflateForShadowTest()
+	}
+
+	ctx := context.Background()
+	startTime := time.Now()
+	diskName := "disk-" + inflater.request.ExecutionID
+
+	cd, err := inflater.createDisk(diskName)
+	if err != nil {
+		return persistentDisk{}, shadowTestFields{}, daisy.Errf("Failed to create disk by api inflater: %v", err)
+	}
+
+	return inflater.getDiskAttributes(ctx, diskName, cd, startTime)
+}
+
+func (inflater *apiInflater) inflateForShadowTest() (persistentDisk, shadowTestFields, error) {
 	inflater.wg.Add(1)
 	defer inflater.wg.Done()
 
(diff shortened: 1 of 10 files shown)


cli_tools/common/image/importer/inflater.go
```diff
@@ -13,40 +13,125 @@
...
+
+	facade.logger.Metric(&pb.OutputInfo{
+		InflationType:   "qemu_failed",
+		InflationTimeMs: []int64{tf.inflationTime.Milliseconds()},
+	})
+	return pd, tf, err
+}
+
+func (facade *inflaterFacade) Cancel(reason string) bool {
+	// No need to cancel apiInflater.
```
Quintonamore, line 129: Why do we explicitly have to cancel Daisy and not the API Inflater? Is this something unique to Daisy?

cli_tools/common/image/importer/inflater.go
```diff
@@ -13,40 +13,125 @@
...
 	"github.com/GoogleCloudPlatform/compute-image-tools/proto/go/pb"
 )
 
-const (
-	inflateFilePath  = "image_import/inflate_file.wf.json"
-	inflateImagePath = "image_import/inflate_image.wf.json"
+// Inflater constructs a new persistentDisk, typically starting from a
+// frozen representation of a disk, such as a VMDK file or a GCP disk image.
+//
+// Implementers can expose detailed logs using the traceLogs() method.
```
EricEdens, line 29:
```suggestion
// Inflater constructs a new persistentDisk, typically starting from a
// frozen representation of a disk, such as a VMDK file or a GCP disk image.
```

Stale comment now that `logging.Logger` is available :)

cli_tools/common/image/importer/api_inflater.go
```diff
@@ -47,22 +47,24 @@ func isCausedByAlphaAPIAccess(err error) bool {
...
-		computeClient: computeClient,
-		storageClient: storageClient,
-		cancelChan:    make(chan string, 1),
-		logger:        logger,
+		request:          request,
+		computeClient:    computeClient,
+		storageClient:    storageClient,
+		cancelChan:       make(chan string, 1),
+		logger:           logger,
+		isShadowInflater: isShadowInflater,
```
EricEdens, line 67: `calculateChecksum`?  (AFAICT that's the only difference between the two options)

cli_tools/common/image/importer/api_inflater.go
```diff
@@ -71,21 +73,31 @@ func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Cl
...
+
+	cd, err := inflater.createDisk(diskName)
+	if err != nil {
+		return persistentDisk{}, shadowTestFields{}, daisy.Errf("Failed to create disk by api inflater: %v", err)
+	}
+
+	return inflater.getDiskAttributes(ctx, diskName, cd, startTime)
+}
+
+func (inflater *apiInflater) inflateForShadowTest() (persistentDisk, shadowTestFields, error) {
```
EricEdens, line 92: `inflateAndCalculateChecksum()` ?

cli_tools/common/image/importer/api_inflater.go
```diff
@@ -125,20 +162,18 @@ func (inflater *apiInflater) Inflate() (persistentDisk, shadowTestFields, error)
...
-	return pd, ii, err
+	return pd, ii, nil
 }
 
 func (inflater *apiInflater) getShadowDiskName() string {
 	return fmt.Sprintf("shadow-disk-%v", inflater.request.ExecutionID)
 }
 
 func (inflater *apiInflater) Cancel(reason string) bool {
+	if !inflater.isShadowInflater {
```
EricEdens, line 173: Can you add a comment why this is always false?

cli_tools/common/image/importer/daisy_inflater_test.go
```diff
@@ -0,0 +1,331 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
```
EricEdens, line 1:
```suggestion
//  Copyright 2021 Google Inc. All Rights Reserved.
```

EricEdens (commented):
Looks great! A few small comments

cli_tools/common/image/importer/api_inflater.go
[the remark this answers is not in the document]
dntczdx: No - the new format supporting from PD team will be rolled out to GA directly, so it's not useful to call beta API.

cli_tools/common/image/importer/inflater.go
dntczdx: The daisy inflater is a workflow and it can be interrupted from the middle. The PD API is one single API call and is pretty fast.  We can cancel it (it's supported), but it's not very valuable to support that - it brings extra complexity while its benefit is small.

Quintonamore (approved):

cli_tools/common/image/importer/inflater.go
dntczdx: done

cli_tools/common/image/importer/api_inflater.go
dntczdx: calculateChecksum is the internal logic which doesn't have to be exposed to the callers. There will be different behaviors based on whwther it's a shadow inflater. So it's a better abstraction to me.
Looking into the details, there are also some other differences - signals and err msgs.

cli_tools/common/image/importer/api_inflater.go
dntczdx: inflateForShadowTest is a better summary to me, since just as mentiuoned in the above comment, there are some other details different: err msgs, signals. "inflateForShaodeTest" summarixzed the purpose of this function, while "calculateChecksum" is a part of the implementation.

cli_tools/common/image/importer/api_inflater.go
dntczdx: done

cli_tools/common/image/importer/api_inflater.go
EricEdens: makes sense

EricEdens (approved):

cli_tools/common/image/importer/api_inflater.go
```diff
@@ -101,6 +113,31 @@ func (inflater *apiInflater) Inflate() (persistentDisk, shadowTestFields, error)
...
+	}
+	return pd, ii, err
+}
+
+func (inflater *apiInflater) createDisk(diskName string) (compute.Disk, error) {
+	cd := compute.Disk{
+		Name:                diskName,
+		SourceStorageObject: inflater.request.Source.Path(),
+		GuestOsFeatures:     inflater.guestOsFeatures,
+	}
```
zoran15, line 135: This is maybe a nit, but input args for this function are mix and match of object attributes (source path, guest OS attributes) and function args (diskName). It would be cleaner if one or the other approach was used consistently. E.g. if inflater fields are used as input, then diskName can become a field on the inflater. Or, source path and guestOsFeatures are passed as args to the func.

There are other functions that do mix and match of input args like this, even though args are for the same object (disk).

dntczdx: I agree that it's overall a correct thing to use the consistent way to get args. For this specific case, there can be multiple disks created, so diskName became a param.

zoran15 (approved):

PR #1521, opened 2021-02-11
Image import: Increase debug support when translation fails

EricEdens:
Recently we had test failures that said "Unable to correct problems, you have held broken packages", but did not include any useful debug information. This occurred since guestfs's `g.sh` and `g.command` hides stdout when a command fails.

https://prow.k8s.io/view/gcs/compute-image-tools-test/logs/ci-images-import-export-cli-e2e-tests/1359454353202614272
This PR does the following:
1. Replaces usages of  `g.sh` and `g.command` with `guestfsprocess.run` throughout import
2. Adds an optional parameter `check` to `guestfsprocess.run`. It  follows the semantics of [subprocess.run](https://docs.python.org/3/library/subprocess.html#subprocess.run), which this module is emulating.

## Testing
- Ran e2e tests for image import
- Ran image builds for workflows that were modified by this PR

### DIFF
--- a/daisy_workflows/image_import/debian/translate.py
+++ b/daisy_workflows/image_import/debian/translate.py
@@ -25,6 +25,7 @@ import logging
 
 import utils
 import utils.diskutils as diskutils
+from utils.guestfsprocess import run
 
 google_cloud = '''
 deb http://packages.cloud.google.com/apt cloud-sdk-{deb_release} main
@@ -51,17 +52,17 @@ def DistroSpecific(g):
     utils.update_apt(g)
     utils.install_apt_packages(g, 'gnupg')
 
-    g.command(
+    run(g,
         ['wget', 'https://packages.cloud.google.com/apt/doc/apt-key.gpg',
         '-O', '/tmp/gce_key'])
-    g.command(['apt-key', 'add', '/tmp/gce_key'])
+    run(g, ['apt-key', 'add', '/tmp/gce_key'])
     g.rm('/tmp/gce_key')
     g.write(
         '/etc/apt/sources.list.d/google-cloud.list',
         google_cloud.format(deb_release=deb_release))
     # Remove Azure agent.
     try:
-      g.command(['apt-get', 'remove', '-y', '-f', 'waagent', 'walinuxagent'])
+      run(g, ['apt-get', 'remove', '-y', '-f', 'waagent', 'walinuxagent'])
     except Exception as e:
       logging.debug(str(e))
       logging.warn('Could not uninstall Azure agent. Continuing anyway.')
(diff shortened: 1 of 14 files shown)


daisy_workflows/linux_common/utils/guestfsprocess.py
```diff
@@ -23,26 +23,31 @@
...
+def run(g: 'GuestFSInterface', command,
+        check: bool = True) -> 'CompletedProcess':
   """Runs a process in a mounted GuestFS instance, ensuring that
   standard output and standard error is always retained.
 
   Args:
     g: Mounted GuestFS instance.
     command (str or List[str]): Script content that will be executed
     by a bash interpeter on the guest.
+    check (bool): When true and the process exits with a non-zero exit code,
```
zoran15, line 42: The name of this flag doesn't seem appropriate. It sounds like it's an instruction to check something but it's actually printing more debugging info and throwing an exception.

daisy_workflows/image_import/debian/translate.py
```diff
@@ -51,17 +52,17 @@ def DistroSpecific(g):
     utils.update_apt(g)
     utils.install_apt_packages(g, 'gnupg')
 
-    g.command(
+    run(g,
```
zoran15, line 55: Which run invocations should have check=True and which should have it False? It's not clear how this choice is made now or for future run invocations we might add.

daisy_workflows/linux_common/utils/guestfsprocess.py
EricEdens: I've modeled this after the standard library so that it'd be familiar to future programmers: https://docs.python.org/3/library/subprocess.html#subprocess.run

That said, any thoughts on something that would be clearer?

daisy_workflows/image_import/debian/translate.py
EricEdens: `check=True` is good for commands where you don't inspect the output and you want the failure to propagate automatically. For example: `yum install package`; if it's successful, you don't need the output from yum.

daisy_workflows/linux_common/utils/guestfsprocess.py
zoran15: I saw the reference to subprocess.run, but I still find it unintuitive. What does check mean there?

EricEdens: My guess is: "check the result". I could prepend this to the docstring, something like: "Whether to check the result of the command. When true..."

zoran15: Sounds to me more like `raiseOnError` than `check`

EricEdens:
sgtm, I'll change to that

daisy_workflows/linux_common/utils/guestfsprocess.py
EricEdens: sgtm -- changed to `raiseOnError`

daisy_workflows/image_import/debian/translate.py
zoran15: This would be useful if documented in the function description so developers working on this in the future know how to use it.

EricEdens: sgtm - done!

zoran15 (approved):

EricEdens:
Recently we had test failures that said "Unable to correct problems, you have held broken packages", but did not include any useful debug information. This occurred since guestfs's g.sh and g.command hides stdout when a command fails.

https://prow.k8s.io/view/gcs/compute-image-tools-test/logs/ci-images-import-export-cli-e2e-tests/1359454353202614272

This PR does the following:
1. Replaces usages of g.sh and g.command with guestfsprocess.run throughout import
2. Adds an optional parameter check to guestfsprocess.run. It follows the semantics of subprocess.run, which this module is emulating.

PR #1534, opened 2021-02-19
add export metadata for windows

gaohannk:
This workflow is compare to linux export-metadata. 

Create a windows instance:
Using startup script run export-metadata powershell script
export image/package information as a json file and upload to GCS

Test: 
```daisy -gcs_path=gs://hanga-testing-daisy-bkt/daisy/hanga -project=hanga-testing \
-var:metadata_dest=gs://metadata-test-hanga \
-var:image_id=test-id \
-var:image_name=test-name \
-var:image_family=test-family \
-var:source_image="projects/windows-cloud/global/images/windows-server-1909-dc-core-v20210212" \
-var:uefi=true \
${HOME}/IdeaProjects/compute-image-tools/daisy_workflows/export_metadata/export_metadata_windows.wf.json
```

daisy log
https://storage.cloud.google.com/hanga-testing-daisy-bkt/daisy/hanga/daisy-export-metadata-20210223-20%3A17%3A52-3mxzr/logs/inst-exporter-export-metadata-3mxzr-serial-port1.log

metadata file

https://storage.cloud.google.com/metadata-test-hanga/metadata.json

### DIFF
--- a/daisy_workflows/export_metadata/export-metadata.ps1
+++ b/daisy_workflows/export_metadata/export-metadata.ps1
@@ -46,16 +46,21 @@ function Export-ImageMetadata {
     [array]::sort($out)
 
     foreach ($package_line in $out) {
-        $name = $package_line.Trim().Split(' ')[0]
-        # Get Package Info for each package
-        $info = & 'C:\ProgramData\GooGet\googet.exe' -root 'C:\ProgramData\GooGet' 'installed' '-info' $name
-        $version = $info[4].Split(":")[1].Trim()
-        $source = $info[8].Split(":")
-        $source = [String]::Concat($source[1..$source.length])
-        $package_metadata = @{'name' = $name;
-                            'version' = $version;
-                            'commmit_hash' = $source}
-        $image_metadata['packages'] += $package_metadata
+        try {
+            $name = $package_line.Trim().Split(' ')[0]
+            # Get Package Info for each package
+            $info = & 'C:\ProgramData\GooGet\googet.exe' -root 'C:\ProgramData\GooGet' 'installed' '-info' $name
+            $version = $info[4].Split(":")[1].Trim()
+            $source = $info[8].Split(":")
+            $source = [String]::Concat($source[1..$source.length])
+            $package_metadata = @{'name' = $name;
+                                'version' = $version;
+                                'commmit_hash' = $source}
+            $image_metadata['packages'] += $package_metadata
+        }
+        catch {
+            Write-Host "Failed to retrieve metadata for $package_line, skiping."
+        }
     }
 
     # Save the JSON image_metadata.
--- a/daisy_workflows/export_metadata/export-metadata.ps1
+++ b/daisy_workflows/export_metadata/export-metadata.ps1
@@ -1,0 +1,100 @@
+function Get-MetadataValue {
+    <#
+    .SYNOPSIS
+      Returns a value for a given metadata key.
+
+    .PARAMETER $key
+      The metadata key to retrieve.
+
+    .PARAMETER $default
+      The value to return if the key is not found.
+
+    .RETURNS
+      The value for the key or null.
+  #>
+    param (
+        [parameter(Mandatory=$true)]
+        [string]$key,
+        [parameter(Mandatory=$false)]
+        [string]$default
+    )
+
+    # Returns the provided metadata value for a given key.
+    $url = "http://metadata.google.internal/computeMetadata/v1/instance/attributes/$key"
+    try {
+        $client = New-Object Net.WebClient
+        $client.Headers.Add('Metadata-Flavor', 'Google')
+        return ($client.DownloadString($url)).Trim()
+    }
+    catch [System.Net.WebException] {
+        if ($default) {
+            return $default
+        }
+        else {
+            Write-Host "Failed to retrieve value for $key."
+            return $null
+        }
+    }
+}
+
+function Run-Command {
+    [CmdletBinding(SupportsShouldProcess=$true)]
+    param (
+        [Parameter(Mandatory=$true, ValueFromPipelineByPropertyName=$true)]
+        [string]$Executable,
+        [Parameter(ValueFromRemainingArguments=$true,
+                ValueFromPipelineByPropertyName=$true)]
+        $Arguments = $null
+    )
+    Write-Host "Running $Executable with arguments $Arguments."
+    $out = &$executable $arguments 2>&1 | Out-String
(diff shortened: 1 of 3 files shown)


daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,87 @@
...
+        $image_metadata['packages'] += $package_metadata
+    }
+
+    # Save the JSON image_metadata.
+    $image_metadata_json = $image_metadata | ConvertTo-Json -Compress
+    $image_metadata_json | & 'gsutil' cp - "${metadata_dest}/metadata.json"
+}
+
+try {
+    Write-Host 'Beginning export windows package metadata'
```
bkatyl, line 74: Since there is a beginning message, I'd also suggest an ending message.

daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,87 @@
...
+    foreach ($package_line in $out) {
+        $name = $package_line.Trim().Split(' ')[0]
+        # Get Package Info for each package
+        $info = & 'C:\ProgramData\GooGet\googet.exe' -root 'C:\ProgramData\GooGet' 'installed' '-info' $name
+        $version = $info[4].Split(":")[1].Trim()
+        $source = $info[8].Split(":")
+        $source = [String]::Concat($source[1..$source.length])
+        $package_metadata = @{'name' = $name;
+                            'version' = $version;
+                            'commmit_hash' = $source}
```
bkatyl, line 64: source seems like a better name for this.

daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,87 @@
...
+        else {
+            Write-Host "Failed to retrieve value for $key."
+            return $null
+        }
+    }
+}
+
+function Export-ImageMetadata {
+    $image_version = Get-Date -Format "yyyyMMdd"
+    $build_date = Get-Date -Format "o"
```
bkatyl, line 42: $image_version feels like it should be a value that is passed into the workflow as it's contained in the disk metadata.

Instead of $build_date, I'd suggest this should be metadata_date as this isn't when the image is built it's when the image is having metadata exported.

gaohannk: since the metadata is help for us auditing all the publish image, the $image_version we want here is more about the yyyyMMdd suffix we used in tar.gz and image name instead the real image version from OS itself (we are less care about). for build_date, we have a borg job running 5 mins periodically to monitor any new published image and we just assume that we always publish immediately after build.

daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,87 @@
...
+                        'build_date' = $build_date;
+                        'packages' = @()}
+
+    # Get Googet packages.
+    $out = & 'C:\ProgramData\GooGet\googet.exe' -root 'C:\ProgramData\GooGet' 'installed'
+    $out = $out[1..$out.length]
+    [array]::sort($out)
+
+    foreach ($package_line in $out) {
+        $name = $package_line.Trim().Split(' ')[0]
```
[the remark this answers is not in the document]
gaohannk: sure, add try/catch is better. we don't want to fail for a single package in the loop

daisy_workflows/export_metadata/export-metadata.ps1
bkatyl: The Windows builds have a 1-2 day lag between build and publish. So if they're build on 20210222, they'll be published on 20210224 if all goes well. From your intent, it sounds like we'd be better off passing in the value to ensure it's accurate instead of derive it from the current datetime. 

This would especially be true for the derived builds as they might be build the day after the OS builds, but they'd still have our build date of v20210222.

daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,86 @@
...
+    $out = & 'C:\ProgramData\GooGet\googet.exe' -root 'C:\ProgramData\GooGet' 'installed'
+    $out = $out[1..$out.length]
+    [array]::sort($out)
+
+    foreach ($package_line in $out) {
+        try {
+            $name = $package_line.Trim().Split(' ')[0]
+            # Get Package Info for each package
+            $info = & 'C:\ProgramData\GooGet\googet.exe' -root 'C:\ProgramData\GooGet' 'installed' '-info' $name
+            $version = $info[4].Split(":")[1].Trim()
```
adjackura, line 53: Just indexing the output by lines can easily break if the output is altered, instead you should proabbly scan the output line by line looking for the correct line based on prefix.

daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,86 @@
...
+}
+
+try {
+    Write-Host 'Beginning export windows package metadata'
+    $metadata_dest = Get-MetadataValue -key 'metadata_dest'
+    $image_id = Get-MetadataValue -key 'image_id'
+    $image_name = Get-MetadataValue -key 'image_name'
+    $image_family = Get-MetadataValue -key 'image_family'
+    Export-ImageMetadata
+    Write-Host 'Endding export windows package metadata'
```
adjackura, line 78: Finished exporting windows image metadata

daisy_workflows/export_metadata/export-metadata.ps1
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    # Save the JSON image_metadata.
+    $image_metadata_json = $image_metadata | ConvertTo-Json -Compress
+    $image_metadata_json | & 'gsutil' cp - "${metadata_dest}/metadata.json"
+}
+
+try {
+    Write-Host 'Beginning export windows package metadata'
```
adjackura, line 72: image metadata

gaohannk: Fix.

daisy_workflows/export_metadata/export-metadata.ps1
gaohannk: Fix

daisy_workflows/export_metadata/export-metadata.ps1
gaohannk: Great catch. fix.

bkatyl (approved):

PR #1543, opened 2021-02-24
Added custom Compute engine service account to OVF import

zoran15:
Added custom Compute engine service account to OVF (instance and machine image) import. Added E2E tests to cover different scenarios around Compute engine service accounts (disabled/lacking permissions for default service account)


dntczdx (approved):
LGTM. It's mostly the same as what we did for  'image import'

cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -127,6 +127,9 @@ func (oi *OVFImporter) buildDaisyVars(bootDiskImageURI string, machineType strin
 	varMap["boot_disk_image_uri"] = bootDiskImageURI
 	if oi.params.IsInstanceImport() {
 		varMap["instance_name"] = oi.params.InstanceNames
+		if oi.params.ComputeServiceAccount != "" {
+			varMap["compute_service_account"] = oi.params.ComputeServiceAccount
+		}
```
EricEdens, line 132: It think you'll want the var for GMI import as well:

daisy_workflows/ovf_import/create_gmi.wf.json

(Unfortunately there's duplication between the two :/ )

cli_tools/gce_ovf_import/main.go
```diff
@@ -63,6 +63,7 @@ var (
 	scratchBucketGcsPath        = flag.String("scratch-bucket-gcs-path", "", "GCS scratch bucket to use, overrides what is set in workflow")
 	oauth                       = flag.String("oauth", "", "path to oauth json file, overrides what is set in workflow")
 	ce                          = flag.String("compute-endpoint-override", "", "API endpoint to override default")
+	computeServiceAccount       = flag.String("compute-service-account", "", "Compute service account to be used by importer Virtual Machine and the resulting VM or Machine Image. When empty, the Compute Engine default service account is used.")
```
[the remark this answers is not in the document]
zoran15: Added trimming.

Regarding validation, they can be emails but they can also be `default`. Maybe other IDs are acceptable as well, e.g. without the domain. 

I don't see much in that docs to be able to build a robust validation that we can trust it won't case false negatives. I suggest we leave it as is, since instance creation will do the validation for us, though a bit delayed.

cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
zoran15: `create_gmi.wf.json` only creates a GMI from a running instance and there's no way to specify service account in this case. I'm not even sure if GMIs preserve this information. If they do, service account info will be inherited from the source instance.

zoran15: Actually, you're right, `create_gmi.wf.json` also creates the instance. It looks like that, even without setting service account, things work, since no operation is performed on this temporary VM, but I'll update it to have the custom service account in case it gets preserved in GMI.

cli_tools/gce_ovf_import/main.go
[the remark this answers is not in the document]
EricEdens: sgtm -- when I did a quick search I didn't see an authoritative answer on format either.

EricEdens (approved):

PR #1547, opened 2021-02-25
OVF Import: added -scope flag to allow custom instance access scopes

zoran15:
OVF Import: added -scope flag to allow custom instance access scopes. This is needed as the resulting instance/GMI will have custom service account set. Without the ability to set scopes as well, instance will be set with no scopes, potentially restricting what can be accessed from instance. 

Also added the check for service account in GMI E2E import test, since service accounts are preserved preserved by GMIs and transfered to any VMs screated from them.

### DIFF
--- a/cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
+++ b/cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/paramhelper"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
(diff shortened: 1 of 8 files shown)


EricEdens (approved):

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
```diff
@@ -591,6 +622,22 @@ func verifyImportedInstance(
 		}
 	}
 
+	if props.instanceAccessScopes != "" {
```
dntczdx, line 625: if the scope is not specified, what's the most proper value? Maybe make it align with the result of "gcloud compute instances create"?

zoran15: Added default scopes if not specified by user

dntczdx (approved):

PR #1549, opened 2021-02-26
Add -byol flag to OVF import

EricEdens:
The `-byol` flag allows users to import systems with existing licenses when OS detection is enabled.

In addition to adding the flag, this adds logic to interpret when a user specifies both `-os` and `-byol`.

### DIFF
--- a/cli_tools/common/image/importer/request.go
+++ b/cli_tools/common/image/importer/request.go
@@ -105,19 +105,25 @@ type ImageImportRequest struct {
 	Zone                  string `name:"zone" validate:"required"`
 }
 
-// MergeBYOLIntoOSID interprets the user's --os and --byol flags, and returns a semantically-equivalent
-// tuple. The returned values follow the invariants from ImageImportRequest's validation where --byol may only
-// be specified when --os is empty (implying that detection is being used).
+// FixBYOLAndOSFlags fixes the user's --os and --byol flags to follow the invariants from ImageImportRequest's
+// validation where --byol may only be specified when --os is empty (implying that detection is being used).
 //
 // For example, given `--byol --os=rhel-8`, this will return `--os=rhel-8-byol`.
-func MergeBYOLIntoOSID(osID string, byol bool) (resolvedOSID string, resolvedBYOL bool) {
-	if osID == "" || !byol {
-		return osID, byol
+func FixBYOLAndOSFlags(osID string, byol bool) (fixedOSID string, fixedBYOL bool) {
+	if osID == "" {
+		// User wants OS detection, so keep the `--byol` flag unchanged.
+		return "", byol
 	}
-	if strings.HasSuffix(osID, "byol") {
-		return osID, false
+
+	// User wants OS override. If they set --byol, clear it, and
+	// update osID to include the `-byol` suffix.
+	fixedBYOL = false
+	fixedOSID = osID
+	if byol && !strings.HasSuffix(osID, "byol") {
+		fixedOSID += "-byol"
 	}
-	return osID + "-byol", false
+
+	return fixedOSID, false
 }
 
 // EnvironmentSettings returns the subset of EnvironmentSettings that are required to instantiate
--- a/cli_tools/common/image/importer/request.go
+++ b/cli_tools/common/image/importer/request.go
@@ -105,6 +105,21 @@ type ImageImportRequest struct {
 	Zone                  string `name:"zone" validate:"required"`
 }
 
+// MergeBYOLIntoOSID interprets the user's --os and --byol flags, and returns a semantically-equivalent
+// tuple. The returned values follow the invariants from ImageImportRequest's validation where --byol may only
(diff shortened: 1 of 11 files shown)


cli_tools/common/image/importer/request.go
```diff
@@ -105,6 +105,21 @@ type ImageImportRequest struct {
 	Zone                  string `name:"zone" validate:"required"`
 }
 
+// MergeBYOLIntoOSID interprets the user's --os and --byol flags, and returns a semantically-equivalent
+// tuple. The returned values follow the invariants from ImageImportRequest's validation where --byol may only
```
zoran15, line 109: nit: a tuple is a list of elements, while only one string is returned (ignoring bool status)

cli_tools/common/image/importer/request.go
```diff
@@ -105,6 +105,21 @@ type ImageImportRequest struct {
...
+// be specified when --os is empty (implying that detection is being used).
+//
+// For example, given `--byol --os=rhel-8`, this will return `--os=rhel-8-byol`.
+func MergeBYOLIntoOSID(osID string, byol bool) (resolvedOSID string, resolvedBYOL bool) {
+	if osID == "" || !byol {
+		return osID, byol
+	}
+	if strings.HasSuffix(osID, "byol") {
+		return osID, false
```
zoran15, line 118: What is the meaning of `resolvedBYOL`? I.e. osID in this case is byol but `false` is returned for `resolvedBYOL` which is unintuitive (not necessarily incorrect depending on how this func is used). 

More documentation and/or better return value names would be helpful here to understand what this func does and why.

cli_tools/common/image/importer/request.go
```diff
@@ -105,6 +105,21 @@ type ImageImportRequest struct {
 	Zone                  string `name:"zone" validate:"required"`
 }
 
+// MergeBYOLIntoOSID interprets the user's --os and --byol flags, and returns a semantically-equivalent
+// tuple. The returned values follow the invariants from ImageImportRequest's validation where --byol may only
+// be specified when --os is empty (implying that detection is being used).
+//
+// For example, given `--byol --os=rhel-8`, this will return `--os=rhel-8-byol`.
+func MergeBYOLIntoOSID(osID string, byol bool) (resolvedOSID string, resolvedBYOL bool) {
```
zoran15, line 113: nit: resolvedOSID->mergedOSID, or rename the func to ResolveBYOLIntoOSID to have naming consistent?

cli_tools/gce_ovf_import/README.md
```diff
@@ -61,7 +61,11 @@ Exactly one of these must be specified:
 + `-no-restart-on-failure` the instance will not be restarted if it’s terminated by Compute Engine.
   This does not affect terminations performed by the user.
 + `-os=OS` Specifies the OS of the image being imported. Execute the tool with `-help` to
-  see the list of currently-supported operating systems.
+  see the list of currently-supported operating systems. When omitted, the tool detects the OS.
++ `-byol` Import using an [existing license](https://cloud.google.com/compute/docs/nodes/bringing-your-own-licenses). These are equivalent:
+  * `-os=rhel-8 -byol`
+  * `-os=rhel-8-byol -byol`
+  * `-os=rhel-8-byol`
```
zoran15, line 68: nit: move -byol flag ahead of -os (or add more spaces) to make it more readable (`-os=rhel-8 -byol` looks very similar to `-os=rhel-8-byol`)

cli_tools/common/image/importer/request.go
EricEdens: That makes sense -- the new version doesn't mention "tuple"

cli_tools/gce_ovf_import/README.md
EricEdens: done!

cli_tools/common/image/importer/request.go
EricEdens: New version uses "fixed" for method name and to prefix return values

cli_tools/common/image/importer/request.go
EricEdens: Updated the method's name, docs, and comments to clarify.

In short: The method standardizes various permutations of `--os` and `--byol` into a "standard" representation to avoid bugs in downstream code.

zoran15: I think the cause for the confusion on my side is that I didn't realize this func consumes and returns flag values. I.e. even now, `fixedBYOL`, is actuall `fixedBYOLFlagValue`. Once it's clear these are flag values, not just generic OS/BYOL variables, returning `false` for `fixedBYOL` makes sense, otherwise it looks like a bug. So, consider renaming these vars to indicate they are flag values.

EricEdens: That's good feedback. Updated the arguments to be pointers and tweaked the method's names and docs.

zoran15 (approved):

PR #1561, opened 2021-03-10
Daisy: Don't panic when workflow is canceled twice

EricEdens:
This PR adds a method called `daisy.Workflow.CancelWorkflow()` that is idempotent, and protects against closing an already-closed channel.

Background: Image import has had panics when manually closing workflows, since we (currently) close a workflow by closing its `Cancel` channel.

Here's an example scenario that would cause a panic:

1. Daisy fails here: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/423907b63e7430a8e9dbe5edb296e4811f73797a/daisy/workflow.go#L302
2. The shadow inflater reaches here: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/b5310952cee40f215182002d869188396582b0c6/cli_tools/common/image/importer/inflater.go#L188

### DIFF
--- a/cli_tools/common/utils/daisy/daisy_utils.go
+++ b/cli_tools/common/utils/daisy/daisy_utils.go
@@ -24,10 +24,11 @@ import (
 	"sort"
 	"strings"
 
-	stringutils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/string"
-	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
+
+	stringutils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/string"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
 
 const (
@@ -382,7 +383,7 @@ func RunWorkflowWithCancelSignal(ctx context.Context, w *daisy.Workflow) error {
 		select {
 		case <-c:
 			w.LogWorkflowInfo("\nCtrl-C caught, sending cancel signal to %q...\n", w.Name)
-			close(w.Cancel)
+			w.CancelWorkflow()
 		case <-w.Cancel:
 		}
 	}(w)
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -89,7 +89,9 @@ func (v *Var) UnmarshalJSON(b []byte) error {
 // Workflow is a single Daisy workflow workflow.
 type Workflow struct {
 	// Populated on New() construction.
-	Cancel chan struct{} `json:"-"`
+	Cancel       chan struct{} `json:"-"`
+	isCanceled   bool
+	inCanceledMx sync.Mutex
 
 	// Workflow template fields.
 	// Workflow name.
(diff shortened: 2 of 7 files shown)


daisy/workflow.go
```diff
@@ -894,7 +887,22 @@ func (w *Workflow) IterateWorkflowSteps(cb func(step *Step)) {
...
 	w.cancelReason = reason
-	close(w.Cancel)
+	w.CancelWorkflow()
+}
+
+// CancelWorkflow cancels the workflow. Safe to call multiple times.
+// Prefer this to closing the w.Cancel channel,
+// which will panic if it has already been closed.
+func (w *Workflow) CancelWorkflow() {
+	w.cleanupHooksMx.Lock()
```
dntczdx, line 897: should be inCanceledMx per my understanding

EricEdens: shoot, that was a typo! Thanks; fixed.

daisy/workflow.go
```diff
@@ -89,7 +89,9 @@ func (v *Var) UnmarshalJSON(b []byte) error {
 // Workflow is a single Daisy workflow workflow.
 type Workflow struct {
 	// Populated on New() construction.
-	Cancel chan struct{} `json:"-"`
+	Cancel       chan struct{} `json:"-"`
```
zoran15, line 92: Can this be non-exported to avoid direct use? It would be a breaking change but using previous package version would work.

daisy/workflow_test.go
```diff
@@ -155,6 +155,43 @@ func TestGetSourceGCSAPIPath(t *testing.T) {
...
+	if !w.isCanceled {
+		t.Error("Expect workflow to be canceled.")
+	}
+}
+
+func TestCancelWithReason_IsIdempotent(t *testing.T) {
+	w := testWorkflow()
+	reason := "reason"
+	w.CancelWithReason(reason)
+	w.CancelWithReason(reason)
```
zoran15, line 174: What if reasons are different? Which one should apply, the first or the second?

zoran15 (approved):

dntczdx (approved):

cli_tools/common/utils/daisy/daisy_utils.go
```diff
@@ -24,10 +24,11 @@ import (
 	"sort"
 	"strings"
 
-	stringutils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/string"
-	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
+
```
adjackura, line 29: Why the separation of imports here?

daisy/workflow.go
adjackura: +1, I'm fine with making this private

daisy/workflow.go
```diff
@@ -894,7 +887,22 @@ func (w *Workflow) IterateWorkflowSteps(cb func(step *Step)) {
...
+}
+
+// CancelWorkflow cancels the workflow. Safe to call multiple times.
+// Prefer this to closing the w.Cancel channel,
+// which will panic if it has already been closed.
+func (w *Workflow) CancelWorkflow() {
+	w.isCanceledMx.Lock()
+	defer w.isCanceledMx.Unlock()
+
+	if !w.isCanceled {
```
adjackura, line 900: Why not just keep the isCanceled() function and not use this bool value so you don't need the recover here?

daisy/workflow_test.go
EricEdens: Good question! There's probably an argument for keeping either the first or the last. My initial thought is first, since the last may be a timeout that occurs while the first is being cleaned up, and therefore the first would be a more specific reason.

(I updated the code and docs to reflect the first reason is kept. )

daisy/workflow.go
EricEdens: I found some legitimate usecases for reading the channel, and given that it'd be a backward-incompatible change, I'm hesitant overall. That said, I overall agree that exposing the channel has downsides.

I'd propose punting this to its own PR if we feel that it's a good idea?

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/1521d107ad566e886a627344a8b1055165de7a9b/cli_tools/common/utils/daisy/daisy_utils.go#L387

cli_tools/common/utils/daisy/daisy_utils.go
EricEdens: In cli_tools, we've been splitting into three groups: standard library, 3p deps, compute-image-tools; eg: https://github.com/golang/tools/commit/ed69e84b1518b5857a9f4e01d1f9cefdcc45246e

daisy/workflow.go
EricEdens: AFAICT,  `isCanceled` indicated that either the channel was closed, or a value was written to it. Given the ambiguity, I thought it was clearer to consolidate to a boolean value.

zoran15 (approved):

adjackura (approved):

PR #1563, opened 2021-03-11
Image import: Disable calling PD API.

EricEdens:
This adds a new filed `ImageImportRequest.FormatsWithParallelInflation` which controls which disk file formats are processed in parallel using the PD API. Currently the map is empty, therefore calling the PD API is disabled.

# Testing

Ran an OVF import using the three-disk OVA that was failing with deadlock.  When run with this PR and https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1561, the import failed as expected rather than deadlocking.

### DIFF
--- a/cli_tools/common/image/importer/inflater.go
+++ b/cli_tools/common/image/importer/inflater.go
@@ -52,7 +52,9 @@ func newInflater(request ImageImportRequest, computeClient daisyCompute.Client, 
 		return nil, err
 	}
 
-	if isImage(request.Source) {
+	// TODO: Remove boolean guard when PD API is ready.
+	tryNativePDInflation := false
+	if isImage(request.Source) || !tryNativePDInflation {
 		return di, nil
 	}
 


cli_tools/common/image/importer/request.go
```diff
@@ -103,6 +103,9 @@ type ImageImportRequest struct {
 	Timeout               time.Duration `name:"timeout" validate:"required"`
 	UefiCompatible        bool
 	Zone                  string `name:"zone" validate:"required"`
+	// FormatsWithParallelInflation contains the file formats (eg: vmdk, vpc) that should be inflated
+	// in parallel with the PD API.
+	FormatsWithParallelInflation map[string]bool
```
dntczdx, line 108: In which place this value should be specified when we want to enable shadow test?

cli_tools/common/image/importer/inflater.go
```diff
@@ -47,88 +48,37 @@ type shadowTestFields struct {
...
-		InflationType:   "qemu_failed",
-		InflationTimeMs: []int64{tf.inflationTime.Milliseconds()},
-	})
-	return pd, tf, err
-}
-
-func (facade *inflaterFacade) Cancel(reason string) bool {
-	// No need to cancel apiInflater.
-	return facade.daisyInflater.Cancel(reason)
+	return &shadowTestInflaterFacade{
```
dntczdx, line 73: I think we should still have the gate to determine whether to return &inflaterFacade to use PD API.

cli_tools/common/image/importer/request.go
zoran15, line 108: Naming for this map is inconsistent with shadowInflater.

cli_tools/common/image/importer/inflater.go
zoran15: +1

We should have a property to easily switch between daisy and API inflaters. This will make our lives easier when we do need to switch and there will be less risk of errors.

zoran15 (commented):
A lot of the code got removed in this PR that we might want to keep for the future. In particular, inflaterFacade. We want to be able to switch between daisy, shadow and API (with fallback) inflaters with minimal code changes. As is, we would have to dig through old commits to get inflaterFacade back.

cli_tools/common/image/importer/request.go
EricEdens: Switching to a quick implementation that disables the PD API

cli_tools/common/image/importer/inflater.go
EricEdens: Switching to a quick implementation that disables the PD API

cli_tools/common/image/importer/request.go
EricEdens: Switching to a quick implementation that disables the PD API

EricEdens:
> We should have a property to easily switch between daisy and API inflaters.

@dntczdx @zoran15 

I switched to a quick implementation that puts a boolean guard around using the PD API. You both had good feedback around API design, as well as which code should be left versus removed. With this smaller change, we can avoid the current timeout issues, while allowing further offline discussions about the items that you talked about.

dntczdx (approved):

zoran15 (approved):

PR #1569, opened 2021-03-12
AlmaLinux 8 image build.

zmarano:


daisy_workflows/image_build/enterprise_linux/almalinux_8.wf.json
```diff
@@ -0,0 +1,59 @@
+{
+  "Name": "build-almalinux-8",
```
gaohannk, line 2: is AlmaLinux has a version number 8 like centos?

zmarano: Yup- its the same versioning scheme - and it will follow RHEL just like CentOS.

PR #1571, opened 2021-03-15
Refactor el

hopkiw:
tested with side-by-side comparison as in:

```
for repo in repo_options:
  for release in release_options:
    for byos in byos_options:
      for sap in sap_options:
        origcontent = orig.BuildKsConfig(repo,release,byos,sap)
        newcontent = new.BuildKsConfig(repo,release,byos,sap)
        if origcontent != newcontent:
          print("failed with %s %s %s %s" % (repo,release,byos,sap))
          return
print("success")
```

though some munging for e.g. whitespace was needed, that i've pruned from the final commit


zmarano:
Two comments: one is that this now assumes names in kickstart files (even more than before). I think that's fine but lets throw a basic note about how these files get rendered into the kickstart config in the README file for future sanity. The other is to take into account more derivatives - which I think should work fine but we're going to have AlmaLinux and Rocky Linux soon. See #1569

zmarano (approved):

PR #1579, opened 2021-03-22
Add SEV_CAPABLE flag to EL8 images.

zmarano:


zoran15 (commented):
Should we do the same for imported images?

zmarano:
> Should we do the same for imported images?

Only if we are updating to the most recent kernel (there were bugs in previous RHEL 8 kernels). For import I suggest we leave it off for now however.

PR #1581, opened 2021-03-24
import_precheck: Fail when root device spans multiple physical disks

EricEdens:
Prior to this, the precheck tool falsely reported that a system was importable when the root device spanned multiple physical disks (typically due to usage of LVM). 

## Testing

- Created a multi-disk LVM system that reproduced the original issue, and confirmed that this PR solves it.
- Added unit tests that run against `lsblk` output from four representative disk layouts (single disk, EFI, multiple disks that should pass, multiple disks that should fail).

### DIFF
--- a/cli_tools/import_precheck/check_disks_linux.go
+++ b/cli_tools/import_precheck/check_disks_linux.go
@@ -15,6 +15,7 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -25,6 +26,9 @@ import (
 // disksCheck performs disk configuration checking:
 // - finding the root filesystem partition
 // - checking if the device is MBR
+// - checking whether the root mount is physically located on a single disk.
+//   The check fails, for example, when the root mount is on an LVM
+//   logical volume that spans multiple disks.
 // - check for GRUB
 // - warning for any mount points from partitions from other devices
 type disksCheck struct {
@@ -55,8 +59,12 @@ func (c *disksCheck) readMounts() (*mountPoints, error) {
 		cmd := exec.Command("lsblk", "--json", "--output", "name,mountpoint,type")
 		lsblkOut, err = cmd.Output()
 		if err != nil {
-			exitErr := err.(*exec.ExitError)
-			err = fmt.Errorf("lsblk: %v, stderr: %s", err, exitErr.Stderr)
+			var exitErr *exec.ExitError
+			if errors.As(err, &exitErr) {
+				err = fmt.Errorf("lsblk: %v, stderr: %s", err, exitErr.Stderr)
+			} else {
+				err = fmt.Errorf("lsblk: %v", err)
+			}
 		}
 	}
 
@@ -73,16 +81,6 @@ func (c *disksCheck) readMounts() (*mountPoints, error) {
 	return mounts, nil
 }
 
-func (c *disksCheck) lsblk() (string, error) {
-	cmd := exec.Command("lsblk", "-i")
-	out, err := cmd.Output()
-	if err != nil {
-		exitErr := err.(*exec.ExitError)
-		return "", fmt.Errorf("lsblk: %v, stderr: %s", err, exitErr.Stderr)
-	}
-	return string(out), nil
-}
-
 func (c *disksCheck) getName() string {
 	return "Disks Check"
 }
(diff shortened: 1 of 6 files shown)


cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -60,47 +87,114 @@ func (c *disksCheck) getName() string {
...
+		if mount == "/" {
+			continue
+		}
+		devices := allMounts.listPhysicalDevicesForMount(mount)
+		switch len(devices) {
+		case 0:
+			// This implies a bug in mountPoints.addAll.
+			panic(fmt.Sprintf("Invalid parse of mount %s", mount))
+		case 1:
+			if devices[0] != rootDevice {
```
schweikert, line 182: I feel like it should hard-fail for some specific mount points like /var or /usr (/usr is unlikely these days, but I think some people mount /var on a separate partition/volume).

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -29,7 +29,7 @@ import (
 // - warning for any mount points from partitions from other devices
```
zoran15, line 29: Should LVM be mentioned in this doc?

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -46,6 +46,33 @@ func (c *disksCheck) getMBR(devName string) ([]byte, error) {
...
+func (c *disksCheck) readMounts() (*mountPoints, error) {
+	var lsblkOut []byte
+	var err error
+	if c.lsblkOverride != nil {
+		lsblkOut, err = c.lsblkOverride()
+	} else {
+		cmd := exec.Command("lsblk", "--json", "--output", "name,mountpoint,type")
+		lsblkOut, err = cmd.Output()
+		if err != nil {
+			exitErr := err.(*exec.ExitError)
```
zoran15, line 58: Is it always ExitError?

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -60,47 +87,114 @@ func (c *disksCheck) getName() string {
...
-		out, err = c.lsblkOverride()
-	} else {
-		out, err = c.lsblk()
+// mountPoints supports listing the mounts on a system, and determining which block device(s)
+// the mount is physically contained. This is helpful since technologies such as LVM allow
+// logical mounts to span physical devices.
+type mountPoints struct {
+	mounts []struct {
+		dir       string
+		hierarchy []string
```
zoran15, line 96: is `hierarchy` a list of block devices? Or a path split into elements? It's not clear what this field represents

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -60,47 +87,114 @@ func (c *disksCheck) getName() string {
...
 	}
-	if err != nil {
-		return nil, err
+}
+
+// listPhysicalDevicesForMount returns the the block device(s) that contain the mount.
+// An empty return value means the mount was not found.
+func (m *mountPoints) listPhysicalDevicesForMount(mount string) (devices []string) {
```
zoran15, line 102: mount -> mountDir?

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -60,47 +87,114 @@ func (c *disksCheck) getName() string {
...
+	set := map[string]struct{}{}
+	for _, mp := range m.mounts {
+		if mp.dir == mount {
+			set[mp.hierarchy[0]] = struct{}{}
```
zoran15, line 106: Why only [0]?

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -60,47 +87,114 @@ func (c *disksCheck) getName() string {
...
+func (m *mountPoints) addAll(elements []DiskElement, basePath []string) {
+	for _, element := range elements {
+		path := make([]string, len(basePath))
+		copy(path, basePath)
+		path = append(path, element.Name)
+		if element.Mountpoint != "" {
+			m.mounts = append(m.mounts, struct {
+				dir       string
+				hierarchy []string
+			}{element.Mountpoint, path})
```
zoran15, line 141: Maybe have this struct as a type, since it's repeated here and in the definition of mountPoints

cli_tools/import_precheck/check_disks_linux.go
```diff
@@ -60,47 +87,114 @@ func (c *disksCheck) getName() string {
...
+
+	rootDevices := allMounts.listPhysicalDevicesForMount("/")
+	switch len(rootDevices) {
+	case 0:
 		r.Fatal("root filesystem partition not found on any block devices.")
 		return r, nil
+	case 1:
+		r.Info(fmt.Sprintf("root filesystem found on device: %s", rootDevices[0]))
+	default:
+		format := "root filesystem spans multiple physical devices (%s). Translation only supports single block device."
```
zoran15, line 166: Customers probably don't know what translation is. We can just say "image import" instead. 

Also, maybe add a bit more context to customers on what caused this situation, e.g. mention LVM (or any other similar technologies).

cli_tools/import_precheck/check_disks_linux.go
EricEdens: done!

cli_tools/import_precheck/check_disks_linux.go
EricEdens: Probably not, good call; I copied that from the original code too quickly. 

Updated to allow for other types of errors.

cli_tools/import_precheck/check_disks_linux.go
EricEdens: Good question! The first entry is the physical device; added a function and docs to help explain.

cli_tools/import_precheck/check_disks_linux.go
EricEdens: I'm worried about failing the check when there are circumstances where the import *could* have worked.

For example when the image is imported without the guest environment. In that case, there'd be no interactions with the package manager, and therefore probably no writing to /var/log.

(During translation, the disk is mounted as a chroot rather than booting.)

zoran15 (approved):

cli_tools/import_precheck/check_disks_linux.go
EricEdens: done

cli_tools/import_precheck/check_disks_linux.go
EricEdens: good question -- described with a comment (it's a list of block devices. the first is the physical device, then partition, then virtual devices)

cli_tools/import_precheck/check_disks_linux.go
EricEdens: done

cli_tools/import_precheck/check_disks_linux.go
EricEdens: done

PR #1584, opened 2021-03-29
Update CentOS 8 import tests: 8.2 to 8.3

EricEdens:


zoran15 (approved):
What's the reason for doing this?

EricEdens:
> What's the reason for doing this?

The 8.2 image was having intermittent XFS filesystem corruption, and my first-attempt fix would be to run package updates. 

AFAICT, when a new minor version comes out, the previous one is no longer supported. So, by updating this image to 8.3, we get the package updates that I was looking for, and we get an image that uses the currently-supported minor version of 8.2.

https://lwn.net/Articles/601915/

PR #1585, opened 2021-03-29
Adding Alpha API support in Daisy for CreateImage/GetImage/DeprecateImage

bkatyl:
Adding Alpha API support for CreateImage/GetImage/DeprecateImage so that gce_image_publish can be updated to use Alpha API for rollout policy


zoran15 (commented):
I thought gce_image_publish uses Daisy instead of making direct calls to GCE API? If so, you'll need to update Daisy steps to utilize these newly created Alpha API functions.

zoran15 (commented):
In general, LGTM. You might need to split this PR in two: first one for Daisy changes and the second one for cli_tools changes. This is because cli_tools depend on Daisy changes and the build will not recognize Daisy changes when building cli_tools within the same PR.

cli_tools/gce_image_publish/main.go
```diff
@@ -49,6 +49,7 @@ var (
 	noConfirm      = flag.Bool("skip_confirmation", false, "don't ask for confirmation")
 	ce             = flag.String("compute_endpoint_override", "", "API endpoint to override default, will override ComputeEndpoint in template")
 	filter         = flag.String("filter", "", "regular expression to filter images to publish by prefixes")
+	rolloutRate 	 = flag.String("rollout_rate","60", "The number of minutes between the image rolling out between zones. 0 minutes will not use a rollout policy.")
```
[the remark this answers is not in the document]
bkatyl: This is removed for now. It will be in the follow-on PR that includes all of the changes to gce_image_publish and library under it.

daisy/step_deprecate_images.go
```diff
@@ -51,8 +54,14 @@ func (d *DeprecateImages) validate(ctx context.Context, s *Step) DError {
 			return Errf("cannot deprecate image %q: project does not exist: %q", di.Image, di.Project)
 		}
 
-		if !strIn(di.DeprecationStatus.State, deprecationStates) {
-			return Errf("DeprecationStatus.State of %q not in %q", di.DeprecationStatus.State, deprecationStates)
+		if di.DeprecationStatusAlpha.State != "" {
+			if !strIn(di.DeprecationStatusAlpha.State, deprecationStates) {
```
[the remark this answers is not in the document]
bkatyl: Great point, done.

daisy/step_deprecate_images.go
```diff
@@ -51,8 +54,14 @@ func (d *DeprecateImages) validate(ctx context.Context, s *Step) DError {
 			return Errf("cannot deprecate image %q: project does not exist: %q", di.Image, di.Project)
 		}
 
-		if !strIn(di.DeprecationStatus.State, deprecationStates) {
-			return Errf("DeprecationStatus.State of %q not in %q", di.DeprecationStatus.State, deprecationStates)
+		if di.DeprecationStatusAlpha.State != "" {
```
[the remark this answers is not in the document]
bkatyl: If it's empty the GA API should be used. deprecationStates includes the empty string as that is ensuring the state is valid.

daisy/step_create_images.go
```diff
@@ -116,15 +141,22 @@ func (ci *CreateImages) run(ctx context.Context, s *Step) DError {
...
+	if imageUsesAlphaFeatures(ci.ImagesAlpha) {
+		for _, i := range ci.ImagesAlpha {
 			wg.Add(1)
 			go createImage(i, i.OverWrite)
 		}
 	} else {
-		for _, i := range ci.Images {
-			wg.Add(1)
-			go createImage(i, i.OverWrite)
+		if imageUsesBetaFeatures(ci.ImagesBeta) {
```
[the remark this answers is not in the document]
bkatyl: Done. Change the logic of pully the GA to be !imageUsesAlphaFeatures(ci.ImagesAlpha) && !imageUsesBetaFeatures(ci.ImagesBeta)

daisy/compute/compute.go
```diff
@@ -1152,6 +1230,31 @@ func (c *client) ListImages(project string, opts ...ListCallOption) ([]*compute.
 	}
 }
 
+// ListImagesAlpha gets a list of GCE Images.
```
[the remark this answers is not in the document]
bkatyl: Done.

daisy/step_deprecate_images.go
```diff
@@ -51,8 +54,14 @@ func (d *DeprecateImages) validate(ctx context.Context, s *Step) DError {
...
+			if !strIn(di.DeprecationStatusAlpha.State, deprecationStates) {
+				return Errf("DeprecationStatusAlpha.State of %q not in %q", di.DeprecationStatusAlpha.State, deprecationStates)
+			}
+		} else {
+			if !strIn(di.DeprecationStatus.State, deprecationStates) {
```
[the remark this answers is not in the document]
bkatyl: Thanks for catching that.

daisy/compute/compute_test.go
```diff
@@ -141,87 +146,103 @@ func TestCreates(t *testing.T) {
...
+			&compute.ForwardingRule{Name: testForwardingRule},
 			fr,
 		},
 		{
 			"FirewallRules",
 			func() error { return c.CreateFirewallRule(testProject, fir) },
 			fmt.Sprintf("/%s/global/firewalls/%s?alt=json&prettyPrint=false", testProject, testFirewallRule),
 			fmt.Sprintf("/%s/global/firewalls?alt=json&prettyPrint=false", testProject),
-			&compute.Firewall{Name: testFirewallRule, SelfLink: "foo"},
+			&compute.Firewall{Name: testFirewallRule},
```
zoran15, line 165: Why are SelfLink fields removed here and for other resources?

daisy/compute/compute_test.go
```diff
@@ -288,8 +309,8 @@ func TestStops(t *testing.T) {
 	}
 	defer svr.Close()
 
-	stopURL = fmt.Sprintf("/%s/zones/%s/instances/%s/stop?alt=json&prettyPrint=false", testProject, testZone, testInstance)
-	opGetURL = fmt.Sprintf("/%s/zones/%s/operations//wait?alt=json&prettyPrint=false", testProject, testZone)
+	stopURL = fmt.Sprintf("/projects/%s/zones/%s/instances/%s/stop?alt=json&prettyPrint=false", testProject, testZone, testInstance)
+	opGetURL = fmt.Sprintf("/projects/%s/zones/%s/operations//wait?alt=json&prettyPrint=false", testProject, testZone)
```
zoran15, line 313: Why are these changed?

daisy/step_copy_gcs_objects_test.go
```diff
@@ -56,10 +56,10 @@ func TestCopyGCSObjectsValidate(t *testing.T) {
 		} else if m == "GET" && u == "/b/bucket2?alt=json&prettyPrint=false&projection=full" {
 			fmt.Fprint(w, `{}`)
 		} else if m == "POST" && (u == "/b/bucket1/o?alt=json&prettyPrint=false&projection=full&uploadType=multipart" ||
-			u == "/upload/storage/v1/b/bucket1/o?alt=json&prettyPrint=false&projection=full&uploadType=multipart") {
+			u == "/upload/storage/v1/b/bucket1/o?alt=json&name=daisy-validate--abcdef&prettyPrint=false&projection=full&uploadType=multipart") {
```
zoran15, line 59: This doesn't seem related to the  feature implemented in this PR. Why this change?

daisy/step_create_images.go
```diff
@@ -116,12 +141,19 @@ func (ci *CreateImages) run(ctx context.Context, s *Step) DError {
 		ci.markCreatedInWorkflow()
 	}
 
+	if imageUsesAlphaFeatures(ci.ImagesAlpha) {
+		for _, i := range ci.ImagesAlpha {
+			wg.Add(1)
+			go createImage(i, i.OverWrite)
+		}
+	}
 	if imageUsesBetaFeatures(ci.ImagesBeta) {
```
zoran15, line 150: Beta could use Alpha features in which case `createImage` would be called twice (once for alpha, once for beta). This will not happen right now as isBeta returns `false` but it could happen in the future if we want to use a beta-only feature.

I think the best would be to have these `if`s separated with `else`. I.e.:

```
if alpha {
   createAlpha
} else if beta {
   createBeta
} else {
    create
}
```

daisy/step_deprecate_images.go
```diff
@@ -76,10 +83,16 @@ func (d *DeprecateImages) run(ctx context.Context, s *Step) DError {
...
+			if di.DeprecationStatusAlpha.State != "" {
+				w.LogStepInfo(s.name, "DeprecateImages", "%q --> %q with DefaultRolloutTime %s.", di.Image, di.DeprecationStatusAlpha.State, di.DeprecationStatusAlpha.StateOverride.DefaultRolloutTime)
+				if err := w.ComputeClient.DeprecateImageAlpha(di.Project, di.Image, &di.DeprecationStatusAlpha); err != nil {
+					e <- newErr("failed to deprecate images", err)
+				}
+			} else {
+				w.LogStepInfo(s.name, "DeprecateImages", "%q --> %q.", di.Image, di.DeprecationStatus.State)
+				if err := w.ComputeClient.DeprecateImage(di.Project, di.Image, &di.DeprecationStatus); err != nil {
+					e <- newErr("failed to deprecate images", err)
+				}
```
zoran15, line 95: Minor: writing to `e` channel can be consolidated in one place outside the if/else

daisy/step_deprecate_images.go
```diff
@@ -76,10 +83,16 @@ func (d *DeprecateImages) run(ctx context.Context, s *Step) DError {
 		wg.Add(1)
 		go func(di *DeprecateImage) {
 			defer wg.Done()
-
-			w.LogStepInfo(s.name, "DeprecateImages", "%q --> %q.", di.Image, di.DeprecationStatus.State)
-			if err := w.ComputeClient.DeprecateImage(di.Project, di.Image, &di.DeprecationStatus); err != nil {
-				e <- newErr("failed to deprecate images", err)
+			if di.DeprecationStatusAlpha.State != "" {
```
zoran15, line 86: I assume the reason to call deprecate alpha is because Alpha API uses the rollout features of the image  while GA doesn't? Otherwise there's no difference in DeprecationStatus between the two.

daisy/compute/compute_test.go
bkatyl: Same as above, when the versions were updated in go.mod for the imported libraries these tests broke. It looked like the URL expectations changes from the library given the same input.

daisy/step_copy_gcs_objects_test.go
bkatyl: Same as above, when the versions were updated in go.mod for the imported libraries these tests broke with name being added to the expected URL output.

daisy/step_deprecate_images.go
bkatyl: Correct, the rollout feature, [StateOverride *RolloutPolicy](https://pkg.go.dev/google.golang.org/api@v0.44.0/compute/v0.alpha#DeprecationStatus.StateOverride), is only available in the Alpha API.

daisy/step_create_images.go
bkatyl: That makes sense, thanks for the feedback. Made this change in [90f4a2d](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1585/commits/90f4a2d7d536a8a77324428d4ba9a2a0757eaf8f)

daisy/step_deprecate_images.go
bkatyl: Made this change in [90f4a2d](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1585/commits/90f4a2d7d536a8a77324428d4ba9a2a0757eaf8f)

zoran15 (approved):

PR #1608, opened 2021-04-22
Updated imageUsesAlphaFeatures and added tests

bkatyl:
- Updated imageUsesAlphaFeatures logic to check each item for nil
- Added imageUsesAlphaFeatures specific tests

### DIFF
--- a/daisy/step_create_images.go
+++ b/daisy/step_create_images.go
@@ -54,11 +54,15 @@ func (ci *CreateImages) UnmarshalJSON(b []byte) error {
 
 func imageUsesAlphaFeatures(imagesAlpha []*ImageAlpha) bool {
 	for _, imageAlpha := range imagesAlpha {
-		if imageAlpha != nil && len(imageAlpha.RolloutOverride.DefaultRolloutTime) > 0 {
-			return true
+		if imageAlpha != nil && imageAlpha.RolloutOverride != nil {
+			if len(imageAlpha.RolloutOverride.DefaultRolloutTime) > 0 {
+				return true
+			}
 		}
-		if imageAlpha != nil && len(imageAlpha.Deprecated.StateOverride.DefaultRolloutTime) > 0 {
-			return true
+		if imageAlpha != nil && imageAlpha.Deprecated != nil && imageAlpha.Deprecated.StateOverride != nil {
+			if len(imageAlpha.Deprecated.StateOverride.DefaultRolloutTime) > 0 {
+				return true
+			}
 		}
 	}
 	return false
--- a/daisy/step_create_images.go
+++ b/daisy/step_create_images.go
@@ -54,15 +54,11 @@ func (ci *CreateImages) UnmarshalJSON(b []byte) error {
 
 func imageUsesAlphaFeatures(imagesAlpha []*ImageAlpha) bool {
 	for _, imageAlpha := range imagesAlpha {
-		if imageAlpha != nil && imageAlpha.RolloutOverride != nil {
-			if len(imageAlpha.RolloutOverride.DefaultRolloutTime) > 0 {
-				return true
-			}
+		if imageAlpha != nil && imageAlpha.RolloutOverride != nil  && len(imageAlpha.RolloutOverride.DefaultRolloutTime) > 0 {
+			return true
 		}
-		if imageAlpha != nil && imageAlpha.Deprecated != nil && imageAlpha.Deprecated.StateOverride != nil {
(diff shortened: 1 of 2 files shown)


daisy/step_create_images.go
```diff
@@ -54,11 +54,15 @@ func (ci *CreateImages) UnmarshalJSON(b []byte) error {
 
 func imageUsesAlphaFeatures(imagesAlpha []*ImageAlpha) bool {
 	for _, imageAlpha := range imagesAlpha {
-		if imageAlpha != nil && len(imageAlpha.RolloutOverride.DefaultRolloutTime) > 0 {
-			return true
+		if imageAlpha != nil && imageAlpha.RolloutOverride != nil {
+			if len(imageAlpha.RolloutOverride.DefaultRolloutTime) > 0 {
```
zoran15, line 58: minor, you can squash these into one `if`

daisy/step_create_images.go
```diff
@@ -54,11 +54,15 @@ func (ci *CreateImages) UnmarshalJSON(b []byte) error {
...
+				return true
+			}
 		}
-		if imageAlpha != nil && len(imageAlpha.Deprecated.StateOverride.DefaultRolloutTime) > 0 {
-			return true
+		if imageAlpha != nil && imageAlpha.Deprecated != nil && imageAlpha.Deprecated.StateOverride != nil {
+			if len(imageAlpha.Deprecated.StateOverride.DefaultRolloutTime) > 0 {
```
zoran15, line 63: here as well

daisy/step_create_images_test.go
```diff
@@ -115,3 +116,84 @@ func TestCreateImagesRun(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			wantResult: true,
+		},
+		{
+			desc: "Don't AlphaAPI, empty",
```
zoran15, line 162: Should this read "Don't use AlphaAPI"?

daisy/step_create_images_test.go
```diff
@@ -115,3 +116,84 @@ func TestCreateImagesRun(t *testing.T) {
...
+			cia: []*ImageAlpha{{Image: computeAlpha.Image{}}},
```
zoran15, line 163: This will test the `imageAlpha.RolloutOverride = nil` but we should also test when `len(imageAlpha.RolloutOverride.DefaultRolloutTime) = 0` and:
* `imageAlpha.Deprecated == nil` or
* `imageAlpha.Deprecated.StateOverride == nil`

daisy/step_create_images_test.go
bkatyl: Absolutely, fixed throughout. Thanks.

daisy/step_create_images.go
bkatyl: Updated in [99c3847](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1608/commits/99c3847ba83eaf50d1fc32d551bd5343d185b45e)

daisy/step_create_images.go
bkatyl: Updated in [99c3847](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1608/commits/99c3847ba83eaf50d1fc32d551bd5343d185b45e)

daisy/step_create_images_test.go
bkatyl: Updated in [99c3847](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1608/commits/99c3847ba83eaf50d1fc32d551bd5343d185b45e)

- len(imageAlpha.RolloutOverride.DefaultRolloutTime) = 0 
  - I believe this is covered by `Do not use AlphaAPI, RolloutOverride present with all default values.` as `RolloutOverride: &computeAlpha.RolloutPolicy{}` initializes the struct and DefaultRolloutTime is a string which has a default value of empty string.

- imageAlpha.Deprecated == nil 
  - Added `Do not use AlphaAPI, Deprecated is nil.` I had thought this would be covered by `Do not use AlphaAPI, empty` since it's not defined there.

- imageAlpha.Deprecated.StateOverride == nil
  -  I added `Do not use AlphaAPI, deprecated with StateOverride = nil.`. I had thought `Do not use AlphaAPI, deprecated but with no StateOverride.` would have covers this as StateOverride should be nil since it wasn't defined and it's a pointer.

zoran15 (approved):

zoran15 (approved):

PR #1609, opened 2021-04-22
cleanerupper: add OSPolicyAssignments and staging osconfig resources

adjackura:


test-infra/prowjobs/cleanerupper/main.go
```diff
@@ -359,6 +362,48 @@ func cleanGuestPolicies(ctx context.Context, client *osconfig.Client, project st
...
+			}
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				op, err := client.DeleteOSPolicyAssignment(ctx, &osconfigv1alphapb.DeleteOSPolicyAssignmentRequest{Name: ospa.GetName()})
+				if err != nil {
+					fmt.Printf("Error deleting OSPolicyAssignment: %v\n", err)
+					return
+				}
+				op.Wait(ctx)
```
manjunathganga, line 400: Will this return when you get 404 not found which is the current behavior after the assignment is deleted

adjackura: Yes, this will wait for operation to be Done, which will never happen as the operation will disappear, so Wait will return an error that we do not check for, so this works fine for this use case.

PR #1612, opened 2021-04-23
error handling in metadata exporter

hopkiw:


daisy_workflows/export_metadata/export-metadata.py
```diff
@@ -90,14 +91,19 @@ def main():
     cmd = cmd_prefix + [package]
     process = None
     try:
-      process = subprocess.run(cmd, stdout=subprocess.PIPE, check=True)
+      process = subprocess.run(cmd, capture_output=True, check=True)
```
gaohannk, line 94: I found rhel default python3 is 3.6. And it doesn't have this argument. https://docs.python.org/3.6/library/subprocess.html#module-subprocess

hopkiw: i ran the workflow before submitting the PR :) 
default on the worker image is 3.7.3

PR #1617, opened 2021-04-28
Replace 'google.com' with 'google_com' in bucket names.

mbernhardt6:
Cloud Bucket policy prevents the use of 'google.com' in the name. Replace it with the friendlier 'google_com' so we don't also chop off the front of the scratch bucket name.

### DIFF
--- a/cli_tools/common/utils/storage/scratch_bucket_creator.go
+++ b/cli_tools/common/utils/storage/scratch_bucket_creator.go
@@ -152,7 +152,8 @@ func (c *ScratchBucketCreator) getBucketAttrsIfInProject(project string, bucketN
 }
 
 func (c *ScratchBucketCreator) formatScratchBucketName(project string, location string) string {
-	bucket := strings.Replace(strings.Replace(project, "google.com", "google_com", -1), ":", "-", -1) + "-daisy-bkt"
+	bucket := strings.Replace(project, "google.com", "elgoog.com", -1)
+	bucket = strings.Replace(bucket, ":", "-", -1) + "-daisy-bkt"
 	if location != "" {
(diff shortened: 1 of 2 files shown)


cli_tools/common/utils/storage/scratch_bucket_creator.go
```diff
@@ -152,7 +152,7 @@ func (c *ScratchBucketCreator) getBucketAttrsIfInProject(project string, bucketN
 }
 
 func (c *ScratchBucketCreator) formatScratchBucketName(project string, location string) string {
-	bucket := strings.Replace(project, ":", "-", -1) + "-daisy-bkt"
+	bucket := strings.Replace(strings.Replace(project, "google.com", "google_com", -1), ":", "-", -1) + "-daisy-bkt"
```
bkatyl, line 155: https://cloud.google.com/storage/docs/naming-buckets#requirements
- Bucket names cannot begin with the "goog" prefix.
- Bucket names cannot contain "google" or close misspellings, such as "g00gle"

If you wanted to change the string from `google.com` to `elgoog_com` I believe it would meet all of the requirements.
I'd also suggest doing this on 2 lines instead of combining it into a single operation, so it's easier to read.
```
bucket := strings.Replace(project, "google.com", "elgoog.com", -1)
bucket = strings.Replace(bucket, ":", "-", -1) + "-daisy-bkt"
```

Also can you include updating a test to verify the behavior.

mbernhardt6: Split into 2 lines and added a test function.

zoran15 (approved):

bkatyl (approved):

PR #1625, opened 2021-05-06
Adding "server sent GOAWAY" as a response that should be retried.

bkatyl:
- Adding "server sent GOAWAY" as a response that should be retried.
- Sleeping 5 seconds instead of 1 second, when a "PENDING", "RUNNING" operation status is received.

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -219,10 +219,15 @@ func shouldRetryWithWait(tripper http.RoundTripper, err error, multiplier int) b
 		}
 	}
 
+
 	apiErr, ok := err.(*googleapi.Error)
 	var retry bool
 	switch {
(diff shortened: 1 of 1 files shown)


daisy/compute/compute.go
```diff
@@ -346,7 +346,7 @@ func (c *client) operationsWaitHelper(project, name string, getOperation operati
 
 		switch op.Status {
 		case "PENDING", "RUNNING":
-			time.Sleep(1 * time.Second)
+			time.Sleep(5 * time.Second)
```
zoran15, line 349: Maybe make this configurable so we keep the existing behavior where existing errors will retry in 1 second while GOAWAY will retry after 5 seconds.

bkatyl: That makes sense. I moved this logic into the should retry function so that specific response doubles the sleep duration multiplier, which is based on the retry attempt count.

zoran15 (approved):

PR #1626, opened 2021-05-07
Changing wait operation retries to 40

bkatyl:
With the updated libraries the wait call, especially for long operations like in-place upgrade, can exceed the fixed number of retries in the retry function were as for most image publish operations 4 retries corrected their issue.
- Added RetryAttempts function which lets the caller specify how many retry attempts should be made for that specific call.
- Updated the 3 wait calls (global/region/zone) to 40 tries.
  - This number of retries ensure that if there is a truly fatal error it eventually gives up, but otherwise changes the behavior to be rather resilience to GOAWAY errors.

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -308,7 +308,7 @@ type operationGetterFunc func() (*compute.Operation, error)
 
 func (c *client) zoneOperationsWait(project, zone, name string) error {
 	return c.operationsWaitHelper(project, name, func() (op *compute.Operation, err error) {
-		op, err = c.Retry(c.raw.ZoneOperations.Wait(project, zone, name).Do)
+		op, err = c.RetryAttempts(40, c.raw.ZoneOperations.Wait(project, zone, name).Do)
 		if err != nil {
 			err = fmt.Errorf("failed to get zone operation %s: %v", name, err)
 		}
@@ -318,7 +318,7 @@ func (c *client) zoneOperationsWait(project, zone, name string) error {
 
 func (c *client) regionOperationsWait(project, region, name string) error {
 	return c.operationsWaitHelper(project, name, func() (op *compute.Operation, err error) {
-		op, err = c.Retry(c.raw.RegionOperations.Wait(project, region, name).Do)
+		op, err = c.RetryAttempts(40, c.raw.RegionOperations.Wait(project, region, name).Do)
 		if err != nil {
 			err = fmt.Errorf("failed to get region operation %s: %v", name, err)
 		}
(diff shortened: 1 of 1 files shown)


daisy/compute/compute.go
```diff
@@ -417,6 +417,54 @@ func (c *client) RetryAlpha(f func(opts ...googleapi.CallOption) (*computeAlpha.
 	return
 }
 
+// RetryAttempts invokes the given function, retrying the specified number times if the HTTP
+// status response indicates the request should be attempted again or the
+// oauth Token is no longer valid.
+func (c *client) RetryAttempts(attempts int, f func(opts ...googleapi.CallOption) (*compute.Operation, error), opts ...googleapi.CallOption) (op *compute.Operation, err error) {
```
illfelder, line 423: Usually it's a good idea to have some wait between retries (potentially with some backoff). I'd be surprised if some library you are using doesn't already have this functionality. Here's an example I found in Python: https://googleapis.dev/python/google-api-core/latest/retry.html

zoran15: To add to this, we already have some examples of retries with backoff implemented here:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/common/imagefile/inspector.go#L72

bkatyl: The wait between attemps is actually in the [shouldRetryWithWait](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/d5d70336cabf120388c02a002ef5ddddd77fde70/daisy/compute/compute.go#L251) function. Retry and RetryAttempts pass in the attempt number and that is used to calcuate how long the delay, specifically it's a multipler for the time to wait.

Sleep is called if the response is determined to be retry-able. If it is not found to be retry-able, the function returns false and Retry/RetryAttempts break out of the for loop and exits Retry/RetryAttempts.

illfelder: I'm satisfied with this. I'll defer to Zoran on approval, but LGTM.

EricEdens: It'd be nice to see an alternate factoring that achieves these goals:
1. Eliminate duplicated code (right now a future reader or maintainer has to update the GA, Alpha, and Beta loops)
2. Add unit tests to ensure the inner function is called the expected number of times (hint: If I called this function with `1` attempt right now, how many times would `f` get invoked?)
3. Unit tests showing that retry eventually stops
4. Log errors, calls, and responses from each retry (or is that already happening somewhere else in the stack?)
I've found these tests to be helpful, since retry code is often tweaked, and can hide bugs that cause nasty behavior (infinite retry, geometric retry, no retry)

One approach to achieve these goals that you might consider:
1. Refactor a function `shouldRetry` out of `shouldRetryWithWait`. This function is given an error and returns whether it should be retried
2. Use this function in conjunction with a library such as https://pkg.go.dev/github.com/cenkalti/backoff/v4 . This will encourage  you to have a single implementation of retry logic across GA, Beta, and Alpha, and isolate the differences inside a small closure

For some examples of unit tests that assert retry behavior, see: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/common/imagefile/inspector_test.go

PR #1627, opened 2021-05-08
Support -source_snapshot flag on gce_vm_image_export cli tool

rueian:
Hi, I though it would be nice if `gce_vm_image_export` tool can export snapshot to GCS and therefore I created this PR.

Thanks to the the Daisy implementation, the only thing I have to do is adding `SourceSnapshot` to the `CreateDisks` step to let the Daisy create disk from the snapshot.

This PR add new `-source_snapshot` flag and also validate only one of `-source_image` and `-source_snapshot` must be provided.


zoran15 (commented):
`gce_vm_image_export` was supposed to perform GCE image export to a file. Disk snapshots are not images so this change breaks the semantics of the tool. However, I can see how this would be useful to customers and the change is minimal (vs. implementing a similar tool for disk snapshots).

One comment about variable naming and help texts: the resource full name is "persistent disk snapshot": https://cloud.google.com/compute/docs/disks/snapshots
We can call it disk snapshot. But calling it just snapshot could be confusing from the perspective of GCE images (i.e. someone might assume it's image snapshot which doesn't exist as a resource).

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -120,6 +120,7 @@ type ImageExportParams struct {
 
 	DestinationURI        string `json:"destination_uri,omitempty"`
 	SourceImage           string `json:"source_image,omitempty"`
+	SourceSnapshot        string `json:"source_snapshot,omitempty"`
```
zoran15, line 123: README should be updated with the new flag and optionality of -source_image.

rueian:
Hi @zoran15,

Thank you for reviewing. I just renamed the snapshot to disk snapshot on variables and help text.

README is also updated.

cli_tools/gce_vm_image_export/README.md
```diff
@@ -20,10 +20,12 @@ go get github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/gce_vm_image
...
   virtual disk file. For example: gs://my-bucket/my-exported-image.vmdk.
-+ `-source_image=SOURCE_IMAGE` An existing Compute Engine image URI from which to 
-  export.
 
-#### Optional flags  
+#### Optional flags
++ `-source_image=SOURCE_IMAGE` An existing Compute Engine image URI from which to
+  export. Exactly one of `-source_image` or `-source_disk_snapshot` must be provided.
++ `-source_disk_snapshot=SOURCE_DISK_SNAPSHOT` An existing Compute Engine disk snapshot URI from which to
+  export. Exactly one of `-source_image` or `-source_disk_snapshot` must be provided.
```
zoran15, line 28: These are not really optional flags, since at least one of them has to be specified.

Have a look at the README for https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/cli_tools/gce_vm_image_import
There, we have `--source_file` and `--source_image` that are similar to image/snapshot flags in the exporter. Use the same README structure for the exporter (under Required flags, add the "Exactly one of these must be specified:" sub-section).

Also, the usage section should be updated as well by adding image/snapshot flag group.

rueian: Thank you for pointing out the example and the missing usage section.
README is updated.

cli_tools/common/utils/logging/service/log_entry.go
```diff
@@ -120,6 +120,7 @@ type ImageExportParams struct {
 
 	DestinationURI        string `json:"destination_uri,omitempty"`
 	SourceImage           string `json:"source_image,omitempty"`
+	SourceDiskSnapshot    string `json:"source_disk_snapshot,omitempty"`
```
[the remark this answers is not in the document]
rueian: ok. is the logging backend also in the repository?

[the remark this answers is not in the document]
EricEdens: The settings that we'll need to make are internal, and they require ~2 weeks to propagate prior to adding a new field.

EricEdens (approved):
lgtm. Thanks @rueian !

zoran15 (approved):

PR #1629, opened 2021-05-10
Add detection and removal of read-only attribute from windows host file

bkatyl:
Add detection and removal of read-only attribute from windows host file

### DIFF
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -146,13 +146,8 @@ function Configure-Network {
   # Register netkvmco.dll.
   Run-Command rundll32 'netkvmco.dll,RegisterNetKVMNetShHelper'
 
-  # Remove ReadOnly file attribute from host_file, if present.
-  if ((Get-ItemProperty -Path $script:hosts_file).attributes -like '*ReadOnly*') {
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -146,6 +146,11 @@ function Configure-Network {
   # Register netkvmco.dll.
   Run-Command rundll32 'netkvmco.dll,RegisterNetKVMNetShHelper'
 
+  # Remove ReadOnly file attribute from host_file, if present.
```
LawrenceHwang, line 149: Might I also share an alternative way which is to use the `-Force` with `Add-Content` [Ref: [Add-Content doc](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/add-content?view=powershell-5.1)]? It might be simpler this way. It also doesn't alter the original file permission. Thanks. 

```powershell
Write-Host "Appending Google Compute Engine metadata server address to host file $($script:hosts_file)."
Add-Content -Path $script:hosts_file -Value "# Google Compute Engine metadata server`n    169.254.169.254    metadata.google.internal metadata" -Force
```

bkatyl: Thanks, I like that simplier solution.

EricEdens (approved):

zoran15 (approved):

PR #1634, opened 2021-05-12
add install package wf

hopkiw:
new workflow and script to create a derivative image with a single package installed
packages are installed from gcs but with repositories available to satisfy dependencies
based on metadata export workflow

### DIFF
--- a/daisy_workflows/image_build/install_package/installpackage.py
+++ b/daisy_workflows/image_build/install_package/installpackage.py
@@ -24,7 +24,7 @@ def get_mount_disk(image):
   # 2. partition 2 is the root mount for the installed system.
   #
   # Except on debian 10, which has out-of-order partitions.
-  if gpt and 'debian-10' not in image:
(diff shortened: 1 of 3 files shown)


daisy_workflows/image_build/install_package/installpackage.py
```diff
@@ -0,0 +1,92 @@
...
+  devname = _GetMetadataParam('disks/1/device-name', raise_on_not_found=True)
+  devicepath = f'/dev/disk/by-id/google-{devname}'
+  gpt = get_part_type(devicepath) == 'gpt'
+
+  # This assumes that, for UEFI systems:
+  # 1. partition 1 is the EFI system partition.
+  # 2. partition 2 is the root mount for the installed system.
+  #
+  # Except on debian 10, which has out-of-order partitions.
+  if gpt and 'debian-10' not in image:
```
zmarano, line 27: Include debian-11 here as well

zmarano (approved):

PR #1637, opened 2021-05-14
Update dependencies for Daisy and dependent tools

bkatyl:
- Updating the 3 go.mod files and all of the tests that need to be changes due to library changes
  - google.golang.org/api v0.44.0
  - cloud.google.com/go v0.81.0
  - cloud.google.com/go/storage v1.14.0 (1.15 requires google.golang.org/api v0.45.0)

Tested by compiling and using daisy and gce_image_publish.
gce_image_publish has the most concurrency, so the test was publishing all SQL images, which in the last attempt has a 100% failure rate. The 28 images published in this test.


zoran15 (approved):
the fifth time is the charm? :)

zmarano (approved):

PR #1641, opened 2021-05-18
Adding Alpha API support to the image create/deprecate processes.

bkatyl:
- Adding Alpha API support to the image create/deprecate processes
- This PR does not update any dependency

### DIFF
--- a/daisy/compute/compute.go
+++ b/daisy/compute/compute.go
@@ -42,8 +42,10 @@ type Client interface {
 	CreateForwardingRule(project, region string, fr *compute.ForwardingRule) error
 	CreateFirewallRule(project string, i *compute.Firewall) error
 	CreateImage(project string, i *compute.Image) error
+	CreateImageAlpha(project string, i *computeAlpha.Image) error
 	CreateImageBeta(project string, i *computeBeta.Image) error
 	CreateInstance(project, zone string, i *compute.Instance) error
+	CreateInstanceAlpha(project, zone string, i *computeAlpha.Instance) error
 	CreateInstanceBeta(project, zone string, i *computeBeta.Instance) error
 	CreateNetwork(project string, n *compute.Network) error
 	CreateSnapshot(project, zone, disk string, s *compute.Snapshot) error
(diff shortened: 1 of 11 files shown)


daisy/compute/compute.go
```diff
@@ -350,7 +364,7 @@ func (c *client) operationsWaitHelper(project, name string, getOperation operati
 
 		switch op.Status {
 		case "PENDING", "RUNNING":
-			time.Sleep(1 * time.Second)
+			time.Sleep(5 * time.Second)
```
zoran15, line 367: do we still need this? If I remember correctly, this was done to reduce the frequence of wait operations? With v.0.44.0 GCP go library, we might not need to do this.

daisy/step_deprecate_images.go
```diff
@@ -51,7 +54,11 @@ func (d *DeprecateImages) validate(ctx context.Context, s *Step) DError {
 			return Errf("cannot deprecate image %q: project does not exist: %q", di.Image, di.Project)
 		}
 
-		if !strIn(di.DeprecationStatus.State, deprecationStates) {
+		// Verify State is one of the deprecated states.
+		// The Alpha check also requires the value to not be emptry string as in that case the GA API will be used.
+		if di.DeprecationStatusAlpha.State != "" && strIn(di.DeprecationStatusAlpha.State, deprecationStates) {
```
zoran15, line 59: I forgot, can di.DeprecationStatusAlpha be null?

daisy/compute/compute.go
bkatyl: No we don't. Updated the PR to revert it back to 1 second.

zoran15 (approved):

daisy/step_deprecate_images.go
bkatyl: The deprecationStates lists empty string as an allowed value, so requiring di.DeprecationStatusAlpha.State to not be empty string ensures that when it's not set in the Alpha API data structure, we use the GA data structure.

PR #1644, opened 2021-05-19
deprecate master branch

gaohannk:


zoran15 (commented):
What is this change about? We use master as our main branch, build all the tools from it etc. We can't just delete it.

gaohannk:
@zoran15  Hi Zoran: We propose a change which deprecate master branch and change to main. Here is the proposal. 
I am working on to make change in guest-agent and test the workflow. 
https://docs.google.com/document/d/1t1wQVseCPoi_4BTIT77RRS0AsygM0dIRtsOH_ZUjPwA/edit#

gaohannk:
Here is the step to make this change. https://docs.google.com/document/d/1ekSMf6u2E9KZ8xOTsoQ7OjtJkZ5gZ-x73kXN0IGVwWY/edit#

PR #1647, opened 2021-05-24
Adding image rollout policy support to GCE_Image_Publish

bkatyl:
Adding image rollout policy support to GCE_Image_Publish
- Added rollout_rate to gce_image_publish, default to 60 minutes
- Migrated GCE_image_publish to use the Alpha API
- Adding image rollout policy support to publish library
  - Added image rollout schedule to printout

Sending for review, however I want to hold on the merge until 6/1


cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -473,22 +491,36 @@ func (p *Publish) deprecatePrintOut(deprecateImages *daisy.DeprecateImages) {
...
 			p.toObsolete = append(p.toObsolete, image)
-		case "":
+		case "ACTIVE", "":
 			p.toUndeprecate = append(p.toUndeprecate, image)
 		}
 	}
 }
 
-func (p *Publish) populateWorkflow(ctx context.Context, w *daisy.Workflow, pubImgs []*compute.Image, img *Image, rb, sd, rep, noRoot bool) error {
+func (p *Publish) rolloutPolicyPrintOut(rp *computeAlpha.RolloutPolicy) {
```
zoran15, line 505: This function is named like it's printing rollout policies but it's actually building a printout. Consider either renaming it to something like `buildRolloutPolicyPrintOut` or change the behavior if this is an oversight.

cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -473,22 +491,36 @@ func (p *Publish) deprecatePrintOut(deprecateImages *daisy.DeprecateImages) {
...
 	for _, di := range *deprecateImages {
 		image := path.Base(di.Image)
-		switch di.DeprecationStatus.State {
+		switch di.DeprecationStatusAlpha.State {
 		case "DEPRECATED":
 			p.toDeprecate = append(p.toDeprecate, image)
 		case "OBSOLETE":
 			p.toObsolete = append(p.toObsolete, image)
-		case "":
+		case "ACTIVE", "":
```
zoran15, line 499: Was this a bug earlier or a downstream dependency changed?

cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -585,3 +624,48 @@ func calculateExpiryDate(deleteAfter string) (*time.Time, error) {
...
+		sort.Strings(value)
+	}
+
+	// zoneList is the ordered list of zones to apply the rollout policy to.
+	var zoneList []string
+
+	// zoneList's order should be the first zone from each region, then second zone from each region, third zone from each region, etc.
+	// us-central1-a, us-central2-b, us-central3-c, us-central1-a, us-central2-b, us-central3-c
+	for zoneCount := 0; zoneCount < maxRegionLength; zoneCount++ {
+		for _, value := range regions {
```
zoran15, line 655: minor: `value` -> `regionZones` or similar

cli_tools/gce_image_publish/publish/publish_test.go
```diff
@@ -518,13 +682,76 @@ func TestDeprecatePrintOut(t *testing.T) {
...
 			}
 			if !reflect.DeepEqual(p.toUndeprecate, tt.toUndeprecate) {
-				t.Errorf("deprecatePrintOut() got2 = %v, want %v", p.toUndeprecate, tt.toUndeprecate)
+				t.Errorf("deprecatePrintOut() toUndeprecate got = %v, want %v", p.toUndeprecate, tt.toUndeprecate)
+			}
+		})
+	}
+}
+
+func TestCreateRollOut(t *testing.T) {
```
zoran15, line 697: This doesn't test for sorting zones since test data is already sorted. Not sure that's a strict requirement though so it might not matter.

cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -585,3 +624,48 @@ func calculateExpiryDate(deleteAfter string) (*time.Time, error) {
...
+				zoneList = append(zoneList, value[zoneCount])
+			}
+		}
+	}
+
+	var rolloutPolicy map[string]string
+	rolloutPolicy = make(map[string]string)
+
+	for i, zone := range zoneList {
+		rolloutPolicy[fmt.Sprintf("zones/%s", zone)] = rolloutStartTime.Add(time.Duration(rolloutRate*i) * time.Minute).Format(time.RFC3339)
```
zoran15, line 667: Is one zone every `rolloutRate`minutes the only strategy we want to use? Have you considered increasing the speed of rollout over time (e.g. first one zone, then two zones, then 3 etc...)?

zoran15 (approved):
Some questions and comments but looks goods.

cli_tools/gce_image_publish/main.go
```diff
@@ -49,6 +49,7 @@ var (
 	noConfirm      = flag.Bool("skip_confirmation", false, "don't ask for confirmation")
 	ce             = flag.String("compute_endpoint_override", "", "API endpoint to override default, will override ComputeEndpoint in template")
 	filter         = flag.String("filter", "", "regular expression to filter images to publish by prefixes")
+	rolloutRate    = flag.Int("rollout_rate", 60, "The number of minutes between the image rolling out between zones. 0 minutes will not use a rollout policy.")
```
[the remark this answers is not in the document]
illfelder: Right - we should default to using the rollout policy (safer rollouts by default).

cli_tools/gce_image_publish/publish/publish.go
illfelder: We will look at alternative strategies over time, but one zone every hour was the initial plan agreed upon with SRE.

cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -236,17 +242,23 @@ func (p *Publish) CreateWorkflows(ctx context.Context, varMap map[string]string,
 		printList(p.toDelete)
 	}
 
+	if len(p.rolloutPolicy) > 0 {
+		fmt.Println("  All images will have the following rollout policy:")
```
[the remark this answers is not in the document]
bkatyl: The extra spaces in the beginning is to match the existing output pattern. It should be useful to everyone that publishes images as it'll share the rollout schedule that will be used for the images they're publishing.

Here's a sample of the output that is generated.
```
["debian"] Preparing workflows from template
  - Creating publish workflow for "debian-10-buster"
  The following images will be created in "my-project":
   - [ debian-10-buster-v20210310: (Debian, Debian GNU/Linux, 10 (buster), amd64 built on 20210310, supports Shielded VM features) ]
  All images will have the following rollout policy:
   - [ Default rollout time: 2021-05-24T14:33:16-06:00 ]
   - [ Zone asia-east1-a at 2021-05-25T04:33:16-06:00 ]
   - [ Zone asia-east1-b at 2021-05-26T06:33:16-06:00 ]
```

cli_tools/gce_image_publish/publish/publish.go
bkatyl: In a rollback we would previously set it to empty string as ACTIVE and empty string in the deprecation policy are the same value.

Since the daisy module switches to the Alpha API when this field is set, I changed rollbacks to set it to active instead of empty set.

cli_tools/gce_image_publish/publish/publish.go
bkatyl: Thanks the more descriptive variable name helps.

cli_tools/gce_image_publish/publish/publish_test.go
bkatyl: I updated the testdata so it's no longer sorted.

zoran15 (approved):

PR #1648, opened 2021-05-24
add kernel-devel

gaohannk:


daisy_workflows/image_build/bare_metal/bare_metal_image_preparer.sh
```diff
@@ -27,7 +27,7 @@ EOF
 # Install Packages.
 echo "Install rpm packages."
 yum -y install mcedaemon google-ecclesia-management-agent
-yum -y install gve
+yum -y install kernel-devel gve
```
hopkiw, line 30: what's the reason for this change?

gaohannk: This is a simple workaround to make gve compile/install correctly. 

here is the context from email thread 
```
The gve RPM does depend on kernel-headers.  However, the kernel-header RPM appears to
dump its headers in /usr/include/linux/ whereas the kernel-devel RPM
(and presumably DKMS) use /usr/src/kernels/... ```

gaohannk: Catherine work on to teach DKMS to consume the headers kernel-headers puts in /usr/include rather
than the ones kernel-devel puts in /usr/src/kernels/....

hopkiw: The `dkms` package in EPEL depends on the `kernel-headers` and `kernel-debug-devel` packages. This is not an issue of file paths,  the contents of these packages are different. `kernel-headers` contains header files (*.h) for developing applications which use kernel interfaces. `kernel-devel` and `kernel-debug-devel` packages contain mostly Makefiles for kernel modules. Depending on what you are trying to accomplish, you may need one or both.

So overall it seems this would be more appropriate as a dependency of your proposed dkms package. Is there a reason that won't work?

zmarano: Yeah- technically you are correct. It should really be a dependency of the dkms gve package. However, given the circumstances here- I'm fine putting it in this (very limited) use case script.

zmarano (approved):

hopkiw:
we worked out the packaging issues.

PR #1653, opened 2021-06-03
OVF import: Apply user-provided tags to final instance

EricEdens:
During OVF import, the user can provide tags using the `--tags` flag. A user reported this feature wasn't working. It's unclear where this functionality was lost. Running an import prior to the recent refactoring doesn't yield an instance with tags applied; eg:

```
gcloud beta compute instances import  tagtest \
   --source-uri=gs://<yourbucket>/ubuntu-ova/ \
   --zone=us-central1-c \
   --tags=tag1,tag2,tag3  \
   --docker-image-tag=8f8fff9d757e70b8a09257e038ed7ecf51ab606f  \
   --os ubuntu-1604
```

## Testing
* Added e2e tests for instance import and GMI import that verify tags were applied
* Manually imported an instance and a GMI and verified that tags were applied

### DIFF
--- a/cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_import_e2e_common.go
+++ b/cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_import_e2e_common.go
@@ -47,6 +47,7 @@ type OvfImportTestProperties struct {
 	Network                   string
 	Subnet                    string
 	Project                   string
+	Tags                      []string
 	ComputeServiceAccount     string
 	InstanceServiceAccount    string
 	NoInstanceServiceAccount  bool
@@ -75,6 +76,12 @@ func BuildArgsMap(props *OvfImportTestProperties, testProjectConfig *testconfig.
 	wrapperArgs = append(wrapperArgs, fmt.Sprintf("-zone=%v", testProjectConfig.TestZone))
 	wrapperArgs = append(wrapperArgs, fmt.Sprintf("-build-id=%v", path.RandString(10)))
 
+	if len(props.Tags) > 0 {
+		tags := strings.Join(props.Tags, ",")
+		gcloudBetaArgs = append(gcloudBetaArgs, fmt.Sprintf("--tags=%v", tags))
+		gcloudArgs = append(gcloudBetaArgs, fmt.Sprintf("--tags=%v", tags))
+		wrapperArgs = append(wrapperArgs, fmt.Sprintf("-tags=%v", tags))
+	}
 	if props.Os != "" {
 		gcloudBetaArgs = append(gcloudBetaArgs, fmt.Sprintf("--os=%v", props.Os))
 		gcloudArgs = append(gcloudBetaArgs, fmt.Sprintf("--os=%v", props.Os))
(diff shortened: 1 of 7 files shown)


cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
```diff
@@ -89,6 +89,8 @@ func TestSuite(
 			testSuiteName, fmt.Sprintf("[%v] %v", testType, "Ubuntu 1604 from Virtualbox"))
 		instanceImportUbuntu16FromAWS := junitxml.NewTestCase(
 			testSuiteName, fmt.Sprintf("[%v] %v", testType, "Ubuntu 1604 from AWS"))
+		instanceImportWithNetworkTags := junitxml.NewTestCase(
```
zoran15, line 92: Add tags to an existing test to avoid expanding the number of tests we run.

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_machine_image_import/ovf_machine_image_import_test_suite.go
```diff
@@ -85,12 +85,15 @@ func TestSuite(
 			testSuiteName, fmt.Sprintf("[%v] %v", testType, "Windows 2012 R2 two disks, Network setting (path)"))
 		machineImageImportStorageLocationTestCase := junitxml.NewTestCase(
 			testSuiteName, fmt.Sprintf("[%v] %v", testType, "Centos 7.4, Storage location"))
+		machineImageImportNetworkTagsTestCase := junitxml.NewTestCase(
```
zoran15, line 88: Same here, merge with an existing test.

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_import_e2e_common.go
```diff
@@ -238,6 +245,22 @@ func VerifyInstance(instance *gcp.InstanceBeta, client daisyCompute.Client,
 		}
 	}
 
+	// Verify that the instance's tags match the user's request (regardless of order).
+	expectedTags := map[string]struct{}{}
+	for _, tag := range props.Tags {
+		expectedTags[tag] = struct{}{}
+	}
+	actualTags := map[string]struct{}{}
+	for _, tag := range instance.Tags.Items {
```
zoran15, line 254: This doesn't test if tags are also attached to transient VMs we use for inflation/translation. Our e2e tests are not set up to validate transient VMs so this would be difficult to implement. 

One way we could validate tags on transient VMs would be to simulate customer's setup that exposed this issue: have a VPC that is allowing traffic only from VMs that have a specific tag and use that VPC for the test.

EricEdens: You're exactly right, and I suspect the approach you're describing of simulating the user's setup will be required.

In this PR, I wanted to focus on applying tags to the final instance/GMI; for the transient VMs, I'll do a follow-up.

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_machine_image_import/ovf_machine_image_import_test_suite.go
EricEdens: done! Added tags to the "network setting" tests, since the tags are network related.

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
EricEdens: done! Added tags to the "network setting" tests, since the tags are network related.

zoran15 (approved):

PR #1655, opened 2021-06-08
Improve error message when detected OS doesn't match specified OS.

EricEdens:
When Linux import fails, and the detected OS doesn't match the specified OS, we show a customized error message (added in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1206). 

This PR provides the following improvements:
1. Support Windows in addition to Linux
2. Differentiate between flavors of SLES

To achieve this, we use the [new OS detection module](https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/daisy_workflows/image_import/inspection) that is run against all bootable disks, rather than the quick detection that was used in #1206 which ran only for Linux.

### DIFF
--- a/cli_tools/common/distro/distro.go
+++ b/cli_tools/common/distro/distro.go
@@ -183,6 +183,17 @@ func newLinuxRelease(distro string, major string, minor string) (Release, error)
 	}
 }
 
+// FromGcloudOSArgumentMustParse parses the argument provided to the `--os` flag of
+// `gcloud compute images import`, and returns a Release if it represents a
+// release we *may* support. If osFlagValue does not parse, the call panics.
+func FromGcloudOSArgumentMustParse(osFlagValue string) Release {
+	r, err := FromGcloudOSArgument(osFlagValue)
+	if err != nil {
+		panic(err)
+	}
+	return r
+}
+
 // FromGcloudOSArgument parses the argument provided to the `--os` flag of
 // `gcloud compute images import`, and returns a Release if it represents a
 // release we *may* support. The caller is responsible for verifying
--- a/cli_tools/common/image/importer/bootable_disk_processor.go
+++ b/cli_tools/common/image/importer/bootable_disk_processor.go
@@ -42,8 +42,7 @@ func (b *bootableDiskProcessor) process(pd persistentDisk) (persistentDisk, erro
 	if err != nil {
 		b.logger.User("Finished making disk bootable")
 		daisy_utils.PostProcessDErrorForNetworkFlag("image import", err, b.request.Network, b.workflow)
-		err = customizeErrorToDetectionResults2(b.request.OS,
-			b.detectedOs, err)
+		err = customizeErrorToDetectionResults(b.request.OS, b.detectedOs, err)
 	}
 	if b.workflow.Logger != nil {
 		for _, trace := range b.workflow.Logger.ReadSerialPortLogs() {
--- a/cli_tools/common/image/importer/bootable_disk_processor.go
+++ b/cli_tools/common/image/importer/bootable_disk_processor.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/distro"
 	daisy_utils "github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/daisy"
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging"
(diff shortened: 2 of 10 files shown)


cli_tools/common/image/importer/errors.go
```diff
@@ -30,3 +30,14 @@ func customizeErrorToDetectionResults(osID, detectedDistro, detectedMajor,
 	}
 	return original
 }
+
+func customizeErrorToDetectionResults2(osID string, detected distro.Release, original error) error {
```
Quintonamore, line 34: If the error message is being improved with usage from `distro`, why will we be using the previous `customizeErrorToDetectionResults` function?

Quintonamore (approved):

cli_tools/common/image/importer/errors.go
EricEdens: Good find!  I didn't finish the refactoring here prior to submitting the PR.

The current version removes the old implementation and updates the tests.

cli_tools/common/distro/distro.go
```diff
@@ -183,6 +183,17 @@ func newLinuxRelease(distro string, major string, minor string) (Release, error)
 	}
 }
 
+// FromGcloudOSArgumentMustParse parses the argument provided to the `--os` flag of
+// `gcloud compute images import`, and returns a Release if it represents a
+// release we *may* support. If osFlagValue does not parse, the call panics.
+func FromGcloudOSArgumentMustParse(osFlagValue string) Release {
```
zoran15, line 189: It seems this is a test utility function, not used for prod code. If so, it should probably live in a test/test util code, unless you expect it to be used in prod code in the future.

cli_tools/common/image/importer/bootable_disk_processor.go
```diff
@@ -40,10 +42,7 @@ func (b *bootableDiskProcessor) process(pd persistentDisk) (persistentDisk, erro
 	if err != nil {
 		b.logger.User("Finished making disk bootable")
 		daisy_utils.PostProcessDErrorForNetworkFlag("image import", err, b.request.Network, b.workflow)
-		err = customizeErrorToDetectionResults(b.request.OS,
-			b.workflow.GetSerialConsoleOutputValue("detected_distro"),
-			b.workflow.GetSerialConsoleOutputValue("detected_major_version"),
-			b.workflow.GetSerialConsoleOutputValue("detected_minor_version"), err)
+		err = customizeErrorToDetectionResults(b.request.OS, b.detectedOs, err)
```
zoran15, line 45: Since this code relies on receiving an error from the processor, does this mean we have to have OS detection in two places: the new inspector as well as in translate? If so, that seems fragile and could potentially lead to misleading results (i.e. translate failed for other reasons but we print an OS mismatch error based on the inspection results in the previous step). 

Can't we fail early, even before translation starts, since we already know about OS mismatch at that point?

EricEdens: The net change of this PR is to reduce two OS detections to one. The results of the first OS detection are *used* by the processor, rather than the processor performing its own detection.

Regarding fragility, if `plan.detectedOs` is nil, then the cusotmized error message won't be generated and the user won't will be shown the original error message.

All of that said, I'm curious about this idea: "Can't we fail early, even before translation starts, since we already know about OS mismatch at that point?". Originally I had envisioned the `--os` flag as an escape hatch, that allowed the user to override the value from OS detection. Without this hatch, I'm worried about creating a situation where the user is stuck and unable to import their image (assuming there's a problem with OS detection).

cli_tools/common/distro/distro.go
EricEdens: It's useful for defining constants, similar to [1], and I envisioned it being used outside test code. That said I don't have an immediate usecase for it. Given all that, I'll leave it up to you --  I'm fine to move it, or leave it. 

1. https://golang.org/pkg/regexp/#MustCompile

cli_tools/common/image/importer/bootable_disk_processor.go
zoran15: My point about having two detections is because `err` is coming from translation, while detectedOs is coming from inspection. This code relies on inspection returning an error, assuming this error is for a mismatched OS, means we detect OS twice. Can we avoid this and just use detected OS and --os flag value before translation even starts?

EricEdens: Talked offline. 

I checked, and the original error message is shown for image import, but not OVF import (since OVF import has stricter filtering in the gcloud). So I updated `customizeErrorToDetectionResults` to always emit the original error at the user level, to ensure that it's shown in OVF output.

zoran15 (approved):

PR #1664, opened 2021-06-15
Adding ability to export fixed-sized VHDs

### LINKED ISSUE #1663: Adding ability to export fixed-sized VHDs
`gce_vm_image_export` tool currently provides a wrapper around `qemu-img` utility and allows users to perform image export + image conversion. The underlying `qemu-img` utility supports passing additional options to control various aspects of the converted image (through `-o` option). It would be great if we can expose it as a flag to the users of `gce_vm_image_export` tool.

redborian:
In this PR, I am adding a new format type called "vpc-fixed" which is an abstraction over fixed size vpc images. In the script layer, it is converted into "vpc" and "-o force_size,subformat=fixed" option.

Go unit tests are passing. However, if you can show me how I can validate this change end-to-end, I will do the same as well.

Fixes #1663


zoran15:
> users better control over the converted image

What kind of control do users need over the converted image? I'm wondering if we can enumerate different options and provide them as first-party flags that we can then map to qemu-img flags.

My concern is that, by exposing qemu-img flags directly, we're corning ourselves for the future when there might be native image export API. In that case, qemu-img flag would have to be deprecated and potentially break existing users.

redborian:
The use case for us is that Azure requires VHDs to be built with `force_size=on,subformat=fixed` flags. As a temporary workaround, we are passing the options along with the format flag (`-format vpc -o force_size=on,subformat=fixed`).

> we're corning ourselves for the future when there might be native image export API

Good point. It would be great if we can expose this somehow, either through a common "options" flag or through specialized flags for `force_size=on` and `subformat=fixed`. Do you want me to add them separately instead?

daisy_workflows/export/export_disk_ext.sh
```diff
@@ -31,6 +31,14 @@ GS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/gcs-path)
 FORMAT=$(curl -f -H Metadata-Flavor:Google ${URL}/format)
 DISK_RESIZING_MON=$(curl -f -H Metadata-Flavor:Google ${URL}/resizing-script-name)
 
+# "vpc-fixed" is a compute-image-tools specific format. It boils down to
+# "vpc" format with "-o force_size=on,subformat=fixed"
+if [ "$FORMAT" = "vpc-fixed" ]
+then
+  FORMAT="vpc"
+  QEMU_OPTIONS="-o force_size=on,subformat=fixed"
+fi
```
zoran15, line 40: IMO, this kind of parameter manipulation should not be done in the export script itself, but in the [exporter](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/gce_vm_image_export/exporter/exporter.go). This script should just deal with the mechanics of coverting a file into a PD given parameters, not with converting user provided parameters. Also, having the code in the script makes it hard to test (much easier to do for the exporter Go binary)

redborian: Thanks for the feedback @zoran15. I have addressed it in the latest commit. PTAL.

redborian:
PTAL @zoran15

zoran15 (approved):
Looks good. We should work with tech writers to update our docs so external customers are aware this format is preferred for importing to Azure.

EricEdens (approved):
lgtm

PR #1668, opened 2021-06-17
Image export: Only validate non-URI images

EricEdens:
This is a follow-up to #1666, which introduced a regression such that exports are blocked when the user specifies the source images as a URI. This PR limits validation to obvious image names. This avoids false negatives where the user specifies a source image that Daisy understands, but hasn't been implemented here. 

Failure message: https://github.com/GoogleCloudPlatform/compute-image-tools/commit/58a99dea0dd1703c93a1b8a7f04a658401528f10#commitcomment-52286130

### DIFF
--- a/cli_tools/gce_vm_image_export/exporter/exporter.go
+++ b/cli_tools/gce_vm_image_export/exporter/exporter.go
@@ -17,7 +17,6 @@ package exporter
 
 import (
 	"context"
-	"fmt"
 	"log"
 	"os"
 	"strings"
@@ -241,7 +240,7 @@ func validateImageExists(computeClient daisyCompute.Client, project string, imag
 	_, err = computeClient.GetImage(project, imageName)
 	if err != nil {
 		log.Printf("Error when fetching image %q: %q.", imageName, err)
-		return fmt.Errorf("Image %q not found", imageName)
+		return daisy.Errf("Image %q not found", imageName)
 	}
 	return nil
 }
--- a/cli_tools/gce_vm_image_export/exporter/exporter.go
+++ b/cli_tools/gce_vm_image_export/exporter/exporter.go
@@ -226,11 +226,22 @@ func Run(clientID string, destinationURI string, sourceImage string, sourceDiskS
 	return w, nil
 }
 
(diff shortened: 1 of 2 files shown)


zoran15 (commented):
What about extracting project/imageName from image URL (if that's what's provided) and validating based on those?

cli_tools/gce_vm_image_export/exporter/exporter.go
```diff
@@ -226,11 +225,22 @@ func Run(clientID string, destinationURI string, sourceImage string, sourceDiskS
...
+// validateImageExists checks whether imageName exists in the specified project.
+//
+// This validates when imageName is a valid image name, and skips validation if
+// the imageName is a URI, or something that's not recognized as an image.
+// The simplistic validation avoids false negatives; Daisy has robust logic for
+// interpreting the various permutations of specifying an image and project,
+// and we don't want to copy that here, since this is a convenience method to create
+// user-friendly messages.
+func validateImageExists(computeClient daisyCompute.Client, project string, imageName string) (err error) {
+	if err := validation.ValidateImageName(imageName); err != nil {
```
EricEdens, line 237: > What about extracting project/imageName from image URL (if that's what's provided) and validating based on those?

@zoran15: Would you be alright going forward with the current version, and leaving parsing as a potential future enhancement?  

I'm concerned about the number number of URI formats that are possible, and the amount of work required to get all of the correct:

- Image using family, eg: `projects/debian-cloud/global/images/family/debian-9`
- Image without family but different project, `projects/debian-cloud/global/images/debian-9-stretch-vYYYYMMDD`
- Potential leading API qualifier, that could be incorrect: **http://www.googleapis.com/compute/v1**
- Partial URI that includes the name but not project: `global/images/image-name`

The cost of failure is blocking an export, while the benefit is a better error message. Furthermore, I wanted to avoid duplicating this parsing logic in our codebase, since Daisy already has this implemented.

zoran15 (approved):

cli_tools/gce_vm_image_export/exporter/exporter.go
zoran15: Sure, let's keep it simple for now. We should create a tracking bug for expanding this functionality.

PR #1679, opened 2021-06-25
avoid interleaved output

dntczdx:
Add "stdbuf -oL" to "echo" for serial-output key so the format won't be broken.
ref to b/167607270.

Tested by running export locally.


cli_tools/common/image/importer/api_inflater.go
```diff
@@ -292,7 +292,7 @@ func (inflater *apiInflater) getCalculateChecksumWorkflow(diskURI string) *daisy
 const (
 	checksumScriptConst = `
 		function serialOutputKeyValuePair() {
```
zoran15, line 294: This is now outputting a prefix in addtion to a key/value pair, so a different function name would be appropriate. E.g. `serialOutputPrefixedKeyValue` or similar

zoran15 (approved):

PR #1687, opened 2021-06-30
Update export_disk_ext.sh

eeaton:
The test on line 22 introduces unexpected errors by testing for access to something different than what's actually needed. I've modified it to test reachability to the Storage API, not www.googleapis.com.

Most enterprise GCP customers will configure firewall and DNS controls for the Restricted VIP. In that environment, the attempt to curl www.googleapis.com will fail, even though Storage and other APIs can be accessed. This forces a weird workaround to change DNS and firewall rules to allow some limited egress to the private VIP, which adds complexity and compromises security. Improving the test condition allows this code to run without forcing weird workarounds.

### DIFF
--- a/daisy_workflows/export/export_disk_ext.sh
+++ b/daisy_workflows/export/export_disk_ext.sh
@@ -19,7 +19,7 @@ function serialOutputPrefixedKeyValue() {
 }
 
 # Verify VM has network access to Storage API by checking a public bucket
-curl --silent --fail "https://storage.googleapis.com/storage/v1/b/gcp-public-data-landsat" &> /dev/null;
+gsutil ls &> /dev/null;
 if [[ $? -ne 0 ]]; then
   echo "ExportFailed: Cannot access Google APIs. Ensure that VPC settings allow VMs to access Google APIs either via external IP or Private Google Access. More info at: https://cloud.google.com/vpc/docs/configure-private-google-access"
   exit
--- a/daisy_workflows/export/export_disk_ext.sh
+++ b/daisy_workflows/export/export_disk_ext.sh
@@ -18,8 +18,8 @@ function serialOutputPrefixedKeyValue() {
   stdbuf -oL echo "$1: <serial-output key:'$2' value:'$3'>"
 }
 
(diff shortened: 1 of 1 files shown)


zoran15:
> The test on line 22 introduces unexpected errors by testing for access to something different than what's actually needed. I've modified it to test reachability to the Storage API, not [www.googleapis.com](http://www.googleapis.com).
> 
> Most enterprise GCP customers will configure firewall and DNS controls for the Restricted VIP. In that environment, the attempt to curl [www.googleapis.com](http://www.googleapis.com) will fail, even though Storage and other APIs can be accessed. This forces a weird workaround to change DNS and firewall rules to allow some limited egress to the private VIP, which adds complexity and compromises security. Improving the test condition allows this code to run without forcing weird workarounds.

zoran15:
> The test on line 22 introduces unexpected errors by testing for access to something different than what's actually needed. I've modified it to test reachability to the Storage API, not [www.googleapis.com](http://www.googleapis.com).
> 
> Most enterprise GCP customers will configure firewall and DNS controls for the Restricted VIP. In that environment, the attempt to curl [www.googleapis.com](http://www.googleapis.com) will fail, even though Storage and other APIs can be accessed. This forces a weird workaround to change DNS and firewall rules to allow some limited egress to the private VIP, which adds complexity and compromises security. Improving the test condition allows this code to run without forcing weird workarounds.

Thanks for the contribution. 

In addition to accessing GCS, image export script also does GCE operations. In particular, disk resize using `gcloud compute disks resize`. This is for temporary disks that hold exported image file before it's being copied to GCS. With this workflow, we still need to ensure we can access GCE API, not only GCS. We can look at doing a more focused API check, i.e. only for Compute Engine.

eeaton:
I've suggested a more robust test that checks both compute.googleapis.com and storage.googleapis.com endpoints by checking HTTP status codes. This test will work regardless of whether an enterprise client has configured Firewall and DNS settings for the Restricted VIP. It also doesn't assume the hack of the public bucket as in my earlier commit. Please review, thanks!

PR #1692, opened 2021-07-12
[import] calculate alloc size with minimum limit 10gb

dntczdx:
[no prod impact] This only impacts the shadow test output.
Minimum 10gb of disks will be allocated. We want to output that number correctly.

### DIFF
--- a/cli_tools/common/image/importer/daisy_inflater.go
+++ b/cli_tools/common/image/importer/daisy_inflater.go
@@ -70,7 +70,7 @@ func (inflater *daisyInflater) Inflate() (persistentDisk, shadowTestFields, erro
 	checksum := inflater.wf.GetSerialConsoleOutputValue("disk-checksum")
 	return persistentDisk{
 			uri:        inflater.inflatedDiskURI,
-			sizeGb:     string_utils.SafeStringToInt(targetSizeGB),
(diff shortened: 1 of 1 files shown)


cli_tools/common/image/importer/daisy_inflater.go
```diff
@@ -195,16 +195,17 @@ func calculateScratchDiskSize(metadata imagefile.Metadata) int64 {
...
+	return getAllocatedDiskSize(padded)
 }
 
 // Ensure a minimum of 10GB (the minimum size of a GCP disk)
 func calculateInflatedSize(metadata imagefile.Metadata) int64 {
-	if metadata.VirtualSizeGB < defaultInflationDiskSizeGB {
+	return getAllocatedDiskSize(metadata.VirtualSizeGB)
+}
+
+func getAllocatedDiskSize(size int64) int64 {
```
EricEdens, line 206:
```suggestion
func enforceMinimumDiskSize(size int64) int64 {
```

Or something like that to differentiate between this method and  `calculateInflatedSize`

dntczdx: thanks, done

zoran15 (approved):

PR #1693, opened 2021-07-13
Daisy: Recursively build child workflows in NewFromFile

EricEdens:
Although `workflow.NewFromFile`'s documentation says that it recursively reads subworkflows, that feature was never implemented. This PR ensures that sub workflows and included workflows are read when `workflow.NewFromFile` is called.

To avoid double reading child workflow files, it also adds a guard in the respective `populate` methods.


daisy/workflow.go
```diff
@@ -793,12 +793,24 @@ func New() *Workflow {
...
-// Recursively reads subworkflow steps as well.
-func NewFromFile(file string) (*Workflow, error) {
-	w := New()
+// Recursively reads sub and included steps as well.
+func NewFromFile(file string) (w *Workflow, err error) {
+	w = New()
 	if err := readWorkflow(file, w); err != nil {
 		return nil, err
 	}
+	for _, step := range w.Steps {
```
dntczdx, line 802: originally, included/sub workflow was read when Validate/populate happens.
Now it happens when NewFromFile was called. The problem is, many attributes were not ready yet. Please ref to "daisycommon.ParseWorkflow": many env settings, for example, OAuth, were populated after NewFromFile is called. Thus, they won't be copied to include/sub workflow.

EricEdens: Thanks for the feedback. However, I'm not sure I completely follow the concern. To help me understand, can you write a test that passes with the current behavior on `master`, but fails with this PR?

EricEdens:
Pausing review for a bit; while running some tests, OVF import is failing with `error populating workflow: Unresolved var "${imported_disk}" found in "${imported_disk}"`.

daisy/workflow.go
dntczdx: IncludeWorkflow.populate() / SubWorkflow.populate() will copy attributes, for example, OAuth, from its parent workflow.

These attributes were initially populated in daisycommon.ParseWorkflow().
So, if you create Include/sub workflow BEFORE those attributes were populated, they won't affect include/sub workflows.

EricEdens: `populate` doesn't happen when the JSON file is loaded; it happens during `Run`. It'd help a lot if you wrote a unit test!

dntczdx: I read the CL again and found my misunderstanding. "populate doesn't happen when the JSON file is loaded; it happens during Run" - correct, that is why attributes can be copied to include/sub workflow.

I thought your new condition "&& i.Workflow == nil" will shortcut return. Actually it just skipped loading json. Thanks.

EricEdens:
Re-opening: Fixed the "Unresolved var" error, added unit test to verify it, and ran additional e2e tests locally.

daisy/common.go
```diff
@@ -159,19 +159,36 @@ func (w *Workflow) substituteSourceVars(ctx context.Context, v reflect.Value) DE
 	})
 }
 
+type traverseAction uint
+
+const (
+	continueTraverse traverseAction = iota
+	prune
+)
```
zoran15, line 167: These should be documented, the type and the constants.

daisy/common.go
```diff
@@ -159,19 +159,36 @@ func (w *Workflow) substituteSourceVars(ctx context.Context, v reflect.Value) DE
...
 // traverseData traverses complex data structures and runs
 // a function, f, on its basic data types.
 // Traverses arrays, maps, slices, and public fields of structs.
 // For example, f will be run on bool, int, string, etc.
 // Slices, maps, and structs will not have f called on them, but will
 // traverse their subelements.
 // Errors returned from f will be returned by traverseDataStructure.
-func traverseData(v reflect.Value, f func(reflect.Value) DError) DError {
+// actions allows the caller to determine which action to take at a node.
+// The default action is 'continueTraverse'.
```
zoran15, line 177: It seems `continueTraverse` is a no op. What is the purpose of having it as sequential integer? Where is the value used?

daisy/step_includeworkflow.go
```diff
@@ -33,7 +33,7 @@ type IncludeWorkflow struct {
 }
 
 func (i *IncludeWorkflow) populate(ctx context.Context, s *Step) DError {
-	if i.Path != "" {
+	if i.Path != "" && i.Workflow == nil {
```
zoran15, line 36: For posterity, add a comment on when `i.Workflow` could be non nil and why this check is needed.

EricEdens: Done!

daisy/common.go
EricEdens: You're right, it's a no-op; the new comments describe this action as the default behavior. 

Regarding sequential integers, I used `iota` since it's my understanding of the current-best option for enums: https://github.com/golang/go/issues/28987

daisy/common.go
EricEdens: Done!

daisy/workflow.go
```diff
@@ -543,6 +533,17 @@ func (w *Workflow) populate(ctx context.Context) DError {
 			return Errf("error populating step %q: %v", name, err)
 		}
 	}
+
```
dntczdx, line 536: In IncludeWorkflow.populate(), there is also a call to validateVarsSubbed() and substituteSourceVars(), and they are still after the "populateStep" loop right now. I guess it should also be moved like what you did here.

EricEdens: done!

zoran15 (approved):

dntczdx (approved):

PR #1696, opened 2021-07-15
Adding detection and remediation of static IP & DNS

bkatyl:
- Adding network.ps1 which is launched by run_startup_scripts.cmd
- Added network.ps1 to workflow files so it's copied into the components GCS location
- Added copying network.ps1 to instance in translate_bootstrap.ps1
- network.ps1 does the following:
  - Verified that the connected adapter is set to DHCP, is set the static IP it will change the adapter to DHCP and set the DNS server to also be obtained by DHCP. If updated it will also restart the network adapter and wait 30 seconds via a ping test.
  - It will always test connectivity to packages.cloud.google.com and output the DNS configuration of the adapter.

### DIFF
--- a/daisy_workflows/image_import/windows/network.ps1
+++ b/daisy_workflows/image_import/windows/network.ps1
@@ -19,36 +19,36 @@ $adapter = Get-NetAdapter | ? {$_.Status -eq "up"}
 $interface = $adapter | Get-NetIPInterface -AddressFamily IPv4
 
 If ($interface.Dhcp -eq "Disabled") {
-  Write-Output "Translate: $($interface.InterfaceAlias) does not have DHCP enabled. Enabling DHCP."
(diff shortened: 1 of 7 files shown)


EricEdens (approved):
Nice!

daisy_workflows/image_import/windows/network.ps1
```diff
@@ -0,0 +1,54 @@
...
+} else {
+  Write-Host "Translate: $($interface.InterfaceAlias) is configured for DHCP."
+}
+
+# Log DNS Server information
+Write-Host 'DNS client configuration:'
+$interface | Get-DnsClientServerAddress
+
+# Test connection to packages.cloud.google.com and log results.
+Write-Host 'Testing connection to packages.cloud.google.com:'
```
zoran15, line 52: If these fail what's the action on our side or on customer's side?

PR #1697, opened 2021-07-19
add target pool resource

gaohannk:
This is also for implement load balancer related image test

### DIFF
--- a/daisy/step_create_target_pool.go
+++ b/daisy/step_create_target_pool.go
@@ -1,0 +1,73 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package daisy
+
+import (
+	"context"
+	"sync"
+)
+
+// CreateTargetPools is a Daisy CreateTargetPools workflow step.
+type CreateTargetPools []*TargetPool
+
+func (c *CreateTargetPools) populate(ctx context.Context, s *Step) DError {
+	var errs DError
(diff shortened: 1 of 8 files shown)


daisy/step_create_target_pool.go
```diff
@@ -0,0 +1,72 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
zoran15, line 1: 2021

daisy/target_pool.go
```diff
@@ -0,0 +1,83 @@
...
+	"net/http"
+	"regexp"
+
+	daisyCompute "github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+)
+
+var (
+	targetPoolURLRegex = regexp.MustCompile(fmt.Sprintf(`^(projects/(?P<project>%[1]s)/)?regions/(?P<zone>%[2]s)/TargetPools/(?P<targetPool>%[2]s)$`, projectRgxStr, rfc1035))
```
zoran15, line 16: Isn't it `/targetPools/`? (lowercase)

daisy/target_pool.go
```diff
@@ -0,0 +1,83 @@
...
+	var errs DError
+	tp.Name, errs = tp.Resource.populateWithGlobal(ctx, s, tp.Name)
+
+	tp.Description = strOr(tp.Description, defaultDescription("TargetPool", s.w.Name, s.w.username))
+	tp.link = fmt.Sprintf("projects/%s/regions/%s/targetPools/%s", tp.Project, getRegionFromZone(s.w.Zone), tp.Name)
+	return errs
+}
+
+func (tp *TargetPool) validate(ctx context.Context, s *Step) DError {
+	pre := fmt.Sprintf("cannot create target-pool %q", tp.daisyName)
```
zoran15, line 46: Use a consistent name for target pools (i.e. either target-pool, TargetPool, target pool, but not all of them)

daisy/target_pool.go
gaohannk: I thought the URL is case insensitive, but to make sure change to make it consistent with cloud document.

daisy/step_create_target_pool.go
gaohannk: Done

daisy/target_pool.go
gaohannk: Using` target pool` as plain text log and use `TargetPool` as ResourceTypeName when explicit indicate type

PR #1699, opened 2021-07-22
CLI tools: Pre-validate network and subnet

EricEdens:
This updates the CLI tools to pre-validate the `-network` and `-subnet` params. It adds two additional compute API calls during the validation phase: one to `GetNetwork` and one to `GetSubnetwork`.

It retains the previous behavior of parsing and interpreting the user's `-network` and `-subnet` params. Specifically:

1. When both are empty, explicitly use the `default` network.
2. If the subnet is empty, leave it empty to allow the compute backend to infer it.
3. Similarly, if only the network is empty, leave it empty to allow inference.
4. Backfill `project` and `region` resource properties if they are omitted in the network and subnet URIs.


cli_tools/common/utils/param/network_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	subnetFields, err := paramhelper.SplitSubnetResource(strings.TrimSpace(originalSubnet))
+	if err != nil {
+		return nil, nil, err
+	}
+	if networkFields.String() == "" && subnetFields.String() == "" {
+		return &paramhelper.NetworkResource{
+				Name:    "default",
+				Project: project,
+			}, &paramhelper.SubnetResource{
+				Name:    "default",
```
zoran15, line 108: Is setting subnet to "default" needed? I think we usually got by with network==default and subnet being empty.

cli_tools/common/utils/param/network_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	client daisyCompute.Client
+}
+
+func (r *computeNetworkResolver) Resolve(
+	originalNetwork, originalSubnet, region, project string) (network, subnet string, err error) {
+
+	// 1. Segment the user's input into component fields such as network name, subnet name, project, and region.
+	// If the URI in originalNetwork or originalSubnet didn't specify project or region, then backfill
+	// those fields using region and project.
+	networkFields, subnetFields, err := parseNetworkAndSubnet(originalNetwork, originalSubnet, region, project)
```
zoran15, line 58: networkFields and subnetFields are of types NetworkResource and SubnetResource. Having var fields match type names would make it easier to follow the code.

cli_tools/common/utils/param/network_resolver_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/api/compute/v1"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/mocks"
+)
+
+func TestNetworkResolver_PassingCases(t *testing.T) {
+	type networkParams struct{ project, name string }
+	type subnetParams struct{ project, region, name string }
```
zoran15, line 30: These type names are easily confused with CLI params. Something like getNetworkComputeParams would be clearer.

cli_tools/common/utils/param/populator.go
```diff
@@ -70,6 +73,13 @@ func (p populator) PopulateMissingParameters(project *string, clientID string, z
 		}
 	}
 
+	if n, s, err := p.NetworkResolver.Resolve(*network, *subnet, *region, *project); err == nil {
```
zoran15, line 76: Minor: the first call that handles an error as well is doing `err != nil` while the last two here do `err == nil`. Not consistent and might lead to bugs in the future.

cli_tools/common/utils/param/populator_test.go
```diff
@@ -456,14 +556,21 @@ func TestPopulator_PopulateMissingParametersPopulatesStorageLocationWithScratchB
...
 		mockResourceLocationRetriever,
 		mockScratchBucketCreator,
-	).PopulateMissingParameters(&project, "gcloud", &zone, &region, &scratchBucketGcsPath, file, &storageLocation)
+	).PopulateMissingParameters(&project, "gcloud", &zone, &region, &scratchBucketGcsPath, file, &storageLocation, &network, &subnet)
 
 	assert.Nil(t, err)
 	assert.Equal(t, "US", storageLocation)
 }
+
+func newLenientNetworkResolver(ctrl *gomock.Controller) NetworkResolver {
```
zoran15, line 572: `lenient` -> `noOp` or similar? 

`lenient` sounds like it still does some resolving but ignores errors. This one, on the other hand, doesn't do anything.

cli_tools/common/utils/paramhelper/network.go
```diff
@@ -0,0 +1,62 @@
...
+
+// String formats the NetworkResource as a GCP-style resource identifier.
+func (r *NetworkResource) String() string {
+	if r.Name == "" {
+		return ""
+	}
+	if r.Project == "" {
+		return fmt.Sprintf("global/networks/%s", r.Name)
+	}
+	return fmt.Sprintf("projects/%s/global/networks/%s", r.Project, r.Name)
```
zoran15, line 43: minor: slight duplication here, up to you if you want to avoid it.

cli_tools/common/utils/paramhelper/subnet.go
```diff
@@ -0,0 +1,68 @@
...
+	if r.Name == "" {
+		return ""
+	}
+	if r.Region == "" {
+		return r.Name
+	}
+	if r.Project == "" {
+		return fmt.Sprintf("regions/%s/subnetworks/%s", r.Region, r.Name)
+	}
+	return fmt.Sprintf("projects/%s/regions/%s/subnetworks/%s", r.Project, r.Region, r.Name)
```
zoran15, line 40: slight duplication here as well

cli_tools/gce_onestep_image_import/onestep_importer/aws_args.go
```diff
@@ -96,8 +96,9 @@ func (args *awsImportArguments) validateAndPopulate(populator param.Populator) e
 		return err
 	}
 
+	networkNotUsed, subnetNotUsed := "", ""
```
zoran15, line 99: I missed the fact one step importer doesn't have the ability to specify custom networks. Maybe this should be added as a feature?

EricEdens: I wondered that too. Here's my reading: it allows specifying a custom network and subnet, however those parameters aren't used in this portion of the code, which has its own subset of parameters, dedicated to working with AWS.

Later, the `network` and `subnet` parameters are sent directly to `gce_vm_image_import` [here](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/8ba90bd33f6b8a56795e1d998a818cde9f0b31fe/cli_tools/gce_onestep_image_import/onestep_importer/utils.go#L110-L109).

cli_tools/common/utils/paramhelper/subnet.go
EricEdens: updated

cli_tools/common/utils/paramhelper/network.go
EricEdens: updated

cli_tools/common/utils/param/populator_test.go
EricEdens: good call - updated

cli_tools/common/utils/param/populator.go
EricEdens: I see -- I was matching the behavior at line 70, but I agree that its best to have all of the checks consistent in the function. Updated this conditional, and the one at line 70 to handle `err != nil` first.

cli_tools/common/utils/param/network_resolver_test.go
EricEdens: updated

cli_tools/common/utils/param/network_resolver.go
EricEdens: updated

cli_tools/common/utils/param/network_resolver.go
EricEdens: ahh good call -- one of my goals was to match the previous behavior. Updated to leave `subnet` empty.

cli_tools/common/utils/param/network_resolver.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+
+	"google.golang.org/api/compute/v1"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/paramhelper"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+	daisyCompute "github.com/GoogleCloudPlatform/compute-image-tools/daisy/compute"
+)
+
+// NetworkResolver standardizes and validates network and subnet fields. It follows the
```
dntczdx, line 27: you may add these info in the resolver code, otherwise it's a hidden logic for others:

When both are empty, explicitly use the default network.
If the subnet is empty, leave it empty to allow the compute backend to infer it.
Similarly, if only the network is empty, leave it empty to allow inference.
Backfill project and region resource properties if they are omitted in the network and subnet URIs.

zoran15 (approved):

cli_tools/common/utils/param/network_resolver.go
EricEdens: done!

dntczdx (approved):

PR #1703, opened 2021-07-27
[cli-tools] fix buffered writer err log prefix so it can be picked up by 'waitForInstanceSignals'

dntczdx:
Motivation: customer faced issue: the failed msg is only in serial log, not in gcloud console. Fixed by allowing setting proper err log prefix.

Also, the err match regex is changed from storage.object.create to to storage.object.* because a customer failed due to the lack of storage.object.get.

There will be another fix on gcloud: the permission issue was not caught by gcloud permission check.

### DIFF
--- a/cli_tools/common/utils/storage/buffered_writer_test.go
+++ b/cli_tools/common/utils/storage/buffered_writer_test.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


dntczdx (commented):
Needs to add unit test.

cli_tools/common/utils/storage/buffered_writer_test.go
```diff
@@ -47,7 +47,7 @@ func TestCreateNewChunkOnFirstWrite(t *testing.T) {
 
 	data := []byte("This is a sample data to write")
 
-	buf := NewBufferedWriter(ctx, bufferSize, workerNum, mockGcsClient, oauth, prefix, bkt, obj)
+	buf := NewBufferedWriter(ctx, bufferSize, workerNum, mockGcsClient, oauth, prefix, bkt, obj, "GCEExport")
```
EricEdens, line 50: The overview mentioned that you're changing the prefix: "Fixed by allowing setting proper err log prefix.", but here it looks like it's still `GCEExport`. Which prefix were you hoping to use?

dntczdx: "GCEExport:" will work. Current released version uses "[export]".

dntczdx: The key point is the line 128 above. For that logic branch, there is only a "log.Fatal(err)", which uses "[export]" as prefix.
There should be a fmt.Printf() to output the msg with "GCEExport:", just as what line 122 did.

dntczdx: Meanwhile, we also don't want to hard-code "GCEExport" as prefix in buffered_writer, because this lib is "common" and is also used by aws importer.

EricEdens (approved):

zoran15 (approved):

PR #1710, opened 2021-08-04
EL Import: Don't require yum

EricEdens:
A customer reported they had a CentOS 7 failed import where the machine didn't have `yum`, even though they used `--no-guest-environment`. This updates the EL logic to check whether `yum` is available prior to using it (for non-guest operations).

For testing, I created a CentOS 7 instance and then removed `yum`. I'm still brainstorming on how to make this an e2e test, since we typically depend on startup scripts. I'll leave that to a follow-up PR.

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate.py
+++ b/daisy_workflows/image_import/enterprise_linux/translate.py
@@ -127,6 +127,13 @@ def yum_is_on_path(spec: TranslateSpec) -> bool:
   return p.code == 0
 
 
+def package_is_installed(spec: TranslateSpec, package: str) -> bool:
+  """Check whether package is installed."""
+  p = run(spec.g, ['rpm', '-q', package], raiseOnError=False)
+  logging.debug('rpm -q: {}'.format(p))
+  return p.code == 0
+
+
 def check_yum_on_path(spec: TranslateSpec) -> str:
   """Check whether the `yum` command is available.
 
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_import/enterprise_linux/translate.py
```diff
@@ -289,7 +295,8 @@ def DistroSpecific(spec: TranslateSpec):
   # Remove NetworkManager-config-server if it's present. The package configures
   # NetworkManager to *not* use DHCP.
   #  https://access.redhat.com/solutions/894763
-  run(g, ['yum', 'remove', '-y', 'NetworkManager-config-server'])
+  if yum_is_on_path(spec):
+    run(g, ['yum', 'remove', '-y', 'NetworkManager-config-server'])
```
zoran15, line 299: Is there a scenario where `NetworkManager-config-server` is installed without yum? Can we check for it's existence without yum and remove it?

dntczdx: that is also my question - yum suppsoe to be the standard way to install/remove a package. However, shall we support the non-standard scenario?

EricEdens: We can use `rpm --erase NetworkManager-config-server` if `yum` isn't available. I'll update the test image to include this package and re-test.

My original intuition: Since the VM's creator removed `yum`, they intended to disable package modification. So if `NetworkManager-config-server` was present, they intentionally guarded against its removal. 

That said, the person importing the VM to GCP expects the system to be configured with DHCP.

zoran15 (approved):

daisy_workflows/image_import/enterprise_linux/translate.py
dntczdx: "Since the VM's creator removed yum, they intended to disable package modification. So if NetworkManager-config-server was present, they intentionally guarded against its removal.

That said, the person importing the VM to GCP expects the system to be configured with DHCP."

I don't know whether that assumption is correct. DId you discuss with Zach?

PR #1717, opened 2021-08-09
Remove workflow.RunWithModifiers

EricEdens:
This is part of our initiative to move daisy out of `compute-image-tools`. It removes the method `RunWithModifiers` from the `daisy.Workflow` struct, and updates its call sites to use the `Run` method.

### DIFF
--- a/cli_tools/daisycommon/daisy_worker.go
+++ b/cli_tools/daisycommon/daisy_worker.go
@@ -36,9 +36,9 @@ type DaisyWorker interface {
 // twice on the same instance.
 func NewDaisyWorker(wf *daisy.Workflow, env EnvironmentSettings, logger logging.Logger) DaisyWorker {
 	worker := &defaultDaisyWorker{wf, env, logger}
-	worker.env.ApplyToWorkflow(worker.wf)
-	worker.env.ApplyWorkerCustomizations(worker.wf)
-	daisy_utils.UpdateAllInstanceNoExternalIP(worker.wf, worker.env.NoExternalIP)
+	worker.env.ApplyToWorkflow(wf)
+	worker.env.ApplyWorkerCustomizations(wf)
+	daisy_utils.UpdateAllInstanceNoExternalIP(wf, env.NoExternalIP)
 	return worker
 }
 
--- a/cli_tools/daisycommon/daisy_worker.go
+++ b/cli_tools/daisycommon/daisy_worker.go
@@ -35,7 +35,11 @@ type DaisyWorker interface {
 // designed to be run once and discarded. In other words, don't run RunAndReadSerialValue
 // twice on the same instance.
 func NewDaisyWorker(wf *daisy.Workflow, env EnvironmentSettings, logger logging.Logger) DaisyWorker {
-	return &defaultDaisyWorker{wf, env, logger}
+	worker := &defaultDaisyWorker{wf, env, logger}
+	worker.env.ApplyToWorkflow(worker.wf)
+	worker.env.ApplyWorkerCustomizations(worker.wf)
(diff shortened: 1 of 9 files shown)


cli_tools/gce_ovf_export/exporter/instance_disks_exporter.go
```diff
@@ -155,10 +155,16 @@ func (ide *instanceDisksExporterImpl) addExportDisksSteps(w *daisy.Workflow, ins
 		if params.ComputeServiceAccount != "" {
 			varMap["compute_service_account"] = params.ComputeServiceAccount
 		}
+		p := path.Join(params.WorkflowDir, "/export/disk_export_ext.wf.json")
 		exportDiskStep.IncludeWorkflow = &daisy.IncludeWorkflow{
```
dntczdx, line 159: this is  a bit anti intuitive - feels like it's initialzied twice. would you wrap it so it looks like a single operation?

EricEdens: makes sense -- refactored into its own method.

dntczdx (approved):

cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -388,6 +335,54 @@ func (oi *OVFImporter) setUpImportWorkflow() (workflow *daisy.Workflow, err erro
...
+			return "gce-ovf-import-tmp"
+		},
+		DiskLabelKeyRetriever: func(disk *daisy.Disk) string {
+			return "gce-ovf-import-tmp"
+		},
+		ImageLabelKeyRetriever: func(imageName string) string {
+			return "gce-ovf-import-tmp"
+		}}
+	rl.LabelResources(workflow)
+	daisyutils.UpdateAllInstanceNoExternalIP(workflow, oi.params.NoExternalIP)
```
zoran15, line 381: These 3 steps (remove privacy hook from above, resource labeler and no external IP) are repeated in all operations (except no external IP for export since it doesn't need, but it would be a no-op). They also all operate on the workflow, even though for e.g. bootable disk processor, it's implemented as a processor's method (it still only operates on the workflow). To avoid code duplication, ideally we would have a master function/class that would do these things for all wrappers. This could be done in a followup PR if you prefer.

There's also applying UEFI compatible which doesn't make sense on exports, but it could be theoretically done by the same master function/class, which would skip over UEFI flag for exports.

zoran15 (approved):

cli_tools/daisycommon/daisy_worker.go
```diff
@@ -35,7 +35,11 @@ type DaisyWorker interface {
 // designed to be run once and discarded. In other words, don't run RunAndReadSerialValue
 // twice on the same instance.
 func NewDaisyWorker(wf *daisy.Workflow, env EnvironmentSettings, logger logging.Logger) DaisyWorker {
-	return &defaultDaisyWorker{wf, env, logger}
+	worker := &defaultDaisyWorker{wf, env, logger}
+	worker.env.ApplyToWorkflow(worker.wf)
+	worker.env.ApplyWorkerCustomizations(worker.wf)
+	daisy_utils.UpdateAllInstanceNoExternalIP(worker.wf, worker.env.NoExternalIP)
```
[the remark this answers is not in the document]
EricEdens: done!

cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
EricEdens: Opened  a bug internally

PR #1719, opened 2021-08-11
[daisy] fix output breakage for waitForInstanceSignal

dntczdx:
We detected broken <serial-output> line, which caused checksum mismatch for disk checksum verification.
After a serial of test, we realized that any serial output streamed to daisy-log may be broken, because each "ComputeClient.GetSerialPortOutput" can get a incomplete output result. Example:

Full output:
123456789
abcdefg
987654321

ComputeClient.GetSerialPortOutput call1 get:
123456789
abc

ComputeClient.GetSerialPortOutput call2 get:
defg
987654321

The possible reason is that, stdout output (like "echo") relies on a buffer, whose size is usually PIPE_BUF=512 bytes, which can't be modified. Once the buffer is full, it will be flushed, even though the whole line hasn't been written.

The checksum output is much longer than others (302 bytes), so it has much higher chance to hit this issue. However, some other output also hit his this during the test with a lower probability (example: target-size-gb).

To fix the root cause, we always check the last line of the ComputeClient.GetSerialPortOutput response. If it's without a "\n", we keep it as a tail string, so it can be concat with the next ComputeClient.GetSerialPortOutput response. Per my testing, it resolved this issue perfectly.

Notice: some other bugs may also be caused by this issue, for example, if the "Import Success" output is split to 2 lines, the instance may not stop. It will cause a import timeout ultimately.

### DIFF
--- a/daisy/step_wait_for_instances_signal.go
+++ b/daisy/step_wait_for_instances_signal.go
@@ -160,8 +160,8 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput, 
 				}
 
 				// If the content is not ended with a "\n", we want to store the last line as tail string, so it can be concat with the next block of content.
-				if i == len(lines)-1 && ln != "" {
-					tailString = lines[len(lines)-1]
+				if i == len(lines)-1 && lines[len(lines)-1] != "" {
+					tailString = ln
 					break
 				}
 
--- a/daisy/step_wait_for_instances_signal.go
+++ b/daisy/step_wait_for_instances_signal.go
@@ -121,6 +121,7 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput, 
 	w.LogStepInfo(s.name, "WaitForInstancesSignal", msg+".")
 	var start int64
 	var errs int
+	tailString := ""
 	tick := time.Tick(interval)
 	for {
 		select {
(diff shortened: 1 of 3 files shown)


daisy/step_wait_for_instances_signal.go
```diff
@@ -150,7 +151,20 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput,
...
 			}
 			start = resp.Next
-			for _, ln := range strings.Split(resp.Contents, "\n") {
+			lines := strings.Split(resp.Contents, "\n")
+			for i, ln := range lines {
+				// If there is a unconsumed tail string from the previous block of content, concat it with the 1st line of the new block of content.
+				if i == 0 && tailString != "" {
+					ln = tailString + ln
+					tailString = ""
+				}
```
EricEdens, line 160: We're doing this concatenation (and serial port reading) in the CreateInstance step. Thoughts on consolidating? 

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/4e113aa505000dbc40935f5f3b93f47241fe565f/daisy/step_create_instances.go#L98

daisy/step_wait_for_instances_signal.go
```diff
@@ -150,7 +151,20 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput,
...
+
+				// If the content is not ended with a "\n", we want to store the last line as tail string, so it can be concat with the next block of content.
+				if i == len(lines)-1 && ln != "" {
+					tailString = lines[len(lines)-1]
+					break
+				}
+
```
EricEdens, line 167: unit tests?

daisy/step_wait_for_instances_signal.go
```diff
@@ -150,7 +151,20 @@ func waitForSerialOutput(s *Step, project, zone, name string, so *SerialOutput,
...
-			for _, ln := range strings.Split(resp.Contents, "\n") {
+			lines := strings.Split(resp.Contents, "\n")
+			for i, ln := range lines {
+				// If there is a unconsumed tail string from the previous block of content, concat it with the 1st line of the new block of content.
+				if i == 0 && tailString != "" {
+					ln = tailString + ln
+					tailString = ""
+				}
+
+				// If the content is not ended with a "\n", we want to store the last line as tail string, so it can be concat with the next block of content.
```
zoran15, line 162: Where is the check that the last line doesn't end with "\n" performed? This `if` block seems to always store last line as `tailString`.

daisy/step_wait_for_instances_signal.go
zoran15: +1

daisy/step_wait_for_instances_signal.go
dntczdx: The lines[n] is from spliting by "\n".If there is a "\n" at the end, there will be always a blank entry as the last one. I tested that. So we jsut check ln == ""

daisy/step_wait_for_instances_signal.go
dntczdx: as discussed offline, it's unnecessary since serial log itself doesn't have any pattern match; also it's non-trivial to make a common loop function here since the actions inside the loop are quite different. Leave it as-is for now.

dntczdx:
added unit test

zoran15 (approved):

PR #1722, opened 2021-08-17
CLI Tools: Prepare daisy worker for wider usage

EricEdens:
This is part of our initiative to move daisy to its own repo.

This PR refactors `DaisyWorker` to support the following:
- Perform common operations used by all CLI tools:
   - Updating worker instances to support non-external IPs.
   - Applying the privacy log handler
   - Handle resource labeling
- Increases test coverage

To limit the blast radius of the change,  `DaisyWorker` is still only used for inspection. As a follow-up PR, I'll update existing call sites of `Daisy.Run` throughout the CLI tools to use the DaisyWorker facade.

Once existing call sites use this facade, then we can remove our "fallback to PD standard" modification in Daisy by implementing it here.


dntczdx:
It looks like "resource labeling" is added in some places - not simply refactoring but added new actions. Is that the case? Is that by design?

cli_tools/common/disk/inspect.go
```diff
@@ -60,14 +60,14 @@ func NewInspector(env daisycommon.EnvironmentSettings, logger logging.Logger) (I
 		logPrefix += "-"
 	}
 	wf.Name = logPrefix + "inspect"
-
-	return &bootInspector{daisycommon.NewDaisyWorker(wf, env, logger), logger}, nil
+	rl := daisyutils.NewResourceLabeler(wf.Name, env.ExecutionID, env.Labels, env.StorageLocation)
+	return &bootInspector{daisyutils.NewDaisyWorker(wf, env, logger, rl), logger}, nil
```
EricEdens: @dntczdx 
> It looks like "resource labeling" is added in some places - not simply refactoring but added new actions. Is that the case? Is that by design?

Thanks, that's a great call-out: I removed that here since you're right I didn't intend to add new features.

dntczdx (approved):

cli_tools/common/utils/daisyutils/daisy_utils.go
```diff
@@ -429,3 +397,96 @@ func GetDiskURI(project, zone, name string) string {
...
+		w.DisableCloudLogging()
+	}
+	if env.DisableStdoutLogs {
+		w.DisableStdoutLogging()
+	}
+}
+
+// UpdateAllInstanceNoExternalIP updates all Create Instance steps in the workflow to operate
+// when no external IP access is allowed by the VPC Daisy workflow is running in.
+func UpdateAllInstanceNoExternalIP(workflow *daisy.Workflow, noExternalIP bool) {
```
zoran15, line 487: Why is no external IP update done from here, but also in NewDaisyWorker?

Btw, it's a bit hard to follow such a large PR. When a lot of code is renamed, moved, it's a better idea to do that step in one PR then add/change functionality in another.

cli_tools/common/utils/daisyutils/traversal.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// To rebuild the mock, run `go generate ./...`
+//go:generate go run github.com/golang/mock/mockgen -package mocks -source $GOFILE -destination ../../../mocks/mock_workflow_traversal.go
+
+// WorkflowTraversal allows for modifying and validating a daisy workflow.
+type WorkflowTraversal interface {
+	Traverse(wf *daisy.Workflow) error
+}
```
zoran15, line 27: Most of the implementations are not really traversals, e.g. `ApplyAndValidateVars` and `ConfigureDaisyLogging` which is confusing.

`WorkflowModifier`, `WorkflowUpdater` or something similar would be a more appropriate name.

cli_tools/common/utils/daisyutils/daisy_utils.go
EricEdens: Good question -- that was for refactoring -- the code from `UpdateAllInstanceNoExternalIP` was copied into `RemoveExternalIPTraversal`, and then `UpdateAllInstanceNoExternalIP` called `RemoveExternalIPTraversal`.

Also, in retrospect, I agree this would have been clearer with the split you suggested. Sorry about that; I'll do better in the future!

cli_tools/common/utils/daisyutils/traversal.go
EricEdens: sgtm -- Updated to `WorkflowModifier`.

zoran15 (approved):

PR #1438, opened 2020-11-18
cli-tools: Switch back to Debian 9 worker

EricEdens:
Rollback of #1431 and #1432, which caused a regression for EL and SLES imports: 
 - https://prow.k8s.io/view/gcs/compute-image-tools-test/logs/ci-images-import-export-cli-e2e-tests/1328876340400623616
 - https://prow.k8s.io/view/gcs/compute-image-tools-test/logs/ci-daisy-e2e/1328882631923732480

PR #1726, opened 2021-08-19
CLI Tools: Use deb10 worker for EL8 import and inspection

EricEdens:
During import, a disk using LVM is failing to mount with guestfs 1.34, but successfully mounts with guestfs 1.40. 

We previously attempted to use the Debian 10 worker in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1438; at that time 

Ultimately we'll need to migrate off Debian 9 completely. This change is a small step in that direction, and unblocks the user who is blocked on their import. 

## Testing

- Ran the daisy e2e tests that had previously failed.

### DIFF
--- a/daisy_workflows/image_import/enterprise_linux/translate_centos_8.wf.json
+++ b/daisy_workflows/image_import/enterprise_linux/translate_centos_8.wf.json
@@ -43,7 +43,7 @@
       "CreateDisks": [
         {
           "Name": "disk-translator",
-          "SourceImage": "projects/compute-image-tools/global/images/family/debian-9-worker",
+          "SourceImage": "projects/compute-image-tools/global/images/family/debian-10-worker",
           "SizeGb": "10",
           "Type": "pd-ssd",
           "FallbackToPdStandard": true
--- a/daisy_workflows/image_import/inspection/boot-inspect.wf.json
+++ b/daisy_workflows/image_import/inspection/boot-inspect.wf.json
@@ -4,7 +4,8 @@
   "Vars": {
     "pd_uri": {
       "Required": true,
-      "Description": "PD that will be inspected (URI)."
+      "Description": "PD that will be inspected (URI).",
+      "Value": "projects/edens-test/zones/us-west1-b/disks/b194189632"
     },
     "network": {
       "Value": "global/networks/default",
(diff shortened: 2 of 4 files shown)


daisy_workflows/image_import/inspection/boot-inspect.wf.json
```diff
@@ -4,7 +4,8 @@
   "Vars": {
     "pd_uri": {
       "Required": true,
-      "Description": "PD that will be inspected (URI)."
+      "Description": "PD that will be inspected (URI).",
+      "Value": "projects/edens-test/zones/us-west1-b/disks/b194189632"
```
zoran15, line 8: Leftover from local testing?

zoran15 (commented):
Will this not cause the regression that we had in last year when we reverted back to Debian 9?

dntczdx (approved):
LGTM except for Zoran's finding

daisy_workflows/image_import/inspection/boot-inspect.wf.json
EricEdens: shoot, thanks for finding that. Removed!

daisy_workflows/image_import/enterprise_linux/translate_centos_8.wf.json
```diff
@@ -43,7 +43,7 @@
       "CreateDisks": [
         {
           "Name": "disk-translator",
-          "SourceImage": "projects/compute-image-tools/global/images/family/debian-9-worker",
+          "SourceImage": "projects/compute-image-tools/global/images/family/debian-10-worker",
```
EricEdens, line 46: @zoran15 
> Will this not cause the regression that we had in last year when we reverted back to Debian 9?

Good question! I re-ran the affected tests and couldn't repro the issue, so my *guess* is that a package update in the Debian 10 repos fixed the issue.

The previous error while running a shell process for `virt-customize`: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/3a6d1ada541e4f53cb7c92dfdc67c120e81228dd/daisy_workflows/image_import/enterprise_linux/translate.py#L358         

(#1438 has links to the failing e2e tests)

zoran15 (approved):

PR #1728, opened 2021-08-23
Add options field to PopulateClients

Quintonamore:
Added an options field to PopulateClients for greater control over the options passed to the compute, storage, and logging clients. Updated gce_image_publish library and cli tool to pass the new field.

### DIFF
--- a/daisy/workflow.go
+++ b/daisy/workflow.go
@@ -236,7 +236,7 @@ func (w *Workflow) SetLogProcessHook(hook func(string) string) {
 
 // Validate runs validation on the workflow.
 func (w *Workflow) Validate(ctx context.Context) DError {
-	if err := w.PopulateClients(ctx); err != nil {
+	if err := w.PopulateClients(ctx, []option.ClientOption{}); err != nil {
 		w.CancelWorkflow()
 		return Errf("error populating workflow: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


daisy/workflow_test.go
```diff
@@ -1092,54 +1092,55 @@ func TestRunStepTimeout(t *testing.T) {
...
-	tryPopulateClients(t, w)
+	tryPopulateClients(t, w, options)
 	if w.cloudLoggingClient == nil {
 		t.Errorf("Did not populate Cloud Logging client.")
 	}
 }
 
-func tryPopulateClients(t *testing.T, w *Workflow) {
-	if err := w.PopulateClients(context.Background()); err != nil {
+func tryPopulateClients(t *testing.T, w *Workflow, options []option.ClientOption) {
```
bkatyl, line 1142: Can you add a test that uses addition of this new options field to exercise that what gets passed is correctly set in the computeOptions, storageOptions, loggingOptions?

daisy/workflow.go
```diff
@@ -377,11 +377,25 @@ func (w *Workflow) getSourceGCSAPIPath(s string) string {
 }
 
 // PopulateClients populates the compute and storage clients for the workflow.
-func (w *Workflow) PopulateClients(ctx context.Context) error {
+func (w *Workflow) PopulateClients(ctx context.Context, options []option.ClientOption) error {
```
EricEdens, line 380: You might consider `PopulateClients(ctx context.Context, options ... option.ClientOption)` The type of options is still []option.ClientOption:

https://yourbasic.org/golang/three-dots-ellipsis/

Issue #1736, opened 2021-08-26
`import_precheck` fails on CentOS 7


charlesomer:
Fails since `lsblk` on CentOS 7 is version `2.23.2` which I believe is quite old and doesn't support `--json`.

`* WARN: Failed to execute lsblk: lsblk: exit status 1, stderr: lsblk: unrecognized option '--json'`

Works fine on CentOS 8 which uses `lsblk` version `2.32.1`

PR #1740, opened 2021-09-01
Import precheck: Refactor to allow mocking

EricEdens:
This refactors `cli_tools/import_precheck` to achieve the following:

1. Move implementation to a different package than `main`, to allow gomock to be used.
2. Add comments and documentation
3. Change the copyright formatting to match the rest of  `compute-image-tools`

This works toward fixing https://github.com/GoogleCloudPlatform/compute-image-tools/issues/1736


cli_tools/import_precheck/precheck/check.go
```diff
@@ -0,0 +1,24 @@
+//  Copyright 2017 Google Inc. All Rights Reserved.
```
zoran15, line 1: minor: 2021. For other new files as well, unless they were just moved.

zoran15 (approved):

cli_tools/import_precheck/precheck/check.go
EricEdens: No new files in this PR -- this was moved from `cli_tools/import_precheck/check.go`, for example.

PR #1742, opened 2021-09-02
Don't install the auto updater package on windows import

adjackura:


zoran15 (commented):
Why are we removing this now and why did we have this step in the first place?

adjackura:
> Why are we removing this now and why did we have this step in the first place?

We stopped including auto updates in images a while ago, apparently we missed updating the importers.

PR #1751, opened 2021-09-14
[rollout] new rollout config

dntczdx:
This is a PR that added new cloudbuild config for new rollout mechanism.
cli_tools_cloudbuild_build.yaml is copied from cli_tools_cloudbuild.yaml but removed those pushing commands to "artifacts registry".
cli_tools_cloudbuild_deploy.yaml responses for pushing to "artifacts registry".
cli_tools_cloudbuild_release_workflow.yaml is new, which simply launches the new rollout workflow.


zoran15 (commented):
Can we have these files in their own subfolder to avoid confusion with the old build YAML file? Also, to have them in one place since it's not one file anymore.

dntczdx:
> Can we have these files in their own subfolder to avoid confusion with the old build YAML file? Also, to have them in one place since it's not one file anymore.

sounds good, moved to "rollout"

zoran15 (approved):

EricEdens (approved):

PR #1754, opened 2021-09-22
Image Import: Use daisy worker

EricEdens:
This updates image import (and therefore a portion of OVF import) to use the DaisyWorker facade. Specifically:

1. Inflating the disk using qemu-img
2. Calculating the checksum of the API-inflated disk
3. Translating the disk

Disk inspection was already using the DaisyWorker facade.

### DIFF
--- a/cli_tools/common/utils/daisyutils/daisy_utils.go
+++ b/cli_tools/common/utils/daisyutils/daisy_utils.go
@@ -438,6 +438,15 @@ Loop:
 	return w, nil
 }
 
+// Tool is used to communicate the tool's name ot the user.
+type Tool struct {
+
+	// HumanReadableName is used for error messages, for example: "image import".
+	HumanReadableName string
+	// URISafeName is used programmatically, eg: "image-import"
+	URISafeName string
+}
+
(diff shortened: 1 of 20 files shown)


cli_tools/common/utils/daisyutils/daisy_utils.go
```diff
@@ -438,6 +438,15 @@ Loop:
 	return w, nil
 }
 
+// Tool is used to communicate the tool's name ot the user.
+type Tool struct {
+
+	// HumanReadableName is used for error messages, for example: "image import".
+	HumanReadableName string
+	// URISafeName is used programmatically, eg: "image-import"
+	URISafeName string
```
zoran15, line 447: How is this name being URI safe and being used programmatically related?

EricEdens: Updated to `ResourceLabelName`

The original intention was to have one for programmatic use, and one for putting in logs, with URL safe being an indicator of which characters to use. That said, it's probably clearest to document exactly where it will be used, as with this new name.

zoran15 (approved):

PR #1759, opened 2021-09-30
Windows upgrade: Use daisy worker

EricEdens:
This changes windows upgrade to use the `DaisyWorker` facade. For testing, I ran the the tests in gce_windows_upgrade.Dockerfile.

### DIFF
--- a/cli_tools/gce_windows_upgrade/upgrader/upgrader.go
+++ b/cli_tools/gce_windows_upgrade/upgrader/upgrader.go
@@ -18,12 +18,17 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
+
+	"google.golang.org/api/option"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/compute"
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/daisyutils"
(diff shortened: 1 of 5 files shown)


cli_tools/gce_windows_upgrade/upgrader/upgrader.go
```diff
@@ -103,11 +109,22 @@ type upgrader struct {
...
-	u := upgrader{InputParams: p}
+	u := upgrader{
+		InputParams: p,
+		logger:      logging.NewToolLogger(logPrefix),
+		derivedVars: &derivedVars{
+			executionID: os.Getenv(daisyutils.BuildIDOSEnvVarName),
+		},
+	}
+	if u.executionID == "" {
+		u.executionID = path.RandString(5)
```
dntczdx, line 126: I wonder why it doesn't have this exec id and why now we need it

zoran15 (approved):

cli_tools/gce_windows_upgrade/upgrader/upgrader.go
EricEdens: The executionID is used when labeling resources (eg: temporary instances and disks):

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/common/utils/daisyutils/daisy_worker.go#L60

When running on cloud build, it will use the BUILD_ID environment variable. If it's running outside cloud build, that environment variable may be empty, so it generates a new one.

PR #1763, opened 2021-10-06
Ubuntu import: When possible, don't install cloud-init 21.3-1

EricEdens:
This updates Ubuntu translate to avoid installing cloud-init 21.3-1, which blocks systemd from starting the guest agent on first boot. 

It achieves this by querying `apt` to find the available package versions, and picking a version such that 21.3-1 is avoided if possible. If a version other than 21.3-1 is found, this falls back to the previous behavior of allowing apt to choose which version to install. In this case, the guest agent will start after the first boot, which is the current behavior.

Testing:
- Ran Ubuntu e2e tests.

### DIFF
--- a/daisy_workflows/image_import/ubuntu/translate.py
+++ b/daisy_workflows/image_import/ubuntu/translate.py
@@ -27,6 +27,7 @@ import re
 
 import guestfs
 import utils
+from utils.apt import Apt
 import utils.diskutils as diskutils
 from utils.guestfsprocess import run
 
@@ -89,7 +90,6 @@ system_info:
          primary: http://ports.ubuntu.com/ubuntu-ports
          security: http://ports.ubuntu.com/ubuntu-ports
 '''
-
 
 # Network configs
 #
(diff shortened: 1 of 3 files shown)


daisy_workflows/image_import/ubuntu/translate.py
```diff
@@ -203,6 +203,44 @@ def install_osconfig_agent(g: guestfs.GuestFS):
...
+  if version_to_install:
+    pkg_to_install = 'cloud-init=' + version_to_install
+  elif curr_version == '':
+    pkg_to_install = 'cloud-init'
+  # If this block doesn't execute, it means that cloud-init was found
+  # on the system, but there wasn't an upgrade candidate. Therefore
+  # leave the version that's currently installed.
+  if pkg_to_install:
+    logging.info(pkg_to_install)
+    utils.install_apt_packages(g, pkg_to_install)
```
zoran15, line 228: Couldn't this possibly still install 21.3-1, since version is not specified?

daisy_workflows/linux_common/utils/apt.py
```diff
@@ -0,0 +1,94 @@
...
+  @staticmethod
+  def determine_version_to_install(current_version: str,
+                                   available_versions: typing.Set[str],
+                                   blocked_versions: typing.Set[str]) -> str:
+    non_blocked_versions = set()
+    for v in available_versions:
+      blocked = False
+      for b in blocked_versions:
+        if v.startswith(b):
+          blocked = True
```
zoran15, line 39: Break out of the loop early?

daisy_workflows/linux_common/utils/apt.py
```diff
@@ -0,0 +1,94 @@
...
+      if not blocked:
+        non_blocked_versions.add(v)
+    candidate = ''
+    for v in non_blocked_versions:
+      if v > current_version and v > candidate:
+        candidate = v
```
zoran15, line 45: Here as well

daisy_workflows/linux_common/utils/apt.py
EricEdens: good call, done!

zoran15 (approved):

PR #1766, opened 2021-10-11
Support RHEL 7.9 for SAP.

zmarano:
RHEL 7.9 for SAP has its own unique configuration because it does not use E4S (Update Services) content and does not use the releasever yum variable to peg to a RHEL point release. Instead, it uses the regular `7Server` release but has a different set of content in the client configuration because it is an SAP SKU.


hopkiw:
looks like it will work
but i think this makes it official that we should move to a flat file system and away from generation, even though the changes are all small

PR #1768, opened 2021-10-12
CLI Tools Refactor: Allow daisy worker modifiers to run after workflow

EricEdens:
Prior to this, DaisyWorker was extensible using a **modifier** which ran before the workflow. This PR changes the **modifier** to a **hook**, allowing extensibility before and after the workflow runs.

This is required to support fallback-to-PD-standard; if the workflow fails, a post-run hook will look at the error to determine whether it failed due to SSD quota issue. If so, the hook signals that it wants to re-run the workflow. On the next invocation of the workflow, the corresponding pre-run hook updates the workflow to convert SSD disks to standard PD disks.

The post-run hook will also be used


cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -62,19 +62,19 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, modifiers []Workflo
...
 
 // Run runs the daisy workflow with the supplied vars.
 func (w *defaultDaisyWorker) Run(vars map[string]string) error {
-	if err := (&ApplyAndValidateVars{w.env, vars}).Modify(w.wf); err != nil {
+	if err := (&ApplyAndValidateVars{w.env, vars}).PreRunHook(w.wf); err != nil {
 		return err
 	}
-	for _, t := range w.modifiers {
-		if err := t.Modify(w.wf); err != nil {
+	for _, t := range w.hooks {
```
zoran15, line 76: What does t stand for here? Not sure what it even stood for when modifiers were used tbh.

zoran15 (commented):
It seems all workflow hooks are preRun, and postRuns are no-op. Maybe consider having a single operation for each hook (e.g. just `Run(w)`) and register them as either pre- or post- run.

cli_tools/common/utils/daisyutils/workflow_hook.go
```diff
@@ -19,9 +19,18 @@ import (
...
 
-// WorkflowModifier allows for modifying and validating a daisy workflow.
-type WorkflowModifier interface {
-	Modify(wf *daisy.Workflow) error
+// WorkflowPreHook executes before the workflow runs.
+type WorkflowPreHook interface {
+	// PreRunHook allows a WorkflowHook to modify a workflow prior to running.
+	PreRunHook(wf *daisy.Workflow) error
+}
+
```
EricEdens, line 29: @zoran15 

> It seems all workflow hooks are preRun, and postRuns are no-op. Maybe consider having a single operation for each hook (e.g. just Run(w)) and register them as either pre- or post- run.

Good call! Updated to split the interface so that implementers don't have to create no-op functions.

zoran15: I think this looks good. Going a step further, have you considered having a single interface for hooks/modifiers (e.g. WorkflowHook) and same hook could be registered as either pre- or post- workflow hook?

zoran15: Yeah, the interface looks different enough that it doesn't make sense to merge them into one, as some args would be ignored in different scenarios.

zoran15 (approved):

PR #1777, opened 2021-10-26
CLI Tools: Allow workflows to be re-run

EricEdens:
This PR updates DaisyWorker to allow workflows to be retried. To achieve this, clients of DaisyWorker supply a closure that *generates* a workflow. On the first invocation, and any retries, this factory is used to create a new workflow. This is required since daisy workflows aren't designed to be run multiple times.

This PR is the first half of removing the fallback-to-PD-standard feature from daisy. The second PR adds a hook that implements the fallback-to-PD-standard behavior.

Testing
* Ran a subset of the OVF and image import e2e tests locally.

### DIFF
--- a/cli_tools/common/image/importer/bootable_disk_processor.go
+++ b/cli_tools/common/image/importer/bootable_disk_processor.go
@@ -65,12 +65,10 @@ func newBootableDiskProcessor(request ImageImportRequest, wfPath string, logger 
 		vars["compute_service_account"] = request.ComputeServiceAccount
 	}
 
-	workflow, err := daisyutils.ParseWorkflow(wfPath, vars,
-		request.Project, request.Zone, request.ScratchBucketGcsPath, request.Oauth, request.Timeout.String(),
-		request.ComputeEndpoint, request.GcsLogsDisabled, request.CloudLogsDisabled, request.StdoutLogsDisabled)
-
-	if err != nil {
-		return nil, err
+	workflowProvider := func() (*daisy.Workflow, error) {
+		return daisyutils.ParseWorkflow(wfPath, vars,
+			request.Project, request.Zone, request.ScratchBucketGcsPath, request.Oauth, request.Timeout.String(),
+			request.ComputeEndpoint, request.GcsLogsDisabled, request.CloudLogsDisabled, request.StdoutLogsDisabled)
 	}
 
 	env := request.EnvironmentSettings()
@@ -80,12 +78,12 @@ func newBootableDiskProcessor(request ImageImportRequest, wfPath string, logger 
 	env.DaisyLogLinePrefix += "translate"
 	diskProcessor := &bootableDiskProcessor{
 		request:    request,
-		worker:     daisyutils.NewDaisyWorker(workflow, env, logger, createResourceLabeler(request)),
+		worker:     daisyutils.NewDaisyWorker(workflowProvider, env, logger, createResourceLabeler(request)),
 		logger:     logger,
 		detectedOs: detectedOs,
 		vars:       vars,
 	}
-	return diskProcessor, err
+	return diskProcessor, nil
 }
 
 func createResourceLabeler(request ImageImportRequest) *daisyutils.ResourceLabeler {
--- a/cli_tools/common/image/importer/bootable_disk_processor.go
+++ b/cli_tools/common/image/importer/bootable_disk_processor.go
@@ -50,7 +50,7 @@ func (b *bootableDiskProcessor) cancel(reason string) bool {
 	return true
 }
 
-func newBootableDiskProcessor(request ImageImportRequest, wfPath string, logger logging.Logger, detectedOs distro.Release) (processor, error) {
+func newBootableDiskProcessor(request ImageImportRequest, wfPath string, logger logging.Logger, detectedOs distro.Release) processor {
 	vars := map[string]string{
 		"image_name":           request.ImageName,
 		"install_gce_packages": strconv.FormatBool(!request.NoGuestEnvironment),
@@ -83,7 +83,7 @@ func newBootableDiskProcessor(request ImageImportRequest, wfPath string, logger 
 		detectedOs: detectedOs,
 		vars:       vars,
 	}
-	return diskProcessor, nil
+	return diskProcessor
 }
 
 func createResourceLabeler(request ImageImportRequest) *daisyutils.ResourceLabeler {
--- a/cli_tools/common/image/importer/daisy_inflater.go
+++ b/cli_tools/common/image/importer/daisy_inflater.go
@@ -109,9 +109,6 @@ func newDaisyInflater(request ImageImportRequest, fileInspector imagefile.Inspec
 		if err != nil {
 			return nil, err
 		}
-		// If there's a failure during inflation, remove the PD that would otherwise
-		// be left for translation.
-		wf.ForceCleanupOnError = true
 		if request.UefiCompatible {
 			addFeatureToDisk(wf, "UEFI_COMPATIBLE", inflationDiskIndex)
 		}
--- a/cli_tools/common/image/importer/daisy_inflater.go
+++ b/cli_tools/common/image/importer/daisy_inflater.go
@@ -101,18 +101,24 @@ func newDaisyInflater(request ImageImportRequest, fileInspector imagefile.Inspec
 		inflationDiskIndex = 1 // First disk is for the worker
 	}
 
-	wf, err := daisyutils.ParseWorkflow(path.Join(request.WorkflowDir, wfPath), vars,
-		request.Project, request.Zone, request.ScratchBucketGcsPath, request.Oauth, request.Timeout.String(), request.ComputeEndpoint,
-		request.GcsLogsDisabled, request.CloudLogsDisabled, request.StdoutLogsDisabled)
-	if err != nil {
-		return nil, err
-	}
+	workflowProvider := func() (*daisy.Workflow, error) {
+		wf, err := daisyutils.ParseWorkflow(path.Join(request.WorkflowDir, wfPath), vars,
+			request.Project, request.Zone, request.ScratchBucketGcsPath, request.Oauth, request.Timeout.String(), request.ComputeEndpoint,
+			request.GcsLogsDisabled, request.CloudLogsDisabled, request.StdoutLogsDisabled)
 
-	if request.UefiCompatible {
-		addFeatureToDisk(wf, "UEFI_COMPATIBLE", inflationDiskIndex)
-	}
-	if strings.Contains(request.OS, "windows") {
-		addFeatureToDisk(wf, "WINDOWS", inflationDiskIndex)
+		if err != nil {
+			return nil, err
+		}
+		// If there's a failure during inflation, remove the PD that would otherwise
+		// be left for translation.
+		wf.ForceCleanupOnError = true
+		if request.UefiCompatible {
+			addFeatureToDisk(wf, "UEFI_COMPATIBLE", inflationDiskIndex)
+		}
+		if strings.Contains(request.OS, "windows") {
+			addFeatureToDisk(wf, "WINDOWS", inflationDiskIndex)
+		}
+		return wf, err
 	}
 
 	env := request.EnvironmentSettings()
@@ -121,7 +127,7 @@ func newDaisyInflater(request ImageImportRequest, fileInspector imagefile.Inspec
 	}
 	env.DaisyLogLinePrefix += "inflate"
 	return &daisyInflater{
-		worker:          daisyutils.NewDaisyWorker(wf, env, logger),
+		worker:          daisyutils.NewDaisyWorker(workflowProvider, env, logger),
 		inflatedDiskURI: fmt.Sprintf("zones/%s/disks/%s", request.Zone, diskName),
 		logger:          logger,
 		source:          request.Source,
--- a/cli_tools/common/utils/daisyutils/daisy_utils.go
+++ b/cli_tools/common/utils/daisyutils/daisy_utils.go
@@ -366,6 +366,8 @@ func RunWorkflowWithCancelSignal(w *daisy.Workflow, cancel <-chan string) error 
 		case reason := <-cancel:
 			if reason != "" {
 				w.CancelWithReason(reason)
+			} else {
+				w.CancelWorkflow()
 			}
 			break
 		case <-c:
@@ -375,6 +377,9 @@ func RunWorkflowWithCancelSignal(w *daisy.Workflow, cancel <-chan string) error 
 		case <-w.Cancel:
 		}
 	}(w)
+	// Daisy doesn't support cancellation through context; if the context that's passed in
+	// is cancelled, then all of its clients die, causing confusing errors and resources
+	// being left that should have been cleaned up.
 	return w.Run(context.Background())
 }
 
--- a/cli_tools/common/utils/daisyutils/daisy_utils.go
+++ b/cli_tools/common/utils/daisyutils/daisy_utils.go
@@ -355,19 +355,27 @@ func PostProcessDErrorForNetworkFlag(action string, err error, network string, w
 	}
 }
 
-// RunWorkflowWithCancelSignal runs Daisy workflow with accepting Ctrl-C signal
-func RunWorkflowWithCancelSignal(ctx context.Context, w *daisy.Workflow) error {
+// RunWorkflowWithCancelSignal runs a Daisy workflow, and allows for cancellation from two sources:
+//   1. The user types Ctrl-C on their keyboard.
+//   2. The caller sends a cancellation reason on the cancel channel (or closes it).
+func RunWorkflowWithCancelSignal(w *daisy.Workflow, cancel <-chan string) error {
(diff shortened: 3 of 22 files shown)


cli_tools/gce_windows_upgrade/upgrader/workflows.go
```diff
@@ -524,7 +528,8 @@ func (u *upgrader) runWorkflowWithSteps(workflowName string, timeout string, pop
 		},
 	}
 
-	return w, daisyutils.NewDaisyWorker(w, env, u.logger).Run(map[string]string{})
+	err := daisyutils.NewDaisyWorker(workflowProvider, env, u.logger).Run(map[string]string{})
+	return wf, err
```
zoran15, line 532: I find the flow around `wf` convoluted: `NewDaisyWorker` will use `workflowProvider` to build a workflow but `wf` is modified by `workflowProvider` closure and this relies on the fact `NewDaisyWorker` will be called. 

Can daisyWorker store the workflow that was built to avoid this complexity?

cli_tools/common/image/importer/bootable_disk_processor.go
```diff
@@ -80,12 +78,12 @@ func newBootableDiskProcessor(request ImageImportRequest, wfPath string, logger
...
 	diskProcessor := &bootableDiskProcessor{
 		request:    request,
-		worker:     daisyutils.NewDaisyWorker(workflow, env, logger, createResourceLabeler(request)),
+		worker:     daisyutils.NewDaisyWorker(workflowProvider, env, logger, createResourceLabeler(request)),
 		logger:     logger,
 		detectedOs: detectedOs,
 		vars:       vars,
 	}
-	return diskProcessor, err
+	return diskProcessor, nil
```
zoran15, line 86: If error returned is always `nil`, remove error from `newBootableDiskProcessor` return value.

cli_tools/common/image/importer/daisy_inflater.go
```diff
@@ -101,18 +101,24 @@ func newDaisyInflater(request ImageImportRequest, fileInspector imagefile.Inspec
...
-		addFeatureToDisk(wf, "UEFI_COMPATIBLE", inflationDiskIndex)
-	}
-	if strings.Contains(request.OS, "windows") {
-		addFeatureToDisk(wf, "WINDOWS", inflationDiskIndex)
+		if err != nil {
+			return nil, err
+		}
+		// If there's a failure during inflation, remove the PD that would otherwise
+		// be left for translation.
+		wf.ForceCleanupOnError = true
```
zoran15, line 114: This seems like new logic being added within a refactor.

cli_tools/common/utils/daisyutils/daisy_utils.go
```diff
@@ -355,19 +355,27 @@ func PostProcessDErrorForNetworkFlag(action string, err error, network string, w
...
+//   2. The caller sends a cancellation reason on the cancel channel (or closes it).
+func RunWorkflowWithCancelSignal(w *daisy.Workflow, cancel <-chan string) error {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 	go func(w *daisy.Workflow) {
 		select {
+		case reason := <-cancel:
+			if reason != "" {
+				w.CancelWithReason(reason)
+			}
```
zoran15, line 369: What is reason is not provided? Should we call `w.CancelWorkflow()`?

cli_tools/common/utils/daisyutils/daisy_utils.go
```diff
@@ -355,19 +355,27 @@ func PostProcessDErrorForNetworkFlag(action string, err error, network string, w
...
+			break
 		case <-c:
 			w.LogWorkflowInfo("\nCtrl-C caught, sending cancel signal to %q...\n", w.Name)
 			w.CancelWorkflow()
+			break
 		case <-w.Cancel:
 		}
 	}(w)
-	return w.Run(ctx)
+	return w.Run(context.Background())
```
zoran15, line 378: No need to reuse the context here?

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -34,12 +35,15 @@ type DaisyWorker interface {
 	Cancel(reason string) bool
 }
 
+// WorkflowProvider should return a new instance of a daisy workflow every time it is called.
```
zoran15, line 38:
```suggestion
// WorkflowProvider returns a new instance of a Daisy workflow.
```

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -74,35 +78,93 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, hooks []interface{}
...
-func (w *defaultDaisyWorker) Run(vars map[string]string) error {
-	if err := (&ApplyAndValidateVars{w.env, vars}).PreRunHook(w.wf); err != nil {
-		return err
+func (w *defaultDaisyWorker) Run(vars map[string]string) (err error) {
+	var wf *daisy.Workflow
+	for attempts := 0; attempts <= 1; attempts++ {
+		if wf, err = w.workflowProvider(); err != nil {
+			break
+		}
+		if err = w.checkIfCancelled(wf); err != nil {
```
zoran15, line 98: Why would a workflow be cancelled after just being created?

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -74,35 +78,93 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, hooks []interface{}
...
 }
 
 // Run runs the daisy workflow with the supplied vars.
-func (w *defaultDaisyWorker) Run(vars map[string]string) error {
-	if err := (&ApplyAndValidateVars{w.env, vars}).PreRunHook(w.wf); err != nil {
-		return err
+func (w *defaultDaisyWorker) Run(vars map[string]string) (err error) {
+	var wf *daisy.Workflow
+	for attempts := 0; attempts <= 1; attempts++ {
+		if wf, err = w.workflowProvider(); err != nil {
```
zoran15, line 95: How will this work for e.g. fallback to pd-standard scenario? Wouldn't workflowProvider need to know it's a retry attempt and generate a slighly different workflow for the retry? For that, it would require some kind of an arg (e.g. `attemp`.

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -74,35 +78,93 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, hooks []interface{}
...
+	cancelGuard sync.Once
 }
 
 // Run runs the daisy workflow with the supplied vars.
-func (w *defaultDaisyWorker) Run(vars map[string]string) error {
-	if err := (&ApplyAndValidateVars{w.env, vars}).PreRunHook(w.wf); err != nil {
-		return err
+func (w *defaultDaisyWorker) Run(vars map[string]string) (err error) {
+	var wf *daisy.Workflow
+	for attempts := 0; attempts <= 1; attempts++ {
```
zoran15, line 94: Should attempt count limit be configurable? I assume one retry attempt is enough for use cases we currently handle, but, since you're anyway adding the concept of retry, it might make sense to make it configurable and default to 2 attempts.

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -74,35 +78,93 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, hooks []interface{}
...
+		}
+		if err = w.checkIfCancelled(wf); err != nil {
+			break
+		}
+		var retryRequested bool
+		retryRequested, err = w.runOnce(wf, vars)
+		if err == nil || !retryRequested {
+			break
+		}
+		w.logger.Debug("Retry requested. err=" + err.Error())
```
zoran15, line 106: If `err` is not `nil`, but `retryRequested` is `false`, the execution will still end up here, so printing _Retry requested_ is not correct.

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -74,35 +78,93 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, hooks []interface{}
 }
 
 type defaultDaisyWorker struct {
-	wf     *daisy.Workflow
-	logger logging.Logger
-	env    EnvironmentSettings
-	hooks  []interface{}
+	workflowProvider WorkflowProvider
+	finished         *daisy.Workflow
```
zoran15, line 82: `finishedWf` ?

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -74,35 +78,93 @@ func createResourceLabelerIfMissing(env EnvironmentSettings, hooks []interface{}
...
 	if err != nil {
-		PostProcessDErrorForNetworkFlag(w.env.Tool.HumanReadableName, err, w.env.Network, w.wf)
+		PostProcessDErrorForNetworkFlag(w.env.Tool.HumanReadableName, err, w.env.Network, wf)
+		for _, hook := range w.hooks {
+			preHook, isPreHook := hook.(WorkflowPostHook)
+			if isPreHook {
+				wantRetry := false
+				wantRetry, err = preHook.PostRunHook(err)
+				retry = retry || wantRetry
+			}
```
zoran15, line 164: This is hard to understand: preHook is run as postHook? Maybe add some comments to clarify the logic here.

cli_tools/common/utils/daisyutils/daisy_worker.go
```diff
@@ -117,18 +179,22 @@ func (w *defaultDaisyWorker) RunAndReadSerialValue(key string, vars map[string]s
...
 
 func (w *defaultDaisyWorker) Cancel(reason string) bool {
-	if w.wf != nil {
-		w.wf.CancelWithReason(reason)
-		return true
-	}
-
-	//indicate cancel was not performed
-	return false
+	// do-once is required to ensure that additional calls
```
zoran15, line 191: What is `do-once`? `Do` is called below.

dntczdx (commented):
just as Zoran said, I don't get how it can help on the pd-fallback scenario. It seems that right now the new feature is added but I don't know how it will be used.

EricEdens:
> just as Zoran said, I don't get how it can help on the pd-fallback scenario. It seems that right now the new feature is added but I don't know how it will be used.

To implement the pd-fallback scenario, the next PR adds a hook that implements both `PreRun` and `PostRun`. If the `PostRun` sees a quota error, then it requests retry of the workflow. On the subsequent retry, the associated `PreRun` method traverses the workflow and changes `ssd` to `pd-standard`.

cli_tools/gce_windows_upgrade/upgrader/workflows.go
EricEdens: Thanks for the feedback.

> Can daisyWorker store the workflow that was built to avoid this complexity?

Unfortunately, no, since (as far as I know) a workflow can't be re-run.

Here's my logic: To retry the workflow using pd-standard disks, then we need a new workflow. Across the CLI tools, there's a myriad of customizations for creating workflows. The `workflowProvider` abstraction allows call sites to describe *how* to make the workflow, leaving `daisy_worker` to execute the workflow.

Do you think a different name would make it clearer? Or perhaps a different factoring?

cli_tools/common/image/importer/bootable_disk_processor.go
EricEdens: done

cli_tools/common/image/importer/daisy_inflater.go
EricEdens: Shoot, thanks for catching that -- that change is required for the *next* PR where the fallback-to-pd-standard hook is added.

cli_tools/common/utils/daisyutils/daisy_utils.go
EricEdens: In the current implementation, the string is empty when the channel is closed. That said, I can see how that would be confusing, both now and in the future. So I added a call to `w.CancelWorkflow()`, which is safe since we added a guard that allows daisy workflow cancellation to be idempotent.

cli_tools/common/utils/daisyutils/daisy_utils.go
EricEdens: Good question -- in short, the context that's passed here shouldn't be cancelled. If cancellation occurs, then all of the clients within daisy die, and cleanup can't occur. I added a comment to explain why we're adding a new context.

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: done

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: There's a race where a caller could call `DaisyWorker.Cancel` before calling `Run`. A typical example would be a timeout that occurs in the period between construction and invocation of the DaisyWorker.

Here's the cancel method:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/85f7e15dfd425f494df6c7353953d92cc925e6fb/cli_tools/common/utils/daisyutils/daisy_worker.go#L190

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: I added a comment to Tony's question below.

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: For the current requirements, I don't think multiple retries would be helpful. That said, I was thinking the refactoring would be easy if someone needed to pull that out later.

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: Ahh, good point. Updated the log message.

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: done

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: Shoot, the variable names are incorrect for the type. Updated.

cli_tools/common/utils/daisyutils/daisy_worker.go
EricEdens: Updated to `once.Do`  (`do-once` is just what made sense in my head, but clearly wasn't that great :)  )

cli_tools/gce_windows_upgrade/upgrader/workflows.go
zoran15: isn't the `wf` returned from here the same as `daisyWorker.finishedWf`?

EricEdens: Got it, thanks! 

I looked some more, and the upgrade tool was exposing the workflow for unit testing. I updated the tests so that they inspected the worker, instead of the workflow.

cli_tools/gce_windows_upgrade/upgrader/upgrader.go
```diff
@@ -102,12 +102,12 @@ type upgrader struct {
...
-	retryUpgradeFn            func() (*daisy.Workflow, error)
-	rebootFn                  func() (*daisy.Workflow, error)
-	cleanupFn                 func() (*daisy.Workflow, error)
-	rollbackFn                func() (*daisy.Workflow, error)
+	prepareFn                 func() (daisyutils.DaisyWorker, error)
+	upgradeFn                 func() (daisyutils.DaisyWorker, error)
+	retryUpgradeFn            func() (daisyutils.DaisyWorker, error)
+	rebootFn                  func() (daisyutils.DaisyWorker, error)
+	cleanupFn                 func() (daisyutils.DaisyWorker, error)
+	rollbackFn                func() (daisyutils.DaisyWorker, error)
```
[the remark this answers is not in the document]
EricEdens: Good question. The interface is implemented with pointer receivers, so the object that's being passed around is a pointer to the struct that implements the interface:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/85f7e15dfd425f494df6c7353953d92cc925e6fb/cli_tools/common/utils/daisyutils/daisy_worker.go#L62

I think this is normal across Go's typical usage; eg in the standard library, the functions that accept implementations of Reader and Writer take the interfaces directly rather than as pointers (which would be a pointer to a pointer to a struct that implements that interface)

https://pkg.go.dev/io#Copy

I'm still kind of new to Go, so my understanding could be wrong here! But that was the intent.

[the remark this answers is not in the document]
zoran15: I missed the fact `DaisyWorker` is an interface. Nevermind.

zoran15 (approved):

dntczdx (approved):

PR #1782, opened 2021-11-09
[safely adopt VMDK API] stage 1

dntczdx:
Logics and tests are all here for stage 1 & 2. Still shadow mode so it's running stage 1 code. 1 line of change to switch to stage 2 later.

Ref to the design doc to see the new logic branches.


cli_tools/common/imagefile/qemu_img.go
```diff
@@ -39,6 +42,7 @@ type ImageInfo struct {
 	Format           string
 	ActualSizeBytes  int64
 	VirtualSizeBytes int64
+	Checksum         string
```
EricEdens, line 45: Can you update the doc for `ImageInfo` to explain checksum? (The current doc refers only to qemu-img info)

cli_tools/common/imagefile/qemu_img.go
```diff
@@ -53,34 +57,112 @@ func NewInfoClient() InfoClient {
...
+
+func (client defaultInfoClient) getFileChecksum(ctx context.Context, filename string, virtualSizeBytes int64) (checksum string, err error) {
+	// Check size = 200000*512 = 100MB
+	checkCount := int64(200000)
+	blockSize := int64(512)
+	blockCount := virtualSizeBytes / blockSize
+	skips := []int64{0, int64(2000000) - checkCount, int64(20000000) - checkCount, blockCount - checkCount}
+	tmpOutFilePrefix := "out" + pathutils.RandString(5)
+	for i, skip := range skips {
+		tmpOutFileName := fmt.Sprintf("%v%v", tmpOutFilePrefix, i)
```
EricEdens, line 115: It looks like test only checks if the md5sum is non-empty, so I think there's an opportunity to improve coverage.

Take a look at `ShellExecutor` which is a replacement for using `exec` directly.
 - Interface: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/common/utils/shell/executor.go
 - Example usage: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/c36c64a28b34f9459d83fbb3dac90f3e6955410e/cli_tools/common/mount/inspector_test.go#L27

Using these, you can verify the arguments passed to the shell command, and also verify that your go code correctly parses the response.

cli_tools/common/imagefile/qemu_img.go
```diff
@@ -53,34 +57,112 @@ func NewInfoClient() InfoClient {
...
+		defer os.Remove(tmpOutFileName)
+
+		if skip < 0 {
+			skip = 0
+		}
+
+		// Write 100MB data to a file.
+		cmd := exec.CommandContext(ctx, "qemu-img", "dd", fmt.Sprintf("if=%v", filename),
+			fmt.Sprintf("of=%v", tmpOutFileName), fmt.Sprintf("bs=%v", blockSize),
+			fmt.Sprintf("count=%v", skip+checkCount), fmt.Sprintf("skip=%v", skip))
```
EricEdens, line 125: It looks like you're writing 100MB to the docker container that's running in cloud build. Are there any modifications required to support this? (For example: mounting a volume or writing  to a special directory?) Or is writing this amount of data supported by the cloud build runtime?

cli_tools/common/imagefile/qemu_img.go
```diff
@@ -53,34 +57,112 @@ func NewInfoClient() InfoClient {
...
+		var out []byte
+		out, err = cmd.Output()
+		err = constructCmdErr(string(out), err, "inspection for checksum failure")
+		if err != nil {
+			return
+		}
+
+		// Calculate checksum for the 100MB file.
+		cmd = exec.CommandContext(ctx, "md5sum", tmpOutFileName)
```
EricEdens, line 134: In case unseen, the Go standard library allows you to calculate MD5s. This might be helpful to avoid having to parse the output of `md5sum`, and making sure that `md5sum` is available at runtime

https://pkg.go.dev/crypto/md5

cli_tools/common/image/importer/api_inflater.go
```diff
@@ -54,41 +54,52 @@ type apiInflater struct {
 	cancelChan       chan string
 	logger           logging.Logger
 	isShadowInflater bool
+	needChecksum     bool
 }
 
-func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Client, storageClient domain.StorageClientInterface, logger logging.Logger, isShadowInflater bool) *apiInflater {
+func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Client, storageClient domain.StorageClientInterface,
```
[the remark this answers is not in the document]
dntczdx: done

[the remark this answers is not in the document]
EricEdens: OK, this makes sense. Are you collecting a metric to see how often the `qemu` checksum fails?

cli_tools/common/imagefile/qemu_img.go
EricEdens: Given that a bug in this checksum logic (either now or in the future) would cause an import to fail, I'd definitely be interested in having unit test coverage. We can talk as a team, though, as well. Maybe there's something I'm missing here!

dntczdx: I modified the tests yesterday. The coverage is that I checked the checksum output follows the format "^[a-zA-Z0-9]{32}-[a-zA-Z0-9]{32}-[a-zA-Z0-9]{32}-[a-zA-Z0-9]{32}$". If the checksum calculation is failed, we won't get this format of output. Do you think that is a good coverage?

cli_tools/common/imagefile/qemu_img.go
```diff
@@ -53,34 +60,111 @@ func NewInfoClient() InfoClient {
...
+		defer os.Remove(tmpOutFileName)
+
+		if skip < 0 {
+			skip = 0
+		}
+
+		// Write 100MB data to a file.
+		cmd := exec.CommandContext(ctx, "qemu-img", "dd", fmt.Sprintf("if=%v", filename),
+			fmt.Sprintf("of=%v", tmpOutFileName), fmt.Sprintf("bs=%v", blockSize),
+			fmt.Sprintf("count=%v", skip+checkCount), fmt.Sprintf("skip=%v", skip))
```
zoran15, line 128: Please document the semantics for each of these values.

cli_tools/common/imagefile/qemu_img.go
```diff
@@ -53,34 +60,111 @@ func NewInfoClient() InfoClient {
...
+		}
+
+		// Calculate checksum for the 100MB file.
+		f, fileErr := os.Open(tmpOutFileName)
+		if err != nil {
+			err = daisy.Errf("Failed to open file '%v' for QEMU md5 checksum calculation: %v", tmpOutFileName, fileErr)
+			return
+		}
+		defer f.Close()
+		h := md5.New()
```
zoran15, line 143: can this be reused within the loop?

cli_tools/common/imagefile/qemu_img.go
```diff
@@ -53,34 +60,111 @@ func NewInfoClient() InfoClient {
...
+	}
+	return &jsonTemplate, err
+}
+
+func (client defaultInfoClient) getFileChecksum(ctx context.Context, filename string, virtualSizeBytes int64) (checksum string, err error) {
+	// Check size = 200000*512 = 100MB
+	checkCount := int64(200000)
+	blockSize := int64(512)
+	blockCount := virtualSizeBytes / blockSize
+	skips := []int64{0, int64(2000000) - checkCount, int64(20000000) - checkCount, blockCount - checkCount}
```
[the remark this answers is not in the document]
dntczdx: Let me add more descriptions.

cli_tools/common/imagefile/qemu_img.go
dntczdx: I exampled in the new description at the begging of "getFileChecksum", please have a look.

cli_tools/common/imagefile/qemu_img.go
dntczdx: no, it keeps accepting written string from IO, but there is no way to reset the content. We new a new obj for each calculation.

cli_tools/common/image/importer/api_inflater.go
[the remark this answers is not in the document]
EricEdens: I searched for `api_success_checksum_skipped` in this PR and in the codebase but didn't see anything -- is that coming later?

cli_tools/common/image/importer/api_inflater.go
```diff
@@ -54,41 +54,57 @@ type apiInflater struct {
...
 }
 
-func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Client, storageClient domain.StorageClientInterface, logger logging.Logger, isShadowInflater bool) *apiInflater {
+// createAPIInflater creates an apiInflater.
+//"isShadowInflater" indicates whether it's used as a shadow inflater.
+// "needChecksum" indicates whether checksum of the output disk needs to be
+// calculated. When QEMU checksum is failed to calculated, we don't need to
+// calculate the disk checksum anymore.
+func createAPIInflater(request ImageImportRequest, computeClient daisyCompute.Client, storageClient domain.StorageClientInterface,
+	logger logging.Logger, isShadowInflater bool, needChecksum bool) *apiInflater {
```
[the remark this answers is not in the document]
zoran15: There's nothing preventing putting all these args in a struct. Or creating a inflater builder were you can set all these flags as builder properties then call build. E.g. `newAPIInflaterBuilder().SetRequest(req).SetComputeClient(client)...SetNeedChecksum(false).build().

cli_tools/common/image/importer/api_inflater.go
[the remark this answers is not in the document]
dntczdx: As you suggested, I split it to 2 fields, so the failure is tracked by "qemu_checksum_missing", Please ref to the latest commit.

cli_tools/common/image/importer/api_inflater.go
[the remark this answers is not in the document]
dntczdx: oh you mean grouping all 6 as a struct, I thought 2. Let me group them.

cli_tools/common/image/importer/inflater.go
```diff
@@ -84,46 +105,79 @@ type inflaterFacade struct {
...
+				facade.logger.Metric(&pb.OutputInfo{
+					InflationType:   "api_failed",
+					InflationTimeMs: []int64{ii.inflationTime.Milliseconds()},
+				})
+				return pd, ii, err
+			}
+			fallbackReason = "unsupported_format"
+		}
+
+		if err == nil {
```
zoran15, line 137:
```suggestion
		} else {
```

cli_tools/common/image/importer/inflater.go
```diff
@@ -84,46 +105,79 @@ type inflaterFacade struct {
...
+			if isChecksumMatch(ii.checksum, facade.qemuChecksum) {
+				facade.logger.Metric(&pb.OutputInfo{
+					InflationType:   "api_success",
+					InflationTimeMs: []int64{ii.inflationTime.Milliseconds()},
+				})
+				return pd, ii, err
+			}
+
+			// If checksum mismatches , delete the corrupted disk.
+			facade.logger.User("Disk checksum mismatch, recreating...")
```
zoran15, line 147: Why does this log state "recreating" when it's deleting a disk? Does that part of the log message apply to the code outside this block (`fallbackToDaisyInflater')? If so, move it after the deleteDisk attempt, since if disk deletion fails, then this log does not apply.

cli_tools/common/image/importer/inflater.go
```diff
@@ -84,46 +105,79 @@ type inflaterFacade struct {
...
+			fallbackReason = "checksum_mismatch"
+		}
 	}
 
-	// Retry with daisy inflater.
-	pd, tf, err = facade.daisyInflater.Inflate()
+	// Now fallback to daisy inflater. As described above, it is due to one of the below reasons:
+	// 1. The API failed because it doesn't support the format of the image file.
+	// 2. Checksum mismatch, which means the API produced a corrupted disk.
+	// 3. QEMU checksum missed, which means we have no way to compare the checksum.
```
zoran15, line 159:
```suggestion
	// 3. QEMU checksum is missing, which means we have no way to compare the checksum.
```

cli_tools/common/image/importer/inflater.go
```diff
@@ -84,46 +105,79 @@ type inflaterFacade struct {
...
+	pd, ii, err = facade.fallbackToDaisyInflater(fallbackReason)
+	return pd, ii, err
```
zoran15, line 161: Can you just `return facade.fallbackToDaisyInflater(fallbackReason)`?

cli_tools/common/image/importer/inflater.go
```diff
@@ -227,18 +284,38 @@ func (facade *shadowTestInflaterFacade) Cancel(reason string) bool {
...
+
+func getDiskName(executionID string) string {
+	return fmt.Sprintf("disk-%v", executionID)
+}
+
+// isChecksumMatch verifies whether checksum matches, excluded useless characters.
+func isChecksumMatch(checksum1, checksum2 string) bool {
+	reg, _ := regexp.Compile("[^a-zA-Z0-9]+")
+	processedChecksum1 := reg.ReplaceAllString(checksum1, "")
+	processedChecksum2 := reg.ReplaceAllString(checksum2, "")
```
zoran15, line 319: Why is removing non alphanumeric ('useless') characters needed?

cli_tools/common/image/importer/inflater.go
dntczdx: done

cli_tools/common/image/importer/inflater.go
dntczdx: you are right, moved

cli_tools/common/image/importer/inflater.go
dntczdx: done

cli_tools/common/image/importer/inflater.go
dntczdx: done

cli_tools/common/image/importer/inflater.go
dntczdx: checksum output format from shell command vs golang lib is slightly different. Comparing to operating the string to enforce the format, it's more reliable and easier to just ignore useless charactors.

zoran15 (approved):

cli_tools/common/imagefile/qemu_img.go
dntczdx: done

PR #1784, opened 2021-11-10
CLI Tools: Implement fallback to PD standard

EricEdens:
This removes implements the `FallbackToPdStandard` feature usinga DaisyWorker. A follow-up PR will remove `FallbackToPdStandard` from daisy.

## Testing

- Added an e2e test that confirms the fallback behavior.

### DIFF
--- a/cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
+++ b/cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
@@ -1,0 +1,62 @@
+//  Copyright 2021 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package daisyutils
+
+import (
+	"strings"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// FallbackToPDStandard detects if a workflow fails due to insufficient SSD
+// quota. If so, it re-runs the workflow and modifies the disks to use PD Standard.
+type FallbackToPDStandard struct {
+	logger   logging.Logger
+	fallback bool
+}
+
+// PreRunHook modifies the workflow to use standard disks if `fallback` is true.
+func (t *FallbackToPDStandard) PreRunHook(wf *daisy.Workflow) error {
+	if t.fallback {
+		fallbackToPDStandard(wf)
+	}
+	return nil
+}
+
+// PostRunHook inspects the workflow error to see if it's related to insufficient SSD quota.
+// If so, it requests a retry that will re-run the workflow using standard disks.
+func (t *FallbackToPDStandard) PostRunHook(err error) (wantRetry bool, wrapped error) {
+	if t.fallback {
+		// A fallback has already occurred; don't request retry.
+		t.fallback = false
+	} else if err != nil && strings.Contains(err.Error(), "SSD_TOTAL_GB") {
+		if t.logger != nil {
+			t.logger.Debug("Workflow failed with insufficient SSD quota. Requesting retry. error=" + err.Error())
+		}
+		t.fallback = true
+	}
+	return t.fallback, err
+}
+
+func fallbackToPDStandard(workflow *daisy.Workflow) {
+	workflow.IterateWorkflowSteps(func(step *daisy.Step) {
+		if step.CreateDisks != nil {
+			for _, disk := range *step.CreateDisks {
+				disk.Disk.Type = "pd-standard"
+			}
+		}
+	})
+}
--- a/cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
+++ b/cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
@@ -24,34 +24,32 @@ import (
 // FallbackToPDStandard detects if a workflow fails due to insufficient SSD
 // quota. If so, it re-runs the workflow and modifies the disks to use PD Standard.
 type FallbackToPDStandard struct {
-	logger   logging.Logger
-	fallback bool
+	logger         logging.Logger
+	shouldFallback bool
 }
 
-// PreRunHook modifies the workflow to use standard disks if `fallback` is true.
-func (t *FallbackToPDStandard) PreRunHook(wf *daisy.Workflow) error {
-	if t.fallback {
-		fallbackToPDStandard(wf)
+// PreRunHook modifies the workflow to use standard disks if `shouldFallback` is true.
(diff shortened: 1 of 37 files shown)


cli_tools/gce_ovf_export/exporter/instance_disks_exporter.go
```diff
@@ -145,7 +145,6 @@ func (ide *instanceDisksExporterImpl) addExportDisksSteps(w *daisy.Workflow, ins
 			"format":                     params.DiskExportFormat,
 			"export_instance_disk_image": "projects/compute-image-tools/global/images/family/debian-9-worker",
 			"export_instance_disk_size":  "200",
-			"export_instance_disk_type":  "pd-ssd",
```
dntczdx, line 148: do we intentionally change the disk type?

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
```diff
@@ -125,10 +128,53 @@ func TestSuite(
...
+		testSuiteName, fmt.Sprintf("[%v] %v", e2e.Wrapper, "Allow import to work when SSD quota is exhausted"))] = fallbackWhenSSDQuotaExhausted
+
 	e2e.CLITestSuite(ctx, tswg, testSuites, logger, testSuiteRegex, testCaseRegex,
 		testProjectConfig, testSuiteName, testsMap)
 }
 
+func fallbackWhenSSDQuotaExhausted(ctx context.Context, testCase *junitxml.TestCase, logger *log.Logger,
+	testProjectConfig *testconfig.Project, testType e2e.CLITestType) {
+	suffix := path.RandString(5)
+	zone := "us-central2-a"
```
dntczdx, line 142: Did you setup this zone's quota to make it happen?

cli_tools/gce_ovf_export/exporter/instance_disks_exporter.go
EricEdens: Good question -- I think I removed that since `pd-ssd` is the default value for that variable. That said, I think removing it here could cause confusion, so I added it back.

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
EricEdens: The quota for that zone was already limited to 500GB. Since the test depends on having insufficient quota, I added the assertion that creating a 600GB disk fails. (The OVF has a 1000GB disk).

dntczdx (approved):
LGTM

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
```diff
@@ -0,0 +1,62 @@
...
+
+// FallbackToPDStandard detects if a workflow fails due to insufficient SSD
+// quota. If so, it re-runs the workflow and modifies the disks to use PD Standard.
+type FallbackToPDStandard struct {
+	logger   logging.Logger
+	fallback bool
+}
+
+// PreRunHook modifies the workflow to use standard disks if `fallback` is true.
+func (t *FallbackToPDStandard) PreRunHook(wf *daisy.Workflow) error {
```
zoran15, line 32: What does `t` stand for? I got it confused for test, so maybe a better name should be used..

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/logging"
+	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
+)
+
+// FallbackToPDStandard detects if a workflow fails due to insufficient SSD
+// quota. If so, it re-runs the workflow and modifies the disks to use PD Standard.
+type FallbackToPDStandard struct {
+	logger   logging.Logger
+	fallback bool
```
zoran15, line 28: Maybe `shouldFallback` or `doFallback`? Just `fallback` sounds like it's a property of the Fallback object, which sounds strange, while above suggestions are obviously instructions to perform fallback.

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	return nil
+}
+
+// PostRunHook inspects the workflow error to see if it's related to insufficient SSD quota.
+// If so, it requests a retry that will re-run the workflow using standard disks.
+func (t *FallbackToPDStandard) PostRunHook(err error) (wantRetry bool, wrapped error) {
+	if t.fallback {
+		// A fallback has already occurred; don't request retry.
+		t.fallback = false
```
zoran15, line 44: Is re-setting this field needed? Is there a risk of some kind of a deadloop?

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
```diff
@@ -0,0 +1,62 @@
...
+	} else if err != nil && strings.Contains(err.Error(), "SSD_TOTAL_GB") {
+		if t.logger != nil {
```
zoran15, line 46: Why would logger be `nil`?

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
```diff
@@ -0,0 +1,62 @@
...
+			t.logger.Debug("Workflow failed with insufficient SSD quota. Requesting retry. error=" + err.Error())
+		}
+		t.fallback = true
+	}
+	return t.fallback, err
+}
+
+func fallbackToPDStandard(workflow *daisy.Workflow) {
```
zoran15, line 54: `fallbackToPDStandard` sounds like fallback is occuring in this func, while it's just modifying the workflow for fallback. Consider a more appropriate name.

zoran15 (commented):
I remember we had to split Daisy and wrapper changes when adding new features to Daisy (submit Daisy PR first, then wrapper). Because this is removing, it might not be needed, but I guess we'll see when you merge this. A safer way would have been to remove the fallback dependency in wrappers, then later on remove the feature from Daisy.

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
EricEdens: Good point. Changed to `f`.

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
EricEdens: SGTM. Changed to `shouldFallback`.

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
EricEdens: I can see the concern about a loop, especially if someone updates daisyworker in the future to support multiple retries. Updated to not mutate that field.

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
EricEdens: Removed that check (it simplified testing a bit; but in retrospect it meant we weren't executing the log in tests)

cli_tools/common/utils/daisyutils/workflow_hook_fallback_pd.go
EricEdens: sgtm. Changed to `useStandardDisks`

zoran15 (approved):

zoran15 (approved):

PR #1785, opened 2021-11-12
Add sqlserver workflows

LawrenceHwang:
The workflows for the Guest OS pipeline migration.

### DIFF
--- a/daisy_workflows/build-publish/sqlserver/sql-2012-enterprise-windows-2012-r2-dc.wf.json
+++ b/daisy_workflows/build-publish/sqlserver/sql-2012-enterprise-windows-2012-r2-dc.wf.json
@@ -1,0 +1,76 @@
+{
+  "Name": "sql-2012-enterprise-windows-2012-r2-dc",
+  "Project": "gce-image-builder",
+  "Zone": "us-central1-b",
+  "GCSPath": "gs://gce-image-build-bucket/daisy/${USERNAME}",
+  "Vars": {
+    "build_date": {
+      "Value": "${TIMESTAMP}",
+      "Description": "Build datestamp used to version the image."
+    },
+    "gcs_url": {
+      "Required": true,
+      "Description": "The GCS url that the image raw file exported to."
+    },
+    "gcs_internal_url": {
+      "Required": true,
+      "Description": "The GCS url that the exported image is saved to for internal use."
+    },
+    "workflow_root": {
+      "Value": "/workflows",
+      "Description": "Root of github workflows, defaults to /workflows in the container."
+    },
+    "install_disk": "disk-install",
+    "publish_project": "${PROJECT}",
+    "sql_server_media": {
+      "Required": true,
+      "Description": "GCS or local path to the SQL Server installer media"
+    },
+    "source_image_project": {
+      "Value": "windows-cloud",
+      "Description": "Project to source base image from."
+    },
+    "ssms_exe": {
+      "Required": true,
+      "Description": "GCS or local path to SSMS installer"
+    },
+    "timeout": {
+      "Value": "2h",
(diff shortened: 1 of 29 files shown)


daisy_workflows/build-publish/sqlserver/sql-2012-enterprise-windows-2012-r2-dc.wf.json
```diff
@@ -0,0 +1,76 @@
...
+  "Vars": {
+    "build_date": {
+      "Value": "${TIMESTAMP}",
+      "Description": "Build datestamp used to version the image."
+    },
+    "gcs_url": {
+      "Required": true,
+      "Description": "The GCS url that the image raw file exported to."
+    },
+    "gcs_internal_url": {
```
bkatyl, line 15: the internal images are only for windows. We don't need that for SQL and Containers.

daisy_workflows/build-publish/sqlserver/sql-2012-enterprise-windows-2012-r2-dc.wf.json
```diff
@@ -0,0 +1,76 @@
...
+          "sql_server_media": "${sql_server_media}",
+          "source_image_project": "${source_image_project}",
+          "ssms_exe": "${ssms_exe}",
+          "timeout": "2h"
+        }
+      }
+    },
+    "sqlserver-export": {
+      "Timeout": "60m",
+      "IncludeWorkflow": {
```
bkatyl, line 60: I don't think we need our own sql-server export process. Since there is a single file output we can use the common export/disk_export.wf.json",

ex:
      "IncludeWorkflow": {
        "Path": "${workflow_root}/export/disk_export.wf.json",
        "Vars": {
          "source_disk": "${install_disk}",
          "destination": "${gcs_path}/${image_prefix}-v${build_date}/root.tar.gz"
        }
      }

daisy_workflows/build-publish/sqlserver/sql-2012-enterprise-windows-2012-r2-dc.wf.json
```diff
@@ -0,0 +1,76 @@
...
+      "Description": "GCS or local path to SSMS installer"
+    },
+    "timeout": {
+      "Value": "2h",
+      "Description": "The timeout to set for the image build."
+    }
+  },
+  "Steps": {
+    "build": {
+      "Timeout": "4h",
```
bkatyl, line 44: A steps timeout should be the same as the timeout we pass to the job. 
I remember changing this to 4 hours, but the timeout in IncludeWorkflow needs to be updated to match.

The timeout variable in the workflow is actually not being used even though it's being set. So I'd suggest changing the following:
2h --> 4h
   "timeout": {
      "Value": "4h",
      "Description": "The timeout to set for the image build."
    }

  "Steps": {
    "build": {
      "Timeout": "${timeout}",
      "IncludeWorkflow": {
        "Path": "${workflow_root}/image_build/sql-2012-enterprise-windows-2012-r2-dc.wf.json",
        "Vars": {
          "build_date": "${TIMESTAMP}",
          "install_disk": "disk-install",
          "publish_project": "${publish_project}",
          "sql_server_media": "${sql_server_media}",
          "source_image_project": "${source_image_project}",
          "ssms_exe": "${ssms_exe}",
          "timeout": "${timeout}"
        }
        
        
This will ensure that if we need to change this value, we can confidently pass in the value from concourse and have it flow through and be honored by all decendant workflows.

bkatyl (commented):
These comments apply to all workflows

LawrenceHwang:
Closing PR for now. Will send a new one with just one image's workflow.

PR #1790, opened 2021-11-18
[vmdk api adoption] Fix 2 bugs

dntczdx:
1. In PR #1754 , "startTime := time.Now()" before the daisy workflow running got removed, so the inflation time didn't get calculated correwctly. Bring it back. Otherwise, the inflation time will be always 0.
2. The gcsfuse was used without arg "--implicit-dirs". Added it. This bug made the file inspection never taking effect. The file inspection works fine on local without gcloud wrapper, because it avoided the nested folder "tmpimage". Once we call it via gcloud, the nested folder is always involved, so the bug was always triggered.

### DIFF
--- a/cli_tools/common/gcsfuse/client.go
+++ b/cli_tools/common/gcsfuse/client.go
@@ -17,8 +17,10 @@ package gcsfuse
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"os"
 	"os/exec"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/cli_tools/common/utils/path"
 )
 
 // Client provides methods for mounting and unmounting FUSE filesystems
@@ -39,7 +41,9 @@ func NewClient() Client {
 type defaultClient struct{}
 
 func (client defaultClient) MountToTemp(ctx context.Context, bucket string) (string, error) {
-	dir, err := ioutil.TempDir("", bucket)
+	//dir, err := ioutil.TempDir("", bucket)
+	dir := bucket + path.RandString(5)
+	err := os.Mkdir(dir, 0755)
 	if err != nil {
 		return "", fmt.Errorf("failed to create a destination directory: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


EricEdens (approved):

cli_tools/common/gcsfuse/client.go
```diff
@@ -43,7 +43,7 @@ func (client defaultClient) MountToTemp(ctx context.Context, bucket string) (str
 	if err != nil {
 		return "", fmt.Errorf("failed to create a destination directory: %w", err)
 	}
-	cmd := exec.CommandContext(ctx, "gcsfuse", bucket, dir)
+	cmd := exec.CommandContext(ctx, "gcsfuse", "--implicit-dirs", bucket, dir)
```
EricEdens, line 46: Can you add some unit tests that confirm the bug, and then show that your fix address the isssue?

cli_tools/common/image/importer/daisy_inflater.go
```diff
@@ -72,7 +73,7 @@ func (inflater *daisyInflater) Inflate() (persistentDisk, inflationInfo, error)
 			sourceType: serialValues[importFileFormatKey],
 		}, inflationInfo{
 			checksum:      serialValues[diskChecksumKey],
-			inflationTime: time.Since(time.Now()),
+			inflationTime: time.Since(startTime),
```
EricEdens, line 76: ditto on unit tests.

dntczdx: done

cli_tools/common/gcsfuse/client.go
EricEdens: Talked offline, and Tony pointed out that gcsfuse would be difficult for unit tests. I suggested creating a module test similar to [OS inspect](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/5bd10d7493e35f6fb3ab3ed440f56ece695568b4/cli_tools_tests/module/diskinspect/disk_inspect_test.go#L341-L340)

In the test, call `NewGCSInspector`, then run `Inspect` with a gcs path, and assert the results are what you'd expect.

dntczdx: added a new file for that

EricEdens (approved):
looks great!

zoran15 (approved):

PR #1792, opened 2021-11-18
OVF import: Fix PD-fallback

EricEdens:
This fixes the failure that we're seeing in the "Allow import to work when SSD quota is exhausted" e2e test.

https://oss-prow.knative.dev/view/gs/compute-image-tools-test/logs/ci-ovf-import-e2e-tests-daily/1460869781400850432

Two changes:
1. Implement pd-standard fallback for attached disks
2. Decouple disk import from final instance creation

## More info

The test is failing with an error "Intermediate image ovf-w1207zz123-1 already exists. Re-run import." This happened since the SSD quota was insufficient to create the imported instance; prior to this PR, OVF import would re-run *everything*, and find that the disks had already been imported to images.

This PR splits OVF import into two phases:
1. Import disks to images.
2. Create an instance from the images.

While working on this, I noticed an existing bug: If creation of the final instance (or GMI fails), the intermediate disk images will be left. I created an internal bug to track this, and will submit a follow-up PR for that.

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
+++ b/cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
@@ -84,6 +84,9 @@ type OVFImporter struct {
 
 	// Populated when disk file import finishes.
 	imageURIs []string
+
+	// Populated after reading OVF file descriptor.
+	machineTypeString string
 }
 
 // NewOVFImporter creates an OVF importer, including automatically populating dependencies,
@@ -285,8 +288,7 @@ func (oi *OVFImporter) getMachineType(
 	return machineTypeProvider.GetMachineType()
 }
 
-func (oi *OVFImporter) setUpImportWorkflow() (workflow *daisy.Workflow, err error) {
-
+func (oi *OVFImporter) importDiskFilesToImages() error {
 	oi.imageLocation = oi.params.Region
 
 	ovfGcsPath, shouldCleanup, err := oi.getOvfGcsPath(oi.params.ScratchBucketGcsPath)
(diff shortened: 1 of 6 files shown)


cli_tools/gce_ovf_import/ovf_importer/ovf_importer.go
```diff
@@ -360,23 +363,27 @@ func (oi *OVFImporter) setUpImportWorkflow() (workflow *daisy.Workflow, err erro
...
+	if err := oi.createWorkerForFinalInstance().Run(map[string]string{}); err != nil {
 		oi.Logger.User(err.Error())
 		return err
 	}
 	oi.Logger.User("OVF import workflow finished successfully.")
 	return nil
 }
 
-func (oi *OVFImporter) setupWorker() daisyutils.DaisyWorker {
+func (oi *OVFImporter) createWorkerForFinalInstance() daisyutils.DaisyWorker {
```
zoran15, line 384: `ResourceLabeler` should label all temporary and final resources. I'm not sure if this feature was potentially lost in the DaisyWorker refactor earlier, but it seems it would be lost after this PR is merged.

EricEdens: All of the workflows in CLI tools are now executed through DaisyWorker, which enforces a resource labeler to be present:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/a9f351a2d3e296536bd521b58afe6604f7f6db21/cli_tools/common/utils/daisyutils/daisy_worker.go#L70

I ran an OVF import with this PR, and here are the labels I'm seeing for the inflation step:

**instance**
```
  "labels": {
    "instance-import-build-id": "ovf-tss8d49yhw-1",
    "instance-import-tmp": "true"
  },
```
**boot disk**
```
  "labels": {
    "instance-import-build-id": "ovf-tss8d49yhw-1",
    "instance-import-tmp": "true"
  },
```

**scratch disk**
```
  "labels": {
    "instance-import-build-id": "ovf-tss8d49yhw-1",
    "instance-import-tmp": "true"
  },
```

**translation disk**
```
  "labels": {
    "instance-import-build-id": "ovf-tss8d49yhw-1",
    "instance-import-tmp": "true"
  },
```

zoran15 (approved):

PR #1809, opened 2021-12-03
OVF e2e: Use Debian 11 for 3 disk test

EricEdens:
The Ubuntu 16.04 three-disk tests have been flaky for the past couple of weeks. While researching this, I found that Ubuntu 16.04 became EOL in 2021: https://ubuntu.com/about/release-cycle

This replaces the three-disk test with Debian 11. The goal is to isolate whether the flakiness comes from Ubuntu 16.04, or having multiple disks during OVF import.

https://oss-prow.knative.dev/view/gs/compute-image-tools-test/logs/ci-ovf-import-e2e-tests-daily/1465254236387282944


dntczdx (approved):
LGTM

PR #1811, opened 2021-12-06
Optional pass image publish template content in CreatePublish

Quintonamore:
Previously the only way to create a Publish object was to call `CreatePublish` with a path to an image publish file that can be read in. This PR adds the option to pass in the image publish template as a string and then continue the template execution without reading in the passed file.


cli_tools/gce_image_publish/publish/publish_test.go
```diff
@@ -758,3 +758,27 @@ func TestCreateRollOut(t *testing.T) {
...
+func TestCreatePublishWithFileAndTemplate(t *testing.T) {
+	tests := []struct {
+		name     string
+		path     string
+		template string
+		wantErr  bool
+	}{
+		{"no valid path", "", "", true},
+		{"pass template", "", "{}", false},
+		{"pass with valid path", "../../../daisy_workflows/build-publish/debian/debian_10.publish.json", "", false},
```
Quintonamore, line 771: For this test, are there existing test image publish templates that I can reuse or will it be best to create a new directory with testing templates?

cli_tools/gce_image_publish/publish/publish.go
```diff
@@ -118,7 +118,7 @@ var (
 )
 
 // CreatePublish creates a publish object
-func CreatePublish(sourceVersion, publishVersion, workProject, publishProject, sourceGCS, sourceProject, ce, path string, varMap map[string]string, imagesCache map[string][]*computeAlpha.Image) (*Publish, error) {
+func CreatePublish(sourceVersion, publishVersion, workProject, publishProject, sourceGCS, sourceProject, ce, path, template string, varMap map[string]string, imagesCache map[string][]*computeAlpha.Image) (*Publish, error) {
```
zoran15, line 121: Being able to provide the same argument in two ways (a file path, and content) in the same function can be dangerous (what if both are provided?) and confusing. Having two separate functions (one can use the other), where one accepts content and another accepts file path is a cleaner way of achieving the same goal.

Quintonamore: I used your advice and moved this functionality into another function `CreatePublishWithTemplate` that calls `createPublish` but doesn't read in a file.

bkatyl (approved):

PR #1813, opened 2021-12-10
Server 2022: Removing BYOL, adding import workflows

bkatyl:
Removing BYOL workflows as per https://cloud.google.com/compute/docs/instances/windows/ms-licensing-faq#license_mobility_faq
- "If the Windows Server version was released before October 1, 2019, you can, subject to the terms of the licensing agreement, use Windows Server licenses acquired before this date on sole-tenant nodes."
Since 2022 was release in 2021, this version of Windows Server cannot use BYOL.

Adding import workflow for Windows Server 2022, to begin testing.

/cc LawrenceHwang


LawrenceHwang (approved):

zoran15 (commented):
What about exceptions listed in the doc:

_Microsoft makes exceptions for Windows Server licenses that meet the following conditions:

- Acquired under existing enrollments with an effective date before October 1, 2019
- Acquired after October 1, 2019 under Microsoft's Enterprise Agreement True-up (PDF download) ordering process
- Purchased for Windows Server versions released before October 1, 2019_

Are these exceptions only for licenses obtained after Oct 1, 2019, or also for versions of Windows released after that date?

daisy_workflows/image_build/windows/windows-server-2022-dc-uefi-byol.wf.json
```diff
@@ -1,106 +0,0 @@
-{
```
zoran15, line 1: What were these two deleted images used for?

bkatyl: Windows Server 2022 was released on Aug 18, 2021, so it cannot have been purchased before October 1, 2019

bkatyl: Windows Server 2022 BYOL, but since no customer can BYOL the images they need to be removed.

PR #1819, opened 2021-12-24
Add containers on windows server 2022 workflows

LawrenceHwang:
Add containers on windows server 2022 workflows

/cc bkatyl


LawrenceHwang:
Close the PR as per discussion with Peter.

PR #1821, opened 2021-12-28
Fix image export failing occasionally for large images by allocating buffer disk size to match raw image size

zoran15:
Image export for non-raw files: fix the issue of temporary buffer disk failing to resize occasionally with a resize2fs error by creating buffer disk of the same size as the image raw size.

Additionally:
* Added disk snapshot name validation
* Fixed raw disk export instance having OS disk match the size of the exported image as this is unnecessary

TODOs:
* Have buffer disk match image size for VMDK/VHD export when image path is provided (instead of a name as for this PR)
* Ensure disk snapshot export works

### DIFF
--- a/cli_tools/gce_vm_image_export/exporter/exporter.go
+++ b/cli_tools/gce_vm_image_export/exporter/exporter.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"log"
 	"os"
+	"strconv"
 	"strings"
 
 	"google.golang.org/api/option"
@@ -55,7 +56,6 @@ const (
 )
 
 func validateAndParseFlags(destinationURI string, sourceImage string, sourceDiskSnapshot string, labels string) (map[string]string, error) {
-
 	if err := validation.ValidateStringFlagNotEmpty(destinationURI, DestinationURIFlagKey); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


cli_tools/gce_vm_image_export/exporter/exporter.go
```diff
@@ -106,8 +106,11 @@ func buildDaisyVars(destinationURI string, sourceImage string, sourceDiskSnapsho
 	}
 
 	if sourceDiskSnapshot != "" {
-		varMap["source_disk_snapshot"] = param.GetGlobalResourcePath(
-			"snapshots", sourceDiskSnapshot)
+		varMap["source_disk_snapshot"] = sourceDiskSnapshot //param.GetGlobalResourcePath(			"snapshots", sourceDiskSnapshot)
```
EricEdens, line 109: Did you intend to leave the commented code?

daisy_workflows/export/disk_export.wf.json
```diff
@@ -47,7 +47,6 @@
         {
           "Name": "disk-${NAME}",
           "SourceImage": "${export_instance_disk_image}",
-          "SizeGb": "${export_instance_disk_size}",
```
EricEdens, line 50: What's the motivation for removing this?

cli_tools/gce_vm_image_export/exporter/exporter.go
zoran15: Nope, good catch. Snapshot export needs some work not related to these changes, but is not urgent since we don't expose this functionality as far as I'm aware.

daisy_workflows/export/disk_export.wf.json
zoran15: This is the OS disk for raw image export. It doesn't have to be set to 200GB (the default) or more if we pass the value based on image size. We don't store any data on it when doing raw file export, it's all written directly to GCS

EricEdens: :+1:

EricEdens (approved):

PR #1822, opened 2021-12-28
Image import inspection: update inspection stage to support archlinux and clearlinux distros

MahmoudNada0:
Image import inspection: update inspection stage to support archlinux and clearlinux distros and new versions of already existing distros
/assign EricEdens


EricEdens (approved):
nice!

PR #1830, opened 2022-01-08
Update Windows Server 2022 Export time to 60 min

LawrenceHwang:
Update Windows Server 2022 Export time to ~90~ 60 min.

/cc bkatyl

### DIFF
--- a/daisy_workflows/build-publish/windows/windows-server-2022-dc-uefi.wf.json
+++ b/daisy_workflows/build-publish/windows/windows-server-2022-dc-uefi.wf.json
@@ -70,8 +70,7 @@
           "build_date": "${build_date}",
           "gcs_url": "${gcs_url}",
           "gcs_internal_url": "${gcs_internal_url}",
-          "workflow_root": "${workflow_root}",
-          "disk_export_time_out": "50m"
+          "workflow_root": "${workflow_root}"
         }
       }
     }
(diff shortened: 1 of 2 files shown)


zoran15 (approved):

daisy_workflows/build-publish/windows/windows-server-2022-dc-uefi.wf.json
```diff
@@ -63,7 +63,7 @@
       }
     },
     "windows-export": {
-      "Timeout": "60m",
+      "Timeout": "90m",
       "IncludeWorkflow": {
         "Path": "./windows_export.wf.json",
```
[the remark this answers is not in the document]
LawrenceHwang: Thank you. I will update as suggested.  👍

daisy_workflows/build-publish/windows/windows_export.wf.json
```diff
@@ -16,11 +16,15 @@
     "workflow_root": {
       "Required": true,
       "Description": "Root of github workflows."
+    },
+    "disk_export_time_out": {
+      "Value": "30m",
```
bkatyl, line 21: I'd suggest changing this to 50 or 60 minutes. Realistically this timeout has only caused a suggestful image to be failed due to it needing 35 minutes.

Since this was likely the root cause, I don't think the overall task needs 90 minutes, but if it does then "disk_export_time_out": "60m" should be 80m as the copy will only get the default 10m timeout.

LawrenceHwang: Update. Thank you.

bkatyl (approved):

PR #1832, opened 2022-01-18
Add GVNIC guestOS feature tag to supported images.

zmarano:


hopkiw (approved):

zoran15:
Do we need to attach `GVNIC` guestOS feature to imported images for these operating systems? What about already imported images?

hopkiw:
We can't change already imported images, customers have to change them. We can set it on import if we know it supports GVNIC, let's discuss by chat.

PR #1840, opened 2022-01-31
OVF Import: Delete images when instance creation fails

EricEdens:
This PR updates OVF import so that temporary images are deleted when instance creation fails. Prior to this, if the final instance couldn't be created, the temporary images would remain in the user's project. There are various reasons the final instance cannot be created; two examples are: insufficient quota, and an invalid machine type flag.

## Testing
 - The PR includes  an e2e test that reproduces the issue and confirms the fix. The test creates an OVF import job that is designed to fail at instance creation, runs it, and then checks whether the images have been deleted.


dntczdx (approved):
LGTM, leave it t Zoran for OVF part

cli_tools/gce_ovf_import/domain/types.go
```diff
@@ -0,0 +1,31 @@
...
+type Image struct {
+	Project, ImageName, URI string
+}
+
+// NewImage constructs an Image instance.
+func NewImage(Project, ImageName string) Image {
+	return Image{
+		Project:   Project,
+		ImageName: ImageName,
+		URI:       param.GetImageResourcePath(Project, ImageName),
```
zoran15, line 29: This should be a method on Image type, as updating Project and/or ImageName should invalidate URI but with this implementation it does not, possibly leading to stale URIs

cli_tools/gce_ovf_import/domain/interfaces.go
```diff
@@ -53,5 +53,10 @@ type ImageImporterInterface interface {
 // If an import fails, all running imports will be cancelled, and images from finished imports
 // will be deleted.
 type MultiImageImporterInterface interface {
-	Import(ctx context.Context, params *OVFImportParams, fileURIs []string) (imageURIs []string, err error)
+	Import(ctx context.Context, params *OVFImportParams, fileURIs []string) (images []Image, err error)
+}
+
+// ImageDeleter checks whether images exist. If so, it deletes them.
+type ImageDeleter interface {
```
zoran15, line 60: Can this be a part of `cli_tools/common/domain`? There's nothing specific to OVF import in this interface and might be useful in other cases (although I don't see one right now).

cli_tools/gce_ovf_import/gce_utils/image_deleter_test.go
```diff
@@ -0,0 +1,59 @@
...
+	mockCompute := mocks.NewMockClient(ctrl)
+	mockCompute.EXPECT().GetImage(project, imageThatExists.ImageName).Return(nil, nil)
+	mockCompute.EXPECT().DeleteImage(project, imageThatExists.ImageName).Return(nil)
+	mockCompute.EXPECT().GetImage(project, imageThatDoesntExist.ImageName).Return(nil, errors.New("image not found"))
+
+	deleter := NewImageDeleter(mockCompute, logging.NewToolLogger("test"))
+	deleter.DeleteImagesIfExist([]ovfdomain.Image{imageThatExists, imageThatDoesntExist})
+}
+
+func TestImageDeleter_LogsMessage_IfDeleteFails(t *testing.T) {
```
zoran15, line 44: It would be nice to have a combo test case where one image is deleted, the other is not.

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
```diff
@@ -129,13 +130,69 @@ func TestSuite(
...
+			SourceURI:      fmt.Sprintf("gs://%v/ova/debian-11-three-disks.ova", ovaBucket),
+			MachineType:    "not-a-machine-type",
+			Os:             "debian-11",
+		}}
+
+	// While the import is running, verify that the temporary images were created
+	// with the names that we expect. This ensures that we're actually verifying
+	// that the correct temporary images are deleted.
+	detectedImages := map[string]bool{}
+	for i := 1; i < 4; i++ {
```
zoran15, line 168: Add a comment this is run 4 times for each image, and it's not a loop to check for image presence itself.  Or add a "constant" image_count or similar.

cli_tools/gce_ovf_import/domain/types.go
EricEdens: Good idea; this feels a lot safer. Updated and moved to `cli_tools/common/image/image.go`, with the interface in `cli_tools/common/domain`.

cli_tools/gce_ovf_import/domain/interfaces.go
EricEdens: Moving sounds good to me; done!

cli_tools/gce_ovf_import/gce_utils/image_deleter_test.go
EricEdens: Done!

cli_tools_tests/e2e/gce_ovf_import/test_suites/ovf_instance_import/ovf_instance_import_test_suite.go
EricEdens: Updated the comment to explain what `i` represents, and to describe how the goroutines work.

zoran15 (approved):

PR #1856, opened 2022-02-10
[os-upgrade] machine image dependency: move from Beta to GA

dntczdx:


EricEdens (approved):
lgtm; can you submit this to https://github.com/GoogleCloudPlatform/compute-daisy as well? You should be able to copy the files from the `daisy/` directory here into the root of `compute-daisy`. (Keep an eye on import paths though since they may be different in the new repo)

zoran15 (approved):

PR #1858, opened 2022-02-21
Image Import: use adaptive timeout in run_startup_scripts.

yswe:
The time to complete the run_startup_scripts can very from image to image, and can be greater than 5 minutes. Therefore the timeout value has been made adaptive to be incremented with every reboot.

### DIFF
--- a/daisy_workflows/image_import/windows/run_startup_scripts.cmd
+++ b/daisy_workflows/image_import/windows/run_startup_scripts.cmd
@@ -16,12 +16,32 @@ REM limitations under the License.
 REM Give the network time to initialize.
 ping 127.0.0.1 -n 60
 
-echo "Translate: Starting image translate..." > COM1:
+REM Specify a location in the registry to store temporary data that must be persistant between system restarts. See the comment about scheduling below for details.
+set image_import_reg_path="HKEY_LOCAL_MACHINE\SOFTWARE\GCEImportProcess"
 
-REM Restart the system in 5 minutes. Translate.ps1 will cancel this restart when it starts.
+for /f "tokens=3" %%a in ('reg query %image_import_reg_path% /v StartupScriptAttempt') do set startup_script_attempt=%%a
+if %ERRORLEVEL% EQU 0 (
+  reg delete %image_import_reg_path% /v StartupScriptAttempt /f
+) else (
+  set startup_script_attempt=0
+)
+set /a startup_script_attempt=%startup_script_attempt% + 1
+reg add %image_import_reg_path% /v StartupScriptAttempt /t REG_DWORD /d %startup_script_attempt%
+
+if %startup_script_attempt% EQU 1 (
+  echo "Translate: Starting image translate..." > COM1:
+) else (
+  echo "Translate: Starting image translate (attempt %startup_script_attempt%)..." > COM1:
+)
+
+REM Schedule the system restart. Translate.ps1 will cancel this restart when it starts.
 REM This is to address initial boot issues, mostly on 2008R2, where the network interface is not yet usable.
-echo "Scheduling restart in 5 minutes. Translate.ps1 will cancel this restart." > COM1:
-shutdown /r /t 300
+REM To avoid being stuck in a boot loop, limit the reboots count to 2.
+set /a script_timeout_m=%startup_script_attempt% * 5
+set /a script_timeout_s=%script_timeout_m% * 60
+if %startup_script_attempt% LEQ 2 (
+  echo "Scheduling restart in %script_timeout_m% minutes. Translate.ps1 will cancel this restart." > COM1: & shutdown /r /t %script_timeout_s%
+)
 
 echo "Running network.ps1 to reconfigure network to DHCP if needed and log DNS and connectivity tests." > COM1:
 PowerShell.exe -NoProfile -NoLogo -ExecutionPolicy Unrestricted -File "%ProgramFiles%\Google\Compute Engine\metadata_scripts\network.ps1" > COM1:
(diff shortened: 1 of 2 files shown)


daisy_workflows/image_import/windows/run_startup_scripts.cmd
```diff
@@ -16,12 +16,28 @@ REM limitations under the License.
...
 
-REM Restart the system in 5 minutes. Translate.ps1 will cancel this restart when it starts.
+for /f "tokens=3" %%a in ('reg query %image_import_reg_path% /v StartupScriptRunsCount') do set startup_script_runs_count=%%a
+if %ERRORLEVEL% EQU 0 (
+  reg delete %image_import_reg_path% /v StartupScriptRunsCount /f
+) else (
+  set startup_script_runs_count=0
+)
+set /a startup_script_runs_count=%startup_script_runs_count% + 1
+reg add %image_import_reg_path% /v StartupScriptRunsCount /t REG_DWORD /d %startup_script_runs_count%
```
EricEdens, line 28: I don't recall seeing registry changes in other parts for image import; I don't have enough Windows development experience to know whether or not this is an idiomatic approach.

I noticed that import creates temporary files, however, and cleans them up after import. So if you wanted to stay consistent with what already exists, then you could use a temporary file. Here's an example of where a temporary file gets removed:

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/91fe21ea4c88821b199fdef40602e3fa65967bf6/daisy_workflows/image_import/windows/translate.ps1#L393

daisy_workflows/image_import/windows/run_startup_scripts.cmd
```diff
@@ -16,12 +16,28 @@ REM limitations under the License.
 REM Give the network time to initialize.
 ping 127.0.0.1 -n 60
 
-echo "Translate: Starting image translate..." > COM1:
+set image_import_reg_path="HKLM\Software\ImageImport"
 
-REM Restart the system in 5 minutes. Translate.ps1 will cancel this restart when it starts.
+for /f "tokens=3" %%a in ('reg query %image_import_reg_path% /v StartupScriptRunsCount') do set startup_script_runs_count=%%a
```
EricEdens, line 21: Have you looked into creating an e2e test that fails prior to this fix, but passes after the fix is applied?

zoran15, line 21: Extra comments here would be useful, describing what's the purpose of this block

daisy_workflows/image_import/windows/run_startup_scripts.cmd
```diff
@@ -16,12 +16,28 @@ REM limitations under the License.
...
+set /a startup_script_runs_count=%startup_script_runs_count% + 1
+reg add %image_import_reg_path% /v StartupScriptRunsCount /t REG_DWORD /d %startup_script_runs_count%
+
+echo "Translate: Starting image translate (attempt %startup_script_runs_count%)..." > COM1:
+
+REM Schedule the system restart. Translate.ps1 will cancel this restart when it starts.
 REM This is to address initial boot issues, mostly on 2008R2, where the network interface is not yet usable.
-echo "Scheduling restart in 5 minutes. Translate.ps1 will cancel this restart." > COM1:
-shutdown /r /t 300
+REM The time to complete the script can vary from image to image, therefore the timeout is inferred incrementally.
```
zoran15, line 34:
```suggestion
REM The time to complete the script can vary from image to image, therefore the timeout is inferred incrementally by adding 5 minutes to the timeout after each restart..
```

daisy_workflows/image_import/windows/run_startup_scripts.cmd
yswe: I updated my PR to use the filesystem instead of the registry (see this [discussion](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1858#discussion_r812175449), please). I added some comments, however, I would be happy if you could have another look.

daisy_workflows/image_import/windows/run_startup_scripts.cmd
yswe: Done.

daisy_workflows/image_import/windows/run_startup_scripts.cmd
zoran15: Comments look good.

bkatyl (commented):
I believe this has gotten too complicated and complexity in this situation can easily lead to unexpected behavior.

The commands in this script should in the "worst case" take 1 minute to complete. In a normal situation it should be likey 10-20 seconds at most. I selected 5 minutes, because if it hits 5 minutes something has gone extremely wrong. 

If you want to limit the reboots, my suggestion would be a single reboot attempt as there is when it has the most utility. 

This is contrary to Eric's comment, but in Windows the registry is the better place to store a value you want to read. Declare a registry path that you know is unique and use that. In these situations where at the end your going to remove it, it's better to have it as a high level in the registry where as if you wanted to persist then you'd place it in a path that represents the company/area/application.

If you're going to delete the keys at the end of the import
HKEY_LOCAL_MACHINE\SOFTWARE\GCEImportProcess

If the keys will exists after the import
HKEY_LOCAL_MACHINE\SOFTWARE\Google\GCEImport

Why the difference? 
- At the end you'd delete HKEY_LOCAL_MACHINE\SOFTWARE\GCEImportProcess
- If you deleted HKEY_LOCAL_MACHINE\SOFTWARE\Google\GCEImport then you may have created a key of Google that wasn't there previously.

If you're unsure of how long these operations are taking to complete then I'd suggest outputing the time or how long the script has been running in the COM1 output so there is data to back up the duration of the operations.

daisy_workflows/image_import/windows/run_startup_scripts.cmd
yswe: Discussed verbally and agreed to test this fix manually only because adding an E2E test for this case sounds tricky.

yswe:
@bkatyl, thank you for your comment! I updated the code according to this comment with a small difference. The count of reboots is limited to 3; the first timeout value is 5 minutes, the second value is 10 minutes, and no limits for the third attempt.

We had a case when 5 minutes was not enough even to complete the `network.ps1` script, if I understood the idea behind scheduling the restart correctly, then this restart should happen after the `network.ps1` script is completed, therefore I decided to leave the auto-incremented timeout value.

EricEdens (approved):

PR #1869, opened 2022-02-28
Windows 10 import: Skip Config-Power

EricEdens:
We observed a Windows 10 64-bit import that failed when running `Configure-Power`. This PR skips `Configure-Power` for all versions of Windows 10 until we find a more granular filter.

### DIFF
--- a/daisy_workflows/image_import/windows/translate.ps1
+++ b/daisy_workflows/image_import/windows/translate.ps1
@@ -192,6 +192,11 @@ function Configure-Network {
 }
 
 function Configure-Power {
+  # The following function blocks some Windows 10 imports, so skip it.
+  if ($script:pn -like '*Windows 10*') {
+    Write-Output 'Skipping Configure-Power for Windows 10'
+    return
+  }
   if (-not (Get-Command Get-CimInstance -ErrorAction SilentlyContinue)) {
     return
   }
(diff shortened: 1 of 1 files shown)


zoran15 (approved):

bkatyl (approved):

daisy_workflows/image_import/windows/translate.ps1
```diff
@@ -356,10 +361,6 @@ try {
   else {
     # Since 32-bit GooGet packages are not provided via repository, the only option is to install them from a local source.
     Install-32bitPackages
```
hopkiw, line 363: there seem to be 2 dimensions, 32vs64? and Windows10 vs !Windows10; did you mean to make the function never run on 32bit as well as not running on Windows10?

EricEdens: Gosh, thanks for spotting this, I've been looking at this for too long today! My intention was to bypass `Configure-Power` for all versions of Windows 10, which will unblock the user that we're working with. In parallel to unblocking them, I'm working to see if we can create a better filter than just "Windows 10", since we're seeing that some versions of Windows 10 work here, while others fail.

hopkiw (approved):

PR #1870, opened 2022-03-01
Windows 10 import: Disable power savings

EricEdens:
We disable power saving in Windows on GCP: it causes the web UI's power button to _wakeup_ the Windows instance.  In Windows 10, we skipped this step since it caused translation to timeout (#1482, #1869).

On further investigation, I found that each call to `Get-CimAssociatedInstance -ResultClassName 'Win32_PowerSetting'` was slow on Windows 10--on the order of seconds--and there were over 1600 such calls.

To avoid this, this PR inverts the search: it finds the single power setting that we need, and then finds the subset of indexes that need to be toggled. It completes in less than a second.

## Testing
- There was an existing Windows 10 x86 test that failed using the previous query. With the new query, it passes.
- Updated the Windows post-translate tests to verify that "Turn off display after" is indeed set to zero.
- Ran all Windows e2e import tests

### DIFF
--- a/daisy_integration_tests/scripts/post_translate_test.ps1
+++ b/daisy_integration_tests/scripts/post_translate_test.ps1
@@ -42,6 +42,23 @@ function Check-Hiberation {
   }
 }
 
+function Check-Power-Settings {
+  $pplan = Get-CimInstance `
+                -Namespace 'root\cimv2\power' `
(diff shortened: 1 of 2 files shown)


bkatyl (approved):
Great job finding the underlying issue.

PR #1871, opened 2022-03-01
Only import the Google GPG keys for EL9.

zmarano:


zmarano:
Currently the `RPM-GPG-KEY-CentOS-SIG-Extras` key does not import and thus all other keys fail to import. It is unclear why it fails but it fails consistently everywhere in CentOS Stream 9. The base ISO should already pre-populate the rpm keyring with the CentOS key so this should be fine. We only need the Google repo keys.

PR #1876, opened 2022-03-08
RHUI: Runner for health checks

EricEdens:
A first pass at health checks for RHUI; for now, it includes a single check. The goal is to validate the overall approach prior to adding more checks.

To facilitate deployment, the script currently exposes the following flags:

```
`health_check.py \`
    --node {cds,rhua}  \
    --result_file RESULT_FILE \
    --nfs_mount NFS_MOUNT_POINT
```

Key points:
- The health checker runs on a timer, and writes the result (`{OK|ERROR}`) to a sentinel file that's read by nginx's health check handler
- Logs are written to stdout, allowing them to be queried with `journalctl`

### DIFF
--- a/daisy_workflows/build-publish/rhui/health_check.py
+++ b/daisy_workflows/build-publish/rhui/health_check.py
@@ -1,0 +1,137 @@
+#  Copyright 2022 Google Inc. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http:#www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Health check for RHUI infrastructure nodes."""
+
+import argparse
+import logging
+import logging.handlers
+import os
+import sys
+import tempfile
+import typing
+
+
+class HealthChecks:
+  nfs_mount: str
+
+  def __init__(
+      self,
+      nfs_mount: str,
+  ) -> None:
+    self.nfs_mount = nfs_mount
+
+  def mount_uses_nfs(self, logger: logging.Logger):
+    logger.info("checking that %s is mounted as NFS", self.nfs_mount)
+    with open("/proc/mounts") as mounts:
+      for line in mounts:
+        parts = line.split()
+        mnt_point, mnt_type = parts[1], parts[2]
+        if mnt_point == self.nfs_mount:
+          if mnt_type == "nfs":
+            return
+          else:
+            raise EnvironmentError("Not mounted as nfs. mount: %s" % line)
+    raise EnvironmentError("%s not in /proc/mounts" % self.nfs_mount)
+
+  def mount_is_writable(self, logger: logging.Logger):
+    logger.info("checking that %s is writable", self.nfs_mount)
+    with tempfile.TemporaryFile(dir=self.nfs_mount) as f:
+      f.write(b"content")
+
+  def mount_is_readable(self, logger: logging.Logger):
+    logger.info("checking that %s is readable", self.nfs_mount)
+    os.listdir(self.nfs_mount)
+
+
+def main(node_type: str, result_file: typing.TextIO, nfs_mount: str):
+  """Executes health check for node_type, and writes the overall
+  result to result_file ('OK' if all checks pass, otherwise 'ERROR').
+  """
+  result_file.truncate()
+  health_checks = HealthChecks(nfs_mount=nfs_mount)
+  checks = [
+    health_checks.mount_is_readable,
+    health_checks.mount_uses_nfs,
+  ]
+  if node_type == "rhua":
+    checks += [
+      health_checks.mount_is_writable,
+    ]
+  success = True
+  for func in checks:
+    logger = logging.getLogger(func.__name__)
+    try:
+      func(logger=logger)
+      logger.info("success")
+    except Exception as e:
+      success = False
+      logger.error(e)
+  if success:
+    result_file.write("HEALTHY\n")
+  else:
+    result_file.write("ERROR\n")
+
+
+if __name__ == "__main__":
+  parser = argparse.ArgumentParser(description="Run health checks.")
+  parser.add_argument(
+    "--node",
+    dest="node",
+    type=str,
+    nargs=1,
+    required=True,
+    choices=["cds", "rhua"],
+    help="type of node where check is running",
+  )
+  parser.add_argument(
+    "--result_file",
+    dest="result_file",
+    required=True,
+    type=argparse.FileType("w"),
+    help="file to write result",
+  )
+  parser.add_argument(
+    "--nfs_mount",
+    dest="nfs_mount",
+    type=str,
+    nargs=1,
+    required=True,
+    help="directory of NFS mount",
+  )
+  parser.add_argument(
+    "--log",
+    dest="log",
+    type=str,
+    nargs=1,
+    choices=["stdout", "syslog"],
+    default="syslog",
+    help="where to write logs",
+  )
+  args = parser.parse_args()
+  if args.log[0] == "stdout":
+    handler = logging.StreamHandler(sys.stdout)
+  else:
+    handler = logging.handlers.SysLogHandler(address="/dev/log")
+    # syslog parses on the colon to determine the tag for the message.
+    handler.ident = "rhui-health-check: "
+  logging.basicConfig(
+    level=logging.DEBUG,
+    format="%(name)s - %(levelname)s - %(message)s",
+    handlers=[handler],
+  )
+  main(
+    node_type=args.node[0],
+    result_file=args.result_file,
+    nfs_mount=args.nfs_mount[0],
+  )
--- a/daisy_workflows/build-publish/rhui/install_cds.sh
+++ b/daisy_workflows/build-publish/rhui/install_cds.sh
@@ -89,6 +89,13 @@ cat $tempdir/rhui-tools.conf > /etc/rhui/rhui-tools.conf
 # Remove enrollment cert and repos from final image.
 subscription-manager remove --all
 
+# Install health checks.
+install -D -t /opt/google-rhui-infra $tempdir/health_check.py
+for unit in rhui-health-check.{service,timer}; do
+  install -m 664 -t /etc/systemd/system $tempdir/$unit
+  systemctl enable $unit
+done
+
 # Delete installer resources.
 rm -rf $tempdir
 
--- a/daisy_workflows/build-publish/rhui/cds.wf.json
+++ b/daisy_workflows/build-publish/rhui/cds.wf.json
@@ -27,6 +27,7 @@
   "Sources": {
     "cds_artifacts": "./cds_artifacts",
     "cds_artifacts/rhui.crt": "${tls_cert_path}",
+    "cds_artifacts/health_check.py": "./health_check.py",
     "install_cds.sh": "./install_cds.sh"
   },
   "Steps": {
--- a/daisy_workflows/build-publish/rhui/cds_artifacts/rhui-health-check.service
+++ b/daisy_workflows/build-publish/rhui/cds_artifacts/rhui-health-check.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=CDS health check
+
+[Service]
+Type=oneshot
+ExecStart=python3 /opt/google-rhui-infra/health_check.py \
+    --node cds \
+    --result_file /var/log/google_rhui_health_check.txt \
+    --nfs_mount /var/lib/rhui/remote_share
+
+[Install]
+WantedBy=multi-user.target
--- a/daisy_workflows/build-publish/rhui/cds_artifacts/rhui-health-check.timer
+++ b/daisy_workflows/build-publish/rhui/cds_artifacts/rhui-health-check.timer
@@ -1,0 +1,10 @@
+[Unit]
+Description=Timer for CDS health check
+Requires=rhui-health-check.service
+
+[Timer]
+Unit=rhui-health-check.service
+OnUnitActiveSec=1m
+
+[Install]
+WantedBy=timers.target
--- a/daisy_workflows/build-publish/rhui/install_rhua.sh
+++ b/daisy_workflows/build-publish/rhui/install_rhua.sh
@@ -99,6 +99,13 @@ rhui-manager --noninteractive --user admin --password "$password" cert upload \
 rhui-manager --noninteractive --user admin --password "$password" repo \
   add_by_repo --repo_ids $(paste -sd "," "${tempdir}/reponames.txt")
 
+# Install health checks.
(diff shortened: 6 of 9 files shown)


daisy_workflows/build-publish/rhui/health_checks/health_check.py
```diff
@@ -0,0 +1,86 @@
...
+
+import argparse
+import logging
+import os
+import sys
+import typing
+
+
+def nfs_is_read_mounted(logger: logging.Logger):
+  nfs_mount = "/var/lib/rhui/remote_share"
```
hopkiw, line 24: can make constant

daisy_workflows/build-publish/rhui/health_checks/health_check.py
```diff
@@ -0,0 +1,86 @@
...
+  logger.info("checking that %s is mounted", nfs_mount)
+  os.statvfs(nfs_mount)
```
hopkiw, line 26: this will just ensure the mount *point* exists, not that it is mounted. i think what we want here is really to check /proc/self/mounts, idk if there's an existing python function for it (there may be, i don't know)

daisy_workflows/build-publish/rhui/health_checks/health_check.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def nfs_is_write_mounted(logger: logging.Logger):
+  nfs_mount = "/var/lib/rhui/remote_share"
+  logger.info("checking that %s is writable", nfs_mount)
+  stat = os.statvfs(nfs_mount)
+  if bool(stat.f_flag & os.ST_RDONLY):
+    raise EnvironmentError("%s is read only" % nfs_mount)
```
hopkiw, line 34: this checks the *permission of the mountpoint*, not the mount status. read-only vs read-write (which i assume is what you want to check). really, the simplest check is to first ensure it is mounted (can layer the other function for that), then write and delete a temp file. the tempfile package has context objects that do this automatically

daisy_workflows/build-publish/rhui/health_checks/rhui-health-check.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Executes RHUI health check
+
+[Service]
+Type=oneshot
+ExecStart=python3 /opt/google-rhui-infra/health_check.py \
+    --node rhua \
```
hopkiw, line 7: the name of this file is generic but it specifies rhua here, do we want to dupe this file for cds? or parameterize it in some way?

hopkiw (changes requested):
looks good overall, just a few points

daisy_workflows/build-publish/rhui/health_checks/rhui-health-check.service
EricEdens: I thought you'd spot that :)  you're right -- each node type should be correctly represented here. do you have a preference on the approach? (dupe versus a template versus something else)

daisy_workflows/build-publish/rhui/health_checks/health_check.py
EricEdens: I like the tempfile idea. In the latest version, if I mount NFS with `mount -o ro`, then the following error is generated:

```
mount_is_writable - INFO - checking that /mnt/nfs is writable
mount_is_writable - ERROR - [Errno 30] Read-only file system: '/mnt/nfs/tmpdi_qfs_d'
```

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+      nfs_mount: str,
+  ) -> None:
+    self.nfs_mount = nfs_mount
+
+  def mount_uses_nfs(self, logger: logging.Logger):
+    logger.info("checking that %s is mounted as NFS", self.nfs_mount)
+    # Using pipes to capture output since RHEL8 has Python 3.6
+    # which doesn't support `capture_output`.
+    p = subprocess.run(
+      ["stat", "-fc", "%T", self.nfs_mount],
```
hopkiw, line 39: it's a file, we really don't need to shell out to check it, python has all the tools to do it. if you want to kick this off now like this it's fine but i'd rather it just do things in a python directly

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+      stdout=subprocess.PIPE,
+      stderr=subprocess.PIPE,
+    )
+    if p.returncode != 0:
+      raise EnvironmentError("Failed to stat %s: %s" % (self.nfs_mount, p))
```
hopkiw, line 44: maybe a nit since i'm suggesting not to shell out but subprocess.run can automatically raise on nonzero

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+    logger.info("checking that %s is writable", self.nfs_mount)
+    with tempfile.TemporaryFile(dir=self.nfs_mount) as f:
+      written_bytes = b"content"
+      f.write(written_bytes)
+      f.seek(0)
+      actual = f.read()
+      if actual != written_bytes:
+        logger.info(
+          "expected file to contain %s actual %s", written_bytes, actual
+        )
```
hopkiw, line 59: we don't need to validate that writes are then read - this is like second guessing the kernel

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+    health_checks.mount_is_readable,
+    health_checks.mount_uses_nfs,
+  ]
+  if node_type == "rhua":
+    checks += [
+      health_checks.mount_is_writable,
+    ]
+  success = True
+  for func in checks:
+    logger = logging.getLogger(func.__name__)
```
hopkiw, line 82: is the idea that we would get named output logs without having the health checks themselves do logging?

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+    try:
+      func(logger=logger)
+      logger.info("success")
+    except BaseException as e:
```
hopkiw, line 86: i'm not familiar with baseexception, but isn't this like catching `Exception` ?

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+      success = False
+      logger.error(e)
+  if success:
+    result_file.write("OK\n")
+  else:
+    result_file.write("ERROR\n")
```
hopkiw, line 92: how does the health check system work? does it treat "ERROR" specially? or is absence of an OK enough?

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+if __name__ == "__main__":
+  logging.basicConfig(
+    stream=sys.stdout,
+    level=logging.DEBUG,
```
hopkiw, line 98: ?

daisy_workflows/build-publish/rhui/install_cds.sh
```diff
@@ -89,6 +89,14 @@ cat $tempdir/rhui-tools.conf > /etc/rhui/rhui-tools.conf
 # Remove enrollment cert and repos from final image.
 subscription-manager remove --all
 
+# Install health checks
+install -g root -o root -D -t /opt/google-rhui-infra $tempdir/health_check.py
```
hopkiw, line 93: nit but this is overkill - the script is already running as root. i also had to look up all these args :laughing:

daisy_workflows/build-publish/rhui/install_cds.sh
```diff
@@ -89,6 +89,14 @@ cat $tempdir/rhui-tools.conf > /etc/rhui/rhui-tools.conf
...
+for unit in health-check.{service,timer}; do
+  install -g root -o root -m 644 -t /etc/systemd/system $tempdir/$unit
+  systemctl enable $unit
+done
+systemctl daemon-reload
```
hopkiw, line 98: nit but not necessary, `systemctl enable` transparently reloads

daisy_workflows/build-publish/rhui/rhua_artifacts/health-check.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Executes RHUA health check
```
hopkiw, line 2: nit but description field is really more of a name field:
```
This may be used by systemd (and other UIs) as a user-visible label for the unit,
so this string should identify the unit rather than describe it, despite the name.
```

hopkiw (changes requested):
minor comments, mostly nits. looking good :sunglasses:

daisy_workflows/build-publish/rhui/install_cds.sh
EricEdens: Updated to remove the user and group

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: Without this, the logs went to standard error, and only debug messages were shown. In the latest version I tweaked it a bit to write directly to syslog, so that we get identifiers and levels.

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: You configure the health checkers to look for a substring in the HTTP response to consider the check healthy. I updated it to `HEALTHY` for the healthy check, and `` for the unhealthy check.

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: BaseException is the root exception in Python's exception hierarchy tree.

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: Yes -- each health check gets a logger that's scoped to its name. Also, this loop handles logging success and failure messages.

Here's some examples from journalctl:

## Prior to mounting NFS:

The error messages appear in read, and they are filterable by with `journalctl -p 3`

```
Mar 24 18:30:46 rhel-8 systemd[1]: Starting RHUA health check...
Mar 24 18:30:46 rhel-8 rhui-health-check[1916]: mount_is_readable - INFO - checking that /var/lib/rhui/remote_share is readable
Mar 24 18:30:46 rhel-8 rhui-health-check[1916]: mount_is_readable - ERROR - [Errno 2] No such file or directory: '/var/lib/rhui/remote_share'
Mar 24 18:30:46 rhel-8 rhui-health-check[1916]: mount_uses_nfs - INFO - checking that /var/lib/rhui/remote_share is mounted as NFS
Mar 24 18:30:46 rhel-8 rhui-health-check[1916]: mount_uses_nfs - ERROR - /var/lib/rhui/remote_share not in /proc/mounts
Mar 24 18:30:46 rhel-8 rhui-health-check[1916]: mount_is_writable - INFO - checking that /var/lib/rhui/remote_share is writable
Mar 24 18:30:46 rhel-8 rhui-health-check[1916]: mount_is_writable - ERROR - [Errno 2] No such file or directory: '/var/lib/rhui/remote_share/tmp20xplcm1'
Mar 24 18:30:46 rhel-8 systemd[1]: rhui-health-check.service: Succeeded.
```

## After mounting NFS:

```
Mar 24 18:33:14 rhel-8 systemd[1]: Starting RHUA health check...
Mar 24 18:33:14 rhel-8 rhui-health-check[1966]: mount_is_readable - INFO - checking that /var/lib/rhui/remote_share is readable
Mar 24 18:33:14 rhel-8 rhui-health-check[1966]: mount_is_readable - INFO - success
Mar 24 18:33:14 rhel-8 rhui-health-check[1966]: mount_uses_nfs - INFO - checking that /var/lib/rhui/remote_share is mounted as NFS
Mar 24 18:33:14 rhel-8 rhui-health-check[1966]: mount_uses_nfs - INFO - success
Mar 24 18:33:14 rhel-8 rhui-health-check[1966]: mount_is_writable - INFO - checking that /var/lib/rhui/remote_share is writable
Mar 24 18:33:14 rhel-8 rhui-health-check[1966]: mount_is_writable - INFO - success
Mar 24 18:33:14 rhel-8 systemd[1]: rhui-health-check.service: Succeeded.
```

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: done

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: Not used anymore after change to `/proc/mounts`

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: sgtm; changed to read `/proc/mounts`

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -0,0 +1,136 @@
...
+  for func in checks:
+    logger = logging.getLogger(func.__name__)
+    try:
+      func(logger=logger)
+      logger.info("success")
+    except BaseException as e:
+      success = False
+      logger.error(e)
+  if success:
+    result_file.write("HEALTHY\n")
```
hopkiw, line 80: should we delete or truncate the file at start of run? i'm thinking of what happens if we write healthy at one point and then fail to write it again afterward.

daisy_workflows/build-publish/rhui/health_check.py
hopkiw: OK, but it still falls under the category of too-broad catch, right?

daisy_workflows/build-publish/rhui/install_cds.sh
```diff
@@ -89,6 +89,15 @@ cat $tempdir/rhui-tools.conf > /etc/rhui/rhui-tools.conf
 # Remove enrollment cert and repos from final image.
 subscription-manager remove --all
 
+# Install health checks.
+install -D -t /opt/google-rhui-infra $tempdir/health_check.py
+for unit in rhui-health-check.{service,timer}; do
+  # RH recommends 664 for unit files
+  # https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_working-with-systemd-unit-files_configuring-basic-system-settings
```
hopkiw, line 96: it's a nit but that's neither backed up by the link nor is RH in any sense a reputable source of truth for things like this. i don't care what permissions we use (it literally just needs to be readable by root) but let's strike the comment

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: sgtm; updated to truncate the file.

daisy_workflows/build-publish/rhui/install_cds.sh
EricEdens: makes sense -- removed the RH link

daisy_workflows/build-publish/rhui/health_check.py
EricEdens: I agree that broad exception catching can be an anti-pattern. 

In this case, it was intentional, to ensure that the health checker continues running, and that any exception that's thrown is attributed (and logged) to the case that failed.

hopkiw (approved):

PR #1878, opened 2022-03-09
Add ARM64 worker image

dntczdx:
We can't simply run "chroot /mnt apt-get install -y /tmp/google-guest-agent_20210927.00-g1_arm64.deb" from a amd64 machine since it's cross-arch.

We need to run it from an arm64 machine. So we support new arg "machine-type" and "wotker-image".

Meanwhile, we need debian-11-worker-arm64 to play this worker role. A regular debian11-arm64 image doesn't have necessary packages.


daisy_workflows/image_build/debian/debian_11_worker_arm64.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
hopkiw, line 1: i don't think we need a completely cloned file for this; can we just introduce some 'if' branching to the main script instead?

hopkiw (changes requested):

PR #1880, opened 2022-03-10
cds build: include security section

hopkiw:
checked by gunicorn-auth service; missing section crashes the service


dntczdx (approved):
LGTM.

PR #1889, opened 2022-03-21
Fix getDeviceForMount call

EricEdens:
In https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1743 we improved mount detection for the precheck tool; when doing so, we used the wrong CLI binary -- it should have been `findmnt`.

Found by @yswe, thanks!

### DIFF
--- a/cli_tools/common/mount/inspector.go
+++ b/cli_tools/common/mount/inspector.go
@@ -66,7 +66,7 @@ func (mi *defaultMountInspector) Inspect(dir string) (mountInfo InspectionResult
 
 // getDeviceForMount returns the path of the block device that is mounted for dir.
 func (mi *defaultMountInspector) getDeviceForMount(dir string) (string, error) {
-	stdout, err := mi.shellExecutor.Exec("getDeviceForMount", "--noheadings", "--output=SOURCE", dir)
+	stdout, err := mi.shellExecutor.Exec("findmnt", "--noheadings", "--output=SOURCE", dir)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


cli_tools/common/mount/inspector.go
```diff
@@ -66,7 +66,7 @@ func (mi *defaultMountInspector) Inspect(dir string) (mountInfo InspectionResult
 
 // getDeviceForMount returns the path of the block device that is mounted for dir.
 func (mi *defaultMountInspector) getDeviceForMount(dir string) (string, error) {
-	stdout, err := mi.shellExecutor.Exec("getDeviceForMount", "--noheadings", "--output=SOURCE", dir)
+	stdout, err := mi.shellExecutor.Exec("findmnt", "--noheadings", "--output=SOURCE", dir)
```
hopkiw, line 69: the only thing i have to say here is that you don't need tools like findmnt when you're in go. there are packages to do this because the list of mounted filesystems is available to every process at /proc/self/mounts. unsure if the change is appropriate to this PR but just as a suggestion, minimizing execs usually produces more reliable code :)

zoran15:
Are there any E2E tests that will pick this issue in the future?

cli_tools/common/mount/inspector.go
EricEdens: I think this is a good suggestion; lacking tests that @zoran15 mentioned, I wouldn't be confident that my custom-parsing logic for `/proc/self/mounts` would work everywhere. (Although I would hope that the format of this file would be consistent across distros and versions.)

EricEdens:
> Are there any E2E tests that will pick this issue in the future?
We don't. I spent about 30 minutes trying to make something work, and from that, I estimate it'd take a couple of days total (between writing a new e2e test suite, wiring up the container builds, wiring up prow, and wiring up testgrid). I created an internal FR to track it however.


@zoran15

zoran15 (approved):

cli_tools/common/mount/inspector.go
hopkiw: i meant to use an existing 3p golang module for it, but yes the format is stable and provided by the kernel directly.

PR #1892, opened 2022-03-24
[windows upgrade e2e] refactor to support upgrading for more new versions

dntczdx:
This PR refactored the code to generalize the code logic of windows_upgrade so the existing logic can be easily applied to new version. It won't be exposed to public by gcloud as a google-supported feature. Instead, it only supports the e2e test: the monthly release of windows-install-media can be verified by this tool.

1. It doesn't change any existing behaviors of 2008r2->2012r2 upgrade
2. It doesn't add any new e2e test for new versions yet.
3. It supports the new e2e test for new versions in the future.


bkatyl (commented):
Since this changes the conditional logic on if an upgrade can occur. 

I think this needs on hold until there is a reviewed and approved design. Also to ensure the billing team agrees with the approach.

dntczdx:
> Since this changes the conditional logic on if an upgrade can occur.
> 
> I think this needs on hold until there is a reviewed and approved design. Also to ensure the billing team agrees with the approach.

Would you tell which conditional logic is changed? I don't change any logic when I refactor it, unless I missed something.

cli_tools/gce_windows_upgrade/upgrade_script.ps1
```diff
@@ -56,14 +85,14 @@ online disk noerr
...
   Set-ExecutionPolicy Unrestricted
-  Set-Location "$($script:install_media_drive)/Windows_Svr_Std_and_DataCtr_2012_R2_64Bit_English"
+  Set-Location "$($script:install_media_drive)/$script:install_folder"
   ./upgrade.ps1
-  $new_ver=[System.Environment]::OSVersion.Version
-  if ("$($ver.Major).$($ver.Minor)" -eq "6.3")
+  $new_ver=(Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").ProductName
+  if ($new_ver -eq $script:expected_new_version)
   {
     Write-Host "GCEMetadataScripts: post-upgrade step is done."
```
bkatyl, line 93: Why does this use a prefix of `GCEMetadataScripts`? Shouldn't it have a prefix of the upgrade process?

cli_tools/gce_windows_upgrade/upgrade_script.ps1
```diff
@@ -12,21 +12,50 @@
...
   # Cleanup garbage files left by the previous failed upgrade to unblock a new upgrade.
   Remove-Item 'C:\$WINDOWS.~BT' -Recurse -ErrorAction SilentlyContinue
 
-  $ver=[System.Environment]::OSVersion.Version
-  if ("$($ver.Major).$($ver.Minor)" -ne "6.1") {
-    if ("$($ver.Major).$($ver.Minor)" -eq "6.3") {
-      Write-Host "GCEMetadataScripts: The instance is already running Windows 2012R2!"
+  $ver=(Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").ProductName
+  if ($ver -ne $script:expected_current_version) {
+    if ($ver -eq $script:expected_new_version) {
```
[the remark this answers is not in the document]
dntczdx: It's not considered since previously we trust that the "license" matches the os version. As you said, to be more secure, we should exclude any unexpected version here as well. Let me add that logic.
We don't use "<" or ">" here since version may not be simply higher or lower, it can just be unexpected. For example, windows 11 vs windows server 2022.

cli_tools/gce_windows_upgrade/upgrade_script.ps1
dntczdx: The whole script is a startup script. This script will output some important status, like "GCEMetadataScripts: Beginning upgrade startup script." Our workflow needs to watch this message to ensure the machine is boot and started executing the upgrade script. We can name it with any prefix, but this prefix indicates everything happens due to the launch of GCEMetadata startup script in my opinion.

bkatyl: If upgrade_script.ps1 is launched using metadata startup scripts then any output from the script is captured by the GCEMetadataScripts binary and it includes the logging prefix of "GCEMetadataScripts: ".

So my concern is that the output in the serial port will look like:
"GCEMetadataScripts: GCEMetadataScripts:  post-upgrade step is done."

cli_tools/gce_windows_upgrade/upgrader/upgrade_paths.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+	return u
+}
+
+var upgradePaths = initUpgradePath(map[string]map[string]*upgradePath{
+	versionWindows2008r2: {
+		versionWindows2012r2: {enabled: true},
+	},
+	versionWindows2012r2: {
```
bkatyl, line 93: Doesn't "enabled: true" enabled upgrades for 2012r2/2016/2019?

cli_tools/gce_windows_upgrade/upgrader/validators.go
```diff
@@ -200,21 +210,23 @@ func validateAndDeriveOSDisk(osDisk *compute.AttachedDisk, derivedVars *derivedV
...
+			if strings.HasSuffix(lic, expectedLic) {
+				matchSourceOSVersion = true
+			} else if strings.HasSuffix(lic, upgradePaths[sourceOS][targetOS].licenseToAdd) {
+				upgraded = true
+			}
 		}
 	}
 	if !matchSourceOSVersion {
-		return daisy.Errf(fmt.Sprintf("Can only upgrade GCE instance with %v license attached", expectedCurrentLicense[sourceOS]))
+		return daisy.Errf(fmt.Sprintf("Can only upgrade GCE instance with any of %v license attached", upgradePaths[sourceOS][targetOS].expectedCurrentLicense))
```
bkatyl, line 226: I'd suggest a clearer error message.

No valid Windows Server PayG license can be found. Any of the following licenses are required %v"

cli_tools/gce_windows_upgrade/upgrader/validators.go
```diff
@@ -200,21 +210,23 @@ func validateAndDeriveOSDisk(osDisk *compute.AttachedDisk, derivedVars *derivedV
...
 	}
 	if upgraded {
-		return daisy.Errf(fmt.Sprintf("The GCE instance is with %v license attached, which means it either has been upgraded or has started an upgrade in the past.", licenseToAdd[sourceOS]))
+		return daisy.Errf(fmt.Sprintf("The GCE instance is with %v license attached, which means it either has been upgraded or has started an upgrade in the past.", upgradePaths[sourceOS][targetOS].licenseToAdd))
```
bkatyl, line 229: The GCE instance has the %v license attached. This likely means the instance has already been upgraded.

bkatyl, line 229: Can we ever have a situation where the upgrade license is attached, but the OS has failed to upgrade?

cli_tools/gce_windows_upgrade/upgrader/workflows.go
```diff
@@ -174,8 +179,8 @@ func populatePrepareSteps(u *upgrader, w *daisy.Workflow) error {
 
 	// If there isn't an original url, just skip the backup step.
 	if u.originalWindowsStartupScriptURL != nil {
-		fmt.Printf("\nDetected an existing metadata for key '%v', value='%v'. Will backup to '%v'.\n\n", metadataKeyWindowsStartupScriptURL,
-			*u.originalWindowsStartupScriptURL, metadataKeyWindowsStartupScriptURLBackup)
+		fmt.Printf("\nDetected an existing metadata for key '%v', value='%v'. Will backup to '%v'.\n\n", metadataWindowsStartupScriptURL,
+			*u.originalWindowsStartupScriptURL, metadataWindowsStartupScriptURLBackup)
```
bkatyl, line 183: The order of execution for metadata startup scripts is here: https://cloud.google.com/compute/docs/instances/startup-scripts/windows#order_of_execution_of_windows_startup_scripts

Since you're using windows-startup-script-url, you'll also want to make sure that windows-startup-script-ps1, windows-startup-script-cmd, and windows-startup-script-bat don't have any values set.

cli_tools/gce_windows_upgrade/upgrade_script.ps1
dntczdx: You are completely right, it's duplicated. Removed it from this script.

cli_tools/gce_windows_upgrade/upgrader/upgrade_paths.go
dntczdx: In the previous commit, they should be disabled; now they all should be enabled for the e2e purpose. The gcloud gated them for the real customers, so it's not a problem to enable them for e2e.

cli_tools/gce_windows_upgrade/upgrader/validators.go
dntczdx: done

cli_tools/gce_windows_upgrade/upgrader/validators.go
dntczdx: yes. We have to attach the license when creating the new boot disk, which is earlier than the actual upgrade.

dntczdx: adopted this words but still keeps "started an upgrade"

cli_tools/gce_windows_upgrade/upgrader/workflows.go
dntczdx: You are right, that part is missing for both manual guide and automation.
It's a bigger behavior change. Since right now my purpose is to enable the higher version upgrade only for e2e purpose (not open to customers), let's not change this behavior right now.

bkatyl (approved):

PR #1893, opened 2022-03-28
nginx for rhua health checks

EricEdens:
- Health checks run on `8080`; RHUA listens on `80`, so I didn't want to interfere with anything that depends on that.
- nginx on the CDS node already includes `/etc/nginx/conf.d` drop-ins. RHUA required patching.
- I moved the results file from the health check to nginx's default HTML directory

### DIFF
--- a/daisy_workflows/build-publish/rhui/install_rhua.sh
+++ b/daisy_workflows/build-publish/rhui/install_rhua.sh
@@ -105,6 +105,7 @@ for unit in rhui-health-check.{service,timer}; do
   install -m 664 -t /etc/systemd/system $tempdir/$unit
   systemctl enable $unit
 done
+install -m 664 -t /etc/nginx/conf.d $tempdir/health_check.nginx.conf
 
 # Add NFS dependency to pulp units
 # We do this instead of patching the Ansible templates, as we need the
--- a/daisy_workflows/build-publish/rhui/rhua_artifacts/rhua.patch
+++ b/daisy_workflows/build-publish/rhui/rhua_artifacts/rhua.patch
@@ -73,3 +73,14 @@ diff -Naur playbooks/rhua-provision.yml playbooks/rhua-provision.yml
  
      - name: Collect static browsable API and media files
        command: "{{ pulpcore_env }} pulpcore-manager collectstatic --noinput"
+
+--- templates/nginx-pulp.conf	2022-01-01 23:56:33.757374995 +0000
++++ templates/nginx-pulp.conf	2022-01-01 23:56:47.768584643 +0000
+@@ -10,6 +10,7 @@
+ }
+
+ http {
++    include /etc/nginx/conf.d/*.conf;
+     include mime.types;
+     # fallback in case we can't determine a type
+     default_type application/octet-stream;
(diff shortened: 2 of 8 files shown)


daisy_workflows/build-publish/rhui/install_rhua.sh
```diff
@@ -105,6 +105,8 @@ for unit in rhui-health-check.{service,timer}; do
   install -m 664 -t /etc/systemd/system $tempdir/$unit
   systemctl enable $unit
 done
+( cd /etc/nginx; patch -b < $tempdir/nginx.conf.patch; )
```
hopkiw, line 108: do you think we should follow our existing paradigm and patch the ansible template, instead?

EricEdens: I'm not sure of the benefit in this case. Were you thinking of inserting a step for [ansible.posix.patch](https://docs.ansible.com/ansible/latest/collections/ansible/posix/patch_module.html) after this step? 

```
    - name: Provide NginX Pulp specific configuration
      template:
        src: "{{ template_prefix }}/nginx-pulp.conf"
        dest: /etc/nginx/nginx.conf
        owner: root
        group: root
        mode: 0644
      notify: Restart NginX
```

What benefits were you thinking?

hopkiw: mainly consistency, so we can see all the changes we're making in one spot. i think we will want to ask for this particular change to be included in the upstream ansible template, too

EricEdens: OK, sgtm. Updated to use ansible for patching `nginx-pulp.conf`

daisy_workflows/build-publish/rhui/rhua_artifacts/rhua.patch
```diff
@@ -73,3 +73,16 @@ diff -Naur playbooks/rhua-provision.yml playbooks/rhua-provision.yml
...
+@@ -402,6 +407,12 @@
+     - name: Set SELinux httpd_can_network_connect boolean and keep it persistent across reboots
+       command: /usr/sbin/setsebool -P httpd_can_network_connect on
+ 
++    - name: Allow drop-ins for NginX
++      lineinfile:
++        dest: "{{ template_prefix }}/nginx-pulp.conf"
++        insertafter: "http {"
++        line: "    include /etc/nginx/conf.d/*.conf;"
++
```
hopkiw, line 85: i think we should change this to a *literal* patch to the file nginx-pulp, included in this file rhua.patch

this approach would add a step to the ansible workflow to patch its own template before use, which is much harder to reason about

hopkiw (changes requested):

daisy_workflows/build-publish/rhui/rhua_artifacts/rhua.patch
```diff
@@ -73,3 +73,16 @@ diff -Naur playbooks/rhua-provision.yml playbooks/rhua-provision.yml
...
+     - name: Provide NginX Pulp specific configuration
+       template:
+         src: "{{ template_prefix }}/nginx-pulp.conf"
```
hopkiw, line 88:
```suggestion

--- templates/nginx-pulp.conf	2022-01-01 23:56:33.757374995 +0000
+++ templates/nginx-pulp.conf	2022-01-01 23:56:47.768584643 +0000
@@ -10,6 +10,7 @@
 }
 
 http {
+    include /etc/nginx/conf.d/*.conf
     include mime.types;
     # fallback in case we can't determine a type
     default_type application/octet-stream;
```

hopkiw (approved):

PR #1895, opened 2022-03-28
e2e test for debian-11-worker-arm64

dntczdx:


daisy_integration_tests/daisy_e2e.test.gotmpl
```diff
@@ -116,5 +116,8 @@
     "Image import allow spaces in disk's filename.": {
      "Path": "./image_import_allow_spaces_in_disk_filename.wf.json"
     }
+    "Test debian worker arm64": {
```
hopkiw, line 119: i can't really say 100% what this test suite is for (half daisy and half image import, looks like), but it's definitely not the one we use for image validations. moreover, i don't think a whole workflow after the build workflow to verify that it has 3 packages installed is valuable. suggest you delete this, or we can discuss where else it can live.

hopkiw (changes requested):

dntczdx:
As Liam said, it's a very limited verification and it's not too helpful. Close it for now.

PR #1896, opened 2022-03-30
ImageImport_Add_support_for_windows_server_2022

MahmoudNada0:
/cc EricEdens
/cc yswe

### DIFF
--- a/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
+++ b/cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
@@ -185,7 +185,7 @@ var ovfOSIDToImporterOSID = map[int16]OsInfo{
 	114: {description: "Microsoft Windows 8", importerOSIDs: []string{"windows-8-x86-byol"}, nonDeterministic: true},
 	115: {description: "Microsoft Windows 8 64-bit", importerOSIDs: []string{"windows-8-x64-byol"}, nonDeterministic: true},
 	116: {description: "Microsoft Windows Server 2012 R2", importerOSIDs: []string{"windows-2012r2", "windows-2012r2-byol"}},
-	117: {description: "Microsoft Windows Server 2016", importerOSIDs: []string{"windows-2016", "windows-2016-byol", "windows-2019", "windows-2019-byol"}}, //as per https://www.ibm.com/support/knowledgecenter/SSZQFR_2.3.2.0/iwd/pct_byos_win_image.html 2016/2019 share the same ID
+	117: {description: "Microsoft Windows Server 2016", importerOSIDs: []string{"windows-2016", "windows-2016-byol", "windows-2019", "windows-2019-byol", "windows-2022", "windows-2022-byol"}}, //as per https://www.ibm.com/support/knowledgecenter/SSZQFR_2.3.2.0/iwd/pct_byos_win_image.html 2016/2019/2022 share the same ID
 	118: {description: "Microsoft Windows 8.1", importerOSIDs: []string{}},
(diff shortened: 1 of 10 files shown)


EricEdens:
/lgtm

:rocket:

cli_tools/gce_ovf_import/ovf_utils/ovf_utils.go
```diff
@@ -185,7 +185,7 @@ var ovfOSIDToImporterOSID = map[int16]OsInfo{
 	114: {description: "Microsoft Windows 8", importerOSIDs: []string{"windows-8-x86-byol"}, nonDeterministic: true},
 	115: {description: "Microsoft Windows 8 64-bit", importerOSIDs: []string{"windows-8-x64-byol"}, nonDeterministic: true},
 	116: {description: "Microsoft Windows Server 2012 R2", importerOSIDs: []string{"windows-2012r2", "windows-2012r2-byol"}},
-	117: {description: "Microsoft Windows Server 2016", importerOSIDs: []string{"windows-2016", "windows-2016-byol", "windows-2019", "windows-2019-byol"}}, //as per https://www.ibm.com/support/knowledgecenter/SSZQFR_2.3.2.0/iwd/pct_byos_win_image.html 2016/2019 share the same ID
+	117: {description: "Microsoft Windows Server 2016", importerOSIDs: []string{"windows-2016", "windows-2016-byol", "windows-2019", "windows-2019-byol", "windows-2022", "windows-2022-byol"}}, //as per https://www.ibm.com/support/knowledgecenter/SSZQFR_2.3.2.0/iwd/pct_byos_win_image.html 2016/2019/2022 share the same ID
```
yswe, line 188: It looks like the attached document contains the information only about Windows 2016 and Windows 2019.

May I ask you to double-check that Windows 2022 shares the same ID and update the comment with a new link?

MahmoudNada0: I think I'm mistaken here, I validated that the osID of windows-2022 is 117 as I see it from an exported ovf file, however I found that ovf-export uses the same mappings (it uses 117 value that i added) :D
Anyway, I think the mapping value for windows-2022 still not mentioned in any of the [vmware docs](https://wiki.abiquo.com/display/doc/Guest+operating+system+definition+for+VMware) or any other docs, so I'd remove these value & create a ticket like this (b/162248830) for windows-2022. what do u think?

yswe: It sounds good to me. Will we fail the export in this case, have you checked?

MahmoudNada0: Yes I check it, the export works normally but it set the OSid=0 (default value) in the ovf file, and if we import it again it should work fine because the OS will be detected from the inspection step of the boot disk.

yswe: I see, can we write `log.warn` in the case of Windows Server 2022, I am afraid that the exported OVF won't be bootable in VMware, therefore at least users will be aware of the issue?

MahmoudNada0: There's no warnings for any "not found" OS, not only windows-2022 ([link](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools/gce_ovf_export/exporter/ovf_descriptor_generator.go#L189)), may be we can make it in another PR as it will require changes in some files not only a print line, what do u think?

EricEdens (approved):

PR #1900, opened 2022-04-11
rhua: update create-dirs.conf

hopkiw:
* create dirs as root
* set ownership after creation


daisy_workflows/build-publish/rhui/rhua_artifacts/create-dirs.conf
```diff
@@ -1,4 +1,11 @@
 [Service]
-ExecStartPre=mkdir -p /var/lib/rhui/remote_share/pulp3/artifact
```
dntczdx, line 2: I dont see a change about "create dirs as root", is it already there?

hopkiw: discussed offline: the `PermissionsStartOnly` directive changes the `ExecStartPre` directives to run as root rather than as the user configured in the service unit

PR #1905, opened 2022-04-18
Update RHUI health checks

hopkiw:
* add cds and rhua service checks
* move nfs mount path to constant
* always log to stdout (systemd will get it to syslog for us)
* always run read & write checks on all node types (mount options aren't different across types)
* simplify nfs mount type check for readability

the final commit refactors the plugin architecture to use classes, but this is just an idea i had. feel free to review excluding that commit. output format is maintained (added fake service name to the check in this example to force a failure):
```
MountIsNFS - INFO - checking that /var/lib/rhui/remote_share is mounted as NFS
MountIsNFS - INFO - success
MountIsWritable - INFO - checking that /var/lib/rhui/remote_share is writable
MountIsWritable - INFO - success
MountIsReadable - INFO - checking that /var/lib/rhui/remote_share is readable
MountIsReadable - INFO - success
ServicesAreActive - INFO - checking that rhua services are active
ServicesAreActive - ERROR - Command '['systemctl', 'is-active', 'fakeservice']' returned non-zero exit status 3.
```

### DIFF
--- a/daisy_workflows/build-publish/rhui/health_check.py
+++ b/daisy_workflows/build-publish/rhui/health_check.py
@@ -20,7 +20,6 @@ import os
 import subprocess
 import sys
 import tempfile
-import typing
 
 NFS_MOUNT = "/var/lib/rhui/remote_share"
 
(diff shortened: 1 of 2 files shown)


daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -101,37 +140,21 @@ def main(node_type: str, result_file: typing.TextIO, nfs_mount: str):
...
-  )
   args = parser.parse_args()
-  if args.log[0] == "stdout":
-    handler = logging.StreamHandler(sys.stdout)
-  else:
-    handler = logging.handlers.SysLogHandler(address="/dev/log")
-    # syslog parses on the colon to determine the tag for the message.
-    handler.ident = "rhui-health-check: "
+
+  handler = logging.StreamHandler(sys.stdout)
```
EricEdens, line 145: If I recall, the syslog handler was helpful since it propagated the error level (so you can query it) and gave a more useful tag name.

daisy_workflows/build-publish/rhui/health_check.py
```diff
@@ -101,37 +140,21 @@ def main(node_type: str, result_file: typing.TextIO, nfs_mount: str):
...
-    node_type=args.node[0],
-    result_file=args.result_file,
-    nfs_mount=args.nfs_mount[0],
-  )
+
+  args.result_file.truncate()
+  if HealthCheckSuite(args.node):
+    args.result_file.write("HEALTHY\n")
+  else:
+    args.result_file.write("ERROR\n")
```
EricEdens, line 156: While you're here, I was thinking it'd be helpful to use `exit(1)` when there's a failure so that systemd's logs show there's a failure for the overall unit. If I remember correctly, in the serial output logs for the instance, when the health check fails, it looks like it's passing since systemd reports that the unit was successful.

daisy_workflows/build-publish/rhui/health_check.py
hopkiw: could be, but i don't think that is enough to merit using syslog directly. if you want to upgrade to structued cloud logging we can discuss that :)

daisy_workflows/build-publish/rhui/health_check.py
hopkiw: the health check *is* passing, though. it is correctly running and determining the health status of the node, which is its job. what are the scenarios where you'd want systemd to see the health check itself as failed? does it have some impact on the timer?

PR #1924, opened 2022-05-12
remove cds to rhua dependency

hopkiw:
Remove the CDS to RHUA fallback, which doesn't work and causes all 404s to timeout
Tested config change on active CDS node, and tested build change with Daisy.


dntczdx (approved):
lgtm

PR #1931, opened 2022-05-27
Move to the real cloud sig repo and arch aware repos.

zmarano:


hopkiw:
> It looks like this also fixes for making the endpoints architecture aware; please update the commit message to also reflect this (or split into separate PRs).

this is actually the norm for this kind of repo file.

zmarano:
> It looks like this also fixes for making the endpoints architecture aware; please update the commit message to also reflect this (or split into separate PRs).

Done. And yes it is the norm to use these macros for repo configs.

PR #1936, opened 2022-06-01
Rewrite kickstart configs for EL image builds

hopkiw:
We have been using static 'snippets' with code to make decisions about merging them into final generated kickstart configs. However, the variances in these files have made this unwieldy. Change to static config files for ease of management.

recommend reading the commits one at a time for reviewing.


daisy_workflows/image_build/enterprise_linux/rhel_7_4_sap.wf.json
```diff
@@ -33,6 +33,7 @@
         "Path": "./enterprise_linux.wf.json",
         "Vars": {
           "el_release": "${rhel_release}",
+          "kickstart_config": "./kickstart/rhel_7_4_sap.cfg",
```
zmarano, line 36: 7.4 is EOL and we can remove it entirely. 7.6 is currently the lowest version.

zmarano (commented):
Since the configs were generated from the old method I'm fairly confident we'll have the same result.

daisy_workflows/image_build/enterprise_linux/rhel_7_4_sap.wf.json
hopkiw: i was trying to keep in line with the 'one change per PR' policy that's been suggested :) follow up changes include removing unused args and pruning now-unused workflows

PR #1948, opened 2022-06-09
CDS: Collect nginx metrics and logs

EricEdens:
This configures the Ops Agent to collect logs and metrics from nginx on the CDS node. I followed the [ops agent onboarding guide for nginx](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/nginx), with a couple of changes:

1. The access logs for client traffic are stored in  `/var/log/nginx/ssl-access.log`
2. I disabled all access to port `80`, except when the connection is from localhost.

## Testing

Manually applied these changes to a CDS image in my testing project, and confirmed that the access logs, error logs, and metrics were propagated to cloud ops.


EricEdens:
Closing since it looks like RHUA needs the configs as well.

PR #1951, opened 2022-06-10
Fix Rocky Linux optimized name/family.

zmarano:


zmarano:
Goes with https://github.com/GoogleCloudPlatform/guest-test-infra/pull/497

PR #1964, opened 2022-07-11
Update windows client publish workflows

jjerger:
Reworks the logic in the publish workflow to add the testing environment, allowing for publishing to bct-prod, which was missing in the original push adding the workflows. This is needed for integration tests; also matches the logic of other publish files. Also updates image families to match those of previous iterations for consistency.

### DIFF
--- a/daisy_workflows/build-publish/windows/windows-11-21h2-ent-x64-uefi.publish.json
+++ b/daisy_workflows/build-publish/windows/windows-11-21h2-ent-x64-uefi.publish.json
@@ -32,7 +32,7 @@
   "Images": [
     {
       "Prefix": "windows-11-ent-x64",
-      "Family": "windows-11-x64",
+      "Family": "windows-11-21h2-x64",
       "Description": "Microsoft, Windows 11 Enterprise, 21h2 Update, x64 built on {{$time}}, supports Shielded VM features",
       "Architecture": "X86_64",
       "Licenses": [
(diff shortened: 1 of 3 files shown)


daisy_workflows/build-publish/windows/windows-11-21h2-ent-x64-uefi.publish.json
```diff
@@ -16,18 +16,23 @@
...
+  "ComputeEndpoint": {{$endpoint}},
+  "DeleteAfter": {{$delete_after}},
   {{- end}}
   {{$guest_features := `["MULTI_IP_SUBNET","UEFI_COMPATIBLE","VIRTIO_SCSI_MULTIQUEUE","GVNIC","WINDOWS"]` -}}
   {{$time := trimPrefix .publish_version "v"}}
   "Images": [
     {
       "Prefix": "windows-11-ent-x64",
-      "Family": "windows-11",
+      "Family": "windows-11-x64",
```
bkatyl, line 35: windows-11-21h2-x64

I think it would be helpful to include what specific release of Windows 11.

There is no x86 version of windows 11, so I'm not sure x64 is needed, but having it maybe helpful if we have an arm64 in the future.

jjerger: Fixed, agreed on specific release. That's what I was thinking as well- even though there is only a 64 bit version of Win11 naming it this way keeps it consistent with the other versions and prevents us from not having to rename it in the future if we do end up getting an arm version in the future.

bkatyl (approved):

PR #1976, opened 2022-07-13
fix typos in d11 worker build

hopkiw:


hopkiw:
close in favor of #1975

PR #1985, opened 2022-07-20
Support windows MTU 1500

KarnVadaliya:


daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
```diff
@@ -160,9 +160,9 @@ function Run-SecondBootSteps {
   # present.
   Restart-Service docker
 
-  Write-Host 'Setting host vEthernet MTU to 1460'
+  Write-Host 'Setting host vEthernet MTU to 1500'
```
bkatyl, line 163: 1500 is the default MTU. So do we need to intentionally set 1500 or can we simply remove setting 1460?

KarnVadaliya: Yes remove the explicit setting to 1460.

bkatyl (approved):

PR #1986, opened 2022-07-21
add arm64 partner exports

hopkiw:
* add arm64 variants for partner export workflows
* add ubuntu 2204 for x86_64 and arm64


jjerger (approved):
/lgtm

PR #1993, opened 2022-08-01
Renaming SSMS variable in SQL 22 publish file

jjerger:
Microsoft is moving away from a universal installer, so each version going forward may have their own SSMS installer version. We're going to have to start naming these to account for pulling/using the correct secrets in Concourse.

This change is a precondition for upcoming changes to the Concourse pipeline; nonbreaking as this currently is not used.


bkatyl:
We talked about this in chat. This change needs to be rolled back.

PR #1999, opened 2022-08-11
Add imagePrefix argument to gce_image_publish

rualark:
Will be tested when we converge on the idea.


hopkiw:
this is not an appropriate change - the very basis of the gce_image_publish tool is to operate off of workflows. you can feel free to fork and modify if you'd like a new tool with a different use case.

PR #2004, opened 2022-09-01
Kernel Commit Queries

dorileo:
This workflow intends to inspect a Linux system to determine if its
kernel contains required patch(es) implementing a given feature.

The initial version implements a "apply & check" strategy where it
will reconstruct the changes chain between two kernel versions and
check if the expected patches are present in the chain.

### V2:
+ Removed `pip_packages` workflow argument;
+ Small code consistency adjustment;

### V3:
+ Fixed `flake8` tests reported issues;

### V4:
+ Simplified the environment configuration steps by:
  + moving it all out of the main program
  + having a shell script for each of "os type"
  + no need of packages and package manager instruction as pass-ins

### DIFF
--- a/daisy_workflows/kernel_commit_query/debian.sh
+++ b/daisy_workflows/kernel_commit_query/debian.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright 2022 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -xeu
+
+apt-get update -y && apt-get upgrade -y && apt-get install -y git linux-source
+
+rm -rf /files/distro_kernel && mkdir -p /files/distro_kernel
+tar --strip-components=1 -xvf $(ls /usr/src/linux-source*.tar.xz) \
+    -C /files/distro_kernel
+
+pushd /files/distro_kernel
+rm -rf {.git,.gitattributes,.gitignore}
+popd
+
+kernel_pkg=$(dpkg -S /boot/vmlinuz-`uname -r` | sed 's/:.*$//g')
+apt-get changelog $kernel_pkg > /files/kernel.changelog
+
+python3 kcq.py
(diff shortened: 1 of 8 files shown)


daisy_workflows/kernel_commit_query/kernel_commit_query.wf.json
```diff
@@ -0,0 +1,122 @@
...
+      "Description": "The package manager used to install required packages."
+    },
+    "packages": {
+      "Required": true,
+      "Description": "List of packages required to run this workflow."
+    },
+    "pip_packages": {
+      "Required": true,
+      "Description": "List of pip3 packages required to run this workflow."
+    },
```
hopkiw, line 24: why are these pass-in and required?

dorileo: `packages` is provided to install 1. this program's runtime dependency(that right now consists of git) and 2. the linux distro's kernel source package, now `pip_packages` is only being used to install `google-cloud-storage`.

I strongly agree that `pip_packages` could be hardcoded and not provided as a pass-in.

For `packages` and `package_manager` I understand I could detect the running system's family and infer all the remaining but I like to keep this kind of intel on the caller rather than the workflow itself and keep it as simple as possible (or at least as objective to its goal as possible). 

(PS: They are all required because without the runtime dependencies the program will fail)

dorileo: Got rid of it, see V4 description.

daisy_workflows/kernel_commit_query/debian.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -xeu
+
+apt-get update -y && apt-get upgrade -y && apt-get install -y git linux-source
```
hopkiw, line 18: since you are using debian to get to its package infrastructure and you aren't necessarily using anything that is specific to the image, you might consider running this directly in a container environment instead of through a daisy workflow. our container images should be just as capable of downloading the `linux-source` package for a given distribution and running python code against it

dorileo: We synced off line, in the "apply & check" strategy we are relying on the currently running kernel.

daisy_workflows/kernel_commit_query/debian.sh
```diff
@@ -0,0 +1,30 @@
...
+
+rm -rf /files/distro_kernel && mkdir -p /files/distro_kernel
+tar --strip-components=1 -xvf $(ls /usr/src/linux-source*.tar.xz) \
+    -C /files/distro_kernel
+
+pushd /files/distro_kernel
+rm -rf {.git,.gitattributes,.gitignore}
+popd
+
+apt-get changelog linux-image-cloud-amd64 > /files/kernel.changelog
```
zmarano, line 28: IIRC, we're using the backports kernel for the ARM64 images- may also need to make a config for that.
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_build/debian/build_fai.py#L108

dorileo: ow yeah! good catch, I should have made it generic enough to assume the installed kernel. Will change. Thanks.

PR #2012, opened 2022-09-15
Update post_install.ps1 to skip KB5007651

jjerger:
This update has appeared in the past and hangs up the build on an update loop. Updated Install-WindowsUpdates to skip it.

### DIFF
--- a/daisy_workflows/image_build/windows/post_install.ps1
+++ b/daisy_workflows/image_build/windows/post_install.ps1
@@ -182,6 +182,18 @@ function Install-WindowsUpdates {
     }
   }
 
+  # Windows 11 may get stuck installing KB5007651. $pn for Windows 11 reports as Win 10 Enterprise.
+  # This is an intended behavior by Microsoft for backwards compatibility.
+  # As such we skip the KB here instead of trying to target by $pn.
+  if ($updates.Count -eq 1) {
+    foreach ($update in $updates) {
+      if ($update.Title -like '*KB5007651*') {
+        Write-Host 'Install-WindowsUpdates: KB5007651 detected as a single update remaining. Skipping known issue KB.'
+        return $false
+      }
+    }
+  }      
+
   foreach ($update in $updates) {
     if (-not ($update.EulaAccepted)) {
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_build/windows/post_install.ps1
```diff
@@ -182,6 +182,18 @@ function Install-WindowsUpdates {
     }
   }
 
+  # Windows 11 may get stuck installing KB5007651. $pn for Windows 11 reports as Win 10 Enterprise.
```
bkatyl, line 185: Windows 11 has a build number of 10.0.22000
https://en.wikipedia.org/wiki/Windows_11_version_history

Windows 10 21h2 has a build number of 10.0.19044.2006
https://en.wikipedia.org/wiki/Windows_10_version_history

So for Windows 11 detection logic:

$productMajorVersion = (Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion' -Name CurrentMajorVersionNumber).CurrentMajorVersionNumber
$productMinorVersion = (Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion' -Name CurrentMinorVersionNumber).CurrentMinorVersionNumber
$productBuildNumber = (Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion' -Name CurrentBuildNumber).CurrentBuildNumber

Windows 11
($productMajorVersion -eq 10 -and $productMinorVersion -eq 0 -and $productBuildNumber -ge 22000)

Windows 10 
($productMajorVersion -eq 10 -and $productMinorVersion -eq 0 -and $productBuildNumber -le 22000 -and $productBuildNumber -ge 10240)

jjerger: I added this to run only if there is 1 update remaining so we don't end up running this more than it needs to be. Question for a potential future PR is if we want to update so that we use these values versus product name for other instances like the Win7 skip. Seems more reliable.

bkatyl (approved):

PR #2021, opened 2022-10-27
Remove LEGACY security config for EL9.

zmarano:


daisy_workflows/image_build/enterprise_linux/kickstart/almalinux_9.cfg
```diff
@@ -116,10 +116,6 @@ exec &> /dev/ttyS0
 # Delete the dummy user account.
 userdel -r gce
 
-# TODO remove this!
```
hopkiw, line 119: this is not the only change that should get reverted in this PR. we also do a loopy two-stage package install because of the different environment at different points in the build, in order to install the old insecure cloud-sdk package (at least). compare rhel 9 to rhel 8 to spot the differences.

zmarano: Oh yes, updated.

thozza:
Thank you for working on this ❤️

wrigri (approved):
/lgtm

PR #2034, opened 2022-11-16
Update sql2022 install parameters

jjerger:
SNAC_SDK became an invalid feature option in the latest RC, blocking install. Removing it to allow installs to continue.


bkatyl (approved):

jjerger:
Just for completeness sake, this feature was [intentionally removed](https://learn.microsoft.com/en-us/sql/relational-databases/native-client/applications/installing-sql-server-native-client?view=sql-server-ver16) by Microsoft so it's safe to drop.

PR #2035, opened 2022-11-18
Almalinux sbom changes and test

koln67:
First pass at SBOM generation in almalinux 9.

### DIFF
--- a/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
+++ b/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
@@ -26,11 +26,11 @@
     },
     "run_sbom_bool": {
       "Value": "false",
-      "Description": "Determines if the SBOM for the image will be generated"
+      "Description": "Determines if the SBOM for the image will be generated."
     },
     "syft_source": {
-      "Value": ".",
-      "Description": "Source url for the syft tar gz file, required if run_sbom_bool=true"
+      "Value": "",
+      "Description": "Source url for the syft tar gz file, required if run_sbom_bool=true."
     }
   },
   "Steps": {
--- a/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
+++ b/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
@@ -44,7 +44,8 @@
         "Path": "${workflow_root}/export/disk_export.wf.json",
         "Vars": {
           "destination": "${gcs_url}",
-          "source_disk": "disk-almalinux-9"
+          "source_disk": "disk-almalinux-9",
+          "run-sbom-bool": "true"
         }
       }
     },
--- a/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
+++ b/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
@@ -24,9 +24,13 @@
       "Required": true,
       "Description": "The AlmaLinux 9 installer ISO to build from."
     },
-    "run-sbom-bool": {
+    "run_sbom_bool": {
       "Value": "false",
       "Description": "Determines if the SBOM for the image will be generated"
+    },
+    "syft_source": {
+      "Value": ".",
+      "Description": "Source url for the syft tar gz file, required if run_sbom_bool=true"
     }
   },
   "Steps": {
@@ -49,7 +53,8 @@
         "Vars": {
           "destination": "${gcs_url}",
           "source_disk": "disk-almalinux-9",
-          "run-sbom-bool": "{run-sbom-bool}"
+          "run_sbom_bool": "${run_sbom_bool}",
+          "syft_source": "${syft_source}"
         }
       }
     },
--- a/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
+++ b/daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
@@ -23,6 +23,10 @@
     "installer_iso": {
       "Required": true,
       "Description": "The AlmaLinux 9 installer ISO to build from."
+    },
+    "run-sbom-bool": {
+      "Value": "false",
+      "Description": "Determines if the SBOM for the image will be generated"
     }
   },
   "Steps": {
@@ -45,7 +49,7 @@
         "Vars": {
           "destination": "${gcs_url}",
           "source_disk": "disk-almalinux-9",
-          "run-sbom-bool": "true"
+          "run-sbom-bool": "{run-sbom-bool}"
         }
       }
     },
--- a/daisy_workflows/export/disk_export.wf.json
+++ b/daisy_workflows/export/disk_export.wf.json
@@ -37,13 +37,18 @@
       "Value": "default",
       "Description": "Service account that will be used by the created worker instance"
     },
-    "run-sbom-bool": {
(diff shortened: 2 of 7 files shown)


daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
```diff
@@ -44,7 +44,8 @@
         "Path": "${workflow_root}/export/disk_export.wf.json",
         "Vars": {
           "destination": "${gcs_url}",
-          "source_disk": "disk-almalinux-9"
+          "source_disk": "disk-almalinux-9",
+          "run-sbom-bool": "true"
```
bkatyl, line 48: This should be a variable about that defaults to false. Then you can pass in the variable as true for images it should be run on. Then the value here would be the variable.

    "run-sbom-bool": {
      "Value": "false",
      "Description": "Determines if a SBOM should be generated."
    }
    
    Then the value here would be
    "run-sbom-bool": "${run-sbom-bool}"

koln67: Fixed, running the image build with the same command as before this change should not generate the SBOM.

daisy_workflows/export/export_disk.sh
```diff
@@ -57,5 +59,23 @@ TARGET_SIZE_BYTES=$(gsutil ls -l "${GCS_PATH}" | head -n 1 | awk '{print $1}')
 TARGET_SIZE_GB=$(awk "BEGIN {print int(((${TARGET_SIZE_BYTES}-1)/${BYTES_1GB}) + 1)}")
 serialOutputPrefixedKeyValue "GCEExport" "target-size-gb" "${TARGET_SIZE_GB}"
 
+# Generate SBOM if flag is set
+function runSBOMGeneration() {
+  gsutil cp gs://koln-bucket/syft_0.59.0_linux_amd64.tar.gz .
+  tar -xf syft_0.59.0_linux_amd64.tar.gz
```
Linskeyd, line 65: Please update this to an existing bucket name or bucket already in use for compute-image-tools (avoiding a personally named bucket).

koln67: Not fixed yet, this will be changed to gs://gce-image-build-resources once I have finished promoting from gs://gce-image-build-resources-rtp using the Artifact Releaser command.

koln67: currently uses a user input path to gcs instead

koln67:
For reference, this is how I ran the almalinux 9 workflow using the syft source stored in gcs.

daisy -project studious-sign-357617 -var:installer_iso=gs://gce-image-build-resources-rtp/linux/AlmaLinux-9.0-x86_64-dvd.iso -var:gcs_url=gs://koln-bucket/img.tar.gz -gcs_path gs://koln-bucket -var:syft_source=gs://gce-image-build-resources-rtp/linux/syft_0.59.0_linux_amd64.tar.gz -var:workflow_root=../.. build-publish/enterprise_linux/almalinux_9.wf.json

This is the command I used to run the testing workflow for enterprise linux

daisy -project studious-sign-357617 -var:workflow_root=../../daisy_workflows -var:syft_source=gs://gce-image-build-resources-rtp/linux/syft_0.59.0_linux_amd64.tar.gz image_test/sbom/enterprise_sbom_test.wf.json

image_test/sbom/test_sbom_export.sh
```diff
@@ -0,0 +1,36 @@
+URL="http://metadata/computeMetadata/v1/instance/attributes"
```
hopkiw, line 1: should this file start with shebang (`#!/bin/bash`) ?

koln67: It might be better to provide some context:

The purpose of this file is so that the steps to generate the SBOM itself (running syft) are all in one location, even though Debian and Enterprise Linux have different mounting methodologies.

From debugging when running this script in #2036, the script needed the shebang when called from a python subprocess, or error 8 would result. For this script to be used in both Debian and Enterprise Linux, it would need to keep the shebang.

daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
```diff
@@ -23,6 +23,14 @@
     "installer_iso": {
       "Required": true,
       "Description": "The AlmaLinux 9 installer ISO to build from."
+    },
+    "run_sbom_bool": {
+      "Value": "false",
+      "Description": "Determines if the SBOM for the image will be generated"
```
dorileo, line 29: I commented this on the other PR, please end the description line with a dot for consistency.

daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
```diff
@@ -23,6 +23,14 @@
...
+    },
+    "syft_source": {
+      "Value": ".",
+      "Description": "Source url for the syft tar gz file, required if run_sbom_bool=true"
```
dorileo, line 33: same here.

daisy_workflows/export/disk_export.wf.json
```diff
@@ -14,7 +14,7 @@
       "Description": "list of GCE licenses to record in the exported image"
     },
     "export_instance_disk_image": {
-      "Value": "projects/compute-image-tools/global/images/family/debian-10-worker",
+      "Value": "projects/compute-image-tools/global/images/family/debian-11-worker",
```
dorileo, line 17: Why do you need to migrate from debian-10 to debian-11? Because of sfdisk?

daisy_workflows/export/export_sbom.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
dorileo, line 1: You are adding the same file in 2 different pull requests?

image_test/sbom/test_sbom_export.sh
```diff
@@ -0,0 +1,36 @@
...
+SBOM_FILE_NAME=$(curl -f -H Metadata-Flavor:Google ${URL}/sbom-file-name)
+OUTS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/outs-path)
```
dorileo, line 3: Also this file is being added by two PR I think.

image_test/sbom/test_sbom_export.sh
koln67: Actually I was looking at the wrong file, I think it should have the shebang.

daisy_workflows/export/disk_export.wf.json
koln67: the reason is because the mounting of the disk runs into a bug with a debian-10 worker, but does not run into a bug with the debian-11 worker.

image_test/sbom/test_sbom_export.sh
koln67: for now, the other PR is being ignored until this PR is merged.

daisy_workflows/export/export_sbom.sh
koln67: yes

daisy_workflows/build-publish/enterprise_linux/almalinux_9.wf.json
koln67: done

dorileo (approved):

PR #2039, opened 2022-12-01
Debian sbom

koln67:
Proposed changes to allow sbom generation for debian 10, if run_sbom_bool is set to true in the daisy workflow. 

By default, an sbom is not generated.


dorileo:
@koln67 can you rebase this PR against head?

PR #2040, opened 2022-12-01
Debian 10 sbom generation changes

koln67:
Add optional functionality to generate the SBOM for debian 10 images, if run_sbom_bool is set to true.

### DIFF
--- a/daisy_workflows/image_build/debian/build_fai.py
+++ b/daisy_workflows/image_build/debian/build_fai.py
@@ -23,6 +23,7 @@ debian_version: The FAI tool debian version to be requested.
 image_dest: The Cloud Storage destination for the resultant image.
 """
 
+import json
 import logging
 import os
 import platform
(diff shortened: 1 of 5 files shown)


daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -49,6 +50,53 @@ def mycopytree(src, dst):
...
+    raise Exception('Could not determine the sector size')
+
+  partitions = ptable.get('partitions', None)
+  if partitions is None:
+    raise Exception('No partitions found in partition table')
+
+  for partition in partitions:
+    partition_type = partition['type']
+    # Get the offset for mounting the disk
+    if partition_type == Linux_Filesystem_GUID:
```
dorileo, line 75: you are not using partition_type anywhere else, so you can make it shorter:
```python
if partition['type'] == Linux_Filesystem_GUI:
...
```

daisy_workflows/image_build/debian/build_fai.py
```diff
@@ -49,6 +50,53 @@ def mycopytree(src, dst):
     return dst
 
 
+# get number of bytes to offset when mounting the raw disk image file
```
hopkiw, line 53: we should not be modifying the build steps for debian; sbom generation should be its own separate process

PR #2042, opened 2022-12-06
Revert the debian worker image for disk_export due to a potential race

zmarano:
in disk enumeration in Debian 11.


KarnVadaliya (approved):
lgtm

PR #2050, opened 2022-12-09
Add Cloud Metrics for RHUA sync status.

wrigri:


daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+  for update in recent_updates:
+    end_time = time.strftime('%Y-%m-%dT%H:%M:%S-00:00', time.gmtime(update))
+    point = {'interval': {'endTime': end_time}, 'value': {'boolValue': True}}
+    points.append(point)
+
+  PublishMetric(metric_type, points, access_token, mig_name)
+
+
+def PublishLastHourCount(hour_count, timestamp, access_token, mig_name):
+  print(hour_count)
```
hopkiw, line 82: is this a debug print?

daisy_workflows/build-publish/rhui/install_rhua.sh
```diff
@@ -89,6 +89,13 @@ for unit in rhui-health-check.{service,timer}; do
 done
 install -m 664 -t /etc/nginx/conf.d $tempdir/health_check.nginx.conf
 
+# Install RHUA sync status.
```
hopkiw, line 92: this block is duplicating the one above including creating the dir; let's just amend line 86, only

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+"""Publish metrics for RHUA sync status."""
+
+import os
+import subprocess
+import time
+
+import requests
+
+
+REPODIR = '/var/lib/rhui/remote_share/symlinks/pulp/content/content'
```
hopkiw, line 23: i think some of this doesn't pass PEP8, can you run that to confirm?

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def GetRepodataList(dirname):
+  # Use "find" here because os.walk takes ~30 minutes,
```
hopkiw, line 27: if this works i'm fine with it, but i think we would not simply say os.walk takes too long but *why* it takes too long. no reason python can't be as fast as find

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+  # while find takes 10-30 seconds.
+  cmd = ['find', dirname, '-name', 'repodata']
+  output = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  return output.stdout.split()
+
+
+def PublishMetric(metric_type, points, access_token, mig_name):
```
hopkiw, line 34: as discussed offline, we either want to use the service library or create a tracking bug to migrate to it in a subsequent pass

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+  end_time = time.strftime('%Y-%m-%dT%H:%M:%S-00:00', time.gmtime(timestamp))
+  metric_type = 'custom.googleapis.com/rhua_sync_age'
+
+  points = [{'interval': {'endTime': end_time},
+             'value': {'int64Value': seconds_since_update}}]
+
+  PublishMetric(metric_type, points, access_token, mig_name)
+
+
+def PublishRecentUpdates(recent_updates, access_token, mig_name):
```
hopkiw, line 68: could you add some function comments indicating what each one actually publishes?

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+    if statinfo.st_mtime > hour_ago:
+      hour_count += 1
+    if statinfo.st_mtime > five_min_ago:
+      recent_updates.append(statinfo.st_mtime)
+    if statinfo.st_mtime > newest:
+      newest = statinfo.st_mtime
+
+  repo_age = int(now - newest)
+  access_token = GetAccessToken()
+  print(access_token)
```
hopkiw, line 142: are these debug prints?

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+  newest = 0.0
+  now = time.time()
+  hour_ago = now - 3600
+  five_min_ago = now - 300
+  hour_count = 0
+  recent_updates = []
+
+  repodata_list = GetRepodataList(REPODIR)
+
+  for repodata in repodata_list:
```
hopkiw, line 131: does this block deserve a function?

hopkiw (changes requested):

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+REPODIR = '/var/lib/rhui/remote_share/symlinks/pulp/content/content'
+
+
+def GetRepodataList(dirname):
+  # Use "find" here because os.walk takes ~30 minutes,
+  # while find takes 10-30 seconds.
+  cmd = ['find', dirname, '-name', 'repodata']
+  output = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
dorileo, line 30: Please pass `check=True` to `subprocess.run()`.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,154 @@
...
+                     'node_id': mig_name}
+  resource = {'type': 'generic_node',
+              'labels': resource_labels}
+
+  time_series = [{'metric': metric,
+                  'resource': resource,
+                  'points': points}]
+
+  data = {'timeSeries': time_series}
+  print(data)
```
dorileo, line 52: Please consider using `logging`.

dorileo (commented):
Looks good overall, just some minor comments.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: The only output I get from flake8 is that it doesn't like the Google-standard 2 space indentation instead of the 4 spaces. The sort-of weird import order is due to the fact that "requests" is not built-in, but is a third-party package (that's already installed on our RHUA machines).

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: Yes, removed.

daisy_workflows/build-publish/rhui/install_rhua.sh
wrigri: Updated.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: So, I'm not really sure why os.walk takes as long as it does. The times listed came from actually trying to use it. I tested both find and os.walk to simply generate a list of "repodata" directories, and while the call to find usually takes less than 20 seconds, using os.walk took about half an hour. It's generally expected that os.walk would be slower than find, since find apparently has some optimizations, but it's possible that there is something specific about this directory that makes it worse. (Apparently pre-Python 3.5, os.walk was particularly bad because it would stat each file. That was fixed in 3.5 with os.scandir, but apparently os.walk is still much slower than find in some cases.)

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: I'll create a tracking bug.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: Done.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: Removed

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: I originally went down that path, but it was confusing since the block needs to return 2 ints and a list. I could return an object, but it didn't seem worth it. I can create an object with this info if you think it would be better.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: Done

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: I removed all of the print calls from the script.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def PublishMetric(metric_type, points, access_token, mig_name):
+  region = GetRegion()
+  url = ('https://monitoring.googleapis.com/v3/projects/'
+         'google.com%3Arhel-infra/timeSeries')
+  headers = {'Authorization': 'Bearer {}'.format(access_token)}
+  metric = {'type': metric_type}
+  resource_labels = {'project_id': '155767908850',
+                     'location': region,
```
hopkiw, line 42: let's detect rather than hardcoding project names and IDs. main can detect and then pass in here.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
hopkiw: i think you have one function now which finds matching pathnames, then main has to act on them. but what you really want are just times here. if you have the lookup function do the stats, then you're just operating a sorted list of integer time values. then the comparisons you need here could be turned into simple comprehensions.

imagine:
```py3
times = GetRepoDataTimes(REPODIR)
in_last_hour = [x for x in times if x > hour_ago]
in_last_five = [x for x in in_last_hour if x > five_min_ago]
PublishRepoAge(times[0], now, access_token, mig_name)  # It can do the math
PublishLastHourCount(len(in_last_hour), now, access_token, mig_name)
PublishRecentUpdates(in_last_five, access_token, mig_name)
```

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: Updated to get project info from metadata.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: Removed the comment.

daisy_workflows/build-publish/rhui/rhua_sync_status.py
wrigri: I like that. I think I'll save it for a future improvement though.

hopkiw (approved):

PR #2071, opened 2023-01-12
Change worker to debian 11

koln67:
As a follow up to the symlink change, use a debian 11 worker instead of debian 10. 

Even if the disks /dev/sda and /dev/sdb are flipped, the symlinks should export the correct disk.


dorileo:
I remember the last time we upgraded the worker to debian 11 we found some issues, has it changed?

koln67:
> I remember the last time we upgraded the worker to debian 11 we found some issues, has it changed?

The issue should be fixed in #2068 

@zmarano for confirmation, some testing was done to ensure that the disk swap issue from last time was fixed.

PR #2072, opened 2023-01-12
EL9 sbom support

koln67:
EL9 sbom support


koln67:
/hold

Doing a quick final test on this PR

PR #2083, opened 2023-01-30
sbom final destination

koln67:
Include export of ${NAME}.sbom.json to the final destination.


daisy_workflows/export/export_disk.sh
```diff
@@ -72,8 +75,9 @@ function runSBOMGeneration() {
   mount -o bind,ro /dev /mnt/dev
   gsutil cp $SYFT_TAR_FILE syft.tar.gz
   tar -xf syft.tar.gz
-  ./syft /mnt -o spdx-json > sbom.json
-  gsutil cp sbom.json $SBOM_PATH
+  ./syft /mnt -o spdx-json > image.sbom.json
+  gsutil cp image.sbom.json $SBOM_PATH
```
zmarano, line 79: why are there two copies? Are they not both going to the same place?

koln67: The second one is copying to the destination, such as the artifact releaser bucket.
@bkatyl

zmarano: Yeah I see, that matches our image behaviors where we retain the artifacts in the build bucket.

PR #2089, opened 2023-02-08
Add sbom destination input to all workflows

koln67:
Allow the workflows to be run with a "SBOM destination" variable, which exports the sbom to the same final location which the disk tar.gz file is exported to. The sbom should be exported with the .sbom.json extension. 

By default, the sbom is not exported to a final url.


koln67:
/uncc 
KarnVadaliya

PR #2098, opened 2023-02-24
export metadata: use gpt's label metadata to identify root

dorileo:
To maintain backward compatibility use gpt's label metadata to identify the root fs - in addition to the guuid logic.


jjerger (commented):
/lgtm

PR #2099, opened 2023-02-24
export metadata: enable gpt lookup for debian-10 only for now.

dorileo:


daisy_workflows/export_metadata/export-metadata.py
```diff
@@ -52,10 +54,17 @@ def main():
...
-    return
+  # This assumes that:
+  # 1. /dev/sdb1 is the EFI system partition.
+  # 2. /dev/sdb2 is the root mount for the installed system.
+  # Except for debian 10, which has out-of-order partitions.
+  if uefi and 'debian-10' not in image_family:
+    mount_disk = '/dev/sdb2'
+  else:
+    mount_disk = unmounted_root_fs()
+    if mount_disk is None:
```
Quintonamore, line 65: Optional nit: The current conditional flow works here, but I think the conditional for checking the mount disk should happen outside of any conditional logic just in case anything changes with how the mount_disk for uefi is defined.

Quintonamore (approved):

PR #2106, opened 2023-03-24
Update Server 2019 container build script

jjerger:
Handles changes required to move from Docker EE -> CE in mid-april.

### DIFF
--- a/daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
+++ b/daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
@@ -21,6 +21,10 @@
 # Docker Community Edition meant for Windows client installations).
 
 $ErrorActionPreference = 'Stop'
+$DockerPath = 'https://master.dockerproject.org/windows/x86_64/docker.exe'
+$DockerDPath = 'https://master.dockerproject.org/windows/x86_64/dockerd.exe'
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_build/windows_for_containers/container_image_install.ps1
```diff
@@ -189,12 +310,17 @@ function Run-SecondBootSteps {
...
+  # docker installation.
+  $dockerStatus = Get-Service -Name 'docker' -ErrorAction SilentlyContinue
   Write-Host "Windows version: $windows_version"
   if (-not $windows_version) {
     throw 'Error retrieving "version" from metadata'
   }
 
-  if (!(Get-Package -ProviderName DockerMsftProvider -ErrorAction SilentlyContinue| Where-Object {$_.Name -eq 'docker'})) {
+  if ($dockerStatus -eq $null) {
     Run-FirstBootSteps
```
jjerger, line 324: I'm already aware this is a misnomer since this will technically run twice per build, but we need to get this operational asap. Renaming/breaking out this block can be something we can do later, since this approach will successfully install and configure the docker runtime.

bkatyl:
As we talked about last night please fork this file so it doesn't impact the april builds

PR #2118, opened 2023-04-19
Make sure we install the GCE guest software.

zmarano:


zmarano:
This changed because of: https://salsa.debian.org/cloud-team/debian-cloud-images/-/blob/master/config_space/bullseye/package_config/GCE

Where for bullseye, the upstream class doesn't include the google-compute-engine package.

PR #2139, opened 2023-05-22
Add skeleton for sbom-util SBOM generation path

koln67:
Add the daisy config variables so sbom-util can be used to generate the sbom, instead of a fixed gcs location where syft is stored.

### DIFF
--- a/daisy_workflows/build-publish/enterprise_linux/almalinux_8.wf.json
+++ b/daisy_workflows/build-publish/enterprise_linux/almalinux_8.wf.json
@@ -31,6 +31,10 @@
     "syft_source": {
       "Value": "",
       "Description": "Source url for the syft tar gz file, if generating SBOM. If empty no SBOM is generated."
+    },
+    "sbom_util_source": {
+      "Value": "",
+      "Description": "Source url for the sbomutil executable, if using sbomutil to generate the SBOM."
     }
   },
   "Steps": {
@@ -54,7 +58,8 @@
           "destination": "${gcs_url}",
           "sbom_destination": "${sbom_destination}",
           "source_disk": "disk-almalinux-8",
-          "syft_source": "${syft_source}"
+          "syft_source": "${syft_source}",
+          "sbom_util_source": "${sbom_util_source}"
         }
       }
     },
--- a/daisy_workflows/build-publish/enterprise_linux/almalinux_8.wf.json
+++ b/daisy_workflows/build-publish/enterprise_linux/almalinux_8.wf.json
@@ -32,9 +32,9 @@
       "Value": "",
       "Description": "Source url for the syft tar gz file, if generating SBOM. If empty no SBOM is generated."
     },
-    "sbom_util_source": {
+    "sbom_util_gcs_root": {
       "Value": "",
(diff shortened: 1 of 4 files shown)


daisy_workflows/build-publish/enterprise_linux/almalinux_8.wf.json
```diff
@@ -31,6 +31,10 @@
     "syft_source": {
       "Value": "",
       "Description": "Source url for the syft tar gz file, if generating SBOM. If empty no SBOM is generated."
+    },
+    "sbom_util_source": {
```
dorileo, line 35: We are using `sbom_util_gcs_root` for packages, it'd be nice if we kept them consistent.

daisy_workflows/export/disk_export.wf.json
```diff
@@ -75,7 +80,9 @@
             "startup-script": "${SOURCE:${NAME}_export_disk.sh}",
             "source-disk-name": "${source_disk}",
             "syft-tar-file": "${SOURCESPATH}/syft.tar.gz",
-            "syft-source": "${syft_source}"
+            "syft-source": "${syft_source}",
+            "sbom-util-source": "${sbom_util_source}",
+            "sbom-util-executable": "${SOURCESPATH}/sbom_util"
```
dorileo, line 85: this `sbom-util-executable` derivation is not exactly true. the binary name is `sbomutil` and between the root and the binary there's the release folder. I believe that carrying over only the sbom-util gcs root and then deriving the full path based on that in the function we are downloading it would be better (like we are doing for first party packages). Another option if you want is to change concourse to derive the full path and pass it in, in that case we wouldn't derive it in the daisy workflow - if you want and understand the latter is the best option we can work it out in a way that we can use it even for the first party packages build, changing the approach I introduced.

daisy_workflows/export/export_disk.sh
```diff
@@ -85,9 +104,13 @@ function runSBOMGeneration() {
 # Always create empty sbom file so workflow copying does not fail
 touch image.sbom.json
 gsutil cp image.sbom.json $SBOM_PATH
+# If the sbom-util program location is passed in, use that instead of SYFT_SOURCE.
+if [ $SBOM_UTIL_SOURCE != "" ]; then
```
dorileo, line 108: good job thinking about the transition path.

dorileo (changes requested):

daisy_workflows/export/disk_export.wf.json
koln67: The second approach would be easier for this daisy workflow, because downloading the sbom-util executable directly to the local daisy sources folder will bypass a lot of IAM permission issues.

daisy_workflows/build-publish/enterprise_linux/almalinux_8.wf.json
koln67: done

daisy_workflows/export/disk_export.wf.json
koln67: done

PR #2167, opened 2023-06-02
Use sbomutil.exe instead of sbomutil on windows

koln67:
Windows executables must end in exe


koln67:
Do not merge until sbomutil.exe has been successfully pushed

PR #2176, opened 2023-06-13
Do not replace windows sbom with empty sbom

koln67:
Windows workflows generate the sbom in the powershell script, before disk export. Without this change, disk export overrides the sbom file with an empty file.


dorileo:
Does it apply only for some image builds?

koln67:
It applies to any build which has something like this line, calling windows_export.wf.json: 

https://github.com/GoogleCloudPlatform/compute-image-tools/blob/defe758f467d093c07afe846a5df02f2897dea7d/daisy_workflows/build-publish/windows/windows-10-21h2-ent-x64-uefi.wf.json#L74

PR #2202, opened 2023-06-28
Do not add sbomutil.exe to the sbom

koln67:
If sbomutil.exe is added to the components directory, it is added to the sbom file by default.

We do not want this, so download sbomutil.exe to C:\ instead


dorileo:
good catch

PR #2204, opened 2023-07-05
Add windows sbom validation workflow

koln67:
Add a workflow which can be used as a presubmit for windows sbom generation validation.

### DIFF
--- a/daisy_workflows/sbom_validation/test_sbom_export.sh
+++ b/daisy_workflows/sbom_validation/test_sbom_export.sh
@@ -2,6 +2,8 @@
 
 URL="http://metadata/computeMetadata/v1/instance/attributes"
 OUTS_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/outs-path)
+# If set to false, do not check if the disk tar file is present
+CHECK_DISK_TAR=$(curl -f -H Metadata-Flavor:Google ${URL}/check-disk-tar)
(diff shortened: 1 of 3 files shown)


daisy_workflows/sbom_validation/test_sbom_export.sh
```diff
@@ -26,14 +28,18 @@ else
...
-status=$?
-if [[ $status -eq 0 ]]; then
-  echo "SBOMTesting: Disk tar file successfully found"
+# Check that the disk export succeeded, if the flag is not false
+if [[ $CHECK_DISK_TAR == "false" ]]; then
+  echo "SBOMTesting: skipping check for Disk tar file"
 else
-  echo "SBOMFailed: Disk tar file not found"
-  exit 1
+  gsutil -q stat $GCS_PATH_OUTDISK
```
dorileo, line 35: I understand the idea but do we really want to validate the disk file? i so we are not actually validating the SBOM right?

koln67: The presubmit was originally written for linux, and I was hoping to check that when generating the SBOM, it did not cause the disk file export to fail.

For windows, the bootstrap script does not export a disk, but it exports an SBOM. The disk file is likely exported in the post_install script. The thinking was that checking for a disk on windows would be good, but it would make the presubmit take too long because it would require running post_install as well.

dorileo: Sounds good.

PR #2214, opened 2023-07-18
Set default windows workflow root for presubmits

koln67:
When running the presubmit, we do not want to have to pass in a workflow_root, and instead want the default workflow_root to work.

This sets the default value correctly.

### DIFF
--- a/daisy_workflows/sbom_validation/windows_sbom_validation.wf.json
+++ b/daisy_workflows/sbom_validation/windows_sbom_validation.wf.json
@@ -36,7 +36,7 @@
       "Description": "The size of disk to provision for the image in GB."
     },
     "workflow_root": {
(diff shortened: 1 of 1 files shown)


bkatyl (approved):

daisy_workflows/sbom_validation/windows_sbom_validation.wf.json
```diff
@@ -36,7 +36,7 @@
       "Description": "The size of disk to provision for the image in GB."
     },
     "workflow_root": {
-      "Value": "/workflows",
+      "Value": "../../daisy_workflows",
```
dorileo, line 39: I'd like it to be an absolute path unless we have a strong limitation not to.

PR #2218, opened 2023-08-17
Switch debian 12+sid to systemd-timesyncd

a-crate:
* Remove chrony config for 12+sid (configs for 10+11 are upstream)

* Add systemd-timesyncd config for GCE

* Select the fai classes appropriately


a-crate:
Test for this change is here: https://github.com/GoogleCloudPlatform/guest-test-infra/pull/753

zmarano (approved):

PR #2220, opened 2023-08-22
Avoid race conditions due to using shortnames in enterprise linux builds

a-crate:
* Pass the name of the gce disk to be installed to the bootstrap stage through metadata

* Modify the kickstart config to reference this disk directly through the /dev/disk/by-id subsystem instead of hoping it appears second on /dev/sdb

### DIFF
--- a/daisy_workflows/image_build/enterprise_linux/build_installer.py
+++ b/daisy_workflows/image_build/enterprise_linux/build_installer.py
@@ -32,6 +32,7 @@ def main():
   # Get Parameters
   release = utils.GetMetadataAttribute('el_release', raise_on_not_found=True)
   savelogs = utils.GetMetadataAttribute('el_savelogs') == 'true'
+  install_disk = 'scsi-0Google_PersistentDisk_' + utils.GetMetadataAttribute('install_disk', raise_on_not_found=True)
 
   logging.info('EL Release: %s' % release)
   logging.info('Build working directory: %s' % os.getcwd())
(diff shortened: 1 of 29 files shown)


daisy_workflows/image_build/enterprise_linux/build_installer.py
```diff
@@ -32,6 +32,9 @@ def main():
   # Get Parameters
   release = utils.GetMetadataAttribute('el_release', raise_on_not_found=True)
   savelogs = utils.GetMetadataAttribute('el_savelogs') == 'true'
+  install_disk = 'scsi-0Google_PersistentDisk_'
```
zmarano, line 35: In this environment, we have symlinks to the device name which come from this [udev rule](https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/rules.d/65-gce-disk-naming.rules). It is probably better to use these symlinks which will exist in this VM environment (it's not quite true in the anaconda build env though) to deal with the NVMe machine types.

a-crate: The disk reference isn't used in this environment, it's used in the anaconda environment. I will copy the udev rule into there and expand this to handle nvme machine types too.

zmarano: If you want to see an NVMe case- you can use the arm build which is NVMe only.

a-crate: Done.

a-crate:
The name of the install disk is now static and the considered alternatives to the udev approach are in comment in build_installer.py

PR #2221, opened 2023-08-24
Remove install_disk var from enterprise linux workflows

a-crate:


a-crate:
No longer used as of https://github.com/GoogleCloudPlatform/compute-image-tools/pull/2220

PR #2231, opened 2023-10-04
Don't install shim-signed on debian arm versions that don't support secure boot

a-crate:


daisy_workflows/image_build/debian/fai_config/package_config/GCE_SPECIFIC
```diff
@@ -30,9 +30,13 @@ shim-signed
 grub-efi-amd64-signed
 
 PACKAGES install ARM64
-# Secure boot on ARM64
-# Will not work on debian 10 or 11, but will not hamper non-secure boot
 grub-efi-arm64
+
+PACKAGES remove ARM64_NO_SECURE_BOOT
+grub-efi-arm64-signed
```
zmarano, line 36: non signed packages?

a-crate: Yep. This was the previous setup when secureboot was disabled on all debian arm builds, removing the signed versions and leaving just unsigned grub.

PR #2244, opened 2023-11-15
Update golang and deps for cli_tools

a-crate:
go 1.13 -> go 1.21
also ran gofmt on everything and fixed lint issues.

This only updates the stdlib and packages which are required to update stdlib as an interim step, updating dependencies will come later.


a-crate:
Added updates for dependencies as well.

a-crate:
Brian and Zach are OoO this week.
/auto-cc

PR #2245, opened 2023-12-04
Update rdp_status.ps1

ekekpe:
Adding commands to verify that:

- Ethernet port is enabled
- DHCP is enabled as static may be set up, and IP address is accurate
- Security Layer authentication value is set to default
- Network-level UserAuthentication value is set to default

### DIFF
--- a/cli_tools/diagnostics/rdp_status.ps1
+++ b/cli_tools/diagnostics/rdp_status.ps1
@@ -39,7 +39,7 @@ $items= @(
     cmd="netsh interface ipv4 show subinterfaces"
   },
   @{
-    description="Verify that client-server seecurity negotiation is set to default value, expect [SecurityLayer REG_DWORD 0x1]:";
+    description="Verify that client-server seecurity negotiation is set to default value, expect [SecurityLayer REG_DWORD 0x0, 0x1, 0x2] depending on Security Layer configuration:";
     cmd="reg query 'HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Terminal Server\WinStations\RDP-Tcp' /v SecurityLayer"
   },
   @{


cli_tools/diagnostics/rdp_status.ps1
```diff
@@ -2,6 +2,14 @@ $RDPTCPpath = 'HKLM:\SYSTEM\CurrentControlSet\Control\Terminal Server\Winstation
 $RDPServerPath = 'HKLM:\SYSTEM\CurrentControlSet\Control\Terminal Server'
 
 $items= @(
+  @{
+    description="Verify that the Ethernet adapter is active, expect [Enabled]:";
+    cmd="netsh interface show interface"
+  },
+  @{
```
bkatyl, line 9: Static IPs are something that we support as they're needed for specific use cases such as domain controllers. 

So obtaining the current IP is great, but the correct evaluation is that the IP matches the instance information, not that it's set to DHCP.

cli_tools/diagnostics/rdp_status.ps1
```diff
@@ -25,6 +33,14 @@ $items= @(
   @{
     description="Verify that MTU size is no greater than 1460, expect [MTU <= 1460, Interface:Ethernet]:";
     cmd="netsh interface ipv4 show subinterfaces"
+  },
+  @{
+    description="Verify that client-server seecurity negotiation is set to default value, expect [SecurityLayer REG_DWORD 0x1]:";
```
bkatyl, line 38: This is at least incorrect for Windows 2022 as it's set to 2.

I would also caution again stating "set to default value", If there is an expected value it's better to state what it is and why.

For these items I'd suggest linking to this:
https://learn.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-terminalservices-rdp-winstationextensions-securitylayer

All 3 values this setting can have are correct depending on the situation. So for this type of setting we should absolutely collect it to understand it's configuration, but there should be no expected value beyond it being between 0 and 3. If you want you can give a short statement to summarize what each value is so going to the link isn't required.

0=Legacy RDP protocol, 1=Negotiate Legacy/TLS, 2=TLS required.

cli_tools/diagnostics/rdp_status.ps1
```diff
@@ -33,4 +49,4 @@ foreach($item in $items){
   Write-Host $item["description"]
   Invoke-Expression $item["cmd"]
   Write-Host
-}
\ No newline at end of file
+}
```
bkatyl, line 52: Identifying if the instance has the Remote Desktop feature enabled (ie it can support more than 2 users) would be very helpful as when that is enable the customer needs to set the "connect to an admin session" for adminstrative operations.

Because it changes the connection setting requirements it's an incredibly important detail to collect. We've had support cases where RDP licensing wasn't configured or out of the grace period and in that state only admin sessions are allowed and the resolution was setting the customer to set that setting in their RDP client.

ekekpe: Added

ekekpe (commented):
All checks done

cli_tools/diagnostics/rdp_status.ps1
bkatyl: I'm not sure that "(Get-Service -Name TSGateway).Status" will accomplish this goal as that's specifically around the Gateway capability.

https://learn.microsoft.com/en-us/windows/win32/termserv/detecting-whether-terminal-services-is-installed

get-windowsfeature RDS-RD-Server
```
Display Name                                            Name                       Install State
------------                                            ----                       -------------
    [ ] Remote Desktop Session Host                     RDS-RD-Server                  Available
```

bkatyl (commented):
Unresolving the comments that had follow ups provided.

cli_tools/diagnostics/rdp_status.ps1
ekekpe: Please I have updated this, thank you so much for your support

cli_tools/diagnostics/rdp_status.ps1
```diff
@@ -25,6 +33,14 @@ $items= @(
   @{
     description="Verify that MTU size is no greater than 1460, expect [MTU <= 1460, Interface:Ethernet]:";
     cmd="netsh interface ipv4 show subinterfaces"
+  },
+  @{
+    description="Verify that client-server seecurity negotiation is set to default value, expect [SecurityLayer REG_DWORD 0x1]:";
+    cmd="reg query 'HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Terminal Server\WinStations\RDP-Tcp' /v SecurityLayer"
+  },
+  @{
+    description="Verify that Network-level UserAuthentication is set to default value, expect [UserAuthentication REG_DWORD 0x0]:";
```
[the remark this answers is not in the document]
ekekpe: Added

cli_tools/diagnostics/rdp_status.ps1
ekekpe: Please I have added this as well, I did not realise the changes early on. My bad

ekekpe (commented):
Kindly review for me, thank you @bkatyl

bkatyl (approved):

PR #2246, opened 2023-12-06
google-cloud-cli -> google-cloud-sdk in rhel sap byos images

a-crate:
rhel sap byos uses a different gcloud package than the rest of enterprise linux, which throws a wrench in our package [testing](http://localhost:8080/teams/guestos/pipelines/linux-image-build/jobs/publish-to-testing-rhel-8-4-sap-byos/builds/58). If there's a preference for one or the other we can switch to either but we should make it consistent to make our testing easier.

/assign @zmarano


zmarano:
This actually should be the other way around. We should accept both cases- but be moving to `google-cloud-cli` as the package name.

PR #2255, opened 2024-01-04
Fix centos-7-predictable-nic publish template name

a-crate:
/cc @zmarano @koln67


a-crate:
Forgot the family. Now it's ready.
/unhold

PR #2260, opened 2024-01-11
Exclude KB5034439 from Server 2022 builds

jjerger:
Likely a temporary exception to allow the build to complete.

Looking at Windows Update logs for a failed test VM it appears the KB is never actually attempted by the Windows Update service, yet the build log shows it repeatedly trying and Succeeding with Errors.


jjerger:
Assigning to Brian, as I think he should be the one to vet that this is a necessary exception to make.

bkatyl (approved):

PR #2269, opened 2024-02-06
Change enterprise linux serial console to 115200

elicriffield:


daisy_workflows/image_build/enterprise_linux/build_installer.py
```diff
@@ -96,13 +96,13 @@ def main():
...
     cfg = re.sub(r'load_video\n',
-           r'serial --speed=38400 --unit=0 --word=8 --parity=no\n'
+           r'serial --speed=115200 --unit=0 --word=8 --parity=no\n'
            'terminal_input serial\nterminal_output serial\n', cfg)
 
     # Change boot args.
     args = ' '.join([
       'inst.text', 'inst.ks=hd:LABEL=INSTALLER:/%s' % ks_cfg,
-      'console=ttyS0,38400n8', 'inst.gpt', 'inst.loglevel=debug'
+      'console=ttyS0,115200', 'inst.gpt', 'inst.loglevel=debug'
```
ajorg, line 105: One does wonder if the `n8` could be important. IIUC it's the default, and has been for decades, so probably not?

elicriffield: from my testing it's unnecessary

PR #2271, opened 2024-02-14
Tag rhel 9.2 with SEV_LIVE_MIGRATABLE and SEV_LIVE_MIGRATABLE_V2

a-crate:
/cc @zmarano @elicriffield 
/hold


a-crate:
Adding a commit to tag SEV_SNP capable things with SEV_SNP_CAPABLE. I tested these manually with a derivative image, and we have a CIT test for booting the image and checking the guest module is loaded too.

drewhli (approved):

PR #2283, opened 2024-03-05
Removing netsh helpers from Windows 11 installations

bkatyl:
Removing netsh helpers from Windows 11 installations


jjerger (commented):
/lgtm

PR #2300, opened 2024-04-17
Remove buster backports.

zmarano:


a-crate:
Didn't upstream drop backports too? Can't we just use the lastest FAI commit from them?

zmarano:
> Didn't upstream drop backports too? Can't we just use the lastest FAI commit from them?

Working on that- there are several other changes that are causing issues. This is temporary.

a-crate (approved):

PR #2304, opened 2024-04-30
Allow the preloading script to create a preloaded COS image in different projects

sejalsharma-google:
By adding a command line arg for '_DEST_PROJECT', we can specify which project the preloaded image should be uploaded to. Previously, it would just use the default project in the gcloud config of whatever/whomever was invoking the dev_cloudbuild.yaml script.

Ideally, this will allow us to upload the preloaded image to the gcp-guest project, like the other preloaded images for other distros during the guest-package-build concourse pipeline.


drewhli:
/ok-to-test

Would you be able to squash all the commits?

PR #2315, opened 2024-05-14
Fix RHEL 9.4 installation errors due to redhat-insights client.

zmarano:
Since RHEL 9.4 (and possibly 9.3), installation fails at package selection due to the redhat-insights packages depending on subscription-manager. This only affects on demand as BYOS images have subscription-manager included.


a-crate (approved):

a-crate:
Does this affect ARM?

zmarano:
> Does this affect ARM?

Oh gosh yes it does! Thanks for noticing that. Fixed.

PR #2323, opened 2024-06-10
Add support for other COS milestones (+109, +105) and also ARM images

sejalsharma-google:
In order to add support for further milestones, we:
1) Take the patches from master and apply those (since these are the most recent patches).
2) Step 1 (above) then removes the requirement of specifying the overlays branch name, so that is removed from all occurrences.

In order to add support for arm images we:
1) make machine type configurable (t2a for arm images)
2) change zone to accommodate ARM (central)
3) replace COS customizer's disable auto update step with our own script (since the default step is not compatible with ARM images) 
4) rewrite the disabling read only root fs script (to not hang after the reboot).

CCing @a-crate, @dorileo, @zmarano.

### DIFF
--- a/daisy_workflows/image_build/install_package/cos/package_replacement/compile_debian_package.sh
+++ b/daisy_workflows/image_build/install_package/cos/package_replacement/compile_debian_package.sh
@@ -26,9 +26,8 @@
 #
 # Most of the code here is sourced from: https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/packagebuild/daisy_startupscript_deb.sh
 #
-# Args: ./compile_debian_package [overlays_branch] [guest_agent_version]
-# Example: ./compile_debian_package release-R113 094ef227ddf92165abcb7b1241ca44728c3086d1
-#     $1 [overlays_branch]: the COS milestone version (to apply the correct patches).
+# Args: ./compile_debian_package [guest_agent_version]
+# Example: ./compile_debian_package 094ef227ddf92165abcb7b1241ca44728c3086d1
 #     $2 [commit_sha]: the guest agent commit sha (to upgrade to).
 
 set -o errexit
@@ -36,9 +35,11 @@ set -o pipefail
 set -o nounset
 
 apply_patches() {
-  # Download the repositories and apply COS specific patches.
+  # Download the repositories and apply COS specific patches. NOTE: The patches will be
+  # pulled from the master branch in COS. This is based on the assumption that master
+  # will have the most recent guest agent version, and therefore the most recent patches.
   echo -e "\nATTENTION: Downloading the board-overlays and guest-agent repos...\n"
-  git clone https://cos.googlesource.com/cos/overlays/board-overlays --branch ${overlays_branch}
+  git clone https://cos.googlesource.com/cos/overlays/board-overlays --branch master
   git clone https://github.com/GoogleCloudPlatform/guest-agent.git
   cd guest-agent
   git checkout ${commit_sha}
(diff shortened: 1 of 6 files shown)


daisy_workflows/image_build/install_package/cos/package_replacement/compile_debian_package.sh
```diff
@@ -26,19 +26,20 @@
...
 set -o nounset
 
 apply_patches() {
-  # Download the repositories and apply COS specific patches.
+  # Download the repositories and apply COS specific patches. NOTE: The patches will be
+  # pulled from the master branch in COS. This is based on the assumption that master
+  # will have the most recent guest agent version, and therefore the most recent patches.
   echo -e "\nATTENTION: Downloading the board-overlays and guest-agent repos...\n"
-  git clone https://cos.googlesource.com/cos/overlays/board-overlays --branch ${overlays_branch}
+  git clone https://cos.googlesource.com/cos/overlays/board-overlays --branch master
```
a-crate, line 42: What if the new agent diverges so much that these patches no longer apply?

daisy_workflows/image_build/install_package/cos/package_replacement/disable_readonly_rootfs.sh
```diff
@@ -32,19 +31,23 @@ disable_vboot() {
...
+    # mount and is sourced at the top of the file.
+    local -r rootdev="$(rootdev -s)"
+    sudo blockdev --setrw "${rootdev}"
+    enable_rw_mount "${rootdev}"
+    umount "${dir}"
+    # Triggers an immediate reboot.
+    reboot
+    # Hang after reboot: the script should not continue executing (return) after
+    # the reboot call due to COS customizer design.
+    while true; do sleep 1; done
```
a-crate, line 44: Can you print something explaining this so that it shows up in build logs

daisy_workflows/image_build/install_package/cos/package_replacement/compile_debian_package.sh
sejalsharma-google: The goal is that we keep the patches required to a minimum. IIUC, there is only one patch still needed for COS: create host key patch. The homedir gid patch was already upstreamed. So if there are significant changes for create host key, it would be nice if we knew so we could update the patches too. What do you think?

Current patch set for reference:
https://cos.googlesource.com/cos/overlays/board-overlays/+/refs/heads/master/project-lakitu/app-admin/google-guest-agent/files/

a-crate: I'm thinking that this will definitely break when we publish the refactor of the guest-agent we are working on because the directory structure will change if nothing else.

When that happens we will have two options:
1) disable COS again (kind of defeats the purpose of doing this)
2) create some way to get a different set of patches (or no patches) for upcoming changes to guest-agent. Branches seem like a good solution for this.

sejalsharma-google: I see, do you mind expanding on the branch solution?

I also remember Leo mentioning, "in this case it's non upstreamable (we can work on adding an extra configuration with the "right semantics" and have one override the other)", is this a possible solution to avoid having this patch?

a-crate: If we discover that some version of the agent in the future needs a patch which can't be upstreamed, or that we need to drop a patch from the next version but not the current version of the agent, COS could provide a `guest-agent-next` branch or something with a new patchset. If there are better solutions for this that's fine too.

I don't think the specifics of the patch are important, because you also use that directory to store your distro configuration which might need new options if we add modules that COS wants to disable or something.
It's more that there should be a way to test the guest-agent using future packaging for COS without needing to affect current COS.

a-crate: Discussed out of band, we can return to this when it's needed.

a-crate (approved):

PR #2333, opened 2024-06-27
Add ARM support for COS preloading

sejalsharma-google:
This PR adds ARM support for COS preloading:

- Adds worker_machine and machine_type as args to the install_package_cos.wf.json: allows us to specify ARM specific machine types i.e. debian-11-worker-arm64, t2a-standard-1
- Moves compiling the debian package (compile_debian_package.sh) out of the cloudbuild and in the worker machine. This is because for ARM image, debian packaging should be compiled in an ARM env and cloudbuild does not support ARM.
- Delete's instance_config file in worker machine to pass all debian tests.

CCing @a-crate, @dorileo, and @zmarano.

### DIFF
--- a/daisy_workflows/image_build/install_package/cos/install_package_cos.wf.json
+++ b/daisy_workflows/image_build/install_package/cos/install_package_cos.wf.json
@@ -20,7 +20,7 @@
       "Description": "worker image"
     },
     "machine_type": {
-      "Value": "e2-medium",
+      "Required": true,
       "Description": "machine type"
     }
   },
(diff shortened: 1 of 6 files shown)


a-crate:
We need to set the new variables in concourse right? The ones that don't have a default anymore?

/ok-to-test
/lgtm

sejalsharma-google:
> We need to set the new variables in concourse right? The ones that don't have a default anymore?

Yes, we would have to pass them in. I haven't made that PR yet since COS is disabled on Concourse right now. When we can re-enable it, I'll re-write the calls

PR #2343, opened 2024-08-01
Update rocky_linux_8_optimized_gcp_arm64.cfg

elicriffield:
add exclude=kernel* to /etc/yum.repos.d/Rocky-BaseOS.repo to insure we don't update the kernel from that repo and get it from Rocky-CloudKernel.repo


a-crate:
Should we do this on rocky 9 too?

elicriffield:
> Should we do this on rocky 9 too?

I don't see why not, i'll make a PL for that too

zmarano:
> > Should we do this on rocky 9 too?
> 
> I don't see why not, i'll make a PL for that too

Yeah we should do this for all of the optimized images (as long as it works for this case).

elicriffield:
> > > Should we do this on rocky 9 too?
> > 
> > 
> > I don't see why not, i'll make a PL for that too
> 
> Yeah we should do this for all of the optimized images (as long as it works for this case).

i just made the rest of the files a new pl, https://github.com/GoogleCloudPlatform/compute-image-tools/pull/2344

a-crate (approved):

PR #2367, opened 2024-09-23
Install ubuntu-drivers-common and the precompiled nvidia driver.

a-crate:
/assign @jjerger
/hold


a-crate:
Added a commit to clean up the .deb afterwards when adding nvidia repos too.

jjerger (commented):
/lgtm

PR #2385, opened 2024-10-30
RHEL 8 for aarch64 changes.

zmarano:
A few  notes:

- Add a RHEL 9 byos arm64 build.
- hyperdisk-balanced disks can't be attached read-only and thus we need a branched disk exporter for now.
- Debian 12's pip3 version will fail if you try to overwrite system python packages (this is used in the bootstrapper).
- We don't need a separate installer workflow for aarch64.
- Moved the partition identification in the installer creation to after the partitions are created and use the symlinks that get created instead of appending a number (which only worked with virtio scsi).
- There is not an issue with aarch64 and x86_64 repo's being enabled - dnf will choose the correct architecture and ignore the other by default. There was however an issue with the version of the client configuration that is being handled separately.


daisy_workflows/export/disk_export_hyperdisk.wf.json
a-crate, line 1: Is this workflow necessary? Can the standard disk export just start using a machine type that's compatible with hyperdisk instead?

zmarano: The issue isn't actually the machine type- it's hyperdisk- balanced which doesn't support attaching as a read-only disk (at this time). We attach the second disk read-only because it is supposed to static at that point. Eventually we should be able to remove this divergence though.

a-crate (approved):

PR #2387, opened 2024-10-31
Cleanup EOL and unused workflows.

zmarano:
/hold


bkatyl (approved):

koln67:
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/51a4d95aa05ea1af2a5e75ccdbc71872e0635d70/daisy_workflows/image_build/debian/debian_12_arm64.wf.json#L19

This looks like the last remaining reference to debian 11 arm 64 which would need to be replaced with debian 12 arm 64.

zmarano:
> https://github.com/GoogleCloudPlatform/compute-image-tools/blob/51a4d95aa05ea1af2a5e75ccdbc71872e0635d70/daisy_workflows/image_build/debian/debian_12_arm64.wf.json#L19
> 
> This looks like the last remaining reference to debian 11 arm 64 which would need to be replaced with debian 12 arm 64.

Ah that was fixed in https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_build/debian/debian_12_arm64.wf.json#L19 ... but I will verify it works correctly-.

a-crate (approved):

PR #2416, opened 2024-11-22
Disable the eclipse module for RHEL 8 aarch64 to avoid warnings from dnf.

zmarano:
Removes warnings for nonexistent eclipse aarch64 module in RHEL 8.

```
Modular dependency problems:

 Problem 1: nothing provides requested module(eclipse:rhel8:8020120200522125903)
 Problem 2: nothing provides requested module(eclipse:rhel8:8030020201023061315)
Dependencies resolved.
```


a-crate (approved):
Is this only in RHEL? Not CentOS?

zmarano:
> Is this only in RHEL? Not CentOS?

It is RHEL 8 only (CentOS 8 and 8 Stream are EOL) and I have not seen this on Rocky Linux either. There are a few discussions about this general issue around https://access.redhat.com/solutions/7044647

PR #2417, opened 2024-11-26
Install dnf-plugin-protected-kmods in Rocky Accelerators

a-crate:
Will filter out kernel updates that are incompatible with critical kmods. Will make things easier for users when upgrading.

/cc @jjerger @zmarano 
/hold


a-crate:
This will be pulled in by a soft dependency. This will make upgrading to RESF repos later where it may or may not be available easier.

PR #2426, opened 2025-01-10
Update server and server core publish workflows for byol

jjerger:
Add support for byol environment publishing recently added to concourse

### DIFF
--- a/daisy_workflows/build-publish/windows/windows-server-2016-dc-core-uefi.publish.json
+++ b/daisy_workflows/build-publish/windows/windows-server-2016-dc-core-uefi.publish.json
@@ -30,13 +30,20 @@
   {{$time := trimPrefix .publish_version "v"}}
   "Images": [
     {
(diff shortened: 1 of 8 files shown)


bkatyl:
Has this been tested? I'm not sure it's going to fix the error as it was relating to the source file in GCS.

jjerger:
> Has this been tested? I'm not sure it's going to fix the error as it was relating to the source file in GCS.

The source file issue was because the name of the prefix set in the publish file. See:
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/build-publish/windows/windows-server-2025-dc-uefi.publish.json#L39

By keeping the prefix the same but changing the destination family name the source file will have the correct name. The caveat to this is that the image names for the byol and non-byol images will be identical but with different image families. Since the publish project is different, there won't be a name collision.

As for testing this- I'm not sure of a good way to do so since it relies on the Concourse method of building the image.

bkatyl (approved):

PR #2453, opened 2025-04-22
Fix CIT tests for COS

chenglongtang1:
Currently the log does not show up in the serial log with echo. So we switch to use `logger daemon.info ""` to print the log.

Also fixed the commands that failed with errors. For example the `apt install git` and `find / -type f`.

Example error of find command:
```
find: â€˜/proc/729/task/729/netâ€™: Invalid argument
find: â€˜/proc/729/netâ€™: Invalid argument
```

To avoid error, we only scan necessary folders, for example, `/usr`, `/etc`, `/opt` and `/var`.

### DIFF
--- a/daisy_workflows/image_build/install_package/cos/replacepackage.sh
+++ b/daisy_workflows/image_build/install_package/cos/replacepackage.sh
@@ -33,7 +33,10 @@ get_vars(){
 }
 
 compile_debian_binaries(){
-  sudo apt install -y git
+  logger -p daemon.info "Attempting to update apt package lists..."
+  sudo apt update 2>&1 | logger -t apt-update -p daemon.info
+  logger -p daemon.info "Attempting to install git..."
+  sudo apt install -y git 2>&1 | logger -t apt-install -p daemon.info
   ./compile_debian_package.sh ${COMMIT_SHA} ${ARCH}
 }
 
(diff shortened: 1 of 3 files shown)


daisy_workflows/image_build/install_package/cos/replacepackage.sh
```diff
@@ -53,11 +56,13 @@ set_machine_and_arch_type(){
 #   _NEW_IMAGE_FAMILY: The new image family for the preloaded image.
 #   _MACHINE_TYPE: The machine type for the source image: default (n1-standard-1) or t2a-standard-1 for ARM.
 create_preloaded_image(){
-  gcloud builds submit . --config=dev_cloudbuild.yaml --disk-size=200 --gcs-log-dir="${DAISY_LOGS_PATH}" --substitutions=_NEW_IMAGE_FAMILY="cos-preloaded-images",_BASE_IMAGE_PROJECT="cos-cloud",_BASE_IMAGE="${SOURCE_IMAGE}",_NEW_IMAGE="${DEST_IMAGE}",_DEST_PROJECT="gcp-guest",_MACHINE_TYPE="${MACHINE_TYPE}"
+  logger -p daemon.info "Attempting to submit gcloud build..."
+  gcloud builds submit . --config=dev_cloudbuild.yaml --disk-size=200 --gcs-log-dir="${DAISY_LOGS_PATH}" --substitutions=_NEW_IMAGE_FAMILY="cos-preloaded-images",_BASE_IMAGE_PROJECT="cos-cloud",_BASE_IMAGE="${SOURCE_IMAGE}",_NEW_IMAGE="${DEST_IMAGE}",_DEST_PROJECT="cloud-kernel-build",_MACHINE_TYPE="${MACHINE_TYPE}" 2>&1 | logger -t gcloud-build -p daemon.info
```
ChaitanyaKulkarni28, line 60: `_DEST_PROJECT="cloud-kernel-build"` -> is this temporary for testing purposes? I don't think it will work in concourse workflow service account might not have access to this project. Additionally, other part of the [workflow](https://github.com/GoogleCloudPlatform/guest-test-infra/pull/1369/files) (I've removed it temporarily since it was failing and blocking other steps) relying on these images would need to be updated as well if these images need to come from a different project.

chenglongtang1: Fixed in the newly pushed commit

ChaitanyaKulkarni28 (approved):

PR #2455, opened 2025-04-23
allow overriding disk type instead of hardcoding

ChaitanyaKulkarni28:
pd-ssd is not compatible with c4a,  allow overriding disk type instead of hardcoding

/cc @dorileo @drewhli

### DIFF
--- a/daisy_workflows/image_build/install_package/install_package.wf.json
+++ b/daisy_workflows/image_build/install_package/install_package.wf.json
@@ -26,6 +26,10 @@
     "machine_type": {
       "Required": true,
       "Description": "machine type"
+    },
+    "disk_type": {
+      "Required": true,
+      "Description": "disk type"
     }
   },
   "Sources": {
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_build/install_package/install_package.wf.json
```diff
@@ -39,12 +39,12 @@
         {
           "Name": "disk-worker",
           "SourceImage": "${worker_image}",
-          "Type": "pd-ssd"
+          "Type": "${disk_type}"
```
drewhli, line 42: Is this something we can set a default for? If possible, maybe we should have this default to `pd-ssd` unless otherwise specified.

ChaitanyaKulkarni28: Similar to machine_type we have a default set where workflow is called https://github.com/GoogleCloudPlatform/guest-test-infra/pull/1375

drewhli (approved):

PR #2456, opened 2025-04-29
Update short descriptions for Accelerator images

jjerger:
Makes them easier to distinguish on short descriptions from the non-accelerated image types

### DIFF
--- a/daisy_workflows/build-publish/enterprise_linux/rocky_linux_8_optimized_gcp_nvidia_570.publish.json
+++ b/daisy_workflows/build-publish/enterprise_linux/rocky_linux_8_optimized_gcp_nvidia_570.publish.json
@@ -35,7 +35,7 @@
     {
       "Prefix": "rocky-linux-8-optimized-gcp-nvidia-570",
       "Family": "rocky-linux-8-optimized-gcp-nvidia-570",
-      "Description": "Rocky Linux, Rocky Linux, 8 optimized for GCP, x86_64 optimized for GCP with the Nvidia 570 driver built on {{$time}}",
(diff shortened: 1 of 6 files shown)


daisy_workflows/build-publish/enterprise_linux/rocky_linux_8_optimized_gcp_nvidia_570.publish.json
```diff
@@ -35,7 +35,7 @@
     {
       "Prefix": "rocky-linux-8-optimized-gcp-nvidia-570",
       "Family": "rocky-linux-8-optimized-gcp-nvidia-570",
-      "Description": "Rocky Linux, Rocky Linux, 8 optimized for GCP, x86_64 optimized for GCP with the Nvidia 570 driver built on {{$time}}",
+      "Description": "Rocky Linux, Rocky Linux, 8 optimized for GCP with the Nvidia 570 driver, x86_64 optimized for GCP with the Nvidia 570 driver built on {{$time}}",
```
ajorg, line 38: Are we duplicating parts of the description then?

jjerger: Based on how the Cloud Console parses this field, I believe this is what we want to do, yes. Perhaps @bkatyl can comment if this is correct?

jjerger: Actually, I redid this a bit to not be redundant but should still have the proper short/long description. Would like @bkatyl to confirm but this should fix our issue.

PR #2480, opened 2025-06-05
Update installpackage resolv.conf check

jjerger:
Adds logic to check for and remove a broken symlink if one exists for resolv.conf while preserving the previous behavior for OSes that do not have a broken symlink or have a file at that location.

Context: Debian-13 moved to a symlink for resolv.conf which causes a terminating error when the builder finds a broken symlink it can't follow.

Verified that Rhel-8, Rhel-9, and Debian-12 can still build with this change.

### DIFF
--- a/daisy_workflows/image_build/install_package/installpackage.py
+++ b/daisy_workflows/image_build/install_package/installpackage.py
@@ -60,8 +60,8 @@ def main():
   logging.info('Mount dev filesystem in chroot')
   run('mount -o bind /dev /mnt/dev')
 
-  # Enable DNS resolution in the chroot, for fetching dependencies. In some cases
-  # the symlink to resolv.conf will break, and needs to be unlinked.
+  # Enable DNS resolution in the chroot, for fetching dependencies. In some
+  # cases the symlink to resolv.conf will break, and needs to be unlinked.
   if os.path.islink('/mnt/etc/resolv.conf'):
     if os.path.isfile('mnt/etc/resolv.conf'):
       os.rename('/mnt/etc/resolv.conf', '/mnt/etc/resolv.conf.bak')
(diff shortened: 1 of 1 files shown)


daisy_workflows/image_build/install_package/installpackage.py
```diff
@@ -60,9 +60,15 @@ def main():
   logging.info('Mount dev filesystem in chroot')
   run('mount -o bind /dev /mnt/dev')
 
-  # Enable DNS resolution in the chroot, for fetching dependencies.
-  if os.path.isfile('/mnt/etc/resolv.conf'):
-    os.rename('/mnt/etc/resolv.conf', '/mnt/etc/resolv.conf.bak')
+  # Enable DNS resolution in the chroot, for fetching dependencies. In some
+  # cases the symlink to resolv.conf will break, and needs to be unlinked.
+  if os.path.islink('/mnt/etc/resolv.conf'):
+    if os.path.isfile('mnt/etc/resolv.conf'):
```
drewhli, line 66: Nit: Should this have a `/` at the front like the others? `/mnt` vs. `mnt`.

daisy_workflows/image_build/install_package/installpackage.py
```diff
@@ -60,9 +60,15 @@ def main():
...
+      os.rename('/mnt/etc/resolv.conf', '/mnt/etc/resolv.conf.bak')
+    else:
+      os.unlink('/mnt/etc/resolv.conf')
+  elif os.path.isfile('mnt/etc/resolv.conf'):
```
drewhli, line 70: Ditto above.

jjerger: Updated accordingly

PR #2487, opened 2025-06-26
feat: externalize VPC and host project config for Windows image workflow

singhal0306:
- Added host_project_id, network, subnetwork, region as external variables
- Updated workflow JSON to use these variables
- Modified README to reflect new usage and GCS-based resource inputs


singhal0306:
@bkatyl, please check this PR.

bkatyl:
The addition of required fields are effectively a breaking change and as such this is not the type of change we can accept.

Unfortuntely, I also don't see a good way to add this support without it being a breaking change this since Daisy doesn't currently support conditional configuration blocks.

At the current time, I'd suggest forking with a local customization.

PR #2499, opened 2025-07-01
clean before installing guest packages to try to avoid checksum problems

lpleahy:


koln67:
you can also try dnf -y clean all in case there's a commnd prompt

PR #2525, opened 2025-07-23
configure /etc/profile.d for centos stream 10 kickstart to ensure google cloud sdk binaries are in the path as well

lpleahy:


jjerger (approved):
/lgtm

PR #2557, opened 2025-09-04
Add IDPF tag to EL ARM images.

drewhli:
The drivers seem to be also present on all the EL ARM images. Given that we added the tag to all the EL x86_64 images, it makes sense to do the same for ARM.

/hold


jjerger (approved):
/lgtm

PR #2575, opened 2025-10-13
Delete Rocky Linux Accelerator workflows

jjerger:
CIQ now owns this process; we don't need these anymore

### DIFF
--- a/daisy_workflows/build-publish/README.md
+++ b/daisy_workflows/build-publish/README.md
@@ -22,8 +22,8 @@ The following directories exist
 
 Note: At one point, [Rocky Linux](https://rockylinux.org/)
 builds were nested under `enterprise_linux` but have since been taken over by
-[CIQ](https://ciq.com/), which can be found
-[here](https://github.com/ctrliq/gcp-public-images/tree/main/daisy).
+[CIQ](https://ciq.com/), which can be found at
+https://github.com/ctrliq/gcp-public-images/tree/main/daisy .
 
 ## Standards
 
--- a/daisy_workflows/image_build/README.md
+++ b/daisy_workflows/image_build/README.md
@@ -21,8 +21,8 @@ The following directories exist (see relevant READMEs for more details):
 
 Note: At one point, [Rocky Linux](https://rockylinux.org/)
 builds were nested under `enterprise_linux` but have since been taken over by
-[CIQ](https://ciq.com/), which can be found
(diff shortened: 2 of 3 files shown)


jjerger:
I'm holding this real quick to confirm with CIQ they don't need these as reference or something.

jdieter:
@jjerger, we're not using any of these anymore.  We're also doing the optimized and vanilla Rocky builds, so if you want to remove _all_ the Rocky configs, go for it.  For reference, the configs are now at https://github.com/ctrliq/gcp-public-images.

bkatyl:
If the goal is to remove them then all of the image build workflows, publish workflows, and any other connected file should be included in this PR.

What I believe is missing is references/pointing to the CIQ build so we're clearly connecting these to the source that is used for the builds.

jjerger:
Would a text file in that directory pointing to the CIQ configs suffice here?

daisy_workflows/build-publish/README.md
```diff
@@ -11,18 +11,20 @@ These are [Daisy] and [`gce_image_publish`] workflows used by Google to build
...
-* [rhui](rhui/README.md) - workflows for internal use images
 * sqlserver - wrapper workflows for [sqlserver](../image_build/sqlserver)
 * windows - wrapper workflows for [windows](../image_build/windows)
 * windows\_container - wrapper workflows for
   [windows\_container](../image_build/sqlserver)
 
+Note: At one point, [Rocky Linux](https://rockylinux.org/)
+builds were nested under `enterprise_linux` but have since been taken over by
+[CIQ](https://ciq.com/), which can be found
+[here](https://github.com/ctrliq/gcp-public-images/tree/main/daisy).
```
bkatyl: I'd suggest not using [here] and instead ", which can be found at https://github.com/ctrliq/gcp-public-images/tree/main/daisy"

jjerger: Updated

SaswatPadhi:
@bkatyl, @jjerger can we add _all_ the Rocky workflow files, as suggested by Jonathan in https://github.com/GoogleCloudPlatform/compute-image-tools/pull/2575#issuecomment-3403267029, and un-hold + merge this?

jjerger:
This was handled with https://github.com/GoogleCloudPlatform/compute-image-tools/pull/2658

SaswatPadhi:
@jjerger I think #2658 only removed a subset of unused files -- it only removed the build-publish files, not the image-build files.

PR #2582, opened 2025-11-05
Add RHEL 9.4 EUS, RHEL 9.6 EUS, & RHEL 10.0 EUS Build Configs

TylerJDao:


bkatyl:
This adds 8 more kickstart configs instead of adding parameters and conditionals in the kickstart files that can change the behavior conditionally and reduce our burden when any part of the build needs to be updated.

I gave Hank the same feedback for the LVM work. I think these can be parameters that are send from the CI/CD or some the workflow file.

Ideally if it's in the CI/CD config then we're also not duplicating the workflow files.

TylerJDao:
Added RHEL 10.0 EUS

As discussed offline, I will continue with the existing new file method to creating these new images while working on the workflow parameterization in a followup PR

daisy_workflows/image_build/enterprise_linux/kickstart/rhel_10_0_eus.cfg
```diff
@@ -0,0 +1,238 @@
...
+name=Google Cloud SDK
+baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el10-x86_64
+enabled=1
+gpgcheck=1
+repo_gpgcheck=0
+gpgkey=https://packages.cloud.google.com/yum/doc/rpm-package-key-v10.gpg
+EOM
+%end
+
+%post --erroronfail
```
lpleahy: I think for these, `--log=/dev/ttyS0` was missed in this like in https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_build/enterprise_linux/kickstart/rhel_8_4_sap.cfg#L128

TylerJDao: Added `--log=/dev/ttyS0` to all instances

lpleahy (approved):

PR #2589, opened 2025-12-02
fix sql_server_2025 configuration file

shenpai35:


daisy_workflows/image_build/sqlserver/sql_install.ps1
```diff
@@ -296,18 +296,53 @@ function Install-SSMS {
     Write-Host "Not installing SSMS for config ${sql_server_config}"
     return
   }
-  Write-Host 'Installing SSMS'
+  
+  Write-Host 'Installing SSMS from layout'
+   
+  $SSMS_SOURCE_ZIP = "C:\SSMS_Installer.zip"
```
bkatyl: How is it obtaining this file? We were obtaining the exe from gcs, however the gsutil call is being removed.

daisy_workflows/image_build/sqlserver/sql_install.ps1
```diff
@@ -296,18 +296,53 @@ function Install-SSMS {
...
 
-  $gs_path = Get-MetadataValue -key 'daisy-sources-path'
```
bkatyl: Replacing this logic means we're fully committing to SMSS 22 with no clean way to revert this change.

The install script should be it's own PR so it can be easily rolled back.

shenpai35: OK, uncommitting this change and will create another PR for this.

bkatyl (approved):

PR #2624, opened 2026-02-10
Creating Python file that will create the Daisy workflow file to create the image & Added RHEL 10.0 for SAP

TylerJDao:
/cc @bkatyl 

This is a script that will be used to generate the image_build workflow files. This is to reduce the need for manual building & editing of these files. This script will need to be updated anytime a new image is to be released or changes are needed.

I've also ran the script and these are the initial changes that it will make to the existing image_build workflow files.
Changes include
1. Formatting changes due to how Python handles workflow writing
2. Adding in Missing GuestOsFeatures. The source of truth I used was the build-publish publish.json files since they are up to date while the build-publish wf.json files are not
3. Changed the build-rhel call to the consolidated image_build workflow which will also call the new consolidated kickstart file
4. Fixed inconsistencies w/ naming of image_name/family
5. Adding Architecture parameter to the image_build.wf.json files. Previously the "Arm" image_build.wf.json files would've created X86 images. We only set the Architecture in the build-publish.publish.json files so this workflow was missing them

Also added the files for RHEL 10.0 SAP & RHEL 10.0 SAP BYOS

### DIFF
--- a/daisy_workflows/image_build/enterprise_linux/rhel_10_0_eus.wf.json
+++ b/daisy_workflows/image_build/enterprise_linux/rhel_10_0_eus.wf.json
@@ -1,6 +1,10 @@
 {
-  "Name": "build-rhel-10_0_eus",
+  "Name": "build-rhel-10-0-eus",
   "Vars": {
+    "auto-generated-file": {
+      "Value": "",
+      "Description": "This file is Generated. Do not edit manually. Modify and run the generator script(write_image_build_workflow.py) instead. This variable is unused as a workaround to Daisy Workflow's lack of support for file-level comments."
+    },
     "google_cloud_repo": {
       "Value": "stable",
       "Description": "The Google Cloud Repo branch to use."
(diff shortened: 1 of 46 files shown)


daisy_workflows/image_build/enterprise_linux/rhel_10_0_eus.wf.json
```diff
@@ -16,18 +20,36 @@
     "publish_project": {
       "Value": "${PROJECT}",
       "Description": "A project to publish the resulting image to."
+    },
+    "el_release": {
+      "Value": "rhel-10",
+      "Description": "The Enterprise Linux (EL) release for the image"
+    },
+    "rhui_package_name": {
```
bkatyl: The customers won't know what the RHUI package name is. 

Since the workflow is for a specific version, we should be setting the variable in all these workflow files.

bkatyl (approved):

daisy_workflows/image_build/enterprise_linux/rhel_10_0_eus.wf.json
cadieuxnicolas51-cloud: > The customers won't know what the RHUI package name is. 
> 
> Since the workflow is for a specific version, we should be setting the variable in all these workflow files.

- [ ] - Yes@ [**__**]()

cadieuxnicolas51-cloud:
Any hot @chingor13

PR #2665, opened 2026-04-06
Replaced usage of pretty.Compare() with cmp.Diff()

CarsonGL:
Recommend using the cmp package for comparing complex structures in tests. `pretty.Compare` is based on visual representation and can miss subtle differences, such as between nil and empty slices.

### DIFF
--- a/gce_image_publish/publish/publish_test.go
+++ b/gce_image_publish/publish/publish_test.go
@@ -21,7 +21,7 @@ import (
 	"time"
 
 	daisy "github.com/GoogleCloudPlatform/compute-daisy"
-	"github.com/kylelemons/godebug/pretty"
+	"github.com/google/go-cmp/cmp"
 	computeAlpha "google.golang.org/api/compute/v0.alpha"
 	"google.golang.org/api/compute/v1"
 )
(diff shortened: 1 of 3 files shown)


gce_image_publish/publish/publish_test.go
```diff
@@ -427,11 +427,11 @@ func TestPublishImage(t *testing.T) {
 			t.Errorf("%s: did not get expected error from publishImage()", tt.desc)
 		}
 
-		if diff := pretty.Compare(dr, tt.wantCI); diff != "" {
-			t.Errorf("%s: returned CreateImages does not match expectation: (-got +want)\n%s", tt.desc, diff)
+		if diff := cmp.Diff(tt.wantCI, dr, protocmp.Transform()); diff != "" {
```
lpleahy: do we need this and similar below? I believe this is only used when we have protobuf-generated types

CarsonGL: You're right—since we aren't using protobuf-generated types, I'll remove those protocmp.Transform() right away.

lpleahy (approved):

TylerJDao (approved):

PR #2676, opened 2026-04-22
Migrate cleanerupper container to compute-images-tools

CarsonGL:


.DS_Store
lpleahy, line 1: Please remove this

lpleahy (changes requested):

.DS_Store
CarsonGL: This has been removed, please let me know if anything else needed.

cleanerupper__container/go.mod
```diff
@@ -0,0 +1,44 @@
+module github.com/GoogleCloudPlatform/compute-image-tools/cleanerupper_job
```
lpleahy: This looks like it is the old name still

lpleahy (changes requested):
The file path has two underscores `cleanerupper__container`, which should not be the case.

cleanerupper__container/go.mod
CarsonGL: Thanks for pointing out, this has been resolved. Please let me know anything else needed.

CarsonGL: This has been resolved, please let me know if anything else needed.

lpleahy (approved):

PR #2682, opened 2026-04-28
Add scripts to add deprecated-public-image tag cloudbuilds for older images and keep only the latest.

shenpai35:


scripts/deprecate-images.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+if [[ "$DRY_RUN" == "true" ]]; then
+  echo "--- DRY RUN MODE ENABLED (No changes will be made) ---"
+fi
+
+echo "Scanning $IMAGE_URL for older versions..."
+
+# Get all SHAs, sorted by upload time descending, skipping the latest one.
+OLD_SHAS=$(gcloud container images list-tags "$IMAGE_URL" --format='get(digest)' --sort-by='~timestamp' --filter="NOT tags:deprecated-public-image-* AND NOT tags:latest" | sed '1d')
```
SaswatPadhi: What about the `release` ones? Do we want to deprecate them here?

SaswatPadhi (approved):
Thanks; LGTM

PR #2683, opened 2026-04-29
update PROJECT_ID setting in cloudbuild yaml

shenpai35:


SaswatPadhi (approved):
Should work.

Confirmed on https://docs.cloud.google.com/build/docs/configuring-builds/substitute-variable-values.

PR #2694, opened 2026-05-13
Add back the RHEL 10 RHUI Beta Client to the RHEL 10.2 Beta image

TylerJDao:
/cc @bkatyl @lpleahy 

We are reversing the RHUI Client change made in [PR](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/2688) to use the RHEL Beta RHUI Client again. We are leaving out the version lock so that it will just use the latest version of each package.


TylerJDao:
We are no longer creating the Beta image now that RHEL 10.2 as GA'd. We will be removing RHEL 10.2 Beta image content in future PR's

PR #2699, opened 2026-05-22
Update rollback logic to deprecate image instead of delete.

nicolenliang:
Updated functionality for publish library allows for rollback image deprecation with introduction of -deprecate flag. Original operations for image deletion still remain intact for smooth migration.

### DIFF
--- a/gce_image_publish/main.go
+++ b/gce_image_publish/main.go
@@ -45,6 +45,7 @@ var (
 	noRoot         = flag.Bool("no_root", false, "with -source_gcs_path, append .tar.gz instead of /root.tar.gz")
 	replace        = flag.Bool("replace", false, "replace any images that already exist, should not be used along with -skip_duplicates")
 	rollback       = flag.Bool("rollback", false, "rollback image publish")
+	deprecate			 = flag.Bool("deprecate", false, "deprecate rollback image, defaults to delete")
 	print          = flag.Bool("print", false, "print out the parsed workflow for debugging")
 	validate       = flag.Bool("validate", false, "validate the workflow and exit")
 	noConfirm      = flag.Bool("skip_confirmation", false, "don't ask for confirmation")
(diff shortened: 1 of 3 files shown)


gce_image_publish/publish/publish.go
```diff
@@ -392,37 +392,55 @@ func publishImage(p *Publish, img *Image, pubImgs []*computeAlpha.Image, skipDup
 	return cis, dis, drs, nil
 }
 
-func rollbackImage(p *Publish, img *Image, pubImgs []*computeAlpha.Image) (*daisy.DeleteResources, *daisy.DeprecateImages) {
+func rollbackImage(p *Publish, img *Image, pubImgs []*computeAlpha.Image) (*daisy.DeprecateImages) {
```
bkatyl: Since the Function call returns are changing. I'd think this would be a breaking change to what's calling it?

I'm not seeing any changes outside of the function where the items that call this function are updated.

nicolenliang: resolved—created separate functions for rollback deprecation and obsoletion operations.

gce_image_publish/main.go
```diff
@@ -45,6 +45,7 @@ var (
 	noRoot         = flag.Bool("no_root", false, "with -source_gcs_path, append .tar.gz instead of /root.tar.gz")
 	replace        = flag.Bool("replace", false, "replace any images that already exist, should not be used along with -skip_duplicates")
 	rollback       = flag.Bool("rollback", false, "rollback image publish")
+	deprecate			 = flag.Bool("deprecate", false, "deprecate rollback image, defaults to delete")
```
bkatyl: There may be tab(s) in here causing the spacing issue.

Since this is a bool, how would we be able to decide if it's deprecate, obsolete, or delete?

bkatyl: https://docs.cloud.google.com/compute/docs/reference/rest/v1/images/deprecate

state | enum - The deprecation state of this resource. This can be ACTIVE, DEPRECATED, OBSOLETE, or DELETED.

Deprecate or Obsolete are the target states. Delete is still an option, but one we prefer not to use as it removes the object from the control plane with no way of undeleting it where as obsolete can be accessed if it is set to deprecated.

